PR #11, opened 2016-06-21
Implement a simple print portal

matthiasclasen:
For now, this just has a PrintFile method.

### DIFF
--- a/data/org.freedesktop.portal.Print.xml
+++ b/data/org.freedesktop.portal.Print.xml
@@ -24,20 +24,20 @@
 -->
 
 <node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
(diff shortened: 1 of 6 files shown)


data/org.freedesktop.portal.Print.xml
```diff
@@ -24,20 +24,20 @@
 -->
 
 <node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <interface name="org.freedesktop.portal.PrintRequest">
+    <signal name="Response">
+      <arg type="u" name="response"/>
+      <arg type="a{sv}" name="options"/>
+    </signal>
```
alexlarsson: I wonder if we should have a "standard" response like this in org.freedesktop.portal.Request.

PR #2, opened 2016-06-21
Add a portal for  network connectivity

matthiasclasen:
Can just proxy the GLibNetworkMonitor functionality over dbus, and make it work in GLib


matthiasclasen:
Note that I'm doing the checking for overall network access by parsing flatpak-info inside the sandbox. That is not a security measure; it is just to avoid reporting irrelevant information to the app.

PR #4, opened 2016-06-22
Add a portal for proxy information

matthiasclasen:
This can probably reuse an existing libproxy backend.

### DIFF
--- a/src/xdg-desktop-portal.c
+++ b/src/xdg-desktop-portal.c
@@ -19,6 +19,7 @@
 #include "print.h"
 #include "request.h"
 #include "network-monitor.h"
+#include "proxy-resolver.h"
 
 static GMainLoop *loop = NULL;
 XdpDocuments *documents = NULL;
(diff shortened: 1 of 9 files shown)


matthiasclasen:
Here too, I do the check for network access inside the sandbox. In this case, we could maybe do the check on the portal side, since we actually do method calls.

src/xdg-desktop-portal.c
```diff
@@ -355,6 +356,29 @@ on_bus_acquired (GDBusConnection *connection,
 
   g_debug ("providing portal %s", g_dbus_interface_skeleton_get_info (skeleton)->name);
   }
+
+  {
+  GDBusInterfaceSkeleton *skeleton = proxy_resolver_create (connection);
+  if (skeleton != NULL)
+    {
+      g_dbus_interface_skeleton_set_flags (skeleton,
+                                           G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
```
alexlarsson: The calls are async so we should perhaps not run these in a thread?

alexlarsson: Actually, the libproxy implementation just uses a thread to make it async, so maybe you should instead just use the sync call and RUN_IN_THREAD.

Issue #10, opened 2016-06-21
Where do we store portal state


matthiasclasen:
The designers favor an approach where we show the annoying dialog a few times, and if the user makes the same choice a few times, we just go ahead and do it automatically from there on. I think we already do something like that for geolocation.

For open uri, that would imply that we a table of

requesting app + content-type -> chosen app + count

In addition, we need a way to let users set a threshold value for the count, including a way to say "always ask".

alexlarsson:
Flatpak has something called the permission store which can be used for this. It stores the document db, as well as the geoclue db for the shell.

mariospr:
I've been looking into using the PermissionStore for storing the information needed for the OpenURI portal right now, but I'm a bit confused on how the Permission Store could be used for that case since it seems to aimed at the use case of the documents portals, but does not seem to fit rightaway (at least "semantically", if that makes any sense) for other, more generic cases.

I've checked https://github.com/GNOME/gnome-shell/blob/master/js/ui/status/location.js#L375 for the particular case and my impression is that the shell is using this permissions-related store as a generic database to store permissions-unrelated stuff, so I think I could do that for the purpose of the OpenURI portal, but I'd like a confirmation / correction on whether that's the right approach, if possible.

matthiasclasen:
Yes, we would use the permission store in a way that is not related to storing permissions, at least not directly: It is more like storing history of past interactions, and the desktop portal will derive decisions about the current interaction from it. Maybe we should be renaming it to something more neutral.

mariospr:
Thanks @matthiasclasen for the clarifications, it makes sense now. I'll continue using the permissions store as it is now, since the 'permissions' concept seems to be very deep inside flatpak and I rather not mess with that for now, if that's ok.

alexlarsson:
Well, the history is what makes it a permission in this usecase.
Anyway, the permission portal is pretty freeform in what you store in it. You either pick a uniqe table name for your use and go wild with how you use the id, data and the permission strings. Or you can share a table between multiple users. For instance, one can share a  permission by using a shared table with the id being the thing your're controlling permission to, say "geo-location" and the permissions being "exact", "fuzzy", etc. That way other users can reuse the table for things other than geo-location.

In fact, this is how gnome-shell uses it for geoclue, and we should share that table if we have a similar layout. I think its called "desktop" or something.

mariospr:
[...]

PR #19, opened 2016-06-23
Use permissions store to implement a simple 'remember latest choice' mechanism

mariospr:
For now, it uses a hardcoded threshold of 5 consecutive executions to determine whether we want to hint the backend, suggesting it to use the default option right away.

This addresses issue https://github.com/flatpak/xdg-desktop-portal/issues/10

### DIFF
--- a/src/Makefile.am.inc
+++ b/src/Makefile.am.inc
@@ -10,6 +10,14 @@ xdp_dbus_built_sources = src/xdp-dbus.c src/xdp-dbus.h
 xdp_impl_dbus_built_sources = src/xdp-impl-dbus.c src/xdp-impl-dbus.h
 BUILT_SOURCES += $(xdp_dbus_built_sources) $(xdp_impl_dbus_built_sources)
 
+FLATPAK_IFACE_FILES =\
+	$(FLATPAK_INTERFACES_DIR)/org.freedesktop.portal.Documents.xml \
+	$(NULL)
+
+FLATPAK_IMPL_IFACE_FILES =\
+	$(FLATPAK_INTERFACES_DIR)/org.freedesktop.impl.portal.PermissionStore.xml \
+	$(NULL)
+
 PORTAL_IFACE_FILES =\
 	data/org.freedesktop.portal.Request.xml \
 	data/org.freedesktop.portal.FileChooser.xml \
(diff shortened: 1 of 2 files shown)


matthiasclasen:
Looks good to me, overall. I agree on adding control-center UI for this eventually.

I think we should send along the last choice in the property bag, so the backend can preselect it (or does it do that based on the last-used api in glib ?)

I think the frontend should do the threshold  handling entirely itself: ie, if you see that the threshold is exceeded, don't even call out to the appchooser, just launch the app directly.

mariospr:
> I think we should send along the last choice in the property bag, so the backend can preselect it (or does it do that based on the last-used api in glib ?)

How it works now is that `g_app_info_set_as_last_used_for_type()` is called in the frontend when a choice has been made in the backend, so the next time that the list of recommended apps is queried there, the latest one will come first in the list before sending it over D-Bus.

Then, when that list of choices is sent to the backend the first one would be the latest used app, thus the one that should be preselected (although I don't think that's being done now anyway)

Another option would be to forget about that glib API and simply rely on the information saved in the permission store (last app and counter) and then send all that over to the backend, which I think is what you propose here? If we do that we'd also need to move the threshold logic there, of course.

> I think the frontend should do the threshold handling entirely itself: ie, if you see that the threshold is exceeded, don't even call out to the appchooser, just launch the app directly.

Hmm... that was my original idea, but the call to the AppChooser doesn't happen in the backend but in the frontend already, so that decision of not using it would have to be made there, right?

That said, I actually was a bit confused at the beginning when I say that the implementation of the org.freedesktop.portal.OpenURI service in the frontend did rely on the org.freedesktop.impl.portal.AppChooser portal, so I might be talking non-sense here. In any case, this is partially why I decided to keep the logic in the backend for now, in case we wanted to add a new org.freedesktop.impl.portal.OpenURI in the backend in case we wanted to skip the AppChooser altogether.

@matthiasclasen could you confirm it's ok then to move all that logic to the backend (in which case I'd appreciate a rough description of how you'd skip the AppChooser)? Thanks!

matthiasclasen:
Just to clarify: 
Frontend = xdg-desktop-portal
Backend = xdg-desktop-portal-gtk

The rough plan that was hashed out last week in Toronto is to keep the backends as simple as possible (just the ui).all portal-specific logic (registering filee, talking to the permission  store, etc) should remain in the frontend, where it can be shared across desktops.

In this paricular case, I expect the following flow:
- frontend receives openuri call
- frontend determines appid and content type 
- looks into permission store
- if threshold is met, just launch the app found there
- otherwise, determine candidate apps for content type
- call choose app on the backend, passing along the candidates and the last choice
- gets result back
- if user canceled, send back response
-else, update permission store
- launch selected app
- send back response

mariospr:
> Just to clarify:
> Frontend = xdg-desktop-portal
> Backend = xdg-desktop-portal-gtk

Yes, that's what I thought. But re-reading previous comments I think I misunderstood things as I thought you were asking to move the threshold logic to the backend, which was confusing, 

> The rough plan that was hashed out last week in Toronto is to keep the backends as simple as possible (just the ui).all portal-specific logic (registering filee, talking to the permission store, etc) should remain in the frontend, where it can be shared across desktops.

Yes, that is what I understood, which is why I implemented the threshold logic in the frontend in the first place. As I said above, the source of the confusion is that I misunderstood one part, but it's all clear now. 

> In this paricular case, I expect the following flow:
>
>     frontend receives openuri call
>     frontend determines appid and content type
>     looks into permission store
>     if threshold is met, just launch the app found there

Ah, ok... we can launch the app directly from there without involving any backend in this case, that was unclear to me until now. Thanks for clarifying.

>     otherwise, determine candidate apps for content type
>     call choose app on the backend, passing along the candidates and the last choice
>     gets result back
>     if user canceled, send back response -else, update permission store
>     launch selected app
>     send back response

This is pretty much what's in place now with the extra bits of sending the latest used app. For that we could use the glib API, but then that would not be specific to every flatpak, so I think we probably want to send the information stored in the permissions store instead.

I'll work on updating these branches today. Thanks for the feedback!

matthiasclasen:
Yes, I agree that sending the last-used info to the backend together with the choices is more in keeping with the overall approach.

mariospr:
PR updated. Sorry for the slow time of response, did not have precisely my best day today, but at least I hope these new 4 commits will address all your comments (Matthias and Alex).

Please review, thanks!

matthiasclasen:
Looks fine to me now. I might suggest moving the latest choice from the options to an actual parameter, but no big deal.

Issue #30, opened 2016-06-27
Don't show the application chooser dialog when only one candidate is available


mariospr:
Even though I know the idea behind desktop portals is that the user will get some kind of visual feedback whenever the app leaves the sandbox (e.g. a dialog), for the particular case of the OpenURI portal it might be interesting to skip that step if there's only one candidate application available.

For instance, in the particular case of Endless's users this might very well be the case when opening `http[s]://` URI schemes, since normally only one browser will be available and we don't want to expose that "detail" at all, so that they get a smooth experience (in a controlled environment).

So perhaps this is something that could be interesting for upstream too?

matthiasclasen:
We've discussed treating http:// in that way. 
I have reservations about not showing the dialog at all, right from the get-go.

mariospr:
I understand, it might very well be a case where it could fit our downstream case nicely, but not that good of a solution for the general case. But still thought I would ask, to be sure.

Feel free to close this and delete the pull request then, unless you want to suggest any changes (e.g. only for `http://` could be acceptable, maybe?), of course. Thanks

matthiasclasen:
I think a possible solution for this can be to make the threshold configurable. This could include values for 'always ask' (the paranoid case) and 'never ask' (your case)...

mariospr:
Configuring the threshold would not be enough for this purpose, since it's not that we don't want to show the dialog ever: we'd like to hide the dialog when there's only one candidate application, but still show it when more options are available.

Still, adding a way to configure the threshold sounds good, but I'm unsure what the best way to do so is at the moment. I assume that using GSettings is not an option here, so perhaps we could implement a simple configuration system that would allow us to specify some options per-portal (maybe using  keyfiles under a drop-in directory in /etc?), or simply extending the command line arguments of flatpak so that different settings could be chosen for different apps.

Let me know what you think and I can file a bug about it, and hopefully find some time to work on it too.

With regard to this issue, I will keep that patch downstream for the time being, and stop pushing :)

alexlarsson:
We *certainly* don't want to skip the dialog in general when there is only one handler, that is a very easy way to attack apps, for instance by using a file type that normally only has one kind of app that handles it.

It *may* be interesting to special case http and https though. We've talked about that before.

mariospr:
Understood, thanks for the clarification. I updated the PR to only consider http and https then for now. Let me know what you think, thanks!

matthiasclasen:
[...]

PR #31, opened 2016-06-27
open-uri: Only show the chooser dialog for http[s] when having several options

mariospr:
Otherwise, simply skip the call to the backend and launch the only candidate
application rightaway, to make the process smoother.

Note that this only applies for http:// and https:// URI schemes. Any other
scheme will show the chooser dialog even if only one candidate exists.

Fixes issue https://github.com/flatpak/xdg-desktop-portal/issues/30


matthiasclasen:
The commit has conflicts; can you redo it ?

mariospr:
Yes, it should be fine now

PR #17, opened 2016-06-28
Autodetect dying client and implicitly close outstanding requests

matthiasclasen:
We should get told by dbus when a client dies, and could then close any outstanding Request objects.

### DIFF
--- a/src/request.c
+++ b/src/request.c
@@ -57,6 +57,8 @@ request_finalize (GObject *object)
 {
   Request *request = (Request *)object;
 
+  xdp_unregister_request (request);
+
   G_LOCK (requests);
   g_hash_table_remove (requests, request->id);
   G_UNLOCK (requests);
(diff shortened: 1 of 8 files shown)


src/xdp-utils.c
```diff
@@ -187,3 +225,26 @@ xdp_connection_track_name_owners (GDBusConnection *connection)
...
                                       NULL, NULL);
 }
+
+void
+xdp_register_request (Request *request)
+{
+  AppIdData *data;
+
+  G_LOCK (app_ids);
+  data = (AppIdData *)g_hash_table_lookup (app_ids, request->sender);
```
alexlarsson: There is a race here, where the lookup could fail if the sender died and we got removed from the hashtable.

alexlarsson: It should perhaps even return FALSE in that case, to handle the fact that we will otherwise miss the client-dying signal.

src/request.c
```diff
@@ -173,6 +188,12 @@ request_unexport (Request *request)
   g_object_unref (request);
 }
 
+void
+request_sender_died (Request *request)
+{
+  g_signal_emit (request, signals[SENDER_DIED], 0);
```
alexlarsson: This doesn't look threadsafe at all. The AppIdData doesn't ref the requests, so it may very well be dead by now. And in general emitting signals cross-threads is fraught with issues.

matthiasclasen:
I pushed a new attempt at this in the sender-died2 branch

matthiasclasen:
Pushed another version which moves the closing to a thread. It depends on the new-backend-api branch, so needs to wait for that to land first.

alexlarsson:
@matthiasclasen New sender-died2 branch looks good, please merge.

PR #33, opened 2016-06-28
Redo the backend api

matthiasclasen:
Following the same pattern that we use for the portal api now.

### DIFF
--- a/src/Makefile.am.inc
+++ b/src/Makefile.am.inc
@@ -40,7 +40,6 @@ $(xdp_dbus_built_sources) : $(FLATPAK_IFACE_FILES) $(PORTAL_IFACE_FILES)
 	$(AM_V_GEN) $(GDBUS_CODEGEN)                            \
 		--interface-prefix org.freedesktop.portal.      \
 		--c-namespace Xdp                               \
-                --c-generate-autocleanup all                      \
 		--generate-c-code $(builddir)/src/xdp-dbus      \
 		--annotate "org.freedesktop.portal.Documents.Add()" "org.gtk.GDBus.C.UnixFD" "true" \
 		--annotate "org.freedesktop.portal.Documents.AddNamed()" "org.gtk.GDBus.C.UnixFD" "true" \
@@ -52,7 +51,6 @@ $(xdp_impl_dbus_built_sources) : $(FLATPAK_IMPL_IFACE_FILES) $(PORTAL_IMPL_IFACE
 	$(AM_V_GEN) $(GDBUS_CODEGEN)                            \
 		--interface-prefix org.freedesktop.impl.portal. \
 		--c-namespace XdpImpl                           \
-                --c-generate-autocleanup all                      \
 		--generate-c-code $(builddir)/src/xdp-impl-dbus \
 		$(FLATPAK_IMPL_IFACE_FILES) \
 		$(PORTAL_IMPL_IFACE_FILES) \
--- a/src/Makefile.am.inc
+++ b/src/Makefile.am.inc
@@ -29,6 +29,7 @@ PORTAL_IFACE_FILES =\
 	$(NULL)
 
 PORTAL_IMPL_IFACE_FILES =\
+	data/org.freedesktop.impl.portal.Request.xml \
 	data/org.freedesktop.impl.portal.FileChooser.xml \
 	data/org.freedesktop.impl.portal.AppChooser.xml \
 	data/org.freedesktop.impl.portal.Print.xml \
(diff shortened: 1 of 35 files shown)


src/Makefile.am.inc
```diff
@@ -39,6 +40,7 @@ $(xdp_dbus_built_sources) : $(FLATPAK_IFACE_FILES) $(PORTAL_IFACE_FILES)
 	$(AM_V_GEN) $(GDBUS_CODEGEN)                            \
 		--interface-prefix org.freedesktop.portal.      \
 		--c-namespace Xdp                               \
+                --c-generate-autocleanup all                      \
```
alexlarsson: This adds a dependency on a *very* recent glib. I'm not sure we want to do that. In flatpak I instead manually define autocleanups for the dbus types.

src/file-chooser.c
```diff
@@ -176,15 +229,27 @@ handle_open_file (XdpFileChooser *object,
...
+                                                  NULL, &error);
+  if (!impl_request)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
+  g_signal_connect (impl_request, "response", (GCallback)handle_open_file_response, request);
+
+  g_object_set_data_full (G_OBJECT (request), "impl-request", g_object_ref (impl_request), g_object_unref);
```
alexlarsson: We could even add a `GDBusProxy *impl_request` to `Request` to avoid the set/get_data() calls.

src/Makefile.am.inc
```diff
@@ -29,6 +29,7 @@ PORTAL_IFACE_FILES =\
 	$(NULL)
 
 PORTAL_IMPL_IFACE_FILES =\
+	data/org.freedesktop.impl.portal.Request.xml \
```
alexlarsson: This file seems to not be in the commit

src/open-uri.c
```diff
@@ -282,105 +322,36 @@ handle_open_uri (XdpOpenURI *object,
...
-  gint latest_chosen_count = 0;
-  g_auto(GStrv) in_permissions = NULL;
-
-  if (get_latest_choice_info (app_id, content_type, &latest_chosen_id, &latest_chosen_count) &&
-      (g_strcmp0 (chosen_id, latest_chosen_id) == 0))
+  impl_request = xdp_impl_request_proxy_new_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (app_chooser_impl)),
+                                                  G_DBUS_PROXY_FLAGS_NONE,
+                                                  g_dbus_proxy_get_name (G_DBUS_PROXY (app_chooser_impl)),
+                                                  request->id,
+                                                  NULL, &error);
```
alexlarsson: This GDbusProxy will:
a) Not handle signal emissions on a thread, you need to call something like set_proxy_use_threads (G_DBUS_PROXY (impl_request));
b) Not verify that the signals its handling are *actually* from the backend. In fact, it will even accept such calls from the sandboxed app. So the app could call the portal, then fake a "allowed" backend response.

I think instead of this signal setup we should use regular dbus method calls for the backend. We already have a reference to the requests now, as we're the one passing that in, so we can use that to close the window even though the dialog request has not returned yet.

src/Makefile.am.inc
matthiasclasen: It is in the first commit

alexlarsson:
The async replies are now running on the main thread. If they need to do any blocking i/o they need to use a thread, and even if they don't they need to grab the request lock.

alexlarsson:
Also, i think the async backend calls need to ref the request they pass as user_data.

alexlarsson:
Also, i think we're not listening to any dbus signals anymore, so the set_proxy_use_thread calls are not needed anymore.

src/request.c
```diff
@@ -58,9 +58,36 @@ request_on_signal_response (XdpRequest *object,
...
+  g_autoptr(GError) error = NULL;
+
+  REQUEST_AUTOLOCK (request);
+
+  if (request->exported)
+    {
+      if (request->impl_request &&
+          !xdp_impl_request_call_close_sync (request->impl_request, NULL, &error))
+        {
+          g_dbus_method_invocation_return_gerror (invocation, error);
```
alexlarsson: If we fail to close the backend here we should probably still unexport the request, and consider it dead.
There is not much else for the client to do, its unlikely closing the requests again would help. This is similar to the close() syscall, which closes the fd even on errors.

src/open-uri.c
```diff
@@ -202,18 +138,130 @@ launch_application_with_uri (const char *choice_id,
...
+
+  launch_application_with_uri (choice, uri, parent_window);
+  update_permissions_store (request->app_id, content_type, choice);
+
+out:
+  if (request->exported)
+    {
+      xdp_request_emit_response (XDP_REQUEST (request), response, options);
+      request_unexport (request);
+    }
```
alexlarsson: This snippet happens a lot, we should probably have a helper for it.

alexlarsson: In fact, a lot of things are still repeated, althought less so than before. I wonder if we can generalize things so that e.g. the request creation and locking is automatic.

alexlarsson:
Those can be handled later though. I think this looks good for a start

PR #53, opened 2016-07-11
Misc fixes

cosimoc:
Some misc fixes.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -293,13 +293,31 @@ handle_open_in_thread_func (GTask *task,
     }
   else
     {
+      g_autoptr(GError) error = NULL;
       g_autoptr(GFile) file = g_file_new_for_uri (uri);
       g_autoptr(GFileInfo) info = g_file_query_info (file,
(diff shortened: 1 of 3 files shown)


src/open-uri.c
```diff
@@ -293,13 +293,23 @@ handle_open_in_thread_func (GTask *task,
...
+
+      if (info != NULL)
+        {
+          content_type = g_strdup (g_file_info_get_content_type (info));
+        }
+      else
+        {
+          g_debug ("failed to fetch content type for uri %s: %s", uri, error->message);
+          return;
+        }
```
matthiasclasen: We already completed the OpenURI call at this point; I think we need to emit a Response signal here (with response==2, probably), and unexport the request. Like this:

      if (request->exported)
        {
          g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
          xdp_request_emit_response (XDP_REQUEST (request), 2, g_variant_builder_end (&opts_builder));
          request_unexport (request);
        }

Issue #32, opened 2016-06-28
Add a way to configure threshold for OpenURI portal


mariospr:
As mentioned in https://github.com/flatpak/xdg-desktop-portal/issues/30#issuecomment-228920715, it would be good to have a way to configure this threshold (and probably other portal-specific options that we might need) so that it can be adapted to different needs.

I initially thought of a key file under `/etc`, but perhaps GSettings is just fine, and cleaner/easier.

matthiasclasen:
yet another alternative is to store the threshold in the permissions db, together with the count, per application.

mariospr:
I suppose, so, but the permissions store is harder to manipulate than a configuration setting and makes it harder to set different options per-deployment, so I'm not sure.

I think GSettings or a key file would be a better way to do this if setting options per application was not a requirement. If it is, perhaps extending the command line options of flatpak to accomodate portal-specific opttions might work, as those could be specified in the json file for the builder too.

What do you think?

matthiasclasen:
Commandline options and metadata is not sufficient.

We need to give users control over this, in the end, with a control-center UI. Looking at the code for the privacy panel, it already reads geolocation settings out of the permission store; so that seems no problem on the control-center side.

mariospr:
Understood, permissions store it is then. I'll try to find some time tomorrow, otherwise next week. Thanks!

mariospr:
I have to apologize for the long time taken to get back to this, but I finally got to do it today. See the pull request here: https://github.com/flatpak/xdg-desktop-portal/pull/56

mariospr:
While running my flatpak'ed version of Thunderbird today, I noticed I made a mistake in the main patch merged last week: when clicking on links, Thunderbird only showed the app chooser dialog once because, after the first time, the threshold was incorrectly set to zero in the permissions store.

The problem is that, while I set sane defaults in `parse_permissions()` thinking of this use case, the truth is that `get_latest_choice_info()` would not even call that if the app was not already registered in the permissions store, and so the defaults set (i.e. zero) by the callers of `get_latest_choice_info()` would prevail, messing everything up.

I've pushed an additional commit in https://github.com/flatpak/xdg-desktop-portal/pull/60 associated to this same issue, but let me know if you'd rather have me open a new one instead.

Sorry for the silly mistake

matthiasclasen:
I think this is fixed now ?

mariospr:
It should, yes.

PR #56, opened 2016-07-28
open-uri: Allow configuring threshold to skip the dialog per application

mariospr:
Add a third field to the data stored to store the threshold used to know
when not to show the dialog for http(s) schemes, so that it can be set
to something different from the hardcoded default value, per application.

https://github.com/flatpak/xdg-desktop-portal/issues/32

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -43,7 +43,6 @@
 #include "documents.h"
 
 #define TABLE_NAME "desktop-used-apps"
-#define USE_DEFAULT_APP_THRESHOLD 5
 
 typedef struct _OpenURI OpenURI;
 
@@ -59,6 +58,13 @@ struct _OpenURIClass
   XdpOpenURISkeletonClass parent_class;
 };
 
+enum {
+  PERM_APP_ID,
+  PERM_APP_COUNT,
+  PERM_APP_THRESHOLD,
+  LAST_PERM
+};
+
 static XdpImplAppChooser *impl;
 static OpenURI *open_uri;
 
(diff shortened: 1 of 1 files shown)


src/open-uri.c
```diff
@@ -107,19 +108,26 @@ get_latest_choice_info (const char *app_id,
               permissions != NULL &&
               permissions[0] != NULL)
             {
-              g_auto(GStrv) permission_detail = g_strsplit (permissions[0], ":", 2);
-              if (g_strv_length (permission_detail) >= 2)
+              guint n_permissions;
+
+              g_auto(GStrv) permission_detail = g_strsplit (permissions[0], ":", 3);
+              n_permissions = g_strv_length (permission_detail);
```
matthiasclasen: I think all of this parsing is a bit annoying; it would be nicer to use the structure of the database instead of stuffing everything into one string. That could be done in several ways:
- just use 3 separate strings, instead of splitting one at ":"
- maybe use a separate table row for the thresholds, to cleanly separate them from the usage data

matthiasclasen:
Sorry, separate row won't work, since we already use one row per content-type in this table. But 3 separate strings would still be a nice improvement.

First string: app id
Second string: use count
Third string: threshold, or "" for 'always ask'

mariospr:
@matthiasclasen: Just updated the pull request with your suggestions addressed in the first commit.

The second commit is not strictly needed and can be dropped without side effects, but I'd still like to propose it since the  handle_open_in_thread_func() is growing too much IMHO.

Let me know what you think, thanks

PR #62, opened 2016-08-10
Updated Brazilian Portuguese translation

josephgbr:


josephgbr:
Sorry, i meant "Added" instead of "Updated".

PR #71, opened 2016-11-09
Add Czech translation

Marv-CZ:


po/cs.po
```diff
@@ -0,0 +1,77 @@
...
+"Language: cs\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;\n"
+"X-Generator: Gtranslator 2.91.7\n"
+
+#: src/device.c:191
+msgid "Turn On Microphone?"
+msgstr "Zapnout mikrofón?"
```
AsciiWolf: The word _microphone_ should be correctly translated as _mikrofon_, not _mikrof**ó**n_.

PR #58, opened 2016-12-10
Supporting different xdg-desktop-portal versions in apps

matthiasclasen:
I've looked a little at the portal's documentation and I think there isn't any documented way for a flatpak application to know which version of xdg-desktop-portal is installed in the system it's running on. I also think the apps can't define in their manifest which xdg-desktop-portal version they require to work properly.

This will be important to maintain compatibility with distros that don't upgrade often. For example, if Debian Stretch is released with xdg-desktop-portal 0.1 and camera support is added in version 0.2, applications using this feature will probably won't work in Debian and users won't know it until they try the app.

In Android they have solved [this way](https://developer.android.com/guide/topics/manifest/uses-sdk-element.html):
- Each Android release has an API version
- The apps define in their manifest the minimum and maximum API Levels under which they are able to run, as well as the preferred API Level that they are designed to support. (Note that the docs recommend [not to set a maxSdkVersion](https://developer.android.com/guide/topics/manifest/uses-sdk-element.html))
- The apps can[ check the platform version at runtime](https://developer.android.com/training/basics/supporting-devices/platforms.html#version-codes), in order to offer a degraded experience for older api versons.

### DIFF
--- a/data/org.freedesktop.portal.Account.xml
+++ b/data/org.freedesktop.portal.Account.xml
@@ -78,5 +78,6 @@
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="handle" direction="out"/>
     </method>
+    <property name="version" type="u" access="read"/>
   </interface>
 </node>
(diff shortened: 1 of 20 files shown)


jgarciao:
I was proposing having one unique version property for the hole xdg-app-portal release (p.e. XDG-APP-PORTAL-VERSION=1 for xdg-app-portal release 0.4.x) but I guess your aproach is more fine grained.

Thank you!

Issue #78, opened 2017-01-12
OpenURI portal does not open local paths 


mariospr:
At the moment the OpenURI portal expects a, well, valid URI to be passed to work, otherwise it won't be able to parse the scheme and content type needed to open the resource outside the sandbox, which is a shame since sometimes that prevents the portal from doing useful things that are left ignored.

For instance, if you have a sandboxed application that tries to open a directory by specifying its path (e.g. `/home/user/mydir`, instead of `file:///home/user/mydir`) it would be nice if the OpenURI portal would handle that seamlessly and offered the user the option to use some application outside the sandbox to handle it (e.g. Nautilus).

This has been something we found out today while making a flatpak of dropbox using the `--extra-data` finish argument: despite of including [Matt Watson's xdg-open shim for flatpak](https://github.com/mattdangerw/flatpak-xdg-open-shim/blob/master/xdg-open) inside the sandbox, Dropbox client's "Open Folder" button (shown right after setting up the account) would refuse to work, which I found was caused because Dropbox was trying to do `xdg-open /home/user/Dropbox`, that did nothing for the reason mentioned above.

hadess:
xdg-open is awful, and this one as well... Fix it to pass URIs, or even better, rewrite it in C.

FWIW, xdg-open upstream launches gvfs-open, which launches gio-open, which does convert the path to a URI by using g_file_new_for_commandline_arg():
https://git.gnome.org//browse/glib/tree/gio/gio-tool-open.c#n81

mariospr:
@hadess Thanks for the input, Bastien. Using a different thing instead of that shim is something I thought of too, but at the same time I think having the OpenURI portal handle local paths gracefully would not be a bad idea either in general.

At least I got quite confused when I realized that trying to use the OpenURI portal to open `file:///my/lovely/path` would work, while `/my/lovely/path` wouldn't, so I still think there's value on supporting that.

hadess:
I still don't find this agreeable. A URI means something particular, a path, especially a relative path is something else.

mariospr:
Maybe you're right, specially about the relative paths. We could limit this to absolute ones too, though

mariospr:
Just updated the PR to consider absolute paths only, @hadess @alexlarsson let me know what you think

mariospr:
After some additional discussion with @cosimoc locally, I think I'm going to close this PR and simply try to get a newer enough version of xdg-utils into our runtime, and that will probably work for us too.

[...]

PR #79, opened 2017-01-12
open-uri: Gracefully handle local (non-URI) paths too

mariospr:
Plus a small commit cleaning up some stray debugging prints I found around the other code.

https://github.com/flatpak/xdg-desktop-portal/issues/78


mariospr:
See my comment in the issue: I think you're right with relative paths, but for absolute ones I still see the benefit.

mariospr:
Just updated the PR to consider absolute paths only

mariospr:
Funny I close the issue but forgot to close this PR. Closing now

PR #86, opened 2017-02-13
Add an email portal

matthiasclasen:
This portal adds a simple method to send an email.

### DIFF
--- a/src/email.c
+++ b/src/email.c
@@ -1,0 +1,212 @@
+/*
+ * Copyright © 2017 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Matthias Clasen <mclasen@redhat.com>
+ */
+
+#include "config.h"
+
+#include <locale.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+
+#include "email.h"
+#include "request.h"
+#include "documents.h"
+#include "xdp-dbus.h"
+#include "xdp-impl-dbus.h"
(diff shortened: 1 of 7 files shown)


src/email.c
```diff
@@ -0,0 +1,212 @@
...
+{
+  Request *request = request_from_invocation (invocation);
+  const char *app_id = request->app_id;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(XdpImplRequest) impl_request = NULL;
+  GVariantBuilder options;
+
+  REQUEST_AUTOLOCK (request);
+
+g_print ("handle send email\n");
```
hadess: Debug.

matthiasclasen:
Alex pointed out that it works just as well to simply feed the mailto: url to the OpenUri portal. So, there is little need for this dedicated email portal after all; we can just wait for the full sharing portal to appear.

hadess:
> Alex pointed out that it works just as well to simply feed the mailto: url to the OpenUri portal.

The URI format is different for pretty much all email clients. Just in the few clients that GNOME users use, there are 4 different formats:
https://git.gnome.org//browse/nautilus-sendto/tree/src/nautilus-sendto.c#n369

matthiasclasen:
I would say: the uri format is defined by rfc 2368, and all the email clients are buggy...

hadess:
The [RFC](https://www.ietf.org/rfc/rfc2368.txt) doesn't mention anything about attachments. The reset of the "options" listed in your API docs are supported though.

matthiasclasen:
Hmm, true. Supporting attachments is somewhat important for my use case. So, would you recommend that we add this dedicated portal anyway ? Then we can copy the support for various mailers from nautilus-sendto into the backend implementation

matthiasclasen:
Alex points out that ComposeEmail might be a better name than SendEmail

hadess:
Ideally the UI part would be provided by a simple dialogue/sheet, and be integrated in the Sharing protocol. Not sure whether you prefer deprecating this portal or getting rid of it when Sharing is ready.

matthiasclasen:
I think we might just keep this around, even after sharing appears. There could be use cases where you really want to send an email, rather than offer a choice of transport, and this portal lets you provide a destination, not sure at this point whether the sharing poral will let you do the same.

src/xdg-desktop-portal.c
```diff
@@ -373,6 +374,11 @@ on_bus_acquired (GDBusConnection *connection,
   if (implementation != NULL)
     export_portal_implementation (connection,
                                   account_create (connection, implementation->dbus_name));
+
+  implementation = find_portal_implementation ("org.freedesktop.impl.portal.Email");
+  if (implementation != NULL)
+    export_portal_implementation (connection,
+                                  email_create (connection, implementation->dbus_name));
```
mariospr: Not for now, but I think it would probably make sense to refactor a bit the code in here, as only the D-Bus interface name and the create function are different in all these cases.

mariospr:
I looked at the code too and it all looks good to me too, with the minor comment that Ithink I agree on that ComposeEmail is probably a better name too.

About the overall scope of the portal itself, I think there's value in having a dedicated portal for email, regardless of the Sharing portal, even if most use cases will be covered by it, as some others will likely not be covered.

PR #98, opened 2017-04-11
notification: Subscribe to the right interface for signals

fmuellner:
Whoops :-)


matthiasclasen:
Strings :-(
Can't get them right.

Issue #100, opened 2017-04-27
request_init_invocation enters infinite loop on request ID conflict


jhenstridge:
In `request_init_invocation()`, the following code is used to allocate a random unique request ID:

    r = g_random_int ();
    do
      {
        g_free (id);
        id = g_strdup_printf ("/org/freedesktop/portal/desktop/request/%u", r);
      }
    while (g_hash_table_lookup (requests, id) != NULL);

Since the random number is only chosen once, if the ID conflicts in one iteration of the loop, it will always conflict.

Presumably a new random number should be chosen for each iteration of the loop.

mariospr:
I think you're right. Could you prepare a PR?

matthiasclasen:
Oops :-)

Yes, please

PR #101, opened 2017-04-28
If the random request ID creates a collision, pick a new random ID.

jhenstridge:
Fix for bug  #100, avoiding an infinite loop on collisions.


mariospr:
Thanks! Merged

PR #105, opened 2017-05-26
Add an OpenFile method

matthiasclasen:


mariospr:
I still think that it would be nice to move the code in the new `else` branch from `handle_open_in_thread_func`  to a separate function, but that won't be reused in any case, so not a big deal. Merged

PR #106, opened 2017-05-29
Send more data to the appchooser

matthiasclasen:
New mockups for the GNOME app chooser dialog require some more
data to display things properly. This commit adds the content type
and the filename.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -460,6 +460,7 @@ handle_open_in_thread_func (GTask *task,
   g_autofree char *scheme = NULL;
   g_autofree char *content_type = NULL;
   g_autofree char *latest_id = NULL;
+  g_autofree char *basename = NULL;
   gint latest_count;
   gint latest_threshold;
   gboolean always_ask;
@@ -569,6 +570,7 @@ handle_open_in_thread_func (GTask *task,
         }
 
       get_content_type_for_file (path, &content_type);
+      basename = g_path_get_basename (path);
 
       scheme = g_strdup ("file");
       uri = g_filename_to_uri (path, NULL, NULL);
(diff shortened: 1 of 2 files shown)


src/open-uri.c
```diff
@@ -600,10 +602,10 @@ handle_open_in_thread_func (GTask *task,
...
       /* Add extra options to the request for the backend */
-      g_variant_builder_add (&opts_builder,
-                             "{sv}",
-                             "last_choice",
-                             g_variant_new_string (latest_id));
+      g_variant_builder_add (&opts_builder, "{sv}", "last_choice", g_variant_new_string (latest_id));
+      g_variant_builder_add (&opts_builder, "{sv}", "content_type", g_variant_new_string (content_type));
+      if (basename)
+        g_variant_builder_add (&opts_builder, "{sv}", "filename", g_variant_new_string (basename));
     }
```
mariospr: Is there any reason why you would not want to send the content type and filename to the backend unless latest_id is set? It seems to me that sending them regardless would be useful to show the right application in the dialog (or to show the right mode of GS using `--search`)

mariospr:
Looks good to me in general, I just have a small comment in https://github.com/flatpak/xdg-desktop-portal/pull/106/files#r120103625 that I think would be nice to address before merging.

mariospr:
Ah! Also, I've run this myself locally to check how the new dialog actually looks, but I think it would be interesting if you could post a link here to the source mockups, for reference

mariospr:
I think these are the ones: https://wiki.gnome.org/Design/OS/UriFileTypePermissions

src/open-uri.c
matthiasclasen: good catch. Fixed

Issue #107, opened 2017-05-31
Can't OpenFile in /app


alexlarsson:
If an app wants to open a file in its own data, it will be a path in /app, this will not be mapped on the host, nor in the target app. This is common for instance if the app wants to spawn a reader for some docs (say a html or pdf) that is included with the app.

I think we need to identify this case and map /app to the actual location of the file is installed on the host system.

alexlarsson:
Same is true for /usr and the runtime i guess.

matthiasclasen:
that requires a fix to your file visibility check, I guess ?

alexlarsson:
I think we need to reverse-map /app to the real location in xdg-desktop-portal, at that point we know the location of these from the .flatpak-info file:
```
[Instance]
app-path=/var/lib/flatpak/app/org.gnome.gedit/x86_64/stable/3602fece3ffa10bc1add4195d36eca0bec2e648634059ee679f000d8387ba8e6/files
runtime-path=/var/lib/flatpak/runtime/org.gnome.Platform/x86_64/3.24/7e0cc2f15bbce94b7123aaa05591c58a69f24294f9544b300588164f80fc22de/files
```

PR #108, opened 2017-05-31
Make OpenFile work from sandbox

alexlarsson:
This makes OpenFile work in the sandbox by fixing various issues that were broken, and by rewriting paths in /app and /usr to the real location outside the sandbox.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -218,6 +218,13 @@ xdp_connection_lookup_app_id_sync (GDBusConnection       *connection,
   return app_id;
 }
 
+GKeyFile *
+xdp_invocation_lookup_cached_app_info (GDBusMethodInvocation *invocation)
+{
+  const gchar *sender = g_dbus_method_invocation_get_sender (invocation);
+  return lookup_cached_app_info_by_sender (sender);
+}
+
 char *
 xdp_invocation_lookup_app_id_sync (GDBusMethodInvocation *invocation,
                                    GCancellable          *cancellable,
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -29,15 +29,15 @@
 #include "xdp-utils.h"
 #include "request.h"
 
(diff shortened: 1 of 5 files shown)


alexlarsson:
This should fix https://github.com/flatpak/xdg-desktop-portal/issues/107

matthiasclasen:
I think the commit message for the first patch should say "Cache entire flatpak info..." ? Instead of "Patch" ?

src/xdp-utils.c
```diff
@@ -118,6 +118,17 @@ parse_app_info_from_fileinfo (int pid, GError **error)
   return g_steal_pointer (&metadata);
 }
 
+static char *
+xdp_get_app_id_from_info (GKeyFile *app_info,
+                          GError **error)
+{
+  const char *group = "Application";
+  if (g_key_file_has_group (app_info, "Runtime"))
+    group = "Runtime";
```
matthiasclasen: This looks like it is fixing an unrelated bug about running bare runtimes ?

src/xdp-utils.c
```diff
@@ -218,6 +218,15 @@ xdp_connection_lookup_app_id_sync (GDBusConnection       *connection,
   return app_id;
 }
 
+GKeyFile *
+xdp_invocation_lookup_cached_app_info (GDBusMethodInvocation *invocation)
+{
+  const gchar *sender = g_dbus_method_invocation_get_sender (invocation);
+  g_autoptr(GKeyFile) app_info = NULL;
```
matthiasclasen: unused variable: app_info

src/xdp-utils.c
alexlarsson: Yeah, it handles the case of running a runtime directly.

src/xdp-utils.c
alexlarsson: pushed a fix

alexlarsson:
Pushed a rebased version with fixed commit message (and merged the unused variable fix)

PR #113, opened 2017-06-09
Improve handling of generated files

smcv:
xdg-desktop-portal tarballs contain a generated resources file which gets regenerated at build time anyway. This branch cleans that up.

It also moves cleanup of files generated by `make` from `make distclean` to `make clean`: the intended symmetry is that `make` and `make clean` go together, and `configure` and `make distclean` go together.

Finally, it updates the `.gitignore` to ignore all generated files.


mariospr:
Another one that looks good to me, and that also builds and passes `make distcheck` fine.

Merged, thanks!

PR #117, opened 2017-07-02
Avoid a race in the portal API

matthiasclasen:


src/request.c
```diff
@@ -156,27 +156,113 @@ request_authorize_callback (GDBusInterfaceSkeleton *interface,
...
+ * information, just hardcode it here for now.
+ */
+static const char *
+get_token (GDBusMethodInvocation *invocation)
+{
+  const char *interface;
+  const char *method;
+  GVariant *parameters;
+  g_autoptr(GVariant) options = NULL;
+  const char *token = "t";
```
mariospr: I think this should be declared as `static`

mariospr: I think this should be declared as static

data/org.freedesktop.portal.Account.xml
```diff
@@ -37,6 +39,14 @@
         Supported keys in the @options vardict include:
         <variablelist>
           <varlistentry>
+            <term>token s</term>
```
mariospr: I think a reference to @options in the description of GetUserInformation is missing between line 34 and line 35

mariospr:
I have looked into this now (sorry for the delay, didn't find the right time until now) and looks good in general, but I found a couple of things in the PR (see comments there) plus what seems to be to issues in the glib and gtk+ branches:

  * In the **gtk+ branch**: both commits for the [filechooser](https://git.gnome.org/browse/gtk+/commit/?h=portal-race&id=d67f0e58c18b479311e3a6a5523705a5ef637f82) and [print](https://git.gnome.org/browse/gtk+/commit/?h=portal-race&id=9ec4a539fec789d906f78aca6146f5c3cce65380) portals are leaking `token`, should probably just `g_autofree`. Maybe make `sender` a `g_autofree` too, so that you don't need to call `g_free`.

  * In the glib branch, in the commit for the [openuri portal](https://git.gnome.org/browse/glib/commit/?h=portal-race&id=054706b270555a8a3aa7ca702e94ea60b1678d7b), you're setting the value of `token` and using it to register for the signal, but forgot to add it to `opt_builder`, so the portal won't be able to use it to generate the path in `get_token()`

Ah! I suppose it will be done when releasing, but I just wanted to point out that this change should be reflected in the NEWS file too, so that packagers are aware of it when managing deps.

Other than these things, it looks good to me.

src/request.c
matthiasclasen: No, I actually mean "handles". That is what we call the out argument everywhere: @handle.

matthiasclasen: I've rewritten the "t" part to make it more clear what that was about: We want get_token() to always return a valid object path element, not NULL.

matthiasclasen:
Looking at the print part closely made me spot a problem: We already take a token in the options vardict there, but it is an integer, not a string.

For now, I've added a commit that doesn't do the token thing for Print - it doesn't matter in the short term since GTK+ is not listening for the Response signal here anyway. Alternatives include: use a different name than token for the argument just in in this case, or use a different name everywhere. Possible names: handle_token, nonce. Thoughts ?

matthiasclasen:
I've fixed the leaks you pointed out in glib and gtk+ branches.

src/request.c
mariospr: > No, I actually mean "handles". That is what we call the out argument everywhere: @handle.

Ah, ok

> I've rewritten the "t" part to make it more clear what that was about: We want get_token() to always return a valid object path element, not NULL.

Ah, I see it better now, thanks for clarifying. However, unless I'm getting something fundamentally wrong (which considering my current state of mind is not unlikely) I don't think the new code is right either, looks to me like you'll be returning a pointer to potentially invalid memory. Let me explain it with a code excerpt, so that I'll make it even more embarrassing if I end up being wrong again:

```
static const char *
get_token (GDBusMethodInvocation *invocation)
{
  [...]
  g_autoptr(GVariant) options = NULL;
  const char *token;

  [...]

  if (strcmp (interface, "org.freedesktop.portal.Account") == 0)
    {
      options = g_variant_get_child_value (parameters, 1);
    }
  else if (strcmp (interface, "org.freedesktop.portal.Device") == 0)
    [...]

  if (options)
    g_variant_lookup (options, "token", "&s", &token);

  return token ? token : "t";
}

void
request_init_invocation (GDBusMethodInvocation *invocation, const char *app_id)
{
  [...]
  const char *token;

  request = g_object_new (request_get_type (), NULL);
  request->app_id = g_strdup (app_id);
  request->sender = g_strdup (g_dbus_method_invocation_get_sender (invocation));
  request->app_info = xdp_invocation_lookup_cached_app_info (invocation);

  token = get_token (invocation);
  sender = g_strdup (request->sender + 1);
  for (i = 0; sender[i]; i++)
    if (sender[i] == '.')
      sender[i] = '_';

  id = g_strdup_printf ("/org/freedesktop/portal/desktop/request/%s/%s", sender, token);

  [...]
}
```

If I'm getting this right, the `token = get_token (invocation)` line in `request_init_invocation()` is getting a `const char*` pointer to either the const string `"t"` in the .data section of the program (so that's ok) or a direct pointer to the string inside the `GVariant options` (see `g_variant_lookup (options, "token", "&s", &token)`) that lives inside the scope of `get_token()`, and gets freed after returning from that function.

So, my understanding is that `token` will be pointing to invalid (already freed) memory when you reference it from the `id = g_strdup_printf(...)` line, which doesn't seem right.

That's why I was proposing so use a `static` qualifier for the `token` variable in `get_token()`: to make sure that the pointer you returned pointed to valid memory. Now I just realized my suggestion was incomplete, since you'd also need to do something else like replace `"&s"` with a `"s"` in the format string when extracting the token, and at that point I wonder whether it wouldn't be simpler to just return a non-const char* in any case, and free it from `request_init_invocation()`, so that we don't mess with static pointers to strings which are probably not the best thing to use here in any case.

Does that make sense? If not, please feel free to point out if I said something silly in my reasoning, I might very well be overlooking something.

data/org.freedesktop.portal.Account.xml
mariospr: This comment still applies

mariospr:
> For now, I've added a commit that doesn't do the token thing for Print - it doesn't matter in the short term since GTK+ is not listening for the Response signal here anyway. Alternatives include: use a different name than token for the argument just in in this case, or use a different name everywhere. Possible names: handle_token, nonce. Thoughts ?

I know it's a bit more cumbersome to change now, but thinking of the long term I think having the same name everywhere is less confusing and makes sense. FWIW, I think `handle_token`, or even `request_token`, would be good names

> I've fixed the leaks you pointed out in glib and gtk+ branches.

I saw those, and also that you added token to the vardict in the OpenUri portal too, thanks!

src/request.c
matthiasclasen: You are right that the token is pulled out of the opions GVariant, which is autofree'd when we leave the scope of get_tokens(). But options is itself part of the parameters GVariant, which will remain perfectly valid until we're done handling this method invocation. Therefore, the token will be good as long as we need it. Convinced ?

mariospr: Ahh! I see it now, thanks for clarifying, I knew I was missing something. A bit too obfuscated for my taste, but I'm convinced now :-)

matthiasclasen:
Renamed the option to handle_token, and used this for the Print method as well now.

mariospr:
Looks good to me, and thanks for adding the extra comment on `get_token()`. It would help people like me understanding what's going on better. I also see you updated the branches for glib and gtk+ to use `handle_token`, so I think this is good to merge now. Doing so...

PR #119, opened 2017-07-12
Change the email portal api to take fds

matthiasclasen:
This not only fixes the issue of guaranteeing that the sandboxed app has access to the file it wants to attach, it also fixes the problem where /tmp inside the sandbox is different from /tmp outside the sandbox, since we do the necessary path translation here to make sure that the portal backend gets a path that it can access.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -317,3 +317,71 @@ xdg_desktop_portal_error_quark (void)
   return (GQuark) quark_volatile;
 }
 
+char *
+xdp_get_path_for_fd (GKeyFile *app_info,
+                     int fd)
+{
+  g_autofree char *proc_path = NULL;
+  int fd_flags;
+  char path_buffer[PATH_MAX + 1];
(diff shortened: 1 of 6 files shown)


src/xdp-utils.c
```diff
@@ -317,3 +317,71 @@ xdg_desktop_portal_error_quark (void)
...
+char *
+xdp_get_path_for_fd (GKeyFile *app_info,
+                     int fd)
+{
+  g_autofree char *proc_path = NULL;
+  int fd_flags;
+  char path_buffer[PATH_MAX + 1];
+  g_autofree char *rewritten_path = NULL;
+  char *path;
+  ssize_t symlink_size;
```
mariospr: I know this was not a problem before, but I think it would be nice to initialize this to zero. It will be used to do `path_buffer[symlink_size] = 0` and not being assigned any value at that point is a possibility, so I think I'd rather not rely on undefined behaviour here.

mariospr:
Just a minor comment to consider when pushing, but LGTM

PR #122, opened 2017-09-19
Add screen cast portal

jadahl:
This adds a screen cast portal, provided via org.freedesktop.portal.ScreenCast. A sample user illustrating how its used can be found here: https://gitlab.gnome.org/snippets/19 . A xdg-desktop-portal-gtk implementation (needing mutter 3.26 + https://bugzilla.gnome.org/show_bug.cgi?id=787715 ): https://github.com/jadahl/xdg-desktop-portal-gtk/commits/wip/screen-cast .

There are some things to note about it, and as is, it is not ready to be merged. More on that below.

1.

This is, as far as I can tell, the first portal API that introduces long lived sessions. I need this so that I can later couple a screen cast session with a remote desktop session. To summarize, doing remote desktop might (most likely) involve screen casting, but we should not pop up two portal dialogs; thus we need a combined "Start" switch.

Sessions here are currently associated by passing a session ID that the application retrieves from the Result vardict from a "CreateSession" method. This is then passed to every method that depend on a given session (SelectSources, Start, Stop).

The "combined" sessions will be done by having the application first create a org.freedesktop.portal.RemoteDesktop session (in the same way as is done in the commit introducing the ScreenCast in this PR). The session ID is then passed as an option when creating the org.freedesktop.portal.ScreenCast session. This will mean that the screen cast session is controlled by the remote desktop session, which in practice means that the application Starts/Stops both sessions atomically using the remote desktop session. Vardict entries from the screen cast Start method will instead be passed via the Result object of the remote desktop Start method.

2.

This requires PipeWire, which is not yet possible to be used from inside a flatpak sandbox. As the idea is to have the portal pass the PipeWire remote file descriptor to the client, this portal API naturally must be changed to also do this. The intention here is to first get feedback on the portal API itself, then make the needed adaptations for passing a file descriptor.

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -1,0 +1,176 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2017 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.ScreenCast:
+      @short_description: Screen cast portal
+  -->
+  <interface name="org.freedesktop.impl.portal.ScreenCast">
+    <!--
+        CreateSession:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session being created
+        @app_id: App id of the application
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Create a screen cast session.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>session_id s</term>
+            <listitem><para>
+              The session id. A string representing the created screen cast session.
+            </para></listitem>
+          </varlistentry>
+      </variablelist>
+    -->
+    <method name="CreateSession">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In3" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+    <!--
+        SelectSources:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @app_id: App id of the application
+        @session_id: Session identifier
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Configure what the screen cast session should record.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>types u</term>
+            <listitem><para>
+              Bitmask of what type of content to record. Default is MONITOR.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>multiple b</term>
+            <listitem><para>
+              Whether to allow selecting multiple sources. Default is no.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+        For available source types, see the AvailableSourceTypes property.
+    -->
+    <method name="SelectSources">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+    <!--
+        Start:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @app_id: App id of the application
+        @parent_window: Identifier for the application window
+        @session_id: Identifier for the screen cast session
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Start the screen cast session. This will typically result the portal presenting
+        a dialog letting the user do the selection set up by SelectSources.
+
+        The @parent_window identifier must be of the form "x11:$XID" for an X11
+        window. Support for other window systems may be added in the future.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>streams a(ua{sv})</term>
+            <listitem><para>
+              An array of PipeWire streams. Each stream consists of a PipeWire
+              node ID (the first element in the tuple, and a Vardict of
+              properties.
+
+              The array will contain a single stream if 'multiple' (see
+              SelectSources) was set to 'false', or at least one stream if
+              'multiple' was set to 'true' as part of the SelectSources method.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+        Stream properties include:
+        <variablelist>
+          <varlistentry>
+            <term>position (ii)</term>
+            <listitem><para>
+              A tuple consisting of the position (x, y) in the compositor
+              coordinate space. Note that the position may not be equivalent to a
+              position in a pixel coordinate space. Only available for monitor
+              streams.
+            </para></listitem>
+          </varlistentry>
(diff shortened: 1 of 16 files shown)


alexlarsson:
The handling of the sessions is not threadsafe. You can have parallel callers of screencast methods, closing of request handles or disappearance of dbus names trigger code, but nothing protects the sessions hash table. The data in that hashtable is never returned from check_session though, so I think all you need is to wrap a mutex around that and where things are added/removed from the hash.

Also, I think the API is a bit weird wrt session lifetimes, or at least not cleanly specified. The common scenario will be the client:
* create session
* select sources
* start
* stop (which terminatest the session)

But this leaves a couple of questions for other scenarios:
* What happens if the client wishes to terminate the session but it was not started. For instance, maybe the start call fails, or the user closed the window before the screencasting was started.
* Can i reuse one session for multiple screencasts
* What happens if i call start multiple times, or stop before start. Or change the sources while a screencast is active.
* What happens if the app creates a session, then dies before it started the recording, but the portal still calls force_stop()

I think you should make session close an explicit operation (and make owner disappearing from bus the same operation as this). You also need to track the state of the screencase (is it active or not, etc) and return errors to the client if its in the wrong state (double start, stop before start, etc). This will validate stuff in the common code so that the portal implementations don't have to care about weird cases. Additionally, it will make it possible for the session close/client exit to automatically do the right thing (stop only if active). Also, you need to decide if multiple start/stop pairs can be done on a single session (and if not return errors if this happens).

jadahl:
The new branch contains the new reworked API, which means "sessions" are outsourced to a generic org.freedesktop.portal.Session API. Another change is that I made it a requirement to call "SelectSources()" before "Start()". The rest is more about clarifications on what may be called when etc.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -0,0 +1,172 @@
...
+        The @parent_window identifier must be of the form "x11:$XID" for an X11
+        window. Support for other window systems may be added in the future.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>streams a(ua{sv})</term>
+            <listitem><para>
+              An array of PipeWire streams. Each stream consists of a PipeWire
+              node ID (the first element in the tuble, and a Vardict of
```
matthiasclasen: s/tuble/tuple/

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -0,0 +1,159 @@
...
+        window. Support for other window systems may be added in the future.
+
+        The following results get returned via the
+        #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>streams a(ua{sv})</term>
+            <listitem><para>
+              An array of PipeWire streams. Each stream consists of a PipeWire
+              node ID (the first element in the tuble, and a Vardict of
```
matthiasclasen: s/tuble/tuple/

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -0,0 +1,159 @@
...
+              properties.
```
matthiasclasen: Is it the case that the array only has one member unless 'multiple' is true ? might be worth pointing out in the docs.

matthiasclasen:
Any update on when the final pipewire api will be available ?

src/session.c
```diff
@@ -0,0 +1,463 @@
...
+    return FALSE;
+
+  g_signal_connect (impl_session, "closed", G_CALLBACK (on_closed), NULL);
+
+  session->id = g_steal_pointer (&id);
+  session->impl_session = g_steal_pointer (&impl_session);
+
+  g_dbus_interface_skeleton_set_flags (G_DBUS_INTERFACE_SKELETON (session),
+                                       G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
+
```
alexlarsson: I think you should add a g-authorize-method callback here that only allows callers where the sender matches the sender part of the object path

data/org.freedesktop.portal.ScreenCast.xml
alexlarsson: Also, does the node id give you some kind of usable for-display identifier for the multiple streams if you were to show the multiple streams in the UI, or should that be in the dbus API?

jadahl:
> Any update on when the final pipewire api will be available ?

Wim added some kind of permission system in https://github.com/PipeWire/pipewire/commit/d5643e84725e928b43e64e6f83e83290229628b3 a few days ago but I have yet to see whether the portal can make use of it.

data/org.freedesktop.portal.ScreenCast.xml
jadahl: I think it'd make sense to add things to the vardict that comes with each stream. The idea is that we can add things as we need them. I guess a 'human readable name' ala *Built in display* or *Dell 24"* might be useful.

alexlarsson:
I had a look over this, and it seems ok to me. One thing I don't quite understand though is how the permissions are going to work under wayland? I mean, the protocol that the portal implementation uses to get the data needs some kind of authentication to not just break all the privacy between wayland clients (sandboxed or not).

jadahl:
> I had a look over this, and it seems ok to me. One thing I don't quite understand though is how the permissions are going to work under wayland? I mean, the protocol that the portal implementation uses to get the data needs some kind of authentication to not just break all the privacy between wayland clients (sandboxed or not).

The idea is that the flow should be something like this:

1. A sandboxed app requests a screen cast
2. The portal shows the UI and the user approves
3. The portal tells mutter to open a PipeWire stream which it does and hands over a stream node id
4. The portal opens a PipeWire remote, "hides" all nodes except the ones just opened by mutter
5. The portal "seals" (does something so that no further permission changes can be done) the PipeWire remote
6. The portal will hand over a file descriptor with the open sealed pw_remote to the app
7. The app opens its pw_remote using the file descriptor, and should now with this handle only be able to see the streams that was opened exactly for the screen cast session it received the pw_remote fd from.

jadahl:
Or what do you mean with data needed? The portal implementation is outside of the sandbox; the apps should only see the communication between the sandox and the non-impl side of just the channel between the portal and the app, right? I mean, two separate apps will not see the portal communication done between the other app and the portal, right? Where do you see any privacy being breached?

alexlarsson:
Ah, the thing I was missing was the "sealing". I.e. what makes the returned fd different from just a connection to the entire pipewire graph.

matthiasclasen:
We are still waiting for pipewire permission api to be integrated here before this can be merged, correct ?

jadahl:
> We are still waiting for pipewire permission api to be integrated here before this can be merged, correct ?

Right. I talked to Wim earlier and can move forward with the API currently on the PipeWire master branch.

jadahl:
This new branch needs PipeWire checked out from master from yesterday (22nd Jan), preferably at least commit f2aafffb2c68a023fd20b08451de20d1e2808fb3 to make the test python app able to work. It also needs an updated mutter; I'll provide a link to the corresponding branch after I cleaned it up a bit.

data/org.freedesktop.impl.portal.Session.xml
```diff
@@ -0,0 +1,52 @@
...
+
+      The Session interface is shared by all portal interfaces that involve
+      long lived sessions.  When a method that creates a session is called, the
+      reply will include a session handle (i.e. object path) for a Session
+      object, which will stay alive for the duration of the session.
+
+      The portal can abort the interaction by calling
+      org.freedesktop.impl.portal.Session.Close() on the Session object.
+  -->
+  <interface name="org.freedesktop.impl.portal.Session">
```
alexlarsson: Can you add a version property here:
```
```
That way we can extend this api later and do runtime checks on it.

data/org.freedesktop.portal.Session.xml
```diff
@@ -0,0 +1,74 @@
...
+
+      The token that the caller provides should be unique and not guessable. To
+      avoid clashes with calls made from unrelated libraries, it is a good idea
+      to use a per-library prefix combined with a random number.
+
+      A client who started a session vanishing from the D-Bus is equivalent to
+      closing all active sessions made by said client.
+  -->
+  <interface name="org.freedesktop.portal.Session">
+
```
alexlarsson: version here too

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -0,0 +1,182 @@
...
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.ScreenCast:
+      @short_description: Screen cast portal
+  -->
+  <interface name="org.freedesktop.portal.ScreenCast">
+    <!--
```
alexlarsson: Want a version property here too, and on the impl.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -0,0 +1,182 @@
...
+
+        Retrieve the file descriptor to the PipeWire remote where the screen
+        cast streams are available. The file descriptor should be used to
+        create a <classname>pw_remote</classname> object, by using
+        <function>pw_remote_connect_fd</function>.
+
+        This function can only be called once, if there was one or more stream
+        shared in the response to starting the session.
+    -->
+    <method name="TakePipeWireRemote">
```
alexlarsson: I think you need:
```
      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/> 
```
here, so that you get the right codegen for the fds.

alexlarsson: Oh, i see you did that in the makefile. We're changing those to inline annotations instead.

data/org.freedesktop.portal.ScreenCast.xml
alexlarsson: Oh, i see you have one at the end.

src/screen-cast.c
```diff
@@ -0,0 +1,1050 @@
...
+      g_dbus_method_invocation_return_error (invocation,
+                                             G_DBUS_ERROR,
+                                             G_DBUS_ERROR_FAILED,
+                                             "No streams were started");
+      return TRUE;
+    }
+
+  fd_list = g_unix_fd_list_new ();
+  fd_id = g_unix_fd_list_append (fd_list,
+                                 pw_remote_steal_fd (remote->remote),
```
alexlarsson: I believe this leaks the fd. fd_list_append() dups the passed in fd and keeps that.

alexlarsson:
Overall this looks good to me, commented on some things.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -0,0 +1,184 @@
...
```
jadahl: I will change this to "OpenPipeWireRemote", and allow it to be called multiple times, having the portal open the PipeWire remote on demand. The reason is that this simplifies app-side by allowing to pass one pw_remote file descriptor per gstreamer object. The alternative would be to add ref-counting to pw_remote and teach the gstreamer things to share.

PR #124, opened 2017-10-06
Handle bubblewrap dropping /newroot prefix

cgwalters:
This is a port of https://github.com/flatpak/flatpak/pull/1063

(And really we should figure out how to share this security-sensitive
 code between the two projects...)

There's an oustanding bubblewrap PR where we'd like to change how we set up the
rootfs; a side effect of this will be that /newroot disappears from the /proc
links: bubblewrap pull 172.


cgwalters:
📞

Issue #128, opened 2017-11-24
Handle directories in the OpenFile method


mariospr:
As mentioned on IRC today, I found that the `OpenFile` method won't handle FDs pointing to directories even in the case where such directories are reachable both inside and outside the sandbox, which is kind of a problem for situation like one we have on Endless, where the Dropbox app (which has `filesystem=home` permissions granted) tries to open the file manager in the "Dropbox directory".

I've discussed this with @alexlarsson and he seems to agree on that we can add an additional check in https://github.com/flatpak/xdg-desktop-portal/blob/master/src/xdp-utils.c#L340, so that fail when the FD passed points to something that is **neither** a regular file nor a directory (instead of checking just for regular files).

This, of course, won't be useful in cases where such directory is not reachable inside the sandbox, since those kind of paths would need to be handled via the documents portal, and FUSE does not support directories, but it can be useful nonetheless in cases like the mentioned here.

I'll be linking to a PR soon later today.

Now see the full IRC log, for reference:
```
<msanchez> alexlarsson: hi, quick question about the OpenFile method if you don't mind me asking...
<alexlarsson> sure
<msanchez> only let me find a link first
<msanchez> https://github.com/flatpak/xdg-desktop-portal/blob/master/src/xdp-utils.c#L340
<msanchez> can you think of any reason why we could not extend that to (mode != S_IFREG && mode != S_IFDIR)? 
<msanchez> the reason I ask is because it would be nice to keep being able to open not just regular files but also directories (assuming you have something to handle that such as Nautilus, of course)
<alexlarsson> Well, at a minimum the fuse filesystem for documents don't handle directories
<alexlarsson> But also, documents are made for "document level" files
<msanchez> I'm not 100% sure that's a problem here. I added that extra check in there to see if it would work and that did the trick for me
<msanchez> when passing the fd for /home/mario (in a sandbox with read permission to my home) I could see the mode matching S_IFDIR
<alexlarsson> That is not the issue though
<msanchez> and just doing that change was the difference between my Dropbox flatpak being able to open the Dropbox directory or not 
<msanchez> hmm... I guess I'm missing things then
<alexlarsson> I mean, you could register one
<alexlarsson> but, could you read the directory in //run/user/1000/doc ?
<alexlarsson> I'd be surprised if it worked at all
<msanchez> ah, right, I see what you mean
<alexlarsson> and its definately not designed to be safe to do that
[...]

PR #130, opened 2017-11-24
xdp-utils: Also resolve file descriptors to paths for directories

mariospr:
This won't be useful if the path passed points to a directory not
reachable inside the sandbox, but it will at least allow opening
them in other cases where that might be the case.

https://github.com/flatpak/xdg-desktop-portal/issues/128


pwithnall:
fwiw, this looks good to me.

Issue #129, opened 2017-11-24
OpenFile method crashes when skipping the app chooser dialog 


mariospr:
When using the `OpenFile` method to deal with paths for which we have determined that [we can skip showing the app chooser](https://github.com/flatpak/xdg-desktop-portal/blob/8676f78/src/open-uri.c#L379), the portal frontend will crash because `uri` will not be set when the following line is reached:

https://github.com/flatpak/xdg-desktop-portal/blob/8676f78/src/open-uri.c#L525

This is fairly uncommon as that situation of skipping the chooser will only happen in practice when opening directories (as for http/https the OpenURI method would be used, and because it's common to have only one file manager around), which has not been a problem since directories were not being handled...  until now :-)

But I just filed issue #128 precisely to support directories too, making this a fairly more common problem.

mariospr:
Ready for review at https://github.com/flatpak/xdg-desktop-portal/pull/131

@matthiasclasen @alexlarsson Please let me know what you think, and merge if needed. Thanks!

PR #131, opened 2017-11-24
open-uri: Don't crash when opening files skipping the app chooser

mariospr:
When OpenFile is used to open a directory and either there's a default
application to handle file:// URIs, or only one appropriate application
is installed (e.g. Nautilus), we would hit a crash with the current code
since launch_default_for_uri() will be passed a NULL uri parameter.

This fixes that by ensuring that uri is always properly set at that point.

https://github.com/flatpak/xdg-desktop-portal/issues/129

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -512,7 +512,8 @@ handle_open_in_thread_func (GTask *task,
       basename = g_path_get_basename (path);
 
       scheme = g_strdup ("file");
-      g_object_set_data_full (G_OBJECT (request), "uri", g_filename_to_uri (path, NULL, NULL), g_free);
+      uri = g_filename_to_uri (path, NULL, NULL);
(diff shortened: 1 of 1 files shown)


src/open-uri.c
```diff
@@ -512,7 +512,8 @@ handle_open_in_thread_func (GTask *task,
       basename = g_path_get_basename (path);
 
       scheme = g_strdup ("file");
-      g_object_set_data_full (G_OBJECT (request), "uri", g_filename_to_uri (path, NULL, NULL), g_free);
+      uri = g_filename_to_uri (path, NULL, NULL);
+      g_object_set_data_full (G_OBJECT (request), "uri", g_strdup(uri), g_free);
```
pwithnall: Nitpick: Missing a space before `(` in the `g_strdup` call. :-)

The rest looks good to me though, fwiw.

mariospr: Argh! I'll change that, thanks

Issue #134, opened 2017-12-07
OpenURI portal can crash when failing to register URI for sandboxed handler 


mariospr:
We found quite an usual case in Endless that led to a crash in the internals of GIO because the OpenURI portal would try open and URI handled by a sandboxed application (Visual Studio), which did not have access to such URI, thus failing to register the path with the document portal (https://github.com/flatpak/xdg-desktop-portal/blob/master/src/open-uri.c#L195).

And because no check is done after that, a GList of one single element, which happened to be NULL, got passed to g_app_info_launch_uris(), which ended up blowing up internally in `expand_macro()`, due to trying to reference that NULL when expanding the macro arguments from the desktop file.

There are a [couple of things that can be fixed in GLib](https://bugzilla.gnome.org/show_bug.cgi?id=791337
) for that (I already have a WIP branch, will submit it for review there later), but regardless I think the portal could use some extra checks here to prevent this situation.

mariospr:
Pull request ready for review here: https://github.com/flatpak/xdg-desktop-portal/pull/135

mariospr:
(pressed the wrong button, sorry for the closing - reopening dance)

pwithnall:
#135 has been merged, so I guess this can be closed.

PR #135, opened 2017-12-07
open-uri: Don't ever pass a list of NULLs to g_app_info_launch_uris()

mariospr:
Add extra checks to never do this when registering the URI with the
documents portal (for sandboxed handler) fails for any reason, and
report to the sandboxed caller accordingly when that happens.

https://github.com/flatpak/xdg-desktop-portal/issues/134

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -485,7 +485,9 @@ handle_open_in_thread_func (GTask *task,
           if (request->exported)
             {
               g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
-              xdp_request_emit_response (XDP_REQUEST (request), 2, g_variant_builder_end (&opts_builder));
+              xdp_request_emit_response (XDP_REQUEST (request),
+                                         XDG_DESKTOP_PORTAL_RESPONSE_OTHER,
+                                         g_variant_builder_end (&opts_builder));
               request_unexport (request);
             }
           return;
@@ -502,7 +504,9 @@ handle_open_in_thread_func (GTask *task,
           if (request->exported)
             {
               g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
-              xdp_request_emit_response (XDP_REQUEST (request), 2, g_variant_builder_end (&opts_builder));
+              xdp_request_emit_response (XDP_REQUEST (request),
+                                         XDG_DESKTOP_PORTAL_RESPONSE_OTHER,
+                                         g_variant_builder_end (&opts_builder));
               request_unexport (request);
             }
           return;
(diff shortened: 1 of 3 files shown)


src/open-uri.c
```diff
@@ -191,8 +191,16 @@ launch_application_with_uri (const char *choice_id,
...
   if (is_sandboxed (info))
     {
+      g_autoptr(GError) error = NULL;
+
       g_debug ("registering %s for %s", uri, choice_id);
-      ruri = register_document (uri, choice_id, FALSE, writable, NULL);
+      ruri = register_document (uri, choice_id, FALSE, writable, &error);
+      if (ruri == NULL)
+        {
+          g_warning ("Error registering %s for %s", uri, choice_id, error->message);
```
pwithnall: This printf has two format placeholders, but takes 3 arguments. I think a `: %s` is missing at the end of the message.

src/open-uri.c
```diff
@@ -191,8 +191,16 @@ launch_application_with_uri (const char *choice_id,
...
+          g_clear_error (&error);
```
pwithnall: Nitpick: The `g_clear_error()` is technically not necessary here, since you’ve used `g_autoptr()` above.

src/open-uri.c
```diff
@@ -522,15 +532,14 @@ handle_open_in_thread_func (GTask *task,
...
-
+      gboolean result = launch_application_with_uri (skip_app_chooser ? choices[0] : latest_id,
+                                                     uri,
+                                                     parent_window,
+                                                     writable);
       if (request->exported)
         {
           g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
-          xdp_request_emit_response (XDP_REQUEST (request), 0, g_variant_builder_end (&opts_builder));
+          xdp_request_emit_response (XDP_REQUEST (request), result ? 0 : 2, g_variant_builder_end (&opts_builder));
```
pwithnall: I assume the `2` here means ‘reject’ or similar? If so, it would be good to have this encoded in an enum (I realise that is a pre-existing problem, so don’t worry about fixing it if you don’t have time).

src/open-uri.c
mariospr: I guess this is what happens when you try to rush a PR just before going for the school pick up... Thanks, I'll fix that

src/open-uri.c
mariospr: Same thing, sloppy mistake, sorry

pwithnall: (I’d say it’s a stylistic issue rather than a mistake!)

src/open-uri.c
mariospr: Yeah, it's an existing problem. I'll try to fix it

mariospr:
@pwithnall Thanks for the review, see the new PR updated.

PR #140, opened 2017-12-17
screenshot: Add an 'interactive' option

matthiasclasen:
This option indicates to the backend that it would
be useful to offer customization before taking a screenshot.
This is just a hint, backends may not have options or may
ignore the interactive option.


mariospr:
This one looks good to me, but waiting for https://github.com/flatpak/xdg-desktop-portal-gtk/pull/106 before merging...

PR #144, opened 2018-01-23
Add DBus annotations

grulja:


src/Makefile.am.inc
```diff
@@ -55,11 +55,6 @@ $(xdp_dbus_built_sources) : $(FLATPAK_IFACE_FILES) $(PORTAL_IFACE_FILES)
 		--interface-prefix org.freedesktop.portal.      \
 		--c-namespace Xdp                               \
 		--generate-c-code $(builddir)/src/xdp-dbus      \
-		--annotate "org.freedesktop.portal.Documents.Add()" "org.gtk.GDBus.C.UnixFD" "true" \
-		--annotate "org.freedesktop.portal.Documents.AddNamed()" "org.gtk.GDBus.C.UnixFD" "true" \
```
alexlarsson: I wonder if we should keep the Add/AddNamed ones. Otherwise we need to depend on an unreleased flatpak for the updated document xml.

PR #145, opened 2018-01-23
Add remote desktop portal

jadahl:
This PR adds a remote desktop portal. By itself it only provides remote control of input devices; combined with the screen cast portal (#122), it can also provide screen sharing in the same session. The branch contains the commits from #122 because there doesn't seem to be a way to pretend the first three are separate, so please refer to #122 for those.


alexlarsson:
I had some comments in #122, but this looks good.

Issue #141, opened 2017-12-21
Return real path for files accessible from sandbox


grulja:
When opening a file using FileChooser portal, you get the file exported through Document portal and the path is something like "/run/user/1001/doc/xxxxxx/foo". While this is not problem most of the time, it becomes a problem when you want to import a project into some IDE (e.g. Qt Creator). To support this, we need to actually allow the application access to $HOME to be able import projects but problem is that FileChooser portal gets always involved. What would solve this issue is returning real path of the file in case the application has direct access to it.

KDE bugreport: https://bugs.kde.org/show_bug.cgi?id=388090

amtlib-dot-dll:
1. When $HOME is accessible, there is no need for the Document Portal
2. If so, this would be a huge security hole, obviously
3. The Document Portal should be redesigned, in my opinion. Usually we only need to open files inside `$HOME`, and one solution is to provide `/home` inside the sandbox as a `tmpfs` and bind mount certain files or directories after the user grants it, eliminating the need for `/run/user/$(id -u)/doc`. Another challenging solution is to provide `/home` as `fuse` 😂

phw:
Probably related: https://github.com/flatpak/xdg-desktop-portal/issues/95

matthiasclasen:
I think this was fixed by the merge

grulja:
It's not fixed completely, still missing to return real path when saving files, see https://github.com/flatpak/xdg-desktop-portal/pull/148.

Alexander-Wilms:
The use case "opening a file" doesn't seem to be covered by the merged PR, right? I want to open a .tex file with xdg-desktop-portal-kde, and need the real path so that lualatex can read included files and write temporary files and the pdf to the same directory, cf. https://github.com/flathub/flathub/pull/381

Running `rm /run/run/user/1000/flatpak-info` fixes this issue, cf. https://github.com/flatpak/flatpak/issues/1276

grulja:
There was a bug recently in flatpak breaking this feature, it should be fixed in the latest version.

Alexander-Wilms:
Do you mean this issue: https://github.com/flatpak/flatpak/issues/1307#issuecomment-384684499 ?

I'm running Fedora 28, which has Flatpak 0.11.7 so maybe that's a different issue?

grulja:
[...]

PR #148, opened 2018-01-24
Return real path when saving files we have access to

grulja:
Make use of new add_named_full() function where we can specify with flags whether
the document portal should return real path to file we have access to or if it
should return path to fuse mountpoint

### DIFF
--- a/src/documents.c
+++ b/src/documents.c
@@ -106,16 +106,35 @@ register_document (const char *uri,
   version = xdp_documents_get_version (documents);
 
   if (for_save)
-    ret = xdp_documents_call_add_named_sync (documents,
-                                             g_variant_new_handle (fd_in),
-                                             basename,
-                                             TRUE,
-                                             TRUE,
-                                             fd_list,
-                                             &doc_id,
-                                             NULL,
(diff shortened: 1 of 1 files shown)


grulja:
This will solve https://github.com/flatpak/xdg-desktop-portal/issues/141.

src/documents.c
```diff
@@ -106,16 +106,35 @@ register_document (const char *uri,
...
-                                             TRUE,
-                                             TRUE,
-                                             fd_list,
-                                             &doc_id,
-                                             NULL,
-                                             NULL,
-                                             error);
+    {
+      if (version >= 3)
+        {
```
matthiasclasen: I don't think there is a version 3 yet:

The flatpak code has:

  xdp_dbus_documents_set_version (XDP_DBUS_DOCUMENTS (dbus_api), 2);

grulja: There hopefully will be, see https://github.com/flatpak/flatpak/pull/1335.

grulja: There is now version 3 as the flatpak patch has been merged.

PR #152, opened 2018-02-07
Import document portal from flatpak

alexlarsson:
This imports the document portal and permissions store from flatpak.
Following this import we should remove it from flatpak master and make it possible to disable in the
0.10.x branch.


matthiasclasen:
Looks good to me

PR #153, opened 2018-02-08
screencast: Document correct vardict option

alexlarsson:
SelectSources() actually looks for the key "types", not "type".

As pointed out by @grulja


grulja:
Thanks for fixing this, I totally forgot to open a PR.

PR #154, opened 2018-02-08
Add application info abstraction

alexlarsson:
This adds an abstraction around how to get the peer application id and other app instance specific things. The goal is to put all the flatpak specific code behind this so that we can also implement a snap version.


alexlarsson:
@jhenstridge @matthiasclasen Can you take a look at this?

PR #155, opened 2018-02-08
Support snap

alexlarsson:
This adds support for snaps, based on the code in https://github.com/flatpak/xdg-desktop-portal/pull/136 by @jhenstridge. I did not actually test this, so it needs testing before it can be commited.

This PR includes the changes from https://github.com/flatpak/xdg-desktop-portal/pull/154, but we should probably land those separately first.

This should handle the basics of portals, but there are two things still missing:
 * xdp_get_app_info_from_pid() needs support to look up a snap AppInfo from the pid
 * app_has_file_access() in document-portal.c currently only supports flatpak, it would be interesting to have a version of this for snaps too.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -556,24 +556,20 @@ xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
 }
 
 static gboolean
-is_valid_initial_name_character (gint c, gboolean allow_dash)
+is_valid_name_character (gint c, gboolean allow_dash)
 {
   return
     (c >= 'A' && c <= 'Z') ||
     (c >= 'a' && c <= 'z') ||
+    (c >= '0' && c <= '9') ||
     (c == '_') || (allow_dash && c == '-');
 }
 
-static gboolean
-is_valid_name_character (gint c, gboolean allow_dash)
-{
-  return
-    is_valid_initial_name_character (c, allow_dash) ||
-    (c >= '0' && c <= '9');
-}
-
+/* This is the same as flatpak apps, except we also allow
+   names to start with digits, and two-element names so that ids of the form
+   snap.$snapname is allowed for all snap names. */
 gboolean
-xdp_is_valid_flatpak_name (const char *string)
+xdp_is_valid_app_id (const char *string)
 {
   guint len;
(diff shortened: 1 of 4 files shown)


alexlarsson:
Rebased this on master with #154 merged

src/xdp-utils.c
```diff
@@ -296,15 +299,154 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
...
+    {
+      g_autofree char *proposed = NULL;
+
+      if (strcmp (mntpt->mnt_type, "securityfs") != 0)
+        continue;
+
+      proposed = g_strdup_printf ("%s/apparmor", mntpt->mnt_dir);
+      if (stat (proposed, &statbuf) == 0)
+        {
+          apparmor_enabled = FALSE;
```
jhenstridge: I needed to change this to `TRUE` for the snap detection code to do anything.

src/xdp-utils.c
```diff
@@ -524,24 +687,20 @@ xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
...
-is_valid_name_character (gint c, gboolean allow_dash)
-{
-  return
-    is_valid_initial_name_character (c, allow_dash) ||
-    (c >= '0' && c <= '9');
-}
-
+/* This is the same as flatpak apps, except we also allow
+   names to start with digits, so that ids of the form
+   snap.pkg.$snapname is allowed for all snap names. */
```
jhenstridge: I used "snap.pkg.*" as the prefix to work around the fact that xdp_is_valid_app_id required at least three dotted components.

I can keep using that, but if we're modifying the validation rules what do you think of allowing two dotted components and shortening it to "snap.$snapname"?

alexlarsson: I'm fine with allowing two components.

src/xdp-utils.c
alexlarsson: Oops.

alexlarsson:
I just merged https://github.com/flatpak/xdg-desktop-portal/pull/163 which could probably be useful for this too.

alexlarsson:
This is just the initial work, but lets merge this for now.

PR #164, opened 2018-02-19
[WIP, don't merge yet] Use o.fd.DBus.Containers1 to identify D-Bus peers

smcv:
This branch follows on from #163 to make use of fd.o#100344 to identify D-Bus peers. If the result of `GetConnectionCredentials()` contains the *container instance ID* from fd.o#100344, then we fetch the rest of the metadata from the dbus-daemon instead of reading `/.flatpak-info`.

In principle this should work for Snap too (see also #155, #136), but it would need to be wired up in the same way as the current Flatpak support (by looking for the well-known string identifying the container type, which I assume might be `io.snapcraft` or something, and storing whatever is needed in the data structure).

I haven't yet tried [reading the container ID from the message header](https://bugs.freedesktop.org/show_bug.cgi?id=101899) via the GDBusMethodInvocation (which will need either minor GDBus enhancements, or a hack). When implemented, that will enable us to avoid calling `GetConnectionCredentials()` at all, if we have access to a `GDBusMethodInvocation` and not just a sender name.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -100,10 +100,24 @@ xdp_mkstempat (int    dir_fd,
   return -1;
 }
 
+typedef enum
+{
+  XDP_APP_INFO_KIND_UNKNOWN = 0,
+  XDP_APP_INFO_KIND_FLATPAK = 1,
+} XdpAppInfoKind;
+
 struct _XdpAppInfo {
   volatile gint ref_count;
   char *id;
-  GKeyFile *flatpak_metadata;
+  XdpAppInfoKind kind;
+
+  union
+    {
+      struct
+        {
+          GKeyFile *keyfile;
+        } flatpak;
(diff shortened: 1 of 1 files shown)


smcv:
Header field support now included. Test instructions: https://bugs.freedesktop.org/show_bug.cgi?id=101899#c18

ebassi:
This MR has been left untouched for 5 years, now; is it still relevant?

swick:
The XdpAppInfo has been completely refactored and there is a more up-to date implementation of the same concept. I think this can be closed.

swick:
The `o.fd.DBus.Containers1` concept didn't make it into dbus, it's questionable if it will, and the code is severely outdated, so I'm closing this PR. Feel free to re-open it or open a new PR when an up-to-date version of it exists.

PR #166, opened 2018-02-23
Do not assume when file doesn't exist that we don't have access to it

grulja:
In scenario when we use --filesystem=host and trying to save a file which doesn't exist yet, then e.g. "flatpak info --file-access=/home/usr/foo org.kde.kate" call will return "hidden", but that's just because this file doesn't exist and we actually have access to it. We shouldn't in this case return false immediately, but call app_has_file_access_fallback() to verify it further.


alexlarsson:
Hmm, i think maybe flatpak info should detect the file-doesn't-exist case and instead check if the app has access to the parent folder. Can you have a look at that?

PR #175, opened 2018-03-26
Add Chinese (Taiwan) translation

zerng07:
I have just done Chinese (Taiwan) translation.

Thanks!


matthiasclasen:
Can you also add zh_TW to po/LINGUAS, please ?

zerng07:
Added. Thanks!

Issue #167, opened 2018-02-27
o.fd.p.OpenURI.OpenFile method can be used to open files the sandboxed app can not read


jhenstridge:
When I was reading through the D-Bus interfaces, my assumption was that the OpenFile method took a file descriptor rather than a file name as a form of proof of access: that the portal should only run an "open" handler for files that the application can already read.

However, the implementation specifically checks to make sure the passed file descriptor was created with in `O_PATH` mode, which doesn't imply access.  For example, I can run the following as a normal user in Python without throwing an exception:

    >>> import os
    >>> fd = os.open('/etc/shadow', os.O_PATH)

The same is true for sandboxes relying on AppArmor or SELinux: you can create `O_PATH` file descriptors for files that the security profile will deny access to.

I realise that the portal will present the user with an access control dialog, so the user could still prevent the app from completing the request, but is this actually the intended behaviour?

alexlarsson:
Hmm, In flatpak, the session would be running in the same AppArmor/SELinux context, so it would have the same permissions to open a file, and the fact that you can see the pathname means you have "similar" access to it as the host.

But, I guess this is not really true for snaps, as you're running in a different security context.

alexlarsson:
So, the fix in #177 fixes at least some of this. For example if the file has different mount flags in the sandbox and the host. It may very well fix others too. That depends on the exact behaviour of selinux/AppArmour here. What we do is call access() on the fd that we got passed, which seems to apply the mount flags from the source sandbox. 

It is possible that it also applies e.g. apparmour rules too, at least if such flags are stored on the vfsmount. Needs more research.

jhenstridge:
I'll do some testing, but I don't think running access() on the `/proc/self/fd/$fd` symlink is going to tell anything about MAC restrictions inside the sandbox.

In the past, I've used an AppArmor specific API to perform access checks, but that API doesn't work correctly with the profiles snapd writes (it performs an access check for a (security_label, path) pair, while some AppArmor rules also depend on the uid).

I'm not sure of a good suggestion other than passing non-O_PATH file descriptors, and I know that sucks because of the code baked into glib/gtk.

alexlarsson:
The reason I chose to use O_PATH file descriptor is that it is easy, safe and non-blocking to verify that a file descriptor is O_PATH. This is a single F_GETFL fcntl that cannot be overridden in the kernel for weird fds (like fuse or some special device driver). 

[...]

PR #177, opened 2018-03-28
Permission fixes for sandboxed access

alexlarsson:
This fixes some issues when using the portal from a sandbox. 

First of all we were not correctly rewriting paths in all cases, so paths from the sandbox ended up with a /newroot prefix in the database which made nothing work.

Secondly, just because you can access a pathname in the sandbox (O_PATH fd) doesn't mean you have full access to it. For instance it may be mounted read-only in the sandbox (but not outside). If you can create a document with write rights then you can use this to modify the file you otherwise could not.

So, we now use access() to verify the access mode to the file and only give the corresponding write rights to the document.

The read-only mount is one aspect of https://github.com/flatpak/xdg-desktop-portal/issues/167 which is fixed by this. It may fix others too, will need looking into.


matthiasclasen:
Does this need to wait for anything on the glib side, or is this ok to merge now ?

alexlarsson:
It works fine with the current glib code, but i need to reconsider this a bit in context of #167

PR #179, opened 2018-04-04
Support non-O_PATH fds for adding existing files

alexlarsson:
As discussed in https://github.com/flatpak/xdg-desktop-portal/issues/167 using O_PATH to pass proof of access to a file is not good enough for e.g. snap, as it only means the sandbox can see the pathname, not that it can e.g. open it. So, if the fd is not O_PATH we use the open flags to detect if the file should be writable or not.

For directories we still use O_PATH because there is no way to prove writability to a directory. However, AddNamed*() is only callable from the host, so that is not really a problem.

We still support O_PATH for host and flatpak callers, because those  are safe and there are still code out there using this.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -605,12 +605,14 @@ portal_add (GDBusMethodInvocation *invocation,
               DOCUMENT_PERMISSION_FLAGS_READ;
 
             if (access_mode & W_OK)
-              perms |= DOCUMENT_PERMISSION_FLAGS_WRITE;
+              {
+                perms |= DOCUMENT_PERMISSION_FLAGS_WRITE;
 
-            /* If its a unique one its safe for the creator to
-               delete it at will */
-            if (!reuse_existing)
-              perms |= DOCUMENT_PERMISSION_FLAGS_DELETE;
+                /* If its a unique one its safe for the creator to
+                   delete it at will */
+                if (!reuse_existing)
+                  perms |= DOCUMENT_PERMISSION_FLAGS_DELETE;
+              }
 
             do_set_permissions (entry, id, app_id, perms);
           }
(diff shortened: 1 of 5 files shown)


document-portal/document-portal.c
```diff
@@ -412,8 +412,6 @@ validate_fd_common (int fd,
   if (fd == -1 ||
       /* Must be able to get fd flags */
       (fd_flags = fcntl (fd, F_GETFL)) == -1 ||
-      /* Must be O_PATH */
-      ((fd_flags & O_PATH) != O_PATH) ||
       /* Must not be O_NOFOLLOW (because we want the target file) */
       ((fd_flags & O_NOFOLLOW) == O_NOFOLLOW) ||
```
jhenstridge: The O_NOFOLLOW check is only really relevant to O_PATH descriptors, but I is probably harmless for other paths.

It looks like the O_NOFOLLOW flag is reported by fcntl(..., F_GETFL) for regular file descriptors opened that way, so the question is whether there are legitimate cases where the client would want to use the flag.

document-portal/document-portal.c
```diff
@@ -435,7 +433,63 @@ validate_fd_common (int fd,
...
+      if (xdp_app_info_supports_opath (app_info))
+        access_mode = get_fd_access (fd);
+      else
+        access_mode = -1;
+    }
+  else if (st_mode == S_IFDIR)
+    {
+      /* For directories we only support O_PATH, which is fine because
+       * they are only used by AddNamed* which is allowed from the host only.
+       */
```
jhenstridge: I'm not sure this is correct.  If I want to open a folder in Nautilus using `OpenURI.OpenFile`, then we'd have a directory descriptor here.

We'd only need to prove read access for that case, which we should be able to handle with non-O_PATH descriptors.  Given that you can't open an O_RDWR or O_WRONLY descriptor for a directory, perhaps the logic for regular files would be sufficient for directories too?

alexlarsson: OpenURI.OpenFile uses a different codepath (xdp_app_info_get_path_for_fd), so this comment doesn't affect this change. However, it is true that we want that to support non-O_PATH fds too. I'll have a look at that.

document-portal/document-portal.c
alexlarsson: The only use for it would be if you only want to export a file if it is not a symlink, which seems not very useful. But I agree, that the check is useless for non-O_PATH, so might as well not keep it.

PR #180, opened 2018-04-12
OpenURI: Fix misleading warning when accessing the permission store

mariospr:
The warning message is wrong since we're not updating the permission
store, but accessing it, at that point so we need to fix that. Besides
the warning message should only shown when the error is other than
NOT_FOUND, since not having a entry recorded in the desktop-apps table
for a particular content-type is a perfectly fine scenario.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -134,7 +134,11 @@ get_latest_choice_info (const char *app_id,
                                                    NULL,
                                                    &error))
     {
-      g_warning ("Error updating permission store: %s", error->message);
+      /* Not finding an entry for the content type in the permission store is perfectly ok */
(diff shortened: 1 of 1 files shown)


src/open-uri.c
```diff
@@ -134,7 +134,11 @@ get_latest_choice_info (const char *app_id,
                                                    NULL,
                                                    &error))
     {
-      g_warning ("Error updating permission store: %s", error->message);
+      /* Not finding an entry for the content type in the permission store if perfectly ok */
```
mcatanzaro: @mariospr if -> is

mariospr: Oops

matthiasclasen:
sure, looks good

Issue #149, opened 2018-01-28
URI opened without user confirmation


amtlib-dot-dll:
https://github.com/flatpak/xdg-desktop-portal/blob/77abb1c65a5ab6d7b206929594a99bf6ca8fd941/data/org.freedesktop.portal.OpenURI.xml#L28
This portal is said to be under the control of the user. However, when used inside the sandboxed environment to open an URI, the browser will immediately open without any user confirmation.
* If the opened site is a phishing site, the user's password may be under risk
* If the opened site have known XSS vulnerability, the user's account may be under risk
* If the browser is known to have some vulnerability and the site is malformed, the system may be under risk

TingPing:
I think the web browser was relaxed on this because it is very commonly used and a bit of a poor UX to prompt every time.

Honestly though unless the URL is extremely obviously bad just asking the user won't necessarily prevent any of those attacks mentioned and last I saw the portal prompt doesn't even show the URI (It probably should)?

amtlib-dot-dll:
Yes, but there should be a configurable switch for users who need more security than convenience. There are [examples (in Chinese)](https://zhuanlan.zhihu.com/p/32674524) on Android that use the Open URI feature to earn money 😂 of course not for the user but for the "attacker". On Win10 UWP platform there are such prompts.

TingPing:
I believe it is configurable somewhere but it wasn't an obvious place...

amtlib-dot-dll:
😂 Well, is it configurable NOW or sometime in the FUTURE? If now, I think I may dig into the conf files 😂

matthiasclasen:
mario, we should maybe document how this is written in the permission store.
Although it won't be very useful without a tool

matthiasclasen:
In fact, we currently hardcode skipping the app chooser for http:, https: and for directories.

Mario, we should consider flipping around the logic from:
 if (skip_app_chooser || (!always_ask && (latest_count >= latest_threshold)))
to 
 if (!always_ask && (skip_app_chooser || (latest_count >= latest_threshold))

And then the permission store can be used to override skip_app_chooser, by setting always_ask.

mariospr:
@matthiasclasen: I just saw this thread now, but I completely missed the comment on Jan 29, sorry about that.

I think that makes sense, I'll prepare a patch for that

mariospr:
[...]

PR #181, opened 2018-04-12
OpenURI: Don't skip the dialog based on content type if a threshold is set

mariospr:
If a threshold is set for a given content-type/schema and app ID in the
permission store, we should make the decision on whether to show the
app chooser dialog or not based only on whether such threshold has been
reached by consistently selecting the same app a big enough amount of
consecutive times, not based on whether we're opening a directory or
a HTTP/HTTPs URI (only cases when we currently might skip the dialog).

If a threshold is not set (this is the default unless the permission store
has been altered with an external tool), the current behaviour is kept
and the dialog will still be skipped for those special cases.

https://github.com/flatpak/xdg-desktop-portal/issues/149


matthiasclasen:
yes, makes sense to me. We still need tooling to set such thresholds in a somewhat user-friendly way, but this is a good first step

PR #182, opened 2018-04-23
Add some quick checklist for releases

matthiasclasen:
This is how I've done releases so far.

### DIFF
--- a/RELEASE_HOWTO.md
+++ b/RELEASE_HOWTO.md
@@ -1,0 +1,14 @@
+# Steps for doing a xdg-desktop-portal release
+
+ - git clean -fxd
+ - bump version in configure.ac
+ - add content to NEWS
+ - git commit -m <version>
+ - git push origin master
+ - ./autogen.sh --enable-docbook-docs 
+ - make all
+ - make distcheck
+ - git tag <version>
+ - git push origin refs/tags/<version>
+ - upload tarball to github as release
+ - edit release, copy NEWS section in


RELEASE_HOWTO.md
```diff
@@ -0,0 +1,14 @@
...
+ - git clean -fxd
+ - bump version in configure.ac
+ - add content to NEWS
+ - git commit -m <version>
+ - git push origin master
+ - ./autogen.sh --enable-docbook-docs 
+ - make all
+ - make distcheck
+ - git tag <version>
+ - git push origin refs/tags/<version>
```
mariospr: I think I'd do `git push <version>`, which should work fine too (unless there's a branch with the same name).

But another option (and this is how I make releases on Endless) that I kind of like a bit more is **atomically** pushing the branch and the tags at the same time: don't do the `git push origin master` before, and simply do here `git push origin master <version>`

RELEASE_HOWTO.md
```diff
@@ -0,0 +1,14 @@
...
+
+ - git clean -fxd
+ - bump version in configure.ac
+ - add content to NEWS
+ - git commit -m <version>
+ - git push origin master
+ - ./autogen.sh --enable-docbook-docs 
+ - make all
+ - make distcheck
+ - git tag <version>
```
mariospr: @matthiasclasen What do you thinkg of signing the tags with  your GPG key using `gpg -a -s <version>`? No strong feelings, but I think it's a nice thing to do when possible.

RELEASE_HOWTO.md
```diff
@@ -0,0 +1,14 @@
...
+ - git push origin refs/tags/<version>
+ - upload tarball to github as release
+ - edit release, copy NEWS section in
```
mariospr: truncated line?

mariospr:
Looks good, but I have a couple of small suggestions + a genuine question about what seems to be a truncated line.

RELEASE_HOWTO.md
matthiasclasen: Feel free to. I don't do gpg

RELEASE_HOWTO.md
matthiasclasen: No, I wrote it like this :-(
I meant: copy the NEWS section for the current release into the github release description and fix up the markup

RELEASE_HOWTO.md
mariospr: Ok, still good to know you're not opposed to that

PR #184, opened 2018-04-27
Include the flatpak portal docs

matthiasclasen:
This is a little different, since the flatpak portal actually lives in the flatpak repository, so we include it here from its installed location in /usr/share/dbus-1/interfaces/

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -1,6 +1,10 @@
 NULL =
 
+org.freedesktop.portal.Flatpak.xml :
+	cp $(datadir)/dbus-1/interfaces/org.freedesktop.portal.Flatpak.xml .
+
 portal_files = 								\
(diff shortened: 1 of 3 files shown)


doc/Makefile.am
```diff
@@ -1,6 +1,7 @@
 NULL =
 
 portal_files = 								\
+	$(datadir)/dbus-1/interfaces/org.freedesktop.portal.Flatpak.xml	\
```
mariospr: I think this will cause `make distcheck` fail, since that file won't be part of the install directory used during the `distcheck`process, which will be in `$(top_srcdir)/xdg-desktop-portal-0.10/_inst`.

Maybe you could add a rule to the Makefile to somehow copy it from `$(datadir)/dbus-1/interfaces/` and add it to `EXTRA_DIST` when creating the tarball, and then you could reference it from `$(top_srcdir)` like the other ones?

PR #198, opened 2018-07-20
Network lockdown for proxy and network monitor portals

matthiasclasen:
Make the proxy portal error out if the app does not have network access.

Do the same for the network monitor portal. To do so, we have to change the api away from properties
and to explicit getters. The corresponding glib changes are here: https://gitlab.gnome.org/GNOME/glib/commits/new-network-portal


mariospr:
Looks good to me. My only concern is that this will break the API and perhaps it could be interesting to keep the properties around for a while as "deprecated", to give callers some room to adapt. Or is GLib the only direct user of this API in practice? (thus the only one needing updating).

matthiasclasen:
As far as I know, glib is the only direct consumer.

The problem with keepng the properties around is that it defeats the purpose of this change, since it continues to leak network information into non-networked sandboxes.

I'd rather break this api now, and have the proper api in 1.0

piotrdrag:
I suppose src/.proxy-resolver.c.swp was added by mistake?

PR #209, opened 2018-08-14
Unlink obsolete doc files (fixes #195)

### LINKED ISSUE #195: document-portal: Creating writeable file a second time fails
```sh
APP=...
doc="$(flatpak document-export --app="$APP" --noexist --allow-write file)"
flatpak run --command=touch "$APP" "$doc"  # works
rm file
flatpak run --command=touch "$APP" "$doc"  # fails
```

Saving a file with the FileChooser portal, moving or deleting the file and then saving another file with the same path fails.

Unrud:
Unlink doc files that are either deleted or point to a different inode on the host file system.

Files deleted on the host file system couldn't be recreated.
Old inodes were reused, when a file got reopened that points to a different inode on the host file system.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -937,6 +937,46 @@ xdp_inode_stat (XdpInode    *inode,
   return 0;
 }
 
+static int
+xdp_inode_unlink_obsolete(XdpInode *parent_inode,
+                          const char *name) {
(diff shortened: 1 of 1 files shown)


alexlarsson:
This is not right. We already check for obsoleted files in lookup(). However, due to a XdpInode leak this was not working. The proper fix is in https://github.com/flatpak/xdg-desktop-portal/pull/216

PR #210, opened 2018-08-20
network monitor: Add GetStatus and CanReach methods

matthiasclasen:
GetStatus lets us avoid multiple roundtrips to get all the values.
CanReach lets us implement g_network_monitor_can_reach().


alexlarsson:
Can the GetStatus method return an a{sv} instead? Then we can easily extend it.

matthiasclasen:
They could, if you think its important.

PR #211, opened 2018-08-26
Trash

matthiasclasen:
Here is a minimal trash implementation. I looked briefly at adding a way to restore files, but we don't load gvfs in the portal frontend, so we can't operate on trash:///. Moving the trash implementation to the backend would be an option, but we currently don't load gvfs there either...


alexlarsson:
I would merge the commits, but otherwise this looks good to me.

wjt:
What's the argument for having this return `0` or `1` rather than nothing on success or a D-Bus error on failure?

PR #219, opened 2018-08-31
network-monitor: Build variants for GetStatus() correctly

smcv:
Passing a boolean on the stack where a GVariant * is expected is just
going to lead to a segfault.

Signed-off-by: Simon McVittie <smcv@collabora.com>


matthiasclasen:
thanks, I used florians patch, before discovering that yours was better :)

PR #220, opened 2018-08-31
network-monitor: Fix variant type

fmuellner:
Vardict entries are of type {sv}, so wrap the actual value in a variant.


matthiasclasen:
man, can't get this thing right :(

fmuellner:
> man, can't get this thing right :(

At least I'm not the only one who gets this wrong :)

PR #222, opened 2018-09-05
File transfer

matthiasclasen:
This branch adds a "FileTransfer" portal to the document portal binary. It can be used to implement dnd and copy/paste. By using the portal as a side-channel to transfer the files (while exporting them as-needed).

This works, and https://gitlab.gnome.org/GNOME/gtk/merge_requests/321 has corresponding gtk changes. Future todo: better handling for directories.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -18,6 +18,8 @@
 #include "permission-db.h"
 #include "permission-store-dbus.h"
 #include "document-portal-fuse.h"
+#include "file-transfer.h"
+#include "document-portal.h"
 
 #include <sys/eventfd.h>
 
@@ -430,85 +432,43 @@ portal_add (GDBusMethodInvocation *invocation,
             GVariant              *parameters,
             XdpAppInfo            *app_info)
 {
+  int fd_id;
+  gboolean reuse_existing, persistent;
+  DocumentAddFullFlags flags = 0;
   GDBusMessage *message;
   GUnixFDList *fd_list;
-  g_autofree char *id = NULL;
-  int fd_id, fd, fds_len;
-  g_autofree char *path = NULL;
-  gboolean writable;
-  const int *fds;
-  struct stat st_buf, real_parent_st_buf;
-  gboolean reuse_existing, persistent;
   GError *error = NULL;
-  const char *app_id = xdp_app_info_get_id (app_info);
+  g_auto(GStrv) ids = NULL;
 
   g_variant_get (parameters, "(hbb)", &fd_id, &reuse_existing, &persistent);
 
+  if (reuse_existing)
+    flags |= DOCUMENT_ADD_FLAGS_REUSE_EXISTING;
+  if (persistent)
+    flags |= DOCUMENT_ADD_FLAGS_PERSISTENT;
+ 
   message = g_dbus_method_invocation_get_message (invocation);
   fd_list = g_dbus_message_get_unix_fd_list (message);
 
-  fd = -1;
   if (fd_list != NULL)
     {
-      fds = g_unix_fd_list_peek_fds (fd_list, &fds_len);
+      int fds_len;
+      const int *fds = g_unix_fd_list_peek_fds (fd_list, &fds_len);
       if (fd_id < fds_len)
-        fd = fds[fd_id];
+        {
+          int fd = fds[fd_id];
+
+          ids = document_add_full (&fd, 1, flags, app_info, "", 0, &error);
+        }
     }
 
-  if (!validate_fd (fd, app_info, &st_buf, &real_parent_st_buf, &path, &writable, &error))
+  if (ids == NULL)
     {
       g_dbus_method_invocation_take_error (invocation, error);
       return;
     }
 
-  if (st_buf.st_dev == fuse_dev)
-    {
-      /* The passed in fd is on the fuse filesystem itself */
-      id = verify_existing_document (&st_buf, reuse_existing);
-      if (id == NULL)
-        {
-          g_dbus_method_invocation_return_error (invocation,
-                                                 XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                                                 "Invalid fd passed");
-          return;
-        }
-    }
-  else
-    {
-      {
-        XDP_AUTOLOCK (db);
-
-        id = do_create_doc (&real_parent_st_buf, path, reuse_existing, persistent);
-
-        if (app_id[0] != '\0')
-          {
-            g_autoptr(PermissionDbEntry) entry = permission_db_lookup (db, id);
-            DocumentPermissionFlags perms =
-              DOCUMENT_PERMISSION_FLAGS_GRANT_PERMISSIONS |
-              DOCUMENT_PERMISSION_FLAGS_READ;
-
-            if (writable)
-              {
-                perms |= DOCUMENT_PERMISSION_FLAGS_WRITE;
-
-                /* If its a unique one its safe for the creator to
-                   delete it at will */
-                if (!reuse_existing)
-                  perms |= DOCUMENT_PERMISSION_FLAGS_DELETE;
-              }
-
-            do_set_permissions (entry, id, app_id, perms);
-          }
-      }
-
-      /* Invalidate with lock dropped to avoid deadlock */
-      xdp_fuse_invalidate_doc_app (id, NULL);
-      if (app_id[0] != '\0')
-        xdp_fuse_invalidate_doc_app (id, app_id);
-    }
-
-  g_dbus_method_invocation_return_value (invocation,
-                                         g_variant_new ("(s)", id));
+  g_dbus_method_invocation_return_value (invocation, g_variant_new ("(s)", ids[0]));
 }
 
 static char *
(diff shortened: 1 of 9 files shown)


document-portal/document-portal.c
```diff
@@ -431,85 +433,43 @@ portal_add (GDBusMethodInvocation *invocation,
...
     {
-      fds = g_unix_fd_list_peek_fds (fd_list, &fds_len);
+      int fds_len;
+      const int *fds = g_unix_fd_list_peek_fds (fd_list, &fds_len);
       if (fd_id < fds_len)
-        fd = fds[fd_id];
+        {
+          int fd = fds[fd_id];
+
+          ids = document_add_full (&fd, 1, flags, app_info, "", 0, &error);
```
alexlarsson: This seems to change the behaviour wrt write access. Before it gave the app write access iff the passed in fd was writable. However, now it never gets it (because we pass 0 as target_perms).

alexlarsson: Ah, no because that is target perms, not caller perms.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,447 @@
...
+G_LOCK_DEFINE (transfers);
+static GHashTable *transfers;
+
+static FileTransfer *
+lookup_transfer (const char *key)
+{
+  FileTransfer *transfer;
+
+  G_LOCK (transfers);
+  transfer = (FileTransfer *)g_hash_table_lookup (transfers, key);
```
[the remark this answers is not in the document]
alexlarsson: Maybe it should return a pre-locked transfer?

document-portal/file-transfer.c
```diff
@@ -0,0 +1,447 @@
...
+
+  transfer = g_new0 (FileTransfer, 1);
+
+  g_mutex_init (&transfer->mutex);
+
+  transfer->app_info = xdp_app_info_ref (app_info);
+  transfer->sender = g_strdup (sender);
+  transfer->writable = writable;
+  transfer->autostop = autostop;
+  transfer->key = g_strdup_printf ("%u", g_random_int ());
```
alexlarsson: Surely you need a loop here to make sure the key isn't already used?

document-portal/file-transfer.c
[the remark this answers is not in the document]
alexlarsson: Actually, that risks creating deadlocks around the lock for the hash table, so we probably need a refcount type thing.

document-portal/document-portal.c
```diff
@@ -1297,6 +1300,15 @@ on_bus_acquired (GDBusConnection *connection,
       g_warning ("error: %s", error->message);
       g_error_free (error);
     }
+
```
alexlarsson: Don't we need a:
```
  g_dbus_interface_skeleton_set_flags (file_transfer,
                                       G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
```
Here? Otherwise all the file transfer calls will happen on the main thread, and the locks they take can block it.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,447 @@
...
+      int fd = -1;
+      g_autofree char *path = NULL;
+      gboolean fd_is_writable;
+
+      if (fd_id < n_fds)
+        fd = fds[fd_id];
+
+      if (fd == -1)
+        {
+          g_dbus_method_invocation_return_gerror (invocation, error);
```
alexlarsson: error is not set here, so you can't return that.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,447 @@
...
+  target_app_id = xdp_app_info_get_id (target_app_info);
+
+  n_fds = transfer->files->len;
+  fds = g_new (int, n_fds);
+  for (i = 0; i < n_fds; i++)
+    {
+      const char *file;
+
+      file = g_ptr_array_index (transfer->files, i);
+      fds[i] = open (file, O_PATH | O_CLOEXEC);
```
alexlarsson: While this works, its kinda weird to have the portal open the paths (which we already verified the calling app could access) and pass the FDs into document_add_full(). Maybe we could have a version of document_add that takes paths?

matthiasclasen: but I think we need fds in document_add_full anyway, to check if the target can access them ?

document-portal/file-transfer.c
```diff
@@ -0,0 +1,447 @@
...
+      path = xdp_app_info_get_path_for_fd (app_info, fd, 0, NULL, &fd_is_writable);
+      if (path == NULL || (transfer->writable && !fd_is_writable))
+        {
+          g_dbus_method_invocation_return_error (invocation,
+                                                 XDG_DESKTOP_PORTAL_ERROR,
+                                                 XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
+                                                 "Can't export file");
+          return;
+        }
+
```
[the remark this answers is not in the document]
alexlarsson: validate_fd() calls xdp_app_info_get_path_for_fd() which yes, does some validation, but then it does one extra check, on the parent directory, to make sure that the actual child name from the parent directory finds the same file.

Assume the app has rw access to /foo. Then there is an attack:
* create /foo/bar/passwd
* fd = open (/foo/bar/passwd)
* transfer.add (fd)
* portal gets the path for /foo/bar/passwd and adds it to the list of paths in transfer->files
* mv /foo/bar /foo/old-bar
* ln -s /etc /foo/bar
* transfer.execute()
* portal now opens /foo/bar/passwd, which now points to /etc/passwd, and makes a document for this
* app reads /etc/passwd

the main portal add_files() avoids this, because at the time of addition it takes the parent directory inode and verifies that the child is actually in that, then it stores the parent inode and requires all later ops to be in the same directory. However, in this new case there is a delay between these two steps, which allows this attack.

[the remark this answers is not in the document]
matthiasclasen: I see. Will fix that up, then

document-portal/document-portal.c
```diff
@@ -718,38 +665,86 @@ portal_add_full (GDBusMethodInvocation *invocation,
...
+        }
 
-  for (i = 0; i < n_args; i++)
+      ids = document_add_full (fd, n_args, flags, app_info, target_app_id, target_perms, &error);
+    } 
+
+  if (ids == NULL)
     {
-      g_autofree char *path = NULL;
+      g_dbus_method_invocation_take_error (invocation, error);
```
alexlarsson: This seems to return a NULL error if ids == NULL due to fd_list being NULL

document-portal/file-transfer.c
```diff
@@ -0,0 +1,544 @@
...
+  transfer->app_info = xdp_app_info_ref (app_info);
+  transfer->sender = g_strdup (sender);
+  transfer->writable = writable;
+  transfer->autostop = autostop;
+  transfer->files = g_ptr_array_new_with_free_func (exported_file_free);
+
+  G_LOCK (transfers);
+  do {
+    g_free (transfer->key);
+    transfer->key = g_strdup_printf ("%u", g_random_int ());
```
alexlarsson: I don't think a 32bit random value is enough to protect against a brute force attack. Calling retrieve_files() 4 billion times seem vaguely realistic, while bumping it to 64bit would make it unthinkable.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,544 @@
...
+          for (j = 0; j < i; j++)
+            close (fds[j]);
+          g_set_error (error,
+                       XDG_DESKTOP_PORTAL_ERROR,
+                       XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
+                       "Can't export file");
+          return NULL;
+        }
+
+      fds[i] = open (file->path, O_PATH | O_CLOEXEC);
```
alexlarsson: This is still racy. The same kind of attack can happen between the fstat() and the time when document_add_full does a stat on the parent dir, because *those* are the parent ino+dev that will be stored in the document portal. To fix the race i think you need to pass  file->parent_dev/ino into document_add_full and ensure that those match exactly what is being stored in the document db.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,544 @@
...
+        {
+          g_dbus_method_invocation_return_error (invocation,
+                                                 XDG_DESKTOP_PORTAL_ERROR,
+                                                 XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
+                                                 "Can't export file");
+          return;
+        }
+
+      dirname = g_path_get_dirname (path);
+      dir_fd = open (dirname, O_CLOEXEC | O_PATH);
```
alexlarsson: Additionally, there is a tiny race here too, where the app could do the attach between getting the path for the passed in fd, and the open+fstate of the parent dir to find the parent inode. In document-portal.c::validate_fd() this is done by opening the parent, verifying that you can fstatat(basename) from the dir fd and find the same inode as the fstat() on the fd that the app passed.

In fact, if you just reuse a call to that here you can drop a bunch of code and fix this race in one go.

alexlarsson:
We could add an "gboolean allow_dir"  to validate_fds?

In fact, shouldn't we similarly be able to handle a directory fd in the regular document creation case with AS_NEEDED_BY_APP in case the app already has access to the directory?

matthiasclasen:
One thing I found working on a test for this: we are currently not cleaning up pending transfers when the creator drops off the bus. I guess we should

document-portal/file-transfer.c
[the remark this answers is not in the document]
matthiasclasen: We now record the parent dev and ino at the time the file is added, so we can prevent this attack.

document-portal/file-transfer.c
```diff
@@ -0,0 +1,552 @@
...
+
+  n_fds = transfer->files->len;
+  fds = g_new (int, n_fds);
+  parent_devs = g_new (int, n_fds);
+  parent_inos = g_new (int, n_fds);
+  for (i = 0; i < n_fds; i++)
+    {
+      ExportedFile *file = (ExportedFile*)g_ptr_array_index (transfer->files, i);
+
+      fds[i] = open (file->path, O_PATH | O_CLOEXEC);
```
matthiasclasen: Could we run out of fds here? Should probably check

PR #224, opened 2018-09-20
Add option to specify app name in screen cast and remote desktop portals

grulja:
Applications using screen cast and remote desktop might not be running in sandbox, which makes
them not able to get application name using flatpak-info file


matthiasclasen:
But this issue (no app name outside the sandbox) is not particular to the the RemoteDesktop portal at all, it affects all portals that want to mention the caller in their UI. If we come up with a better solution that saying 'an application', it should ideally work for all portals.

grulja:
I know this is problem for all portals, but only screen cast and remote desktop are used by non-sandboxed apps. I can of course add this to all of them.

I understand, but I have no idea how to make this more secure. Still I think this doesn't make difference if we trust the name passed by the application or if we just show "some application wants to get access to ...".

matthiasclasen:
I just wonder if we can do some heuristics portal-side, without having to ask the app to self-report - like, doing the same sort of things gnome-shell does to find desktop files ?

grulja:
Do you have any idea where in gnome-shell such heuristic can be found so I can get inspiration?

grulja:
I only found mapping based on windows, not sure same approach can be used from xdg-desktop-portal.

alexlarsson:
Hmm, maybe we can call org.freedesktop.DBus.ListNames() and see if the calling unique dbus id owns some name on the bus. Then we can use that as the default?

Obviously this can break in various ways and is problematic if you own multiple or no names. 
In general though, there is no reliable way to do this.

Another approach is to add a function to org.freedesktop.Application to report the desktop id.

alexlarsson:
Another way is to look in proc for the executable name and try to reverse map to a desktop file Exec= line. That seems even less robust though.

hadess:
Why wouldn't this be good enough?
https://github.com/alexlarsson/xdg-document-portal/blob/master/xdp-util.c#L33

The app name matches the named app that's inside the container.

alexlarsson:
@hadess we already have a way to detect the name for flatpaked apps, this is about host-side ones.

alexlarsson:
So, another way is to look for GIO_LAUNCHED_DESKTOP_FILE in /proc/$pid/environ. Although that is kinda glib specific, but maybe it can be made a generic thing that other desktops set too.

hadess:
The easy is to change the API so that out-of-the-sandbox apps provide that information. Which is no security at all, but that wouldn't be any different from any other files.

Or you wouldn't use the app name at all, because you'd just allow any app outside the sandbox to access the feature (which would be similar to what recent changes to geoclue do).

alexlarsson:
@hadess yes, we're aware its no security at all. But, if you trust an app (run it outside sandbox) then you trust it to provide correct info, and having the name available for presenting in dialogs is nice.

hadess:
> @hadess yes, we're aware its no security at all. But, if you trust an app (run it outside sandbox) then you trust it to provide correct info, and having the name available for presenting in dialogs is nice.

Then change the API to have the app provide it?

alexlarsson:
@hadess yes, that is what this PR currently does. We're just discussing alternative options.

matthiasclasen:
So, looking at this some more, we can't rely on GIO_LAUNCHED_DESKTOP_FILE - it is only set for half the launch paths in gio. The only possible avenues seem to be:

pid -> exe -> match Exec line in desktop file
unique name -> well-known name -> find desktop file by name that matches with DBusActivatable=true

I don't really want to write this code, it seems just as horrible as the self-reporting proposal in this PR.

I would suggest that we just scrap this and live with the fact that non-sandboxed applications will show up as "an application" in portal dialogs. We can make it even more explicit and say "a host application" or "an unsandboxed application".

If people complain, the right answer is: make your app run in a sandbox.

grulja:
I don't see any disadvantage in having apps to report their name. Answering to people that apps should run in a sandbox is also not a good approach. For example, I don't think we will soon have Chrome/Chromium packaged for flatpak and I think that screen sharing in browsers is pretty big usecase. 

It's up to you whether you want to merge this, I just thought it would be nice to let users know which unsandboxed application is requesting screen sharing.

hadess:
> it will not work transparently for portal apis that are hidden behind existing library apis, since there is no good way for those libraries to know what to report

Agreed with the rest, but in this case, wouldn't you want/expect to get a reverse-domain name from g_get_prgname(), or is that too much to expect as well?

But this particular problem shows the difficulty of minimising the consumed API, avoiding having 2 separate APIs (one sandboxed, one not) for applications to use, especially when the API doesn't have a library of its own.

Maybe, in this case, an API entry point which provides the application name should be available, but only to unsandboxed applications, with the sandboxed version not having it?

matthiasclasen:
> Maybe, in this case, an API entry point which provides the application name should be available, but only to unsandboxed applications, with the sandboxed version not having it?

From the api perspective, making it an option as this patch does is unobtrusive enough - libraries can just ignore it. But it is unfortunate that we will get a mix of portals that have the name (when the unsandboxed app calls the dbus api directly and remembers to pass a name) and that don't (when the app forgets or uses the portal indirectly, via a library).

dinhokusanagi:
Captura de tela de 2020-04-12 17-24-15
Captura de tela de 2020-05-05 14-28-00

Issue #225, opened 2018-09-20
file chooser: clarify pattern filters


matthiasclasen:
We've hit a case where gtk and kde backends seem to interpret file filters differently:

https://gitlab.gnome.org/GNOME/gtk/issues/1347

We should a) clarify the docs around filters for expected behavior, and b) figure out the best way to make this work with existing apps.

PR #226, opened 2018-09-20
Clarify the docs on filters in the file chooser

matthiasclasen:
Make it explicit that patterns are case-sensitive.

Closes https://github.com/flatpak/xdg-desktop-portal/issues/225

### DIFF
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -80,11 +80,18 @@
             <para>
               Each item in the array specifies a single filter to offer to the user.
               The first string is a user-visible name for the filter. The a(us)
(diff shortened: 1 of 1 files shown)


mcatanzaro:
LGTM

matthiasclasen:
We were discussing changing gtk to be case insensitive, but no change has been merged yet, which is why this is still open

mcatanzaro:
For the record, that's https://gitlab.gnome.org/GNOME/gtk/merge_requests/359

hadess:
I'm guessing the plan would be to offer a way to make matches case insensitive?

matthiasclasen:
We've  added GtkFileFilter api to create case-insensitive filters, using suffix matches. Those will be translated into case-insensitive patterns for the native (portal) filechooser.

Issue #230, opened 2018-10-03
Gio File Copy: g-io-error-quark, Software caused connection abort


FalkAlexander:
Moved from https://github.com/flatpak/flatpak/issues/2177

## Linux distribution and version
Fedora 28/29, Ubuntu 18.04, Arch Linux
## Flatpak version
1.0.2
## Description of the problem
Using the copy function on a Gio File object for copying into another directory throws
`g-io-error-quark: Error opening file “/run/user/1000/doc/291bd18e/Safe.kdbx”: Software caused connection abort (0)`
https://lazka.github.io/pgi-docs/#Gio-2.0/classes/File.html#Gio.File.copy

Independent of the filesystem permission, tried without, with home and host.
## Steps to reproduce
1. Optionally get my app at https://gitlab.gnome.org/World/PasswordSafe
2. Start Password Safe
3. Click on "New..." for creating a new database
4. Select a directory for the database with GtkFileChooserNative
5. Click Save
6. Close the app
7. Look into the directory, there is a database file which size is > 0kb
8. Delete the database file
9. Reopen the app and recreate the database with the same filename
10. Gio throws g-io-error-quark and the database file size is 0 kb

Does not necessarily have to be the identical folder, sometimes this happens without previous deletion of the database file. Couldn't find an exact schema.

This does only happen with the Flatpak package.

I've discussed already with @bilelmoussaoui about this, maybe he has some information he can add here also.
@bertob experienced this behaviour also while using Password Safe.

bilelmoussaoui:
I can confirm this issue when I was trying to use Gtk.NativeFileChooser on my application. I had to switch to a Gtk.FileChooser instead & added `--filesystem=home` in order to fix it. 
I can provide a test env if you want :+1:

matthiasclasen:
Thats not a 'fix', thats just disabling the sandboxing :(

bilelmoussaoui:
@matthiasclasen Never said it was a fix. I just switched to that before reporting the issue! once it's fixed I will switch back to the NativeFileChooser

BrainBlasted:
I'm trying to write an export function for an app, using FileChooserNative. `--filesystem=home` did not work:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 103, kind: ConnectionAborted, message: "Software caused connection abort" }', libcore/result.rs:945:5
```

alexlarsson:
[...]

PR #232, opened 2018-10-09
document portal tests and various fixes to make them complete

alexlarsson:
This fixes:
https://github.com/flatpak/xdg-desktop-portal/issues/230
But also some other stuff that I found when adding the tests to find the issues.

Additionally it adds some docs on the ownership of XdpInodes.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -617,6 +617,9 @@ xdp_inode_rename_child (XdpInode   *dir,
             }
 
           src_inode->is_doc = TRUE;
+          /* Drop keep-alive-until-unlink ref, since we're now is_doc */
+          xdp_inode_unref (src_inode);
+
           g_free (src_inode->filename);
           src_inode->filename = g_strdup (dst_filename);
           g_free (src_inode->backing_filename);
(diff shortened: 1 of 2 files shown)


tests/test-doc-portal.c
```diff
@@ -455,14 +455,78 @@ global_setup (void)
   g_assert (mountpoint != NULL);
 }
 
+static gboolean
+rm_rf_dir (GFile         *dir,
```
matthiasclasen: Alternatively, we could bring in libglnx and use glnx_shutil_rm_rf_at

alexlarsson: Lets do that when we use it for more than this?

document-portal/document-portal-fuse.c
```diff
@@ -1753,16 +1753,8 @@ xdp_fuse_create (fuse_req_t             req,
         }
```
matthiasclasen: I think the commit message was meant to say "...are not really *under* our control..."

tests/test-doc-portal.c
matthiasclasen: ok

document-portal/document-portal-fuse.c
alexlarsson: fixed

PR #245, opened 2018-11-02
Add Settings portal

TingPing:
This is a minimal portal that effectively just replaces XSettings on Wayland without requring DConf access.

@grulja @aleixpol - I'd like your input on if you can use this to replace your `--filesystem=xdg-config/kdeglobals:ro` permission also.

### DIFF
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2018 Igalia S.L.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Patrick Griffis <pgriffis@igalia.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+    org.freedesktop.portal.Settings:
+    @short_description: Settings interface
+
+    This interface provides read-only access to a small number
+    of host settings required for toolkits similar to XSettings.
+    It is not for general purpose settings.
+
+    Currently the valid keys are entirely implementation details
+    that are undocumented. If you are a toolkit and want to use
+    this please open an issue.
+  -->
+  <interface name="org.freedesktop.portal.Settings">
+
+    <!--
+      ReadAll:
+      @namespace: Namespace to filter results by, supports simple globbing explained below.
+      @value: Dictionary of namespaces to its keys and values.
+
+      If @namespace is an empty string matches all. Globbing is supported but only for
+      trailing sections, e.g. "org.example.*".
+    -->
+    <method name='ReadAll'>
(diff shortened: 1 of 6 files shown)


grulja:
Yes, something like this can be used, but your implementation seems to be a bit different from other portals. I would expect you to have org.freedesktop.portal.impl.Settings interface where backends would implement the logic, you have all in one place, which doesn't allow us to actually use this at all.

grulja:
As discussed on irc, I now understand how this is supposed to work and will look into it ASAP.

data/org.freedesktop.portal.Settings.xml
```diff
@@ -0,0 +1,64 @@
...
+    that are undocumented and how they are stored on the host is
+    undefined. If you are a toolkit and want to use this please
+    open an issue.
+  -->
+  <interface name="org.freedesktop.portal.Settings">
+
+    <!--
+        ReadAll:
+        @namespace: Namespace to filter results by, supports simple globbing explained below.
+
```
matthiasclasen: Should also document the out arguments here

data/org.freedesktop.portal.Settings.xml
```diff
@@ -0,0 +1,64 @@
...
+      <arg name='namespace' type='s'/>
+      <arg name='key' type='s'/>
+      <arg name='value' direction='out' type='v'/>
+    </method>
+
+    <signal name='SettingChanged'>
+      <arg name='namespace' direction='out' type='s'/>
+      <arg name='key' direction='out' type='s'/>
+      <arg name='value' direction='out' type='v'/>
+    </signal>
```
matthiasclasen: Can you document Read and SettingsChanged as well ?

data/org.freedesktop.portal.Settings.xml
```diff
@@ -0,0 +1,64 @@
...
+  </interface>
```
matthiasclasen: Please add a "version" property like all the other portal interface have.

src/settings.c
```diff
@@ -0,0 +1,273 @@
...
+{
+  Settings *self = (Settings *)object;
+
+  g_debug ("Read %s %s", arg_namespace, arg_key);
+
+  // TODO: Handle kdeglobals via same interface
+  if (!g_hash_table_contains (self->settings, arg_namespace))
+    {
+      g_debug ("Attempted to read from unknown namespace %s", arg_namespace);
+      g_dbus_method_invocation_return_error_literal (invocation, G_IO_ERROR, G_IO_ERROR_FAILED,
```
matthiasclasen: I think we should be using FLATPAL_PORTAL_ERROR here

data/org.freedesktop.portal.Settings.xml
```diff
@@ -0,0 +1,64 @@
...
+
+    <!--
+        ReadAll:
+        @namespace: Namespace to filter results by, supports simple globbing explained below.
+
+        Returns a dictionary of each namespace to all of their keys and values in everything
+        matching @namespace.
+
+        If @namespace is an empty string matches all. Globbing is supported but only for
+        trailing sections, e.g. "org.example.*".
```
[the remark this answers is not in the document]
TingPing: I thought it was fine to not be overly strict about it but I just imagined nobody would actually use it in any way other than my example. `'*'` does work to match all which I thought about documenting though.

src/settings.c
TingPing: `XDG_DESKTOP_PORTAL_ERROR` you probably mean? Done.

src/settings.c
```diff
@@ -0,0 +1,278 @@
...
+  init_settings_table (self, self->settings);
+
+  xdp_settings_set_version (XDP_SETTINGS (self), 1);
+}
+
+static void
+settings_finalize (GObject *object)
+{
+  Settings *self = (Settings*)object;
+  g_hash_table_destroy (self->settings);
```
matthiasclasen: finalize needs to chain up

data/org.freedesktop.portal.Settings.xml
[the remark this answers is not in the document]
matthiasclasen: its fine as-is

PR #248, opened 2018-11-04
settings: Support the fontconfig timestamp

matthiasclasen:
This is not a gsetting, so we treat it separately.
To make things easy for clients, we put it in a
fake org.gnome.fontconfig namespace.

The fontconfig monitor code here is copied from
gnome-settings-daemon.

### DIFF
--- a/src/fc-monitor.c
+++ b/src/fc-monitor.c
@@ -19,6 +19,8 @@
  * Author:  Behdad Esfahbod, Red Hat, Inc.
  */
 
+/* NOTE: This file is copied from gnome-settings-daemon, please keep it in sync */
+
 #include "fc-monitor.h"
 
 #include <gio/gio.h>
--- a/src/fc-monitor.c
+++ b/src/fc-monitor.c
@@ -1,0 +1,321 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2008 Red Hat, Inc.
+ * Copyright (C) 2017 Jan Alexander Steffens (heftig) <jan.steffens@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author:  Behdad Esfahbod, Red Hat, Inc.
+ */
+
+#include "fc-monitor.h"
+
+#include <gio/gio.h>
+#include <fontconfig/fontconfig.h>
+
+#define TIMEOUT_MILLISECONDS 1000
+
+static void
+fontconfig_cache_update_thread (GTask *task,
+                                gpointer source_object G_GNUC_UNUSED,
+                                gpointer task_data G_GNUC_UNUSED,
+                                GCancellable *cancellable G_GNUC_UNUSED)
+{
+        if (FcConfigUptoDate (NULL)) {
+                g_task_return_boolean (task, FALSE);
+                return;
+        }
+
+        if (!FcInitReinitialize ()) {
+                g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                         "FcInitReinitialize failed");
+                return;
+        }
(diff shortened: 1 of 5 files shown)


src/settings.c
```diff
@@ -138,31 +153,32 @@ settings_handle_read (XdpSettings           *object,
...
-      else
-        {
-          g_debug ("Attempted to read unknown key %s", arg_key);
-          g_dbus_method_invocation_return_error_literal (invocation, XDG_DESKTOP_PORTAL_ERROR,
-                                                         XDG_DESKTOP_PORTAL_ERROR_NOT_FOUND,
-                                                         _("Requested setting not found"));
+          g_dbus_method_invocation_return_value (invocation, g_variant_new ("(v)", variant));
+          return TRUE;
         }
     }
```
matthiasclasen: I've turned around the logic here to use early returns for successful cases, and have a catch-all error case at the end.

src/settings.c
```diff
@@ -238,6 +254,22 @@ init_settings_table (Settings   *self,
...
+fontconfig_changed (FcMonitor *monitor,
+                    Settings *self)
+{
+  const char *namespace = "org.gnome.fontconfig";
+  const char *key = "timestamp";
+  
+  g_debug ("Emitting changed for %s %s", namespace, key);
+
+  self->fontconfig_timestamp = time (NULL);
+
```
matthiasclasen: This is the same that gsd does: use use time()

src/settings.c
```diff
@@ -138,31 +153,32 @@ settings_handle_read (XdpSettings           *object,
...
-      g_debug ("Attempted to read from unknown namespace %s", arg_namespace);
-      g_dbus_method_invocation_return_error_literal (invocation, XDG_DESKTOP_PORTAL_ERROR,
-                                                     XDG_DESKTOP_PORTAL_ERROR_NOT_FOUND,
-                                                     _("Requested setting not found"));
+      if (strcmp (arg_key, "timestamp") == 0)
+        {
+          g_dbus_method_invocation_return_value (invocation,
+                                                 g_variant_new ("(v)", g_variant_new_int32 (self->fontconfig_timestamp)));
+          return TRUE;
+        }
```
TingPing: else error?

src/settings.c
TingPing: I'm admittedly ignorant where this timestamp is used, but would it be a breaking change to make it monotonic?

src/fc-monitor.c
```diff
@@ -0,0 +1,321 @@
...
+                g_debug ("Got %-38s for %s: restarting fontconfig update", event_name, path);
+                self->state = UPDATE_RESTART;
+                break;
+
+        case UPDATE_RESTART:
+                g_debug ("Got %-38s for %s: waiting on fontconfig update", event_name, path);
+                break;
+        }
+
+	g_free (path);
```
TingPing:
```suggestion
        g_free (path);
```

src/fc-monitor.c
```diff
@@ -0,0 +1,321 @@
...
+}
+
+static void
+update_done (GObject *source_object G_GNUC_UNUSED,
+             GAsyncResult *result,
+             gpointer data)
+{
+        FcMonitor *self = FC_MONITOR (data);
+        gboolean restart = self->state == UPDATE_RESTART;
+        GError *error = NULL;
```
TingPing: We already use `g_autoptr()` in this project so might as well use it throughout here.

src/settings.c
```diff
@@ -36,6 +38,8 @@ struct _Settings
   XdpSettingsSkeleton parent_instance;
 
   GHashTable *settings;
+  FcMonitor *fontconfig_monitor;
```
TingPing: Maybe not ideal to have the exact same `Fc` prefix as real fontconfig types.

src/settings.c
```diff
@@ -121,6 +125,17 @@ settings_handle_read_all (XdpSettings           *object,
 
       g_variant_builder_add (&builder, "{s@a{sv}}", key, g_variant_dict_end (&dict));
     }
+
+  if (namespace_matches ("org.gnome.fontconfig", arg_namespace))
+    {
+      GVariantDict dict;
+
+      g_variant_dict_init (&dict, NULL);
+      g_variant_dict_insert_value (&dict, "timestamp", g_variant_new_int32 (self->fontconfig_timestamp));
```
TingPing: Probably best to keep timestamps in int64.

matthiasclasen: can't do that, since xsettings only knows string, bool and int, and we're still kinda limited by those constraints on the gtk side

TingPing: The GTK MR for it already used int64: https://gitlab.gnome.org/GNOME/gtk/merge_requests/241

Even if it gets truncated by GTK in the short term might as well make the DBus API use the right type.

matthiasclasen: I'd rather rename it. The thing is not a real timestamp in any way

matthiasclasen: I've done that now

hadess:
Any chance you could feed the changes to that copy/paste code back to g-s-d?

PR #251, opened 2018-11-05
Add note about byte array properties in filechooser portal

grulja:
Properties which are send as byte array need to be null-terminated.

### DIFF
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -194,11 +194,19 @@
         </varlistentry>
         <varlistentry>
           <term>current_folder ay</term>
-          <listitem><para>Suggested folder to save the file in.</para></listitem>
+          <listitem>
+            <para>
+              Suggested folder to save the file in. The byte array is expected to be null-terminated.
+            </para>
+          </listitem>
         </varlistentry>
         <varlistentry>
           <term>current_file ay</term>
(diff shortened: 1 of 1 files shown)


grulja:
See https://codereview.qt-project.org/#/c/243069/ for example. We had to explicitly add a terminating character in Qt, xdg-desktop-portal-gtk cannot handle those options without it, while xdg-desktop-portal-kde can.

matthiasclasen:
Sounds like a bug we should fix in xdg-desktop-portal-gtk?

Although, the GVariant docs have this:

/**
 * G_VARIANT_TYPE_BYTESTRING:
 *
 * The type of an array of bytes.  This type is commonly used to pass
 * around strings that may not be valid utf8.  In that case, the
 * convention is that the nul terminator character should be included as
 * the last character in the array.
 **/

So maybe we can be excused for not noticing this oversight earlier.

grulja:
I was trying to fix it first in xdg-desktop-portal-gtk first, but given my basic knowledge of GLib and especially GVariant, I failed. You seem to read that paramater as string, which makes it obvious why it doesn't work. In Qt I read it as byte array and when constructing a new one, the nul terminating character is appended, making this to work for me even without the terminating character and that's why I didn't spot this issue before. Knowing now that GVariant expects the terminating character, I would still suggest to include this information, as there might be people without this knowledge, they might not be using GLib at all.

There was also a similar bug about same problem here: https://github.com/flatpak/flatpak/issues/319

PR #256, opened 2018-11-12
WIP: settings: Support backend implementation

TingPing:
I have not tested this with an actual backend but this should be the general idea.

Do not merge until tested.

CC @grulja

### DIFF
--- a/data/Makefile.am.inc
+++ b/data/Makefile.am.inc
@@ -32,4 +32,5 @@ dist_introspection_DATA = \
 	data/org.freedesktop.impl.portal.Email.xml \
 	data/org.freedesktop.impl.portal.ScreenCast.xml \
 	data/org.freedesktop.impl.portal.RemoteDesktop.xml \
+	data/org.freedesktop.impl.portal.Settings.xml \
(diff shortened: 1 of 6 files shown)


matthiasclasen:
Looks good to me, modulo testing.

grulja:
There is a typo in the signal name. The introspection file says it is "SettingChanged", while you are trying to connect to "SettingsChanged", actually in the code it is "settings-changed" because of the auto-generated code. I will report other issues (if any) once I proceed next with my backend implementation.

grulja:
Update:
Now when returning my results, I get following error:
g_variant_builder_add_value: assertion '!GVSB(builder)->expected_type || g_variant_is_of_type (value, GVSB(builder)->expected_type)' failed

TingPing:
@grulja All issues should have been fixed now.

grulja:
Still doesn't seem to work correctly, when I return more than one namespace with their properties, it adds just the first one to the resulting map. I see a long list of namespaces I returned when watching it using dbus-monitor so it is definitely correct on my side.

Issue #244, opened 2018-11-02
Only one fork()


setharnold:
https://github.com/flatpak/xdg-desktop-portal/blob/70ad5e4a05239177f14b8bea27c3a9a7f746862c/document-portal/document-portal.c#L1511

Hello, I noticed that document-portal.c only performs one `fork(2)` when daemonizing. The usual approach is to `fork(2)`, call `setsid(2)`, then `fork(2)` again. The second `fork(2)` call would prevent the final process from gaining a controlling terminal by accident.

Thanks

alexlarsson:
Actually, i'm not even sure why we have the --daemon option. In testing, you run non-daemonize, and in deployment the portal is always launched by dbus (directly or indirectly via systemd), and that daemonizes by itself.

matthiasclasen:
So, shall we remove it, then ?

alexlarsson:
Yeah, i think so, i can't find any place we use it. Both the dbus and the systemd service files don't pass it.

PR #257, opened 2018-11-13
document portal: Remove --daemon option

matthiasclasen:
We don't generally daemonize in session services.
Tests run the service in the foreground, and in
regular use, it gets activated on the bus, which
already does all thats necessary.

Closes: #244


alexlarsson:
Looks good to me

PR #262, opened 2018-11-18
Icon validation

matthiasclasen:
Validate icons. Its the right thing to do.


matthiasclasen:
We probably should run the validator in a sandbox, for good measure.

src/notification.c
```diff
@@ -382,6 +363,122 @@ check_notification (GVariant *notification,
   return TRUE;
 }
 
+static gboolean
+validate_icon_more (GVariant *v)
+{
+  g_autoptr(GIcon) icon = g_icon_deserialize (v);
+  GBytes *bytes;
+  g_autofree char *name = NULL;
+  const char *argv[] = { LIBEXECDIR "/xdg-desktop-portal-validate-icon", "name", NULL };
```
TingPing: Do we want to wrap it in `bwrap` just to completely eliminate a ton of possible exploits? I mean we know the system has it might as well use it.

src/notification.c
```diff
@@ -382,6 +363,122 @@ check_notification (GVariant *notification,
...
+  if (G_IS_THEMED_ICON (icon))
+    {
+      g_autofree char *a = g_strjoinv (" ", (char **)g_themed_icon_get_names (G_THEMED_ICON (icon)));
+      g_debug ("Icon validation: themed icon (%s) is ok", a);
+      return TRUE;
+    }
+
+  if (!G_IS_BYTES_ICON (icon))
+    {
+      g_warning ("Unexpected icon type: %s", G_OBJECT_TYPE_NAME (icon));
```
TingPing: I think we can support `GFileIcon` just fine. We would just want the validator to spit out a validated file on the host or something.

matthiasclasen: Our stance on passing files to portals so far has been: everything must be passed as fd. If we accept a path, we have to worry about validating that it doesn't point to some host location.

TingPing: Pointing to a host location is safe though because this is only consumed by the host itself.

src/notification.c
```diff
@@ -382,6 +363,122 @@ check_notification (GVariant *notification,
...
+  g_autoptr(GError) error = NULL;
+
+  if (G_IS_THEMED_ICON (icon))
+    {
+      g_autofree char *a = g_strjoinv (" ", (char **)g_themed_icon_get_names (G_THEMED_ICON (icon)));
+      g_debug ("Icon validation: themed icon (%s) is ok", a);
+      return TRUE;
+    }
+
+  if (!G_IS_BYTES_ICON (icon))
```
TingPing: Do we want to convert bytes icons into a local file? I mentioned it in the other comment that this would fix `xdg-desktop-portal-kde` which doesn't support these.

src/notification.c
matthiasclasen: it is still a possible information leak - it can cause your secret photos to appear in a notification on your screen, if the app figures out the right filename to pass...

src/notification.c
matthiasclasen: does the kde backend support files ?

src/notification.c
TingPing: Hmm, alright.

src/notification.c
TingPing: Docs are kinda bad: https://api.kde.org/frameworks/knotifications/html/classKNotification.html#a9533d08af1499da4da7e1c92b8cd3b70

@grulja Does `KNotification::setIconName()` only support an icon theme name or will a file path work?

src/notification.c
```diff
@@ -382,6 +363,122 @@ check_notification (GVariant *notification,
...
+  bytes = g_bytes_icon_get_bytes (G_BYTES_ICON (icon));
+  fd = g_file_open_tmp ("iconXXXXXX", &name, &error); 
+  if (fd == -1)
+    {
+      g_debug ("Icon validation: %s", error->message);
+      return FALSE;
+    }
+
+  stream = g_unix_output_stream_new (fd, TRUE);
+  written = g_output_stream_write_bytes (stream, bytes, NULL, &error);
```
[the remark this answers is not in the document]
matthiasclasen: From the default session bus config:

  

  
  <limit name="max_incoming_bytes">1000000000</limit>
  <limit name="max_incoming_unix_fds">250000000</limit>
  <limit name="max_outgoing_bytes">1000000000</limit>
  <limit name="max_outgoing_unix_fds">250000000</limit>
  <limit name="max_message_size">1000000000</limit>

Can our dbus proxy enforce a smaller limit ?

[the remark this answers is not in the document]
TingPing: From what I can tell the proxy has no limits itself, so that probably isn't ideal.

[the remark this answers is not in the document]
matthiasclasen: Right. I'm saying: if we wanted to address bus DOS, the proxy would be the place to do it

TingPing:
If we don't want the portal to handle `GFileIcon`s does it make sense for Gio to auto-convert them to `GBytesIcon`s?

matthiasclasen:
I don't think passing by fd is going to work unless we do major work on the notification serialization. Converting to a bytes icon should be doable

src/notification.c
grulja: It seems to work also with a file path. I just tested it and it works.

TingPing: Does it handled `file://` URIs? (which the freedesktop spec says to use)

grulja: Sorry for not responding earlir, I somehow missed this question. Yes, it does handle file:// URIs.

matthiasclasen: @grulja: the more important question is: do you handle GBytesIcon, with embedded png content?
 If not, could you add it?

grulja: Unfortunately no, we don't handle it yet. KNotification API allows to set QPixmap so I guess in this case if QPIxmap is set, I should convert it to GBytesIcon and send over DBus. What is GBytesIcon, would something like this https://wiki.qt.io/Converting_QPixmap_to_QByteArray work? Just converting QPixmap to QByteArray and sending it over DBus?

TingPing: Yes that looks like it could work, but we don't guarantee that it is PNG so it might be JPEG or SVG, I assume Qt has APIs to figure that out. You can get the bytes via `g_bytes_icon_get_bytes()`.

grulja:
I tried to add support for GBytesIcon and test whether I pass the validation, but I get following crash:
```
** (/usr/libexec/xdg-desktop-portal:24596): WARNING **: 10:28:14.158: Error getting permissions: GDBus.Error:org.freedesktop.portal.Error.NotFound: No entry for notification
(/usr/libexec/xdg-desktop-portal:24596): GLib-DEBUG: 10:28:14.158: posix_spawn avoided (fd close requested)
[Detaching after fork from child process 25664]
error: **
ERROR:src/xdp-utils.h:119:xdp_close_fd: assertion failed: (errno != EBADF)
 
 
Thread 7 "pool" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffe9466700 (LWP 25663)]
0x00007ffff796353f in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff796353f in raise () at /lib64/libc.so.6
#1  0x00007ffff794d895 in abort () at /lib64/libc.so.6
#2  0x00007ffff7b2ede3 in  () at /lib64/libglib-2.0.so.0
#3  0x00007ffff7b896be in g_assertion_message_expr () at /lib64/libglib-2.0.so.0
#4  0x00000000004110b7 in xdp_close_fd (fdp=0x7fffe9465bac) at src/xdp-utils.h:119
#5  0x0000000000412249 in validate_icon_more (v=0x7fffdc02ac30) at src/notification.c:376
#6  0x0000000000412301 in maybe_remove_icon (notification=0x7fffd401a290) at src/notification.c:448
#7  0x0000000000412447 in handle_add_in_thread_func (task=0x7fffd400dab0, source_object=0x50fd00, task_data=0x4e1750, cancellable=0x0) at src/notification.c:475
#8  0x00007ffff7d1aa07 in  () at /lib64/libgio-2.0.so.0
#9  0x00007ffff7b8ae93 in  () at /lib64/libglib-2.0.so.0
#10 0x00007ffff7b8a48a in  () at /lib64/libglib-2.0.so.0
#11 0x00007ffff7af958e in start_thread () at /lib64/libpthread.so.0
#12 0x00007ffff7a28653 in clone () at /lib64/libc.so.6
```

matthiasclasen:
@grulja I believe I've fixed that

matthiasclasen:
@TingPing I've added bwrap now

matthiasclasen:
I'll double-check

src/notification.c
matthiasclasen: I've added docs about the supported formats now

src/notification.c
```diff
@@ -382,6 +359,189 @@ check_notification (GVariant *notification,
...
+add_bwrap (GPtrArray *array,
+           const char *input)
+{
+  const char *bwrap;
+
+  bwrap = g_getenv ("FLATPAK_BWRAP");
+  if (bwrap == NULL)
+    bwrap = "/usr/bin/bwrap";
+
+  add_args (array,
```
TingPing: Probably a good idea to: `"--tmpfs", "/tmp", "--unsetenv", "TMPDIR",`

src/notification.c
```diff
@@ -382,6 +359,189 @@ check_notification (GVariant *notification,
...
+
+static void
+add_bwrap (GPtrArray *array,
+           const char *input)
+{
+  const char *bwrap;
+
+  bwrap = g_getenv ("FLATPAK_BWRAP");
+  if (bwrap == NULL)
+    bwrap = "/usr/bin/bwrap";
```
TingPing: Not a fan of hardcoding `/usr/bin`, maybe match the configured prefix of flatpak.

TingPing: Alternatively just get the default at configure time, which is probably the most reliable.

PR #267, opened 2018-11-30
Lockdown, part 2

matthiasclasen:
Redo lockdown properties as a separate interface, to avoid problems with the lack of backends for some portals that may grow lockdown.

### DIFF
--- a/data/org.freedesktop.impl.portal.Lockdown.xml
+++ b/data/org.freedesktop.impl.portal.Lockdown.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2018 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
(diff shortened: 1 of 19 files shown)


src/location.c
```diff
@@ -718,10 +739,13 @@ location_class_init (LocationClass *klass)
 
 GDBusInterfaceSkeleton *
 location_create (GDBusConnection *connection,
-                 const char *dbus_name)
+                 const char *dbus_name,
+                 gpointer lockdown_proxy)
 {
   g_autoptr(GError) error = NULL;
 
+  lockdown = g_object_ref (lockdown_proxy);
```
TingPing: Odd to take refs and not release them.

data/org.freedesktop.impl.portal.Lockdown.xml
```diff
@@ -0,0 +1,42 @@
...
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+    org.freedesktop.portal.Settings:
+    @short_description: Settings interface
```
TingPing: Old description.

TingPing:
When in the future more lockdowns are added would it be a cleaner interface to take a string for which lockdown to lookup vs handling non-existing properties?

matthiasclasen:
I went for properties here, since it makes for a very neat backend implementation with g_settings_bind, and we get change notification and caching for free.

src/location.c
matthiasclasen: That did occur to me too. We're leaking stuff all over the place here, anyway. Let me fix this up

PR #286, opened 2019-01-16
file-chooser: Add support for proxying current_filter

mcatanzaro:
This will allow xdg-desktop-portal-gtk to call
gtk_file_chooser_set_filter() to make file filters not suck to use.

https://gitlab.gnome.org/GNOME/gtk/issues/1492

### DIFF
--- a/data/org.freedesktop.impl.portal.FileChooser.xml
+++ b/data/org.freedesktop.impl.portal.FileChooser.xml
@@ -67,6 +67,13 @@
             </para></listitem>
           </varlistentry>
           <varlistentry>
+            <term>current_filter (sa(us))</term>
+            <listitem><para>
+              Request that this filter be set by default at dialog creation.
+              See org.freedesktop.portal.FileChooser.OpenFile() for details.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
             <term>choices a(ssa(ss)s)</term>
             <listitem><para>
               A list of serialized combo boxes.
@@ -139,6 +146,13 @@
             <term>filters a(sa(us))</term>
             <listitem><para>
               A list of serialized file filters.
+              See org.freedesktop.portal.FileChooser.OpenFile() for details.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>current_filter (sa(us))</term>
+            <listitem><para>
+              Request that this filter be set by default at dialog creation.
               See org.freedesktop.portal.FileChooser.OpenFile() for details.
             </para></listitem>
           </varlistentry>
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -89,6 +89,16 @@
           </listitem>
         </varlistentry>
         <varlistentry>
+          <term>current_filter (sa(us))</term>
+          <listitem><para>
+            Request that this filter be set by default at dialog creation. If
+            the filters list is nonempty, it should match a filter in the
+            list to set the default filter from the list. Alternatively, it
+            may be specified when the list is empty to apply the filter
+            unconditionally.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
           <term>choices a(ssa(ss)s)</term>
           <listitem>
             <para>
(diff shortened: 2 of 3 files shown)


mcatanzaro:
Question: do we need to bump the version property? https://github.com/flatpak/xdg-desktop-portal/commit/5ae006913e45bcea10f6e88db6fabaf80257ff48 says yes, because this adds a new option, but there's not actually any reason that checking the version would be required at any side of the protocol: if the option is supported, great you can pull it out of the dict, and if not, no harm done.

matthiasclasen:
I'm not a fan of duplicating a full filter variant in so many places. If we ever change how filters are encoded, all these places need to be updated. I'd much rather have an index into the filter array. Sorry

mcatanzaro:
> I'm not a fan of duplicating a full filter variant in so many places. If we ever change how filters are encoded, all these places need to be updated. I'd much rather have an index into the filter array. Sorry

I can do that, but then `gtk_file_chooser_set_filter()` will still not work properly in all cases. The documentation says:

> Sets the current filter; only the files that pass the filter will be displayed. If the user-selectable list of filters is non-empty, then the filter should be one of the filters in that list. Setting the current filter when the list of filters is empty is useful if you want to restrict the displayed set of files without letting the user change it.

And that's not possible to implement unless we use a variant here. With the variant, it should Just Work. And almost all of the variant parsing code is nicely reused. I'll certainly change it to an int if you want, but variant seems better to me.

mcatanzaro:
To be clear:

> Setting the current filter when the list of filters is empty is useful if you want to restrict the displayed set of files without letting the user change it.

If we use an index, setting the current filter when the list of filters is empty will do nothing. (Which is not awful, but also not the documented behavior.)

matthiasclasen:
Hmm, you have a point. Let me think about this for a minute

data/org.freedesktop.impl.portal.FileChooser.xml
```diff
@@ -66,6 +66,13 @@
               See org.freedesktop.portal.FileChooser.OpenFile() for details.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>current_filter (sa(us))</term>
+            <listitem><para>
+              Filter to set at dialog creation.
+              If the filters list is nonempty, it should be present there.
```
matthiasclasen: This bit of documentation is somewhat unclear (admittedly already the case in gtk): Does this mean it is an error if there is not an identical variant in the filter list, or that the caller can expect this filter to be added to the list that is displayed in the UI ?

matthiasclasen:
> Question: do we need to bump the version property? 5ae0069 says yes, because this adds a new option, but there's not actually any reason that checking the version would be required at any side of the protocol: if the option is supported, great you can pull it out of the dict, and if not, no harm done.

I believe that sending a current_filter option to and old xdg-desktop-portal-release will probably end in an error from the option validation ? If thats the case, bumping the version would seem advisable, so callers have a chance to check and only send the new option when it won't yield an error.

mcatanzaro:
> I believe that sending a current_filter option to and old xdg-desktop-portal-release will probably end in an error from the option validation ? If thats the case, bumping the version would seem advisable, so callers have a chance to check and only send the new option when it won't yield an error.

It would be questionable if xdg-desktop-portal were to complain about unrecognized options, since that would make it difficult to add new options and there would be little benefit. I checked the implementation of `xdp_filter_options()`, where the validation occurs in master, and it looks like it fortunately only validates recognized options. I've also manually tested with the 1.0.3 release and it doesn't notice or complain about the new option. Seems fine.

data/org.freedesktop.impl.portal.FileChooser.xml
mcatanzaro: For GTK+, it's the former. The filter will not be added to the list displayed in the UI.

For Qt... the documentation is not at all clear. I guess we can document the GTK+ behavior, and if Qt wants to do something different it can just add a new option to the dict.

matthiasclasen:
> It would be questionable if xdg-desktop-portal were to complain about unrecognized options, since that would make it difficult to add new options and there would be little benefit. I checked the implementation of xdp_filter_options(), where the validation occurs in master, and it looks like it fortunately only validates recognized options. I've also manually tested with the 1.0.3 release and it doesn't notice or complain about the new option. Seems fine.

There is some value to it. If you silently ignore unrecognized options then
- apps who pass them now will get burned in the future when they start doing something
- typos in option names will go unnoticed
- apps expecting them to do something will not get the functionality they expect when running against an older portal

mcatanzaro:
Do you want me to bump the version, then (even though there is not currently any complaint for unrecognized options)?

matthiasclasen:
I've changed the validate functions to take the other options as argument as well.
If you rebase this work, you can do the filter validation between the two options.

mcatanzaro:
D'oh, of course passing all the options is better than my proposal. I'll update momentarily.

mcatanzaro:
OK, ready for review again.

PR #289, opened 2019-01-18
Update bwrap invocation from libgnome-desktop to fix #288

### LINKED ISSUE #288: notifications portal invokes bubblewrap in a way that won't work on pure 32-bit
The notifications portal uses `bwrap ... --ro-bind /lib64 /lib64`, which will only work if you have `/lib64`. Pure 32-bit systems (such as i386 with no x86_64 biarch support, and 32-bit ARM with no aarch64 biarch support) typically won't have that directory.

I'm testing a patch for this, based on fixes previously made in libgnome-desktop.

smcv:
We need to handle source directories not existing, for example /lib64 on a pure 32-bit system (#288).

Because relative symlinks in a directory that is itself a symlink do not have the same meaning as the same relative symlink in a bind-mount of that directory, ideally we want to use --symlink for /lib and /lib64 on merged-/usr systems, but --ro-bind on non-merged-/usr systems.

Changes originally made in libgnome-desktop by Iain Lane (see <https://bugzilla.gnome.org/show_bug.cgi?id=787072>).

This branch also applies a later libgnome-desktop change from @leio to bind-mount `/etc/ld.so.cache` into the container, which is useful if libraries are in non-standard directories enabled by `/etc/ld.so.conf.d/`.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -432,6 +432,7 @@ add_bwrap (GPtrArray *array,
   add_args (array,
             BWRAP,
             "--ro-bind", "/usr", "/usr",
+            "--ro-bind", "/etc/ld.so.cache", "/etc/ld.so.cache",
             NULL);
 
   /* These directories might be symlinks into /usr/... */
--- a/src/notification.c
+++ b/src/notification.c
@@ -23,6 +23,7 @@
 #include <string.h>
 #include <gio/gio.h>
 #include <gio/gunixoutputstream.h>
+#include <glib/gstdio.h>
 
 #include "notification.h"
 #include "request.h"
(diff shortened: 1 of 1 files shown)


matthiasclasen:
Thanks, I'll go #290, which copies the entire sandbox setup from the flatpak code that is invoking the icon-validator, and also switches to using the flatpak-icon-validator binary.

matthiasclasen:
So maybe this is needed after all. Can we somehow share the setup code between flatpak and xdg-desktop-portal ? Maybe the icon-validator can sandbox itself ?

smcv:
Every time there's a new micro-module I have to wait an unpredictable length of time (could be anywhere from 1 day to several months) to get it into Debian, so please don't be too hasty to introduce those.

matthiasclasen:
Filed https://github.com/flatpak/flatpak/issues/2617

PR #290, opened 2019-01-18
Use flatpak-icon-validator

matthiasclasen:
The icon validator is now shipped with flatpak. Lets use that one.


src/notification.c
```diff
@@ -480,27 +448,42 @@ validate_icon_more (GVariant *v)
...
 
-  add_bwrap (array, name);
-  add_args (array,
-            LIBEXECDIR "/xdg-desktop-portal-validate-icon", name,
+  add_args (args,
+            BWRAP,
+            "--unshare-ipc",
+            "--unshare-net",
+            "--unshare-pid",
+            "--ro-bind", "/", "/",
```
smcv: I thought the purpose of only bind-mounting `/usr` was to stop a sufficiently malformed icon from causing arbitrary code execution that can exfiltrate user files from `/home`, `/srv`, etc.? I see networking and IPC are disallowed here, but a compromised validator could still use D-Bus or something.

matthiasclasen: we can disallow dbus too, for sure

matthiasclasen: and we can also mount less, but we should keep the sandbox setup in sync between the src/notifications.c here and app/flatpak-builtings-build-export.c in flatpak

smcv: I think there's a difference in scope: if you're *building* a Flatpak app, hopefully you trust it not to be actively malicious, whereas if you have merely installed a Flatpak app, you don't necessarily trust it not to be trying to exploit the host system.

PR #294, opened 2019-01-20
inhibit: Add session state to the StateChanged signal

matthiasclasen:
This will be used to inform monitoring applications about
the impending end of the session.


matthiasclasen:
From irc discussion:

Make this less racy by having sessions ack the query-end state with a call, and have a short timeout waiting for the acks before completing the query-end

matthiasclasen:
I have added a QueryEndResponse method.

Issue #299, opened 2019-01-30
add a portal for autostart


matthiasclasen:
This has been brought up in https://github.com/flatpak/flatpak/issues/118

PR #300, opened 2019-02-03
Add a background portal

matthiasclasen:
This is a simple portal to set a 'run in background' permission.

Closes: #299


matthiasclasen:
An alternative to creating autostart files that was mentioned on irc is systemd user services.

Downside: that'll only work with systemd --user, so not quite as universal.

darkdragon-001:
> landed in another pr

#319

PR #310, opened 2019-03-22
document-portal: make xdp_inode_kernel_unref check that kernel_ref_count > 0

jhenstridge:
We're in the process of backporting xdg-desktop-portal 1.0.3 to older Ubuntu releases, and ran into document portal test failures on Ubuntu 18.04.  Here's one example of the test run:
```
Running test: xdg-desktop-portal/test-doc-portal.test
# random seed: R02S2a471573ded9c5f7b07769440c61b09b
outdir: /tmp/xdp-test-O0ZGYZ
dbus-daemon[1003]: [session uid=1000 pid=1003] Activating service name='org.freedesktop.portal.Documents' requested by ':1.0' (uid=1000 pid=1001 comm="/usr/libexec/installed-tests/xdg-desktop-portal/te" label="unconfined")
dbus-daemon[1003]: [session uid=1000 pid=1003] Activating service name='org.freedesktop.impl.portal.PermissionStore' requested by ':1.1' (uid=1000 pid=1008 comm="/usr/libexec/xdg-document-portal " label="unconfined")
dbus-daemon[1003]: [session uid=1000 pid=1003] Successfully activated service 'org.freedesktop.impl.portal.PermissionStore'
dbus-daemon[1003]: [session uid=1000 pid=1003] Successfully activated service 'org.freedesktop.portal.Documents'
1..4
# Start of db tests
ok 1 /db/create_doc
path: /tmp/xdp-test-O0ZGYZ/doc/c2c340f/recursive-file
ok 2 /db/recursive_doc
ok 3 /db/create_docs
ok 4 /db/add_named
# End of db tests
error: **
xdg-desktop-portal:ERROR:document-portal/document-portal-fuse.c:244:xdp_inode_destroy: assertion failed: (inode->children == NULL)

**
xdg-desktop-portal:ERROR:tests/test-doc-portal.c:838:global_teardown: assertion failed: (i != 1000)
Bail out! xdg-desktop-portal:ERROR:tests/test-doc-portal.c:838:global_teardown: assertion failed: (i != 1000)
cleaning up pid 1003
FAIL: xdg-desktop-portal/test-doc-portal.test (Child process killed by signal 6)
```

The xdg-document-portal instance spawned by the test is crashing on an assertion failure.  The inode in question is the root inode:
```
(gdb) print *inode
$1 = {ref_count = 0, kernel_ref_count = -1, ino = 1, type = XDP_INODE_ROOT,
  parent = 0x0, app_id = 0x0, doc_id = 0x0, basename = 0x0, dirname = 0x0,
  dir_dev = 0, dir_ino = 0, children = 0x556c93bad6a0,
  filename = 0x556c93bb0e00 "/", is_doc = 0, mutex = {p = 0x0, i = {0, 0}},
  open_files = 0x0, dir_fd = -1, fd = -1, backing_filename = 0x0,
  trunc_filename = 0x0, trunc_fd = -1, truncated = 0}
```

The one remaining child is the "by-app" subdirectory, but the real problem seems to be `kernel_ref_count = -1`.  With a bit more instrumentation, it looks like the kernel is sending a forget() call for the root inode with no matching lookup().

This is almost certainly a kernel bug, but adding an explicit check that kernel_ref_count is actually positive in this case seems like a useful sanity check.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -321,7 +321,19 @@ static void
 xdp_inode_kernel_unref (XdpInode *inode)
 {
   if (inode)
-    g_atomic_int_dec_and_test (&inode->kernel_ref_count);
+    {
+      gint old_ref;
+
+    retry_atomic_decrement1:
+      old_ref = g_atomic_int_get (&inode->kernel_ref_count);
+      if (old_ref <= 0)
+        {
+          g_warning ("Can't kernel_unref inode with no kernel refs");
+          return;
+        }
+      if (!g_atomic_int_compare_and_exchange (&inode->kernel_ref_count, old_ref, old_ref - 1))
+        goto retry_atomic_decrement1;
+    }
   xdp_inode_unref (inode);
 }
 


jhenstridge:
I've continued investigating this to see if I could find a root cause with no luck.  One possibility is the evict_inode() code path, but I'm not sure why that would be getting triggered.

I think there is still value in taking this patch as is though, since (a) it is corruption that we can easily detect, and (b) it is corruption that can be triggered externally.

alexlarsson:
Looks good to me. I mean, ideally we figure out whats causing this, but better safe than sorry.

PR #313, opened 2019-04-29
OpenURI: add option to open files with default associated app

grulja:
This should allow backends to automatically open files with apps associated with given mimetype. Main use-case for this is file browsers (e.g. Dolphin or Nautilus), where double-click should automatically open selected file with the default application and not always present a dialog with list of apps. The dialog should be used for example when "Open with" option from file browsers is used. I made this option to be "FALSE" by default to keep current behavior in case this is not implemented everywhere so users are not forced to have files opened with just the default application.

### DIFF
--- a/data/org.freedesktop.impl.portal.AppChooser.xml
+++ b/data/org.freedesktop.impl.portal.AppChooser.xml
@@ -62,6 +62,14 @@
             <listitem><para>The filename to choose an application for. Note that this
                is just a basename, without a path.</para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>use_associated_app b</term>
+            <listitem><para>
+              Whether to automatically use default application associated with the mime type
+              of the file we want to open.
+            </para></listitem>
+          </varlistentry>
+
         </variablelist>
 
         The following results get returned via the @results vardict:
--- a/data/org.freedesktop.portal.OpenURI.xml
+++ b/data/org.freedesktop.portal.OpenURI.xml
@@ -101,6 +101,13 @@
               is sandboxed itself.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>use-associated-app b</term>
+            <listitem><para>
+              Whether to automatically use default application associated with the mime type
+              of the file we want to open.
+            </para></listitem>
+          </varlistentry>
         </variablelist>
 
         The OpenFile method was introduced in version 2 of the OpenURI portal API.
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -518,11 +518,13 @@ handle_open_in_thread_func (GTask *task,
   gboolean skip_app_chooser = FALSE;
   int fd;
   gboolean writable = FALSE;
+  gboolean use_associated_app = FALSE;
 
   parent_window = (const char *)g_object_get_data (G_OBJECT (request), "parent-window");
   uri = g_strdup ((const char *)g_object_get_data (G_OBJECT (request), "uri"));
   fd = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (request), "fd"));
   writable = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (request), "writable"));
+  use_associated_app = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (request), "use-associated-app"));
 
   REQUEST_AUTOLOCK (request);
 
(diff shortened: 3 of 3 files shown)


hadess:
This means that bugs associated with automatically opening a particular filetype could be used, without the user knowing about it.

An application could programmaticly launch the default PDF viewer with a specially crafted file. If the default application used is a common enough one isn't sandboxed, this could be used to a privilege escalation. Eg. an attacker could add a call in a Flatpak'ed application that launches a PDF that triggers a security bug in poppler. If the user's default PDF viewer uses poppler (which is pretty likely on Linux systems, whether GNOME or KDE), and this viewer isn't sandboxed, then it might be possible to have code run outside of the sandbox.

I don't think this API should be available to any applications but file managers, which would mean an upfront declaration that the application can do this. An API could then be used to force the chooser to show up.

grulja:
How can you make sure that this API is used only by file managers? This is basically a similar problem as we had with double dialogs for screensharing in browsers. You cannot guarantee it's really the application which is supposed to have access to certain API or to behave a bit differently.

Alternatively we can do it the other way around. You can by default present the dialog to choose an app, but allow the user to check "Always use this application for this type of files" option so it will be his choice whether the default app is used or not. This could be done for each application separately so first run of e.g. Nautilus would ask whether that choice should be remembered for given mimetype and next time it would present the dialog or open the file automatically. It doesn't completely avoid the potential issue you mentioned, but gives at least users a choice. File browsers can then make a different call when "Open with" option is used to enforce showing the dialog again.

hadess:
> How can you make sure that this API is used only by file managers? This is basically a similar problem as we had with double dialogs for screensharing in browsers. You cannot guarantee it's really the application which is supposed to have access to certain API or to behave a bit differently.

That'd be a new `finish` argument in the manifest. It's not 100%, but it means that the functionality was advertised at least once to the user.

hadess:
The other problem is that changing the default application for a mime-type in the file manager doesn't change the default application outside the sandbox. I think that a new API should be created for the use by file managers.

See also this nautilus problem: https://gitlab.gnome.org/GNOME/nautilus/issues/117

grulja:
> > How can you make sure that this API is used only by file managers? This is basically a similar problem as we had with double dialogs for screensharing in browsers. You cannot guarantee it's really the application which is supposed to have access to certain API or to behave a bit differently.
> 
> That'd be a new `finish` argument in the manifest. It's not 100%, but it means that the functionality was advertised at least once to the user.

You mean that it would be advertised as a permission when user is installing a file manager app? Like "Application Nautilus requests file manager access" or something like that? I think that's a great idea, permissions can be verified and such calls would be authorized. It's not nice that file manager apps would need to use a different API and not utilize what is already being used, but it's not like rewritting majority of the codebase to make it work properly sandboxed so perfectly fine.

hadess:
> You mean that it would be advertised as a permission when user is installing a file manager app?

Yes, exactly this. And I would expect the internal/portal API to implement this to be completely transparent to applications using the standard GLib/Qt APIs. Eventually :)

grulja:
@hadess is anyone already working on this or is it free to take?

hadess:
> @hadess is anyone already working on this or is it free to take?

I don't think anyone's working on this yet.

waltmck:
I am running into this issue, and find it annoying enough that I am maintaining a patch to work around it.

> This means that bugs associated with automatically opening a particular filetype could be used, without the user knowing about it.
> 
> An application could programmaticly launch the default PDF viewer with a specially crafted file. If the default application used is a common enough one isn't sandboxed, this could be used to a privilege escalation. Eg. an attacker could add a call in a Flatpak'ed application that launches a PDF that triggers a security bug in poppler. If the user's default PDF viewer uses poppler (which is pretty likely on Linux systems, whether GNOME or KDE), and this viewer isn't sandboxed, then it might be possible to have code run outside of the sandbox.
> 
> I don't think this API should be available to any applications but file managers, which would mean an upfront declaration that the application can do this. An API could then be used to force the chooser to show up.

I don't buy the security argument, since a hardcoded list of mime types is already allowed to open without a prompt: http, https, ftp, mailto, webcal, calendar, and inode/directory. You could do plenty of damage by opening arbitrary URLs with the default browser. In my opinion, either this is a concern---in which case `xdg-desktop-portal` should also drop the hardcoded allowlist---or it is not a concern and you should make it configurable. I can't fathom any reason why opening an HTTP link is less dangerous than opening a PDF.

PR #319, opened 2019-05-13
Implement a background monitor

matthiasclasen:
Implement a facility that monitors running flatpaks,
and notifies the user if they run in the background.
Add a permission table to store users responses
persistently.

The platform-dependent parts of this (getting the
state of running applications, and notifying the user)
are handled by a portal backend.

There is currently no application API.

### DIFF
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2019 Red Hat, Inc
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+    org.freedesktop.impl.portal.Background:
+    @short_description: Background portal backend interface
+
+    This interface provides APIs related to applications
+    that are running in the background.
+  -->
+  <interface name="org.freedesktop.impl.portal.Background">
+
+    <!--
+      EnableAutostart:
+      @app_id: App id of the application
+      @enable: TRUE to enable autostart, FALSE to disable it
+      @commandline: The commandline to use in the autostart file
+      @activatable: Whether to use D-Bus activation in the autostart file
+      @result: TRUE if autostart was enabled, FALSE otherwise
+
+      Enables or disables autostart for an application.
+    -->
+    <method name='EnableAutostart'>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="b" name="enable" direction="in"/>
+      <arg type="as" name="commandline" direction="in"/>
+      <arg type="b" name="activatable" direction="in"/>
+      <arg type="b" name="result" direction="out"/>
+    </method>
+  </interface>
+</node>
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -65,6 +65,7 @@
             <simplelist>
               <member>0: Forbid background activity by this app</member>
               <member>1: Allow background activity by this app</member>
+              <member>2: Allow this instance of background activity</member>
             </simplelist>
           </para></listitem>
         </varlistentry>
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -29,6 +29,51 @@
   <interface name="org.freedesktop.impl.portal.Background">
 
     <!--
+      GetAppState:
+      @apps: an array with information about running apps
+
+      Gets information about running apps. Each entry has
+      an application ID as key. The returned values are of
+      type u and have the following meaning:
+      <simplelist>
+        <member>0: Background (no open window)</member>
+        <member>1: Running (at least one open window)</member>
+        <member>2: Active (in the foreground)</member>
+      </simplelist>
+    -->
+    <method name='GetAppState'>
+      <arg type="a{sv}" name="apps" direction="out"/>
+    </method>
+
+    <!--
+      NotifyBackground:
+      @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+      @app_id: App id of the application
+      @name: A display name for the application
+      @response: Numeric response, not used
+      @results: Vardict with results of the call
+
+      Notifies the user that an application is running in the background.
+
+      The following results get returned via the @results vardict:
+      <variablelist>
+        <varlistentry>
+          <term>allow b</term>
+          <listitem><para>
+            TRUE to allow the app to run in the background, FALSE to stop it.
+          </para></listitem>
+        </varlistentry>
+      </variablelist>
+    -->
+    <method name='NotifyBackground'>
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="s" name="name" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <!--
       EnableAutostart:
       @app_id: App id of the application
       @enable: TRUE to enable autostart, FALSE to disable it
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -58,9 +58,14 @@
       The following results get returned via the @results vardict:
       <variablelist>
         <varlistentry>
-          <term>allow b</term>
+          <term>result u</term>
           <listitem><para>
-            TRUE to allow the app to run in the background, FALSE to stop it.
+            The choice that the user made regarding the background
+            activity:
+            <simplelist>
+              <member>0: Forbid background activity by this app</member>
+              <member>1: Allow background activity by this app</member>
+            </simplelist>
           </para></listitem>
         </varlistentry>
       </variablelist>
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -78,16 +78,21 @@
       @app_id: App id of the application
       @enable: TRUE to enable autostart, FALSE to disable it
       @commandline: The commandline to use in the autostart file
-      @activatable: Whether to use D-Bus activation in the autostart file
+      @flags: Flags influencing the details of the autostarting
       @result: TRUE if autostart was enabled, FALSE otherwise
 
       Enables or disables autostart for an application.
+
+      The following flags are understood:
+      <simplelist>
+        <member>1: Use D-Bus activation</member>
+      </simplelist>
     -->
     <method name='EnableAutostart'>
       <arg type="s" name="app_id" direction="in"/>
       <arg type="b" name="enable" direction="in"/>
       <arg type="as" name="commandline" direction="in"/>
-      <arg type="b" name="activatable" direction="in"/>
+      <arg type="u" name="flags" direction="in"/>
       <arg type="b" name="result" direction="out"/>
     </method>
   </interface>
--- a/data/org.freedesktop.portal.Background.xml
+++ b/data/org.freedesktop.portal.Background.xml
@@ -23,10 +23,9 @@
       org.freedesktop.portal.Background:
       @short_description: Portal for requesting autostart and background activity
 
-      This simple interface lets sandboxed applications request that the
-      application is allowed to run in the background. If the application installs
-      an autostart file, it will be used to run the application when the user
-      logs in.
+      This simple interface lets sandboxed applications request that
+      the application is allowed to run in the background or started
+      automatically when the user logs in.
 
       This documentation describes version 1 of this interface.
   -->
@@ -76,7 +75,7 @@ k linkend="parent_window">Common Conventions</link>
           <varlistentry>
             <term>dbus-activatable b</term>
             <listitem><para>
-              If TRUE, add DBusActivatable=true to the autostart file.
+              If TRUE, use D-Bus activation for autostart.
             </para></listitem>
           </varlistentry>
         </variablelist>
--- a/data/org.freedesktop.portal.Background.xml
+++ b/data/org.freedesktop.portal.Background.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2019 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.Background:
+      @short_description: Portal for requesting autostart and background activity
+
+      This simple interface lets sandboxed applications request that the
+      application is allowed to run in the background. If the application installs
+      an autostart file, it will be used to run the application when the user
+      logs in.
+
+      This documentation describes version 1 of this interface.
+  -->
+  <interface name="org.freedesktop.portal.Background">
+    <!--
+        RequestBackground:
+        @parent_window: Identifier for the application window, see <lin
+k linkend="parent_window">Common Conventions</link>
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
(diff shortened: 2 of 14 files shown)


matthiasclasen:
See https://github.com/flatpak/xdg-desktop-portal-gtk/pull/198 for the backend

matthiasclasen:
I've added the background portal to this PR, so applications can now request run-in-background permissions ahead of time.

hadess:
I had a quick look through this:
- the `org.freedesktop.portal.Background` interface doc doesn't make it clear why it needs a window identifier, and it took me a while to realise that this was to show a dialogue to ask for permission, not for the functionality itself.
- "Rename access impl" should probably be squashed into the original commit

Is there something that terminates the Flatpak when "running in the background" is disabled and the application is not otherwise running?

src/background.c
```diff
@@ -0,0 +1,347 @@
...
+get_permission (const char *app_id)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) out_perms = NULL;
+  g_autoptr(GVariant) out_data = NULL;
+  const char **permissions;
+
+  if (!xdp_impl_permission_store_call_lookup_sync (get_permission_store (),
+                                                   PERMISSION_TABLE,
+                                                   PERMISSION_ID,
```
alexlarsson: Why are you using this index here? It seems like we're basically returning the entire table every time. Is that going to scale? Can't you use the app id as index instead?

src/background-monitor.c
```diff
@@ -0,0 +1,505 @@
...
+
+static gboolean
+enforce_background_permissions (gpointer data)
+{
+  g_autoptr(GTask) task = NULL;
+
+  task = g_task_new (NULL, NULL, NULL, NULL);
+  g_task_run_in_thread (task, check_background_apps);
+
+  return G_SOURCE_CONTINUE;
```
alexlarsson: This seems risky to me. You run this in a regular 30sec timeout and then the timeout spawns a thread. However, the thread relies on global state (the applications hashtable). So, you risk some weird behaviour if ever check_background_apps takes > 30 secs for whatever reason. I think it would be better if enforce_background_permissions returned FALSE, and the thread instead queued a new timeout.

src/background.c
alexlarsson: I guess you need to get the permissions for all the apps in the checking case, to avoid a lot of dbus calls. Still, it seems non-ideal.

alexlarsson:
In some earlier permutation of this we had some leeway for being in the background to avoid the "race condition" where an app was just starting in parallel with the app background check. Like, allow the app to be in the background for one cycle, and only ask the second time we check. I think we need something like that to be robust.

src/background.c
```diff
@@ -324,13 +354,15 @@ background_class_init (BackgroundClass *klass)
 
 GDBusInterfaceSkeleton *
 background_create (GDBusConnection *connection,
-                   const char *dbus_name)
+                   const char *dbus_name_access,
+                   const char *dbus_name_background)
 {
   g_autoptr(GError) error = NULL;
 
+g_print ("background create %s %s\n", dbus_name_access, dbus_name_background);
```
alexlarsson: leftover?

src/xdg-desktop-portal.c
```diff
@@ -188,7 +188,7 @@ sort_impl_by_name (gconstpointer a,
 static void
 load_installed_portals (void)
 {
-  const char *portal_dir = PKGDATADIR "/portals";
+  const char *portal_dir = "portals";
```
alexlarsson: leftover?

matthiasclasen:
>     * the `org.freedesktop.portal.Background` interface doc doesn't make it clear why it needs a window identifier, and it took me a while to realise that this was to show a dialogue to ask for permission, not for the functionality itself.

I copied the usual docs over now.
> 
>     * "Rename access impl" should probably be squashed into the original commit
> 

Will do.

> 
> Is there something that terminates the Flatpak when "running in the background" is disabled and the application is not otherwise running?

The background monitor checks every few minutes and kills apps that are running in the background and are not allowed. We could probably add a way to trigger that, so the control-center can call it when you change permissions there.

src/background-monitor.c
matthiasclasen: > This seems risky to me. You run this in a regular 30sec timeout and then the timeout spawns a thread. However, the thread relies on global state (the applications hashtable). So, you risk some weird behaviour if ever check_background_apps takes > 30 secs for whatever reason. I think it would be better if enforce_background_permissions returned FALSE, and the thread instead queued a new timeout.

I've changed this - no need for a timeout at all

src/background.c
matthiasclasen: > In some earlier permutation of this we had some leeway for being in the background to avoid the "race condition" where an app was just starting in parallel with the app background check. Like, allow the app to be in the background for one cycle, and only ask the second time we check. I think we need something like that to be robust.

Still the case. I added a comment to make that clear.

matthiasclasen:
> In some earlier permutation of this we had some leeway for being in the background to avoid the "race condition" where an app was just starting in parallel with the app background check. Like, allow the app to be in the background for one cycle, and only ask the second time we check. I think we need something like that to be robust.

Still the case. I added a comment to make that clear.

data/org.freedesktop.impl.portal.Background.xml
```diff
@@ -0,0 +1,49 @@
...
+    that are running in the background.
+  -->
+  <interface name="org.freedesktop.impl.portal.Background">
+
+    <!--
+      EnableAutostart:
+      @app_id: App id of the application
+      @enable: TRUE to enable autostart, FALSE to disable it
+      @commandline: The commandline to use in the autostart file
+      @activatable: Whether to use D-Bus activation in the autostart file
```
alexlarsson: I think this should be a flag instead of a boolean. That both makes the code using this easier to read, and allow it to be easily extended later.

data/org.freedesktop.portal.Background.xml
```diff
@@ -0,0 +1,82 @@
...
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.Background:
+      @short_description: Portal for requesting autostart and background activity
+
+      This simple interface lets sandboxed applications request that the
+      application is allowed to run in the background. If the application installs
+      an autostart file, it will be used to run the application when the user
+      logs in.
```
alexlarsson: The reference to autostart file here is a bit outdated, it should not say "installs an autostart file" but that is not true anymore. It should say something like "also allows applications to enable automatically starting on login".

src/background.c
```diff
@@ -0,0 +1,348 @@
...
+  gboolean allowed;
+  const char *reason = NULL;
+
+  REQUEST_AUTOLOCK (request);
+
+  options = (GVariant *)g_object_get_data (G_OBJECT (request), "options");
+  g_variant_lookup (options, "reason", "s", &reason);
+
+  app_id = xdp_app_info_get_id (request->app_info);
+  permission = get_permission (app_id);
```
alexlarsson: I think this needs to special case the xdp_app_info_is_host() case and always return TRUE, because we want to support using portals unconditionally while "just working" when not sandboxed.

alexlarsson: Probably pass the app_info instead of the app_id into get_permission and have the special case there.

src/background.c
```diff
@@ -0,0 +1,348 @@
...
+      guint32 response = 2;
+      g_autoptr(GVariant) results = NULL;
+      g_autoptr(GError) error = NULL;
+      g_autoptr(GAppInfo) info = NULL;
+
+      if (app_id[0] != 0)
+        {
+          g_autofree char *desktop_id;
+          desktop_id = g_strconcat (app_id, ".desktop", NULL);
+          info = (GAppInfo*)g_desktop_app_info_new (desktop_id);
```
alexlarsson: Could you make a xdp_app_info_load_app_info() utility for this?

src/background.c
```diff
@@ -0,0 +1,348 @@
...
+        subtitle = g_strdup (reason);
+      else
+        subtitle = g_strdup_printf (_("%s requests to run in the background."), info ? g_app_info_get_display_name (info) : app_id);
+      body = g_strdup (_("The ‘run in background’ permission can be changed at any time from the application settings."));
+
+      g_debug ("Calling backend for background access for: %s", app_id);
+
+      g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
+      g_variant_builder_add (&opt_builder, "{sv}", "deny_label", g_variant_new_string (_("Don't allow")));
+      g_variant_builder_add (&opt_builder, "{sv}", "grant_label", g_variant_new_string (_("Allow")));
```
alexlarsson: Do we never want the notification to be able to reply "always ask"? (or "allow once")

data/org.freedesktop.impl.portal.Background.xml
```diff
@@ -28,6 +28,51 @@
...
+      @results: Vardict with results of the call
+
+      Notifies the user that an application is running in the background.
+
+      The following results get returned via the @results vardict:
+      <variablelist>
+        <varlistentry>
+          <term>allow b</term>
+          <listitem><para>
+            TRUE to allow the app to run in the background, FALSE to stop it.
```
alexlarsson: Should we allow replying "allow once"?

src/background.c
```diff
@@ -111,6 +112,70 @@ get_permission (const char *app_id)
...
+      g_debug ("No background permissions found: %s", error->message);
+      return NULL;
+    }
+
+  return g_steal_pointer (&out_perms);
+}
+
+static Permission
+get_one_permission (const char *app_id,
+                    GVariant   *perms)
```
alexlarsson: Can get_permission() not use this helper too?

alexlarsson: In fact, it could probably call get_all_permissions() first and then get_one_permission().

src/background.c
```diff
@@ -409,6 +474,372 @@ background_class_init (BackgroundClass *klass)
...
+start_background_monitor (void)
+{
+  g_autoptr(GTask) task = NULL;
+
+  applications = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+  g_debug ("Starting background app monitor");
+
+  task = g_task_new (NULL, NULL, NULL, NULL);
+  g_task_run_in_thread (task, background_monitor);
```
alexlarsson: I don't think you should use a GTask for this. That will constantly use one of the threads in the GTask threadpool. Just spawn a regular thread?

src/background.c
```diff
@@ -409,6 +474,372 @@ background_class_init (BackgroundClass *klass)
...
+}
+
+static void
+kill_app (const char *app_id)
+{
+  g_autoptr(FlatpakInstance) instance = NULL;
+
+  g_debug ("Killing app %s", app_id);
+
+  instance = find_instance (app_id);
```
alexlarsson: What if there are multiple instances? This seems to be a way to escape this. We should kill them all.

alexlarsson:
What about commandline apps. I wonder if there is a way to tie an app tty to a terminal and mark that as not backgrounded.

src/background.c
```diff
@@ -144,6 +145,60 @@ set_permission (const char *app_id,
...
+maybe_quote (const char *str)
+{
+  if (need_quotes (str))
+    return g_shell_quote (str);
+  return g_strdup (str);
+}
+
+static char **
+rewrite_commandline (const char *app_id,
+                     const char * const *commandline)
```
alexlarsson: Why is this quoting the commandline, that doesn't seem right. If we pass the commandline to the backend as an argv there is no need to escape anything. Any escaping should be done by the backend.

src/background.c
```diff
@@ -222,12 +288,30 @@ handle_request_background_in_thread_func (GTask *task,
   else
     allowed = permission == YES ? TRUE : FALSE;
 
+  g_debug ("Setting autostart for %s to %s", app_id,
+           allowed && autostart_requested ? "enabled" : "disabled");
+
+  commandline = rewrite_commandline (app_id, autostart_exec);
```
alexlarsson: We should not rewrite if xdp_app_info_is_host(). In fact, things are probably nicer if we added a xdg_app_info_rewrite_commandline() function. Then we can plug in e.g. snap support later.

src/background.c
matthiasclasen: The portal we use here is just asking a deny/grant question, atm. And the designers don't like the 'always ask' much. They prefer to have permissions that time out.

src/background.c
matthiasclasen: I'll leave this to a future cleanup to have better permission plumbing, across all portals

data/org.freedesktop.impl.portal.Background.xml
matthiasclasen: Needs some cleanup for the application tracking first, to avoid notifying multiple times

hadess:
I haven't looked at the code more, but one thing that came to mind:
- does the portal remove the autostart symlink when the permission store is updated another way (through the Settings for example)?
- what happens when if the portal isn't running at that time? (eg. do we need to modify the Settings so it goes through the portal to modify that, or at least ensure that it's running)

Would it be easier to implement parts of this at the Flatpak level, with the autostart desktop files always being installed but with the .desktop files rewritten so that the background bit only ever get started when the permissions allow it?

We might still need to figure out how the session is supposed to know about all this, to shut down background programs that got the permission removed. Maybe @benzea has an idea?

matthiasclasen:
> I haven't looked at the code more, but one thing that came to mind:
> 
>     * does the portal remove the autostart symlink when the permission store is updated another way (through the Settings for example)?
>     * what happens when if the portal isn't running at that time? (eg. do we need to modify the Settings so it goes through the portal to modify that, or at least ensure that it's running)

Currently not. But autostart is not a permission either. Just a file we put in place or remove. If you get autostarted and run in the background without permission for that, you will get killed by the monitor after a little while.

> 
> Would it be easier to implement parts of this at the Flatpak level, with the autostart desktop files always being installed but with the .desktop files rewritten so that the background bit only ever get started when the permissions allow it?

We decided to keep flatpak out of it. autostart files are no longer installed by the app, they are created by the portal, and we control what goes in them (ie no conditions, etc). This will make it easier to switch to systemd units later, if we choose to do so.

> We might still need to figure out how the session is supposed to know about all this, to shut down background programs that got the permission removed. Maybe @benzea has an idea?

The portal has a monitor that kills flatpaks that are running in the background without the permission.

matthiasclasen:
> What about commandline apps. I wonder if there is a way to tie an app tty to a terminal and mark that as not backgrounded.

Haven't thought about that. For now, just giving them the permission to run in the background seems easy enough

matthiasclasen:
I'm happy with this now.

src/background.c
```diff
@@ -679,69 +643,92 @@ check_background_apps (void)
...
 }
 
 static gpointer
 background_monitor (gpointer data)
 {
+  applications = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                        g_free, instance_data_free);
   while (1)
     {
       check_background_apps ();
```
alexlarsson: You could pass applications here instead of relying on a global maybe?

src/background.c
```diff
@@ -0,0 +1,817 @@
...
+
+static gpointer
+background_monitor (gpointer data)
+{
+  applications = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                        g_free, instance_data_free);
+  while (1)
+    {
+      check_background_apps ();
+      sleep (20);
```
alexlarsson: The docs say each minute, but this is 20 sec. I think every minute is a better value, as there is no need for this to be polling so often.

src/background.c
```diff
@@ -0,0 +1,817 @@
...
+typedef enum { UNSET, NO, YES, ASK } Permission;
+
+static GVariant *
+get_all_permissions (void)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) out_perms = NULL;
+  g_autoptr(GVariant) out_data = NULL;
+
+  if (!xdp_impl_permission_store_call_lookup_sync (get_permission_store (),
```
alexlarsson: It would be nice if this used a utility that used notifications to get a constantly up-to-date GVariant for a permission query  without having to actually query it all the time. However, that can be done separately.

src/background.c
```diff
@@ -0,0 +1,817 @@
...
+  REQUEST_AUTOLOCK (request);
+
+  options = (GVariant *)g_object_get_data (G_OBJECT (request), "options");
+  g_variant_lookup (options, "reason", "&s", &reason);
+  g_variant_lookup (options, "autostart", "b", &autostart_requested);
+  g_variant_lookup (options, "commandline", "^a&s", &autostart_exec);
+  g_variant_lookup (options, "dbus-activatable", "b", &activatable);
+
+  if (activatable)
+    autostart_flags |= AUTOSTART_FLAGS_ACTIVATABLE;
```
alexlarsson: Should we fail on unknown flags?

src/background.c
```diff
@@ -0,0 +1,817 @@
...
+    {
+      g_debug ("Allow this instance of %s to run in background without permission changes", ddata->app_id);
+    }
+  else
+    g_debug ("Unexpected response from NotifyBackground: %u", result);
+
+  if (ddata->perm != UNSET)
+    set_permission (ddata->app_id, ddata->perm);
+
+  idata = g_hash_table_lookup (applications, ddata->id);
```
alexlarsson: This is accessing the applications hashtable on the main thread, so it will need some locking!

src/background.c
```diff
@@ -0,0 +1,817 @@
...
+        ddata->perm = YES;
+    }
+  else if (result == FORBID)
+    {
+      g_debug ("Forbid app %s to run in background", ddata->app_id);
+
+      if (ddata->perm != ASK)
+        ddata->perm = NO;
+
+      kill_instance (ddata->id);
```
alexlarsson: kill_instance() will access the applications hashtable, and this is the main thread, so need locking here.

alexlarsson:
I think this looks generally good to go, but the thread issues have to be fixed.

src/background.c
matthiasclasen: we validate options when we receive them from the caller. These here are already validated.

src/background.c
matthiasclasen: agreed

src/background.c
matthiasclasen: I brought the locking back

src/background.c
matthiasclasen: with the locking back, it is now more clearly a global

src/background.c
```diff
@@ -0,0 +1,857 @@
...
+      if (permission == UNSET)
+        set_permission (app_id, allowed ? YES : NO);
+    }
+  else
+    allowed = permission == YES ? TRUE : FALSE;
+
+  g_debug ("Setting autostart for %s to %s", app_id,
+           allowed && autostart_requested ? "enabled" : "disabled");
+
+  commandline = xdp_app_info_rewrite_commandline (request->app_info, autostart_exec);
```
alexlarsson: This should check for NULL (not supported)

alexlarsson:
Otherwise looks good to me!

PR #320, opened 2019-05-15
Add an updates portal

matthiasclasen:
This portal lets apps monitor for updates and update themselves.

### DIFF
--- a/data/org.freedesktop.portal.Updates.xml
+++ b/data/org.freedesktop.portal.Updates.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2019 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Matthias Clasen <mclasen@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <interface name="org.freedesktop.portal.Updates">
+    <!--
+        InstallUpdate:
+        @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Asks to install an update of the calling app.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>handle_token s</term>
+            <listitem><para>
+              A string that will be used as the last element of the @handle. Must be a valid
+              object path element. See the #org.freedesktop.portal.Request documentation for
+              more information about the @handle.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+      The returned object implements the org.freedesktop.portal.Request and the org.freedesktop.portal.Updates.Update interfaces and will
+      emit #org.freedesktop.portal.Updates.Update::Progress and
+      #org.freedesktop.portal.Request::Response signals.
+    -->
+    <method name="InstallUpdate">
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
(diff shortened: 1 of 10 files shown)


data/org.freedesktop.portal.Updates.xml
```diff
@@ -0,0 +1,78 @@
...
+      emit #org.freedesktop.portal.Updates.Update::Progress and
+      #org.freedesktop.portal.Request::Response signals.
+    -->
+    <method name="InstallUpdate">
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <method name="CreateUpdateMonitor">
```
alexlarsson: No docs?

src/xdp-utils.c
```diff
@@ -193,6 +193,44 @@ xdp_app_info_get_id (XdpAppInfo *app_info)
   return app_info->id;
 }
 
+char *
+xdp_app_info_get_ref (XdpAppInfo *app_info)
```
alexlarsson: get_flatpak_ref() (since this is flatpak specific)

src/xdp-utils.c
```diff
@@ -193,6 +193,44 @@ xdp_app_info_get_id (XdpAppInfo *app_info)
...
+  char *p;
+
+  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+    return NULL;
+
+  path = g_key_file_get_string (app_info->u.flatpak.keyfile,
+                                FLATPAK_METADATA_GROUP_INSTANCE,
+                                FLATPAK_METADATA_KEY_APP_PATH, NULL);
+
+  p = strstr (path, "/app/");
```
alexlarsson: This seems a bit non-robust? Some prefix could randomly be "app". Can't you just go dirname() 5 times (or reverse search for 5 slashes), that should work too.

src/updates.c
```diff
@@ -0,0 +1,706 @@
...
+        }
+    }
+
+  return G_SOURCE_CONTINUE;
+}
+
+static void
+start_monitoring (UpdatesSession *session)
+{
+  check_for_updates (session);
```
alexlarsson: Maybe we could have a shared poll, rather than one per monitor? Because the poll is going to do some network i/o.

src/updates.c
```diff
@@ -0,0 +1,706 @@
...
+check_for_updates (gpointer data)
+{
+  Session *session = data;
+  UpdatesSession *usession = data;
+  g_autoptr(GPtrArray) updates = NULL;
+  int i;
+
+  g_debug ("Checking for updates for %s", usession->ref);
+
+  updates = flatpak_installation_list_installed_refs_for_update (usession->installation, NULL, NULL);
```
alexlarsson: This is a bit overkill as it will scan all remotes and stuff. I think what you want is:

installed_ref = flatpak_installation_get_installed_ref(installation, ...);
remote_ref = (flatpak_installation_fetch_remote_ref_sync(installation, flatpak_installed_ref_get_origin (installed_ref), ...);
should_update = g_strcmp0 (flatpak_ref_get_commit (installed_ref, remote_ref));

In fact, I'm not quite sure this is right either. flatpak_installation_get_installed_ref() will get you the latest *installed* version. But maybe you want to look at the app-commit= line in the .flatpak-info (i.e. the actual running version)? Its a bit weird, as it doesn't fit the notion of "update", but maybe the API should do:
 * If there is a newer version remote than what is installed, install that
 * If there is a newer version installed than what is running, pretend to install that and immediately succeed.
 * Otherwise do nothing.

Its a bit unclear really what is the expected behaviour here if we start racing with e.g. gnome-software doing an update in the background.

alexlarsson: Maybe we should add a flag to the signal actually, whether it needs installing or if its already installed.

src/updates.c
```diff
@@ -0,0 +1,706 @@
...
+    goto out;
+
+  g_debug ("Installing update for %s", app_id);
+
+#if 1
+  ref = xdp_app_info_get_ref (request->app_info);
+  installation = xdp_app_info_get_installation (request->app_info);
+#else
+  ref = g_strdup ("app/org.gnome.Documents/x86_64/master");
+  installation = flatpak_installation_new_user (NULL, &error);
```
alexlarsson: debug leftovers?

src/updates.c
```diff
@@ -0,0 +1,706 @@
...
+  guint response = 2;
+
+  REQUEST_AUTOLOCK (request);
+
+  d = g_new0 (UpdatesData, 1);
+  d->request = g_object_ref (request);
+
+  g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
+
+  app_id = xdp_app_info_get_id (request->app_info);
```
alexlarsson: This should verify that the app is sandboxed, and is actually a flatpak app.

matthiasclasen:
from irc discussion: This may be better off in flatpak, to keep direct flatpak dependencies out of the xdg-desktop-portal code that is shared with snap

matthiasclasen:
Moved to flatpak: https://github.com/flatpak/flatpak/pull/2905

PR #321, opened 2019-05-15
settings: Don't use G_VARIANT_BUILDER_INIT

alexlarsson:
This is glib 2.50 or higher, which doesn't exist on xenial.


alexlarsson:
Had this commit lying around from the ubuntu packages.

PR #328, opened 2019-05-20
configure.ac: Make Flatpak interface directory configurable

tdz:
The current build scripts look in $datadir for Flatpak interface
descriptions. This does not work if these files are installed under
/usr/share (the default) and the package's $datadir is located under
/usr/local/share (also the default). With the patch, the Flatpak
directory is either autodetected from pkg-config or set on configure's
command line.

Signed-off-by: Thomas Zimmermann <tzimmermann@suse.de>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -47,6 +47,12 @@ AC_ARG_WITH(dbus_service_dir,
 DBUS_SERVICE_DIR=$with_dbus_service_dir
 AC_SUBST(DBUS_SERVICE_DIR)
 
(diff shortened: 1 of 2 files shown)


matthiasclasen:
No need for an option. Picking it up from the pkgconfig file should be sufficient

tdz:
I just saw that you merged the PR. Sorry for not getting back to you earlier to address the comment. I've been quite busy during the last days. Anyway, thanks for merging. I hope you can still make use of the change.

matthiasclasen:
sure, thanks!

PR #331, opened 2019-05-21
Camera portal

jadahl:
This adds a camera portal. It also includes fixes to the screen-cast portal, including one that will make it work with pipewire-0.2.6.

The portal works the following way:

There is a method `AccessCamera`, that will query the permission store and maybe the `Access` to query the user.

There is a method `OpenPipeWireRemote` that will open a pw_remote file descriptor. This method will only work if "yes" is already in the permission store. The PipeWire remote will be prepared with portal related metadata using `pipewire.access.portal.*` properties. All permissions will be set to `---` meaning no nodes will be available via the remote initially. PipeWire daemon side, there is a `module-portal` that then will handle keeping permissions up to date by querying and listening to the permission store.

In order to allow an application to only request camera permission would there be any cameras available, there is a `IsCameraPresent` boolean property that will expose whether there are any cameras available at all.

Note that pipewire-0.2.6 doesn't exist yet, but I bumped the required version anyway.

### DIFF
--- a/data/org.freedesktop.portal.Camera.xml
+++ b/data/org.freedesktop.portal.Camera.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2018 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.Camera:
+      @short_description: Camera portal
+
+      The camera portal enables applications to access camera devices, such as
+      web cams.
+  -->
+  <interface name="org.freedesktop.portal.Camera">
+    <!--
+        AccessCamera:
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>handle_token s</term>
+            <listitem><para>
+              A string that will be used as the last element of the @handle. Must be a valid
+              object path element. See the #org.freedesktop.portal.Request documentation for
+              more information about the @handle.
+            </para></listitem>
+          </varlistentry>
+
+        Following the #org.freedesktop.portal.Request::Response signal, if
+        granted, #org.freedesktop.org.Camera.OpenPipeWireRemote can be used to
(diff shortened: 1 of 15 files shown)


src/pipewire.c
```diff
@@ -166,7 +166,8 @@ ensure_pipewire_is_initialized (void)
 }
```
matthiasclasen: typo in the commit message: certian

It would be good if the commit message mentioned that this is new in pipewire 0.2.6. It is, right ?

src/camera.c
```diff
@@ -0,0 +1,244 @@
...
+  Request *request = (Request *)task_data;
+  const char *app_id;
+  const char *device;
+  gboolean allowed;
+  g_autoptr(GError) error = NULL;
+
+  REQUEST_AUTOLOCK (request);
+
+  app_id = (const char *)g_object_get_data (G_OBJECT (request), "app-id");
+  device = (const char *)g_object_get_data (G_OBJECT (request), "device");
```
matthiasclasen: Seems a little silly to funnel the string "camera" through object data when the function is explicitly called handle_access_camera_...

I would just hardcode "camera" here, in the device_query_permission call, like you do in handle_open_pipewire_remote

src/camera.c
```diff
@@ -0,0 +1,244 @@
...
+}
+
+const char *
+get_app_id (XdpAppInfo *app_info)
+{
+  const char *app_id;
+
+  app_id = xdp_app_info_get_id (app_info);
+  if (app_id[0] == '\0' && getenv("TEST_CAMERA"))
+    app_id = "org.gnome.Terminal";
```
matthiasclasen: leftover test code

src/camera.c
```diff
@@ -0,0 +1,244 @@
...
+  PipeWireRemote *remote;
+
+  app_info = xdp_invocation_lookup_app_info_sync (invocation, NULL, &error);
+  app_id = get_app_id (app_info);
+  permission = device_get_permission_sync (app_id, "camera");
+  if (permission != PERMISSION_YES)
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             G_DBUS_ERROR,
+                                             G_DBUS_ERROR_ACCESS_DENIED,
```
matthiasclasen: We have XDG_DESKTOP_PORTAL_ERROR and should probably use that here.

data/org.freedesktop.portal.Camera.xml
```diff
@@ -0,0 +1,72 @@
...
+        This method will only succeed if the application already has permission
+        to access camera devices.
+    -->
+    <method name="OpenPipeWireRemote">
+      <annotation name="org.gtk.GDBus.C.Name" value="open_pipewire_remote"/>
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="h" name="fd" direction="out"/>
+    </method>
+  </interface>
```
matthiasclasen: To follow our portal conventions, there should be a version property here.

jadahl:
Was talking to Wim about robustness, and we discovered that the retry-connect is not enough when systemd socket activation is enabled, as we'll never see the inotify events. To mitigate this, we'll add two things: 1) immediate reconnect when error is received, 2) limit number of reconnects within a period of time to avoid potential busy loop trying to reconnect but receiving an error during or after having connected

jadahl:
There is now a 0.2.6 release of PipeWire available.

dustymabe:
how do I use org.freedesktop.portal.Camera? do I need #83 first?

muelli:
yeah, it'd be nice to have some documentation or example of how to use this portal.

PR #332, opened 2019-05-24
Add support for GameMode 1.4 APIs

gicmo:
GameMode 1.4 added a set of new APIs that mirror the existing ones
but where once process can act on behalf of the another. The only
difference of the new function calls to the existing one is the
additional "requester" argument (a process id). In case of faltpak
this new pid argument also needs to be to be mapped from the pid
namespace of the sandbox to the pid namespace of the host.


gicmo:
I have updated [GameMode Tester](https://github.com/gicmo/gamemode-tester) to include direct GameMode calls so the new APIs (and error handling) can be manually tested.

PR #344, opened 2019-06-25
screen-cast: fix permission setup

### LINKED ISSUE #342: screen cast is broken with pipewire 0.2.6
Pipewire changed the permission handling in the latest release (Commit PipeWire/pipewire@83bc033837f7525d898f1de91119f669f9bf97f5).

Now the permissions of all parent nodes are checked as well. Because of this screen casting fails because the parent node of the node factory is not accessible.

See also: PipeWire/pipewire#158

michaelolbrich:
Without this, the permissions for the parent node of the node factory are
missing. Use append_parent_permissions() here as well to fix this.

Fixes #342

### DIFF
--- a/src/screen-cast.c
+++ b/src/screen-cast.c
@@ -591,6 +591,7 @@ open_pipewire_screen_cast_remote (const char *app_id,
   char *node_factory_permission_string;
   GList *string_stash = NULL;
   struct spa_dict *permission_dict;
+  PipeWireGlobal *node_global;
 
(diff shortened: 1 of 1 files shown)


jadahl:
lgtm, with nit: Commit message could be made more descriptive, e.g. `screen-cast: Set permissions of client node factory parents`

michaelolbrich:
better?

jadahl:
Yes! I'll wait for @matthiasclasen to push the merge button.

PR #347, opened 2019-06-28
utils: use namespace info from bwrap, if available

gicmo:
Newer versions of bubblewrap write the namespace information to the
bwrapinfo.json. Try reading that information and thus avoiding the
extra work to look up the pid namespace information via /proc.
Fallback to this old mechanism if the bwrapinfo.json does not contain
the information.

This needs projectatomic/bubblewrap#323

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -1546,18 +1546,16 @@ map_pids (DIR     *proc,
   return TRUE;
 }
 
-static pid_t
-xdp_app_info_get_child_pid (XdpAppInfo *app_info,
-                            GError    **error)
+static JsonNode *
+xdg_app_info_load_brwap_info (XdpAppInfo *app_info,
+                              GError    **error)
 {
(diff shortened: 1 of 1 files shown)


src/xdp-utils.c
```diff
@@ -1625,7 +1657,22 @@ xdg_app_info_ensure_pidns (XdpAppInfo  *app_info,
...
     return TRUE;
 
-  pid = xdp_app_info_get_child_pid (app_info, error);
+  root = xdg_app_info_load_brwap_info (app_info, error);
+  if (root == NULL)
+    return FALSE;
+
+  /* newer versions of bubblewrap contain the namespace
+   * information directly, so we don' thave to go via the
+   * child-pid; if this fails, we fallback to the old way */
```
matthiasclasen: Pet peeve of mine: I *really* prefer the closing */ on the next line, with the * lining up to the rest

PR #348, opened 2019-07-03
utils: Recognize applications in Toolbox containers from the host

debarshiray:
Toolbox [1] containers are development environments whose primary goal
isn't secure containment of processes, but to offer a safe haven for
development without touching the host operating system. This is
particularly useful on OSTree based systems like Fedora Silverblue,
where it's significantly difficult to set up a development prefix
directly on the host OS. These containers are rootless OCI containers,
not Flatpak containers, and are designed to blur the divide between the
host and the container as much as possible - full access to $HOME, the
D-Bus sockets, the Wayland and X11 sockets, host network access, no
PID namespace, etc..

Therefore, conceptually, running an application inside a Toolbox
container is the same thing as running it on the host.

Identifying the runtime environment for applications starts with
opening /proc/<app-pid>/root. For an application running inside a
toolbox and a portal on the host, the PID is visible to the portal, but
it fails to open it because /proc/<app-pid>/root points to the
container's root directory, not the host's. The openat(2) call fails
with EACCES.

So far, any failure to open /proc/<app-pid>/root was treated as an
indication that the application has died. This specific error
condition is now special-cased to denote applications running on the
host.

[1] https://github.com/debarshiray/toolbox

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -30,6 +30,7 @@
 #include <stdio.h>
 #include <mntent.h>
 #include <unistd.h>
+#include <sys/vfs.h>
 
 #include <gio/gdesktopappinfo.h>
 
@@ -382,9 +383,16 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
     {
       if (errno == EACCES)
         {
-          /* Access to the root dir isn't allowed => inside a Toolbox
-             container, return NULL with no error */
-          return NULL;
+          struct statfs buf;
+
+          /* Access to the root dir isn't allowed. This can happen if the root is on a fuse
+           * filesystem, such as in a toolbox container. We will never have a fuse rootfs
+           * in the flatpak case, so in that case its safe to ignore this and
+           * continue to detect other types of apps.
+           */
+          if (statfs (root_path, &buf) == 0 &&
+              buf.f_type == 0x65735546) /* FUSE_SUPER_MAGIC */
+            return NULL;
         }
 
       /* Otherwise, we should be able to open the root dir. Probably the app died and
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -380,7 +380,14 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
   root_fd = openat (AT_FDCWD, root_path, O_RDONLY | O_NONBLOCK | O_DIRECTORY | O_CLOEXEC | O_NOCTTY);
   if (root_fd == -1)
     {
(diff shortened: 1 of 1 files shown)


alexlarsson:
Hmm, this seems somewhat risky. Are we sure the app can't somehow force this to return EACCESS, thus escaping the sandbox?

alexlarsson:
I'd like to know more about this exact situation. Why does stating the container root fs return EACCES? It seems like we *should* be able to read the root.

giuseppe:
I think this could be happening because `/` in the container is a FUSE process that is not accessible from the host mount namespace (even root cannot access it).  If you join the correct namespaces before, e.g. podman unshare, where the FUSE mount point is accessible, then you can open `/proc/PID/root`

cgwalters:
> I think this could be happening because / in the container is a FUSE process that is not accessible from the host mount namespace (even root cannot access it).

Right, that's exactly why.

>  If you join the correct namespaces before, e.g. podman unshare, where the FUSE mount point is accessible, then you can open /proc/PID/root

So the first thing we should do is `statfs` and check whether it's a FUSE filesystem; we shouldn't just ignore `EACCESS` I don't think.  The next question is - what do we do with this information?

I guess what things hinge on is the variant of https://github.com/flatpak/xdg-desktop-portal/pull/348#issuecomment-521975644 for this - can a confined app make a FUSE mount in a way that would trick us?

alexlarsson:
@cgwalters If stat() fails, are we sure statfs works? Anyway, we rely on sandboxed apps not being able to tweak their namespace setup for other things (like the /.flatpak-info file) so if we *know* its a fuse rootfs I think we can rely on it not being a flatpak.

cgwalters:
> @cgwalters If stat() fails, are we sure statfs works? 

Yes (you can test this as easily as I can):

```
# id -u
0
# stat /run/user/1000/gvfs
stat: cannot stat '/run/user/1000/gvfs': Permission denied
# stat -f /run/user/1000/gvfs
  File: "/run/user/1000/gvfs"
    ID: 0        Namelen: 0       Type: fuseblk
Block size: 0          Fundamental block size: 0
Blocks: Total: 0          Free: 0          Available: 0
Inodes: Total: 0          Free: 0
```

alexlarsson:
Ok. I added a check that on EACCES that the root is a fuse, and if so we can be pretty sure that whatever it is it is not a flatpak, so we can continure to other forms of detection (ending up at host for toolbox).

debarshiray:
Thanks for taking care of this, @alexlarsson !

PR #353, opened 2019-07-19
Fix leak of .flatpak-info fd

### LINKED ISSUE #352: Crash due to fd leak
See https://bugzilla.redhat.com/show_bug.cgi?id=1730327. xdg-desktop-portal has opened the ./flatpak-info file >1000 times and is unable to allocate more fds.

I think the problem is in parse_app_info_from_flatpak_info(), in xdp-utils.c.

mcatanzaro:
Fixes #352

Because xdg-desktop-portal is kinda hard to set up and run, this is a low-effort **untested** merge request. No warranty. But it looks right. We might need to keep info_fd open until we're done with the GMappedFile that's using it -- not sure -- but at this point we are already done with it, so should be OK.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -441,7 +441,12 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
 
   id = g_key_file_get_string (metadata, group, "name", error);
   if (id == NULL)
-    return NULL;
+    {
+      close (info_fd);
+      return NULL;
+    }
+
+  close (info_fd);
 
   app_info = xdp_app_info_new (XDP_APP_INFO_KIND_FLATPAK);
   app_info->id = g_steal_pointer (&id);


src/xdp-utils.c
```diff
@@ -440,8 +440,12 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
     group = "Runtime";
 
   id = g_key_file_get_string (metadata, group, "name", error);
-  if (id == NULL)
+  if (id == NULL) {
```
matthiasclasen: { on the next line here, please.

matthiasclasen:
Looks correct, other than that.

src/xdp-utils.c
mcatanzaro: Oops, sorry. Long day yesterday....

Issue #335, opened 2019-05-31
Portal for setting backgrounds


allanday:
Allowing a user to set a background on the system is a fairly common feature in apps - it would be good to provide a portal for this.

Design considerations:

 - Size requirements for the image
   - Handling of images that do not fit the aspect of ratio of the screen. On the GNOME side, we've tried to avoid the detail of how an image is fit to the display, in an effort to reduce the amount of work for the user and make things Just Work. However, in some cases an image might be used where it is difficult to automatically do the right thing. For example, fitting a portrait orientation image to a landscape display.
 - Format requirements for the image - I think we should be fairly restrictive here, and maybe limit it to PNGs and JPGs.
 - In the GNOME case, a user can set a desktop and/or a lock screen background (this could potentially change in the future though)
 - It would be good if the portal could store a copy of the image, so we're not dependent on the original file being available

From a UX perspective it is probably good to provide a preview of what the background image will look like in use.

allanday:
A quick sketch of what the portal UI could look like in a GNOME context:

![image](https://user-images.githubusercontent.com/213068/58701587-41e1d700-839b-11e9-91dd-642f91c07d60.png)

This doesn't cover cases where the image size or format are problematic.

felipeborges:
I started implementing this some time ago, but without a UI, so I guess I could propose an implementation for this.

@allanday would you also like to expose wallpaper options such as tiling or zoom? This currently can be set in GNOME Tweaks.

Also, from an implementation point of view, where would be the best place to store the image? XDG_CONFIG_DIRS/wallpaper.png ?

I wonder whether KDE has requirements/specifics that they would like to have in the portal. cc @grulja 

p.s.: We should probably use the word **Wallpaper** from now on when it comes to this portal, so it is not confused with the Backgrounds portal that handles the "Run in the background" feature for apps.

grulja:
I don't think that from KDE side there is a need for this portal or specific requirement, because the only way how to change wallpaper is through Plasma configuration (when you click with the mouse right button on a desktop). Given the Plasma config file is quite complicated and the wallpaper is basically a plugin, then I don't think any app would support changing Plasma wallpaper.

grulja:
Ok, I was wrong, there is a way to change wallpaper through Plasma framework, but I don't think I have seen this implemented. Anyway, I assume the requirements will be most likely same as for Gnome in this case.

allanday:
[...]

PR #354, opened 2019-07-22
Add Wallpaper portal

felipeborges:
The implementation is at https://github.com/flatpak/xdg-desktop-portal-gtk/pull/216

It basically has two similar APIs:

**SetWallpaper** and **ShowWallpaperPreview**.

SetWallpaper requires permissions and enables apps to set a wallpaper without any UI (after its been authorized).

ShowWallpaperPreview doesn't interact with the permission store because we assume the user is indicating consent when clicking in the dialog.

Design discussions were held in https://github.com/flatpak/xdg-desktop-portal/issues/335

PS.: Many of the UI labels refer to Wallpapers as Backgrounds. "Backgrounds" is the adopted terminology in GNOME but I went for "Wallpaper" for the portal code part to avoid confusion with the "Backgrounds" portal/monitor that does something else.

### DIFF
--- a/src/wallpaper.c
+++ b/src/wallpaper.c
@@ -1,0 +1,267 @@
+/*
+ * Copyright © 2019 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Felipe Borges <feborges@redhat.com>
+ */
+
+#include "config.h"
+
+#include <string.h>
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+
+#include "wallpaper.h"
+#include "request.h"
+#include "xdp-dbus.h"
+#include "xdp-impl-dbus.h"
+#include "xdp-utils.h"
+
+typedef struct _Wallpaper Wallpaper;
+typedef struct _WallpaperClass WallpaperClass;
+
+struct _Wallpaper
+{
+  XdpWallpaperSkeleton parent_instance;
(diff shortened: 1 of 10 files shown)


felipeborges:
To test the portal I wrote this little snippet https://gist.github.com/felipeborges/dafb921ead2dced34acf4094e0c277d4 (build instructions in the first line).

matthiasclasen:
Sorry it took me so long to look at this. I've finally trried it out.

I must say I'm not a fan of the super-wide dialog with double preview. A single preview would be enough, as far as I'm concerned. And I think the switches are awkward here.

I also don't like the ShowWallpaperPreview method name (since it is really about setting the wallpaper too), and the fact that it is a separate method in the first place. I think we should just
have a single SetWallpaper method, and add options to it - 'show-preview' could be an option
to give a hint to the portal, but it should really be up to the portal whether to show a dialog or not.

matthiasclasen:
Options could also be used to let the application hint at whether it wants to set the lock-screen background or the desktop background, or both.

matthiasclasen:
The api I would suggest:

    
    <method name="SetWallpaperURI">
      <arg type="s" name="uri" direction="in"/>
    </method>

    <method name="SetWallpaperFile">
      <arg type="h" name="fd" direction="in"/>
    </method>

felipeborges:
@matthiasclasen thanks for your tips! I modified the implementation to carry your suggestions.

src/wallpaper.c
```diff
@@ -0,0 +1,429 @@
...
+                                                  G_DBUS_PROXY_FLAGS_NONE,
+                                                  g_dbus_proxy_get_name (G_DBUS_PROXY (impl)),
+                                                  request->id,
+                                                  NULL, &error);
+  request_set_impl_request (request, impl_request);
+
+  g_variant_builder_init (&opt_builder, G_VARIANT_TYPE_VARDICT);
+  xdp_filter_options (options, &opt_builder,
+                      wallpaper_options, G_N_ELEMENTS (wallpaper_options),
+                      NULL);
```
matthiasclasen: This will quietly drop the 'set-on' option if it is invalid. Does the backend handle its absence ok? What is the behavior? Might be good to document that.

matthiasclasen:
Apart from that one question about set-on, this looks good to me now.

matthiasclasen:
Oh, and the portal and impl xml need to be added to the docs too.

AsciiWolf:
_po/POTFILES.in_ needs to be updated to make the new strings available to translate. :-)

felipeborges:
> _po/POTFILES.in_ needs to be updated to make the new strings available to translate. :-)

Thanks for commit 7ddb3b8a6d067b21492b598e219f058b5dc18063

PR #359, opened 2019-08-19
Add a secret portal

### LINKED ISSUE #14: Add a portal for keyring access
Not clear what do to about kernel keyring, need to talk to domain experts. We believe that libsecret should use the kernel keyring underneath.

ueno:
This adds a portal interface to provide sandboxed applications with master secret to locally encrypt credentials.  It works as follows: the application first creates a pipe, sends the write end to the portal, and the portal sends back a secret value through the FD. That way, we can avoid any transport encryption.

If the master secret doesn't exist in keyring, the portal generates a random secret. This step needs to be done with a standard conforming crypto library as the secret is used for encryption. Therefore the backend implementation resides in gnome-keyring.

See also:
- https://gitlab.gnome.org/GNOME/libsecret/merge_requests/6
- https://gitlab.gnome.org/GNOME/gnome-keyring/merge_requests/18

Fixes #14

### DIFF
--- a/data/org.freedesktop.impl.portal.Secret.xml
+++ b/data/org.freedesktop.impl.portal.Secret.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2019 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Daiki Ueno <dueno@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.Secret:
+      @short_description: Secret portal backend interface
+
+      The Secret portal allows sandboxed applications to retrieve a
+      per-application master secret.
+  -->
+  <interface name="org.freedesktop.impl.portal.Secret">
+
+    <!--
+        RetrieveSecret:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @app_id: App id of the application
+        @fd: Writable file descriptor for transporting the secret
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Retrieves a master secret for a sandboxed application.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>token s</term>
+            <listitem><para>
+	      An opaque string associated with the retrieve secret.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="RetrieveSecret">
+      <annotation name="org.gtk.GDBus.C.Name" value="retrieve_secret"/>
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="h" name="fd" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In3" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+    <property name="version" type="u" access="read"/>
+  </interface>
+</node>
--- a/data/org.freedesktop.portal.Secret.xml
+++ b/data/org.freedesktop.portal.Secret.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2019 Red Hat, Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Daiki Ueno <dueno@redhat.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.Secret:
+      @short_description: Portal for retrieving application secret
+
+      The Secret portal allows sandboxed applications to retrieve a
+      per-application secret.  The secret can then be used for
+      encrypting confidential data inside the sandbox.
+
+      This documentation describes version 1 of this interface.
+  -->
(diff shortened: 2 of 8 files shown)


ueno:
Here is a quick demo https://asciinema.org/a/9gAd1DMQuucd3brCGp3SynBtj

data/org.freedesktop.impl.portal.Secret.xml
```diff
@@ -0,0 +1,66 @@
...
+
+      The Secret portal allows sandboxed applications to retrieve a
+      per-application master secret.
+  -->
+  <interface name="org.freedesktop.impl.portal.Secret">
+
+    <!--
+        RetrieveSecret:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @app_id: App id of the application
```
matthiasclasen: Why do we need to pass the app_id as an argument ? The portal should determine it from the sandbox, so it can be trusted.

data/org.freedesktop.impl.portal.Secret.xml
```diff
@@ -0,0 +1,66 @@
...
+        @fd: Writable file descriptor for transporting the secret
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Retrieves a master secret for a sandboxed application.
+
```
matthiasclasen: We should document some guarantees about the 'master secret' here:

- It is unique per application
- It does not change as long as the application is installed (once it has been created)
- It is stored in the users keyring, under the application ID

- What is the format of this secret ? It would be useful to know what data is actually returned here. Or is this always used via a library api that hides the details ?

data/org.freedesktop.portal.Secret.xml
```diff
@@ -0,0 +1,61 @@
...
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="RetrieveSecret">
+      <annotation name="org.gtk.GDBus.C.Name" value="retrieve_secret"/>
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="h" name="fd" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
```
matthiasclasen: Given that there is no user interaction here (we never show a dialog), this could probably be simplified by omitting the reponse object, and just making this a plain method call. The secret will be returned via the fd anyway.

data/org.freedesktop.portal.Secret.xml
```diff
@@ -0,0 +1,61 @@
...
+  <interface name="org.freedesktop.portal.Secret">
+
+    <!--
+        RetrieveSecret:
+        @fd: Writable file descriptor for transporting the secret
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Retrieves a master secret for a sandboxed application.
+
```
matthiasclasen: We should document some guarantees about the 'master secret' here:

- It is unique per application

- It does not change as long as the application is installed (once it has been created)

- It is stored in the users keyring, under the application ID

- What is the format of this secret ? It would be useful to know what data is actually returned here. Or is this always used via a library api that hides the details ?

data/org.freedesktop.portal.Secret.xml
ueno: There is actually an interaction: if the login keyring is locked, the dialog will be shown. However, the situation is exceptional and it might make more sense to to just return error in that case.

matthiasclasen: ah, ok. in that case, lets leave things as-is.

data/org.freedesktop.portal.Secret.xml
matthiasclasen: If libsecret has some docs about this, then just saying something like "in the format understood by libsecret" might be good enough

matthiasclasen:
What is WIP about this ? It looks basically good to merge, to me

data/org.freedesktop.portal.Secret.xml
TingPing: Also the options it accepts should be documented.

TingPing:
> If the master secret doesn't exist in keyring, the portal generates a random secret. This step needs to be done with a standard conforming crypto library as the secret is used for encryption. Therefore the backend implementation resides in gnome-keyring.

How unresonable would it be to de-couple this from gnome-keyring. I know it isn't a crazy dep but I personally don't want to have to tell all flatpak users it depends on `gnome-kerying` now.

matthiasclasen:
> > If the master secret doesn't exist in keyring, the portal generates a random secret. This step needs to be done with a standard conforming crypto library as the secret is used for encryption. Therefore the backend implementation resides in gnome-keyring.
> 
> How unresonable would it be to de-couple this from gnome-keyring. I know it isn't a crazy dep but I personally don't want to have to tell all flatpak users it depends on `gnome-kerying` now.

The gnome-keyring dependency is in the portal backend, though. Do you think that is problematic ?

matthiasclasen:
> > If the master secret doesn't exist in keyring, the portal generates a random secret. This step needs to be done with a standard conforming crypto library as the secret is used for encryption. Therefore the backend implementation resides in gnome-keyring.
> 
> How unresonable would it be to de-couple this from gnome-keyring. I know it isn't a crazy dep but I personally don't want to have to tell all flatpak users it depends on `gnome-kerying` now.

It is a somewhat soft dep though - we depend on a implementation for org.freedesktop.impl.portal.Secret to be present, and as things are now, the only one will be in gnome-keyring. It could be moved out to xdg-desktop-portal-gtk, if that is feasible.

But I don't think it is such a big problem. If the backend is not there, the secret portal interface will simply not be provided.

TingPing:
> But I don't think it is such a big problem. If the backend is not there, the secret portal interface will simply not be provided.

And every application depending on it won't work... Its a hard dep IMO. It has been how many years since the Secret service API was made and `gnome-keyring` is still the only shipped implementation AFAIK.

I don't think its a blocker to be merged but I think moving it somewhere else would be a win.

matthiasclasen:
> I don't think its a blocker to be merged but I think moving it somewhere else would be a win.

Yes. Daiki, would it be hard to separate the backend implementation from gnome-keyring, and move it to xdg-desktop-portal-gtk? It would of course still be the sole implementation and likely still depend on gnome-keying, so you trade a hard direct dependency on gnome-keyring with a hard indirect dependency....

ueno:
> > I don't think its a blocker to be merged but I think moving it somewhere else would be a win.
> 
> Yes. Daiki, would it be hard to separate the backend implementation from gnome-keyring, and move it to xdg-desktop-portal-gtk? It would of course still be the sole implementation and likely still depend on gnome-keying, so you trade a hard direct dependency on gnome-keyring with a hard indirect dependency....

What I'm most concerned is that that would bring in a direct dependency on a crypto library (OpenSSL, GnuTLS, libgcrypt; each has pros and cons) from xdg-desktop-portal-gtk to generate secure random bytes that can be used as encryption key.

And also, the generated key must be sent to gnome-keyring through [the custom wire encryption](https://specifications.freedesktop.org/secret-service/ch07.html), which I want to dismiss in the next few years.

TingPing:
> What I'm most concerned is that that would bring in a direct dependency on a crypto library (OpenSSL, GnuTLS, libgcrypt; each has pros and cons) from xdg-desktop-portal-gtk to generate secure random bytes that can be used as encryption key.

I don't think that is really a concern of `xdg-desktop-portal-gtk`. The license of `gnutls` isn't a problem and `glib-networking` in practice always uses `gnutls`.

If it always has to talk to `gnome-keyring` its a moot point anyway.

ueno:
> If it always has to talk to `gnome-keyring` its a moot point anyway.

I'm afraid that's the case, unless xdg-desktop-portal-gtk implements what gnome-keyring is doing to protect secrets on the host side (e.g., encrypting secrets with login password through pam).

TingPing:
I guess we'll just have to live with it then. Note to future @matthiasclasen, mention in release notes for packagers of the added runtime dep.

matthiasclasen:
Daiki, is there anything left here that is WIP ?

Is this ready to merge from your perspective ?

ueno:
I marked this as WIP because the other parts (libsecret and gnome-keyring) are still under review. If it is fine I'll update the doc and remove WIP.

matthiasclasen:
> I marked this as WIP because the other parts (libsecret and gnome-keyring) are still under review. If it is fine I'll update the doc and remove WIP.

Yes, please. I think this probably needs to land first, right ? Since both of the other changes will need the D-Bus interfaces defined here.

runiq:
> It has been how many years since the Secret service API was made and gnome-keyring is still the only shipped implementation AFAIK.

I'd like to point out that, with the recent KeePassXC release, the above is no longer true. KeePassXC 2.5 has [an implementation of the Secret Storage API](https://github.com/keepassxreboot/keepassxc/pull/2726).

PR #364, opened 2019-09-18
background: Don't spam the logs

matthiasclasen:
Only warn once if we don't find the shell api.

Closes: https://github.com/flatpak/xdg-desktop-portal-gtk/issues/215


dimyme:
it was a bit of a nuisance while it lasted…

PR #365, opened 2019-09-19
memory-monitor: Add portal implementation for GMemoryMonitor

hadess:


data/org.freedesktop.portal.MemoryMonitor.xml
```diff
@@ -0,0 +1,67 @@
...
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>2</term>
+            <listitem><para>
+              Same as "low" but the device has even less free memory, so processes should try harder to free up unneeded resources. If your process does not need to stay running, it is a good time for it to quit.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>3</term>
```
pwithnall: I think it would be a bit more future-proof to space these values out over (say) the range `[0, 255]`, then more values can be inserted in future while retaining the ability to do open-ended comparisons against existing values (`if (warning_level > LOW) drop_caches ();`).

matthiasclasen:
Looks fine to me. If you want to change the enum values as Phillip suggests, I'll merge it

hadess:
> Looks fine to me. If you want to change the enum values as Phillip suggests, I'll merge it

I've made those changes, but I still need to figure out the docs, and I'd rather hold off until I have some tests.

hadess:
Docs are done, we're waiting on GMemoryMonitor landing in GLib now.

matthiasclasen:
Looks like the ci failed. Can you fix that ?

hadess:
> Looks like the ci failed. Can you fix that ?

Still waiting on it landing in GLib... It's getting closer!

hadess:
(The CI passes because the memory monitor gets disabled with the too old GLib that's available, we still want GMemoryMonitor to land in GLib before merging this, to eventually adjust the version number needed)

hadess:
The version of glib that includes GMemoryMonitor will be 2.63.3 due out tomorrow, so I bumped the version number to a not yet existent version of glib.

PR #368, opened 2019-09-24
Implementation of org.freedesktop.portal.FileChooser::SaveFiles

### LINKED ISSUE #158: Cannot save multiple files via org.freedesktop.portal.FileChooser
It's not currently possible to save multiple files at once from within a sandbox using the org.freedesktop.portal.FileChooser portal. This manifests as [GtkFileChooserNativePortal not implementing SELECT_FOLDER](https://bugzilla.gnome.org/show_bug.cgi?id=785884) and silently falling back to a non-native chooser when an application attempts to use it, which works fine if you have RW permissions for $HOME, but not when sandboxed.

Pitivi breaks out of the sandbox to work around this, but it has also broken Geary's Save All Attachments feature, as [reported by a user](https://github.com/flathub/org.gnome.Geary/issues/5) against Flathub, and I'm reluctant to take Pitivi's approach there.

This could maybe be implemented by adding a org.freedesktop.portal.FileChooser SaveFiles method that expects to be passed a list of file names to be saved, and with SaveFilesResponse returning absolute URIs for each file. A select-folder file chooser could then be presented to the user, which can be used to generate the absolute URIs based on the selected folder and the given filenames. This would need some GTK+ API additions to implement, however.

mjog:
Adds definition and implementation of org.freedesktop.portal.FileChooser::SaveFiles

The general idea here is to enable applications saving multiple files at once. This is typically done by opening a Save file chooser in "select a folder" mode, but that's not possible to implement securely for sandboxed applications. To address this, the `SaveFiles` method allows the application to specify all of the file names that need to be saved, then forwards these to the implementation portal to have it open a Save file chooser in folder mode. When a folder is selected, the portal sends back a list of absolute URIs being the folder selected joined with the names of each of the given files. The application then is returned a list of files it can write to.

Fixes #158

### DIFF
--- a/data/org.freedesktop.impl.portal.FileChooser.xml
+++ b/data/org.freedesktop.impl.portal.FileChooser.xml
@@ -211,5 +211,101 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+    <!--
+      SaveFiles:
+      @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
+      @title: Title for the file chooser dialog
+      @options: Vardict with optional further information
+      @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+      Asks for a folder as a location to save one or more files. The
+      names of the files will be used as-is and appended to the
+      selected folder's path in the list of returned files. If the
+      selected folder already contains a file with one of the given
+      names, the portal may prompt or take some other action to
+      construct a unique file name and return that instead.
+
+      Supported keys in the @options vardict include:
+      <variablelist>
+        <varlistentry>
+          <term>handle_token s</term>
+          <listitem><para>
+            A string that will be used as the last element of the
+            @handle. Must be a valid object path element. See the
+            #org.freedesktop.portal.Request documentation for more
+            information about the @handle.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>accept_label s</term>
+          <listitem><para>
+            Label for the accept button. Mnemonic underlines are allowed.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>modal b</term>
+          <listitem><para>
+            Whether the dialog should be modal. Default is yes.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>choices a(ssa(ss)s)</term>
+          <listitem><para>
+            List of serialized combo boxes.
+            See org.freedesktop.portal.FileChooser.OpenFile() for details.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>current_folder ay</term>
+          <listitem>
+            <para>
+              Suggested folder to save the files in. The byte array is
+              expected to be null-terminated.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+        <term>files aay</term>
+          <listitem>
+            <para>
+              An array of file names to be saved. The array and byte
+              arrays are expected to be null-terminated.
+            </para>
+          </listitem>
+        </varlistentry>
+      </variablelist>
+
+      The following results get returned via the
+      #org.freedesktop.portal.Request::Response signal:
+      <variablelist>
+        <varlistentry>
+          <term>uris as</term>
+          <listitem><para>
+            An array of strings containing the uri corresponding to
+            each file given by @options, in the same order. Note that
+            the file names may have changed, for example if a file
+            with the same name in the selected folder already exists.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>choices a(ss)</term>
+          <listitem><para>
+            An array of pairs of strings, corresponding to the passed-in choices.
+            See org.freedesktop.portal.FileChooser.OpenFile() for details.
+          </para></listitem>
+        </varlistentry>
+      </variablelist>
+    -->
+    <method name="SaveFiles">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="s" name="title" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In4" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
   </interface>
 </node>
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -250,6 +250,97 @@
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="handle" direction="out"/>
     </method>
+    <!--
+      SaveFiles:
+      @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
+      @title: Title for the file chooser dialog
+      @options: Vardict with optional further information
+      @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+      Asks for a folder as a location to save one or more files. The
+      names of the files will be used as-is and appended to the
+      selected folder's path in the list of returned files. If the
+      selected folder already contains a file with one of the given
+      names, the portal may prompt or take some other action to
+      construct a unique file name and return that instead.
+
+      Supported keys in the @options vardict include:
+      <variablelist>
+        <varlistentry>
+          <term>handle_token s</term>
+          <listitem><para>
+            A string that will be used as the last element of the
+            @handle. Must be a valid object path element. See the
+            #org.freedesktop.portal.Request documentation for more
+            information about the @handle.
+          </para></listitem>
+        </varlistentry>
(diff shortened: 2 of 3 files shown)


mjog:
This is currently untested, but AFAICT code-complete. Am mostly looking for feedback on the additional API at the moment, but code review also welcome since I'm not a native C speaker.

src/file-chooser.c
```diff
@@ -603,11 +603,115 @@ handle_save_file (XdpFileChooser *object,
   return TRUE;
 }
 
+static XdpOptionKey save_files_options[] = {
+  { "accept_label", G_VARIANT_TYPE_STRING, NULL },
+  { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "current_name", G_VARIANT_TYPE_STRING, NULL },
+  { "current_folder", G_VARIANT_TYPE_BYTESTRING, NULL },
+  { "files", G_VARIANT_TYPE_BYTESTRING, NULL },
+  { "choices", (const GVariantType *)"a(ssa(ss)s)", validate_choices  }
```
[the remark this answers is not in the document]
mjog: That fails with a "error: initializer element is not constant", which is probably why the other instantiations of that struct are doing the same thing.

mjog:
GTK implementation going up over at https://github.com/flatpak/xdg-desktop-portal-gtk/pull/236

mjog:
Who is a good person to ping RE a review from the KDE perspective?

mjog:
@grulja maybe? :)

matthiasclasen:
I'm not opposed to adding something like this. But the bulk of the work in this case will be behind the backend. Neither GTK nor Qt currently have a save-files dialog that could be used for this.

mjog:
Thanks for the feedback!

> I'm not opposed to adding something like this. But the bulk of the work in this case will be behind the backend. Neither GTK nor Qt currently have a save-files dialog that could be used for this.

I mentioned over in #158 what I thought the GTK API addition might look like, just a single new ctor for the GtkFileChooserNative class that accepts the list of files to be saved.

> First step here would probably be some design input on how such a dialog would look, e.g for the case of multiple attachments - will it just be a directory chooser, or do we allow saving the files in different directories ? does that need to be an application choice ? do we allow changing the names ?

Per the bug above, Geary's use case is to save all selected attachments in a single directory. It currently attempts to show a directory chooser, then starts saving individual files in the selected directory. If there's a name clash it currently prompts about overwriting and stops if that is not confirmed (which is pretty suboptimal). I've pinged the Pitivi people who were having the same issue in the bug above already to ask them for some feedback as well.

The name clash is an obvious wrinkle in this approach that I hadn't put any thought into yet. Initially the portal could just generate a unique name when there is a collision and send that back to the application instead. I'd certainly prefer that to Geary's current approach of prompting and falling over. A more robust approach would be to display something like the same UI nautilus does when there is a name collision, and update return URIs according to the choice made (rename, skip, etc). That's probably out of scope of anything I'm going to be able to implement this cycle, however.

What's the use case for saving files in different dirs that wouldn't be covered by prompting for each directory separately? Can't the application can already change the file names up front by just sending a different name to the portal?

matthiasclasen:
>
> I mentioned over in #158 what I thought the GTK API addition might look like, just a single new ctor for the GtkFileChooserNative class that accepts the list of files to be saved.
> 

GtkFileChooserNative is just the tip of the iceberg. You have to add support for this to all the file chooser implementations. I think adding this as a portal method is good, but I'm not in a rush to plumb this all the way through GTK

mjog:
Okay, not problem. Will finish off the implementation over at xdg-desktop-portal-gtk and test it out in a week or two.

matthiasclasen:
Thanks, I'll take a look and try it out, this week

matthiasclasen:
I've finally tried this. While it technically works (modulo a few small fixes), I think at the very least, we need to
- show the filenames to save in the file chooser dialog
- handle name conflicts somehow

data/org.freedesktop.impl.portal.FileChooser.xml
```diff
@@ -211,5 +211,95 @@
...
+
+      The following results get returned via the
+      #org.freedesktop.portal.Request::Response signal:
+      <variablelist>
+        <varlistentry>
+          <term>uris as</term>
+          <listitem><para>
+            An array of strings containing the uri of each given
+            file. The path name of each will have a name matching the
+            given list of files.
```
matthiasclasen: Should probably spell out here that the returned uris are in 1-1 correspondence to the passed-in files

data/org.freedesktop.portal.FileChooser.xml
```diff
@@ -250,6 +250,91 @@
...
+      The following results get returned via the
+      #org.freedesktop.portal.Request::Response signal:
+      <variablelist>
+        <varlistentry>
+          <term>uris as</term>
+          <listitem><para>
+            An array of strings containing the uri of each given
+            file. The path name of each will have a name matching the
+            given list of files.
+          </para></listitem>
```
matthiasclasen: And here too

src/file-chooser.c
```diff
@@ -603,11 +603,115 @@ handle_save_file (XdpFileChooser *object,
   return TRUE;
 }
 
+static XdpOptionKey save_files_options[] = {
+  { "accept_label", G_VARIANT_TYPE_STRING, NULL },
+  { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "current_name", G_VARIANT_TYPE_STRING, NULL },
+  { "current_folder", G_VARIANT_TYPE_BYTESTRING, NULL },
+  { "files", G_VARIANT_TYPE_BYTESTRING, NULL },
```
matthiasclasen: This needs to be G_VARIANT_TYPE_BYTESTRING_ARRAY

data/org.freedesktop.impl.portal.FileChooser.xml
mjog: I've just pushed update that include improved doc comments. What do you think?

data/org.freedesktop.portal.FileChooser.xml
mjog: Ditto.

src/file-chooser.c
mjog: Fixed. Thanks!

mjog:
Thanks for the review!

> I've finally tried this. While it technically works (modulo a few small fixes), I think at the very least, we need to
> 
> * show the filenames to save in the file chooser dialog

I'm not sure how to do this in a scalable way (what if there are 10's of files to save?). I'm not convinced it's necessary, since both the application can (should?) indicate this before Save All gets clicked (it's obvious in Geary, at least), and no data loss can occur since the updated GTK portal impl (see below) now ensures the file names handed back to the app don't exist (at selection time, anyway I guess it's a bit racy as-is?)

> * handle name conflicts somehow

I've just updated the related -gtk PR over at https://github.com/flatpak/xdg-desktop-portal-gtk/pull/236 to add `(n)` for 1..n to the filename in the same way Nautilus does to ensure it is unique. Is that sufficient for now?

I've also got a PR for libportal that I'll push next, between these three PRs I've tested it out and seems to work fine, so barring any other review comments I consider them ready to be merged.

mjog:
The libportal MR is here: https://github.com/flatpak/libportal/pull/4

matthiasclasen:
> Thanks for the review!
> 
> > I've finally tried this. While it technically works (modulo a few small fixes), I think at the very least, we need to
> > 
> > * show the filenames to save in the file chooser dialog
> 
> I'm not sure how to do this in a scalable way (what if there are 10's of files to save?). I'm also not convinced it's necessary, since both the application can (should?) indicate this before Save All gets clicked (it's obvious in Geary, at least), and no data loss can occur since the updated GTK portal impl (see below) now ensures the file names handed back to the app don't exist (at selection time, anyway I guess it's a bit racy as-is?)
> 
> > * handle name conflicts somehow
> 
> I've just updated the related -gtk PR over at [flatpak/xdg-desktop-portal-gtk#236](https://github.com/flatpak/xdg-desktop-portal-gtk/pull/236) to add `(n)` for 1..n to the filename in the same way Nautilus does to ensure it is unique. Is that sufficient for now?

The situation I was thinking of here is something like this:
- The application wants to save file1.txt, file2.txt and file3.txt
- The user picks a folder that already contains file2.txt

I can see a few ways to handle this:
- Just fail and return an error
- Pop up an overwrite confirmation dialog for file2.txt when the user closes the file chooser
- Do the same, but allow the user to change the file name in the dialog (need to worry about him/her changing file2.txt to file3.txt then)
- Display the list of filenames in the filechooser, with some 'conflict indicator'
- Do the same, but allow the user to change the file names to resolve the conflict

mjog:
> The situation I was thinking of here is something like this:
> 
> * The application wants to save file1.txt, file2.txt and file3.txt
> * The user picks a folder that already contains file2.txt

With these PRs as-is, this will currently result in the second saved file being being created as "file2(1).txt".

> I can see a few ways to handle this:
> 
> * Just fail and return an error
> * Pop up an overwrite confirmation dialog for file2.txt when the user closes the file chooser
> * Do the same, but allow the user to change the file name in the dialog (need to worry about him/her changing file2.txt to file3.txt then)
> * Display the list of filenames in the filechooser, with some 'conflict indicator'
> * Do the same, but allow the user to change the file names to resolve the conflict

The third option is best from my perspective - emulating emulate what Nautilus does and ideally with the same UI. The first doesn't seem helpful since there's no way for people to recover without going and doing something about it in the file manager, and I can't see the UX working for the last two options.

I'd be happy to look into something like the second or third option, but it's a substantial addition and I'd like to get on with working on the Contacts portal. Since this is likely to be infrequently used until there is baked-in toolkit support, and since there is no dataloss with the behaviour outlined above as-is, would you be happy if I picked up the confirm/rename dialog work later?

matthiasclasen:
Yes, that seems ok

PR #370, opened 2019-10-02
Allow overriding portals directory

alexlarsson:
Allow overriding $datadir/xdg-desktop-portal/portals using the
$XDG_DESKTOP_PORTAL_DIR env var. We will need this in the test suite when
we want to pass in our own (faked) portal implementation.

This sync this file to the latest version in flatpak.


matthiasclasen:
thanks, already done

PR #373, opened 2019-10-07
gamemode: add pidfd based API

gicmo:
I created a patchset adding a set of pidfd based D-Bus APIs to gamemode. See PR FeralInteractive/gamemode#173 for more details but the gist of it is that we can avoid the costly mapping operation from the container pidns to the host pidns.

### DIFF
--- a/data/org.freedesktop.portal.GameMode.xml
+++ b/data/org.freedesktop.portal.GameMode.xml
@@ -42,7 +42,7 @@
       'UnregisterGame' method, GameMode will automatically un-register
       the client. This might happen with a (small) delay.
 
-      This documentation describes version 2 of this interface.
+      This documentation describes version 3 of this interface.
     -->
     <interface name="org.freedesktop.portal.GameMode">
       <!--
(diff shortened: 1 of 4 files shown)


data/org.freedesktop.portal.GameMode.xml
```diff
@@ -149,6 +149,55 @@
...
+          @requester: Pidfd of the process requesting the information
+          @result: The GameMode status for @target
+
+          Query the GameMode status for a process.
+	  Like org.freedesktop.portal.GameMode.QueryStatusByPid() but @requester
+          and @target are pidfds representing the processes.
+      -->
+      <method name="QueryStatusByPIDFd">
+        <arg type="h" name="target" direction="in"/>
+	<arg type="h" name="requester" direction="in"/>
```
matthiasclasen: Indentation mishap here. More seriously, the args need an 'fd' annotation.

data/org.freedesktop.portal.GameMode.xml
```diff
@@ -149,6 +149,55 @@
...
+          @target: Pidfd of the game to register
+          @requester: Pidfd of the process requesting the registration
+          @result: Status of the request: 0 for success, -1 indicates an error
+
+          Register a game with GameMode.
+	  Like org.freedesktop.portal.GameMode.RegisterGameByPid() but @requester
+          and @target are pidfds representing the processes.
+      -->
+      <method name="RegisterGameByPIDFd">
+        <arg type="h" name="target" direction="in"/>
```
matthiasclasen: Annotation missing here too

data/org.freedesktop.portal.GameMode.xml
```diff
@@ -149,6 +149,55 @@
...
+          @target: Pidfd of the game to un-register
+          @requester: Pidfd of the process requesting the un-registration
+          @result: Status of the request: 0 for success, -1 indicates an error
+
+          Register a game with GameMode.
+	  Like org.freedesktop.portal.GameMode.UnregisterGameByPid() but @requester
+          and @target are pidfds representing the processes.
+      -->
+      <method name="UnregisterGameByPIDFd">
+        <arg type="h" name="target" direction="in"/>
```
matthiasclasen: and here

src/gamemode.c
```diff
@@ -280,6 +317,40 @@ handle_call_thread (GTask        *task,
...
+{
+  g_autoptr(GTask) task = NULL;
+  GDBusMessage *msg;
+  GUnixFDList *fdlist;
+  XdpAppInfo *app_info;
+  Request  *request;
+  CallData *call;
+
+  msg = g_dbus_method_invocation_get_message (invocation);
+  fdlist = g_dbus_message_get_unix_fd_list (msg);
```
matthiasclasen: once you add the fd annotations, the fd_list will come as a parameter here.

src/gamemode.c
```diff
@@ -280,6 +317,40 @@ handle_call_thread (GTask        *task,
...
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_INVALID_ARGS,
+                                             "File descriptor number is incorrect");
+      return;
+    }
+
+  request = request_from_invocation (invocation);
+  app_info = request->app_info;
+
+  call = call_data_new (invocation, app_info, method);
+  call->fdlist = g_object_ref (fdlist);
```
matthiasclasen: Here is a question: can we recognize if an fd is a pidfd ? Because the portal frontend is supposed to validate arguments before passing them on to the backend, so it would be nice to check here, and reject 'bad' fds

data/org.freedesktop.portal.GameMode.xml
gicmo: Ah of course, totally forgot about the 'fd' annotation. Thanks.

src/gamemode.c
gicmo: That is a good point. We can actually do that, by checking that `/proc/self/fdinfo/<fd>` contains a `Pid: X` entry, and additionally that X is not zero. I will add that check.

matthiasclasen:
Does this need to wait until the gamemode changes are landed ?

gicmo:
The pidfd base API has landed upstream :tada:.

Issue #371, opened 2019-10-02
remove gsettings code from frontend


matthiasclasen:
We have a settings backend in xdg-desktop-portal-gtk now, so this code is no longer strictly needed.

What we should think about though is a way to merge the returns from the kde and gtk backends.

TingPing:
> What we should think about though is a way to merge the returns from the kde and gtk backends.

The problem is that the settings are toolkit specific not desktop specific. Generally we want all platforms to have access to the same settings and I don't think it makes sense to require installing other platform portals.

matthiasclasen:
I'm not saying we should require it, but there may be a case for merging the settings from all available backends

PR #374, opened 2019-10-11
settings: Drop GSettings and font config integration

jadahl:
It's dealt with by xdg-desktop-portal-gtk now.


TingPing:
I mentioned this in #371 but this breaks `xdg-desktop-portal-kde` users right? They will lose all GTK settings.

matthiasclasen:
What about the other way around ?

GNOME users don't have settings for kde apps, not much better.

We should figure out a way to address this properly.

PR #395, opened 2019-10-29
Keyboard settings

matthiasclasen:


matthiasclasen:
this turns out to be not needed - wayland protocol already provides key repeat info

PR #397, opened 2019-11-14
build: Add checks for new enough Flatpak version

hadess:
Otherwise building will fail when creating the docs with some missing
files.

```
warning: failed to load external entity "/builddir/xdg-desktop-portal/doc/portal-org.freedesktop.portal.Flatpak.UpdateMonitor.xml"
/builddir/xdg-desktop-portal/doc/portal-docs.xml:116: element include: XInclude error : could not load /builddir/xdg-desktop-portal/doc/portal-org.freedesktop.portal.Flatpak.UpdateMonitor.xml, and no fallback was found
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,30 @@
 #!/bin/sh
 # Run this to generate all the initial makefiles, etc.
 
+REQUIRED_FLATPAK_VERSION=1.5.0
+
+check_flatpak() {
+        # This check is only needed when building from git, as
+        # the tarball carries a copy of a new enough file
+        if ! pkg-config --exists flatpak; then
+                echo "*** flatpak $REQUIRED_FLATPAK_VERSION required ***"
+                exit 1
+        fi
+        if ! pkg-config --atleast-version $REQUIRED_FLATPAK_VERSION flatpak; then
+                FLATPAK_VERSION=$(pkg-config --modversion flatpak)
+                echo "*** flatpak $REQUIRED_FLATPAK_VERSION required, $FLATPAK_VERSION found ***"
+                exit 1
+        fi
+}
+
+# Check whether docs are enabled, and if so, require the newest
+# Flatpak so that docs can be built
+for i in "$@"; do
+        if [ x"$i" = x"--enable-docbook-docs" ]; then
+                check_flatpak
+        fi
+done
+
 test -n "$srcdir" || srcdir=`dirname "$0"`
 test -n "$srcdir" || srcdir=.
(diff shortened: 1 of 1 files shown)


hadess:
The build fails because:
- Flatpak doesn't look like it's being installed at all (?)
- There's no new enough Flatpak [in Ubuntu](https://packages.ubuntu.com/search?keywords=flatpak&searchon=names&suite=all&section=all) or in [Alex' PPA](https://launchpad.net/~alexlarsson/+archive/ubuntu/flatpak)

matthiasclasen:
I don't think that is right though.
We need to build with every flatpak version going back to 1.0.

hadess:
>  We need to build with every flatpak version going back to 1.0.

That's for the git version, to bring in the necessary xml D-Bus interface file. The dist'ed version should still compile with older versions. You cannot compile xdg-desktop-portal against an old flatpak. I wasted an hour trying to find out what was missing. As per the error message in the commit message.

hadess:
For `portal-org.freedesktop.portal.Flatpak.UpdateMonitor.xml` to be created, you need `org.freedesktop.portal.Flatpak.xml` to exist, and to contain an `UpdateMonitor` interface.

`org.freedesktop.portal.Flatpak.xml` is part of `EXTRA_DIST` so a new enough version of it will be in the tarball, whichever version of Flatpak is installed on the system.

If however a git version is built, then `org.freedesktop.portal.Flatpak.xml` will be fetched from the user's flatpak installation. Which might be older than 1.5 and thus not contain the `UpdateMonitor` interface, leading to the error mentioned in the commit message.

In short:
- you can build against an older flatpak if you're building from a tarball
- you can't build against an older flatpak if you're building from git

Hence the flatpak version check in `autogen.sh` rather than in the configure script.

Another possible solution would be to ship org.freedesktop.portal.Flatpak.xml in git and update it when Flatpak grows a new interface that you need to process.

matthiasclasen:
Ah, right. Build from git is affected.

matthiasclasen:
Sadly, this breaks the ci builds, which use the stable flatpak from the ppa.

Could we arrange for the version check to only happen if the doc build is enabled ?

hadess:
> Could we arrange for the version check to only happen if the doc build is enabled ?

I can do that (or something good enough for our use) but dist would be broken then.

hadess:
Note that I ran the script through shellcheck to make sure that it would work under `sh` rather than just under `bash`, just to be sure.

PR #403, opened 2019-11-20
Make document portal exit on unmount

matthiasclasen:
Make the document portal exit when the fuse filesystem
is unmounted externally. This should fix 'suddenly stops
working' issues after suspend.


alexlarsson:
Looks good to me. Did you try it?

matthiasclasen:
Yes, I tried it. I verified that Ctrl-C still exits with status 0, and fusermount -u ... makes the document portal exit with status 21.

PR #413, opened 2019-11-28
WIP: Contacts Portal

mjog:
Adds a Contacts Portal.

This portal is designed to be used as a backend data-source for contact aggregation APIs such as Folks and KPeople, acting as an intermediary to access lower-level data stores such as EDS or Akonadi. Aggregators typically load all contacts at initialisation to provide aggregation, and the portal API supports this by providing a method for loading all contacts in different modes (en-masse, in batches, etc).

The search method returns ids rather than complete contacts on the assumption that they have all been pre-loaded.

Minimalist support for collections is included, mostly at the moment so as to provide a means of consumers determining if the underlying default data store has changed. There should probably be a `ListCollections` method to flesh that out.

It uses VCard 3 as the serialisation format for contact data as a lowest-common denominator, and access to photos encoded in their current format (JPEG, etc) so as to avoid decoding overhead when not immediately needed for presentation, and avoiding introducing artifacts if re-encoded.


mjog:
This is currently just a RFC on the public API - comments welcome.

It really is going to need signals for updating contacts if it's going to fly however - we can't expect people to know or remember to quit their apps to refresh their contacts after their address was just auto-sync'ed with their phone or laptop or whatever.

@matthiasclasen, you said signals were doable but complex, given the API is currently pretty minimal,  would you be happy with those being added up front?

matthiasclasen:
Uh, I think this is a bit much. I would start with GetContact, and not much else.

mjog:
Hmm, well my use case is to make Geary and Contacts viable when installed via Flatpak, which implies enough API so as to be able implement a Folks backend, and the full set of CRUD operations.

On reflection, SearchContacts can be dropped (and likely hence GetContact as well), and it's perhaps arguable if GetDefaultCollection and the collection id arguments are absolutely necessary (I think Contacts will need them), but the rest definitely are required.

To be clear though, are you suggesting just a single method as a first step in a series of PRs, or one method should be enough for most use cases?

jsparber:
> This portal is designed to be used as a backend data-source for contact aggregation APIs such as Folks and KPeople, acting as an intermediary to access lower-level data stores such as EDS or Akonadi. Aggregators typically load all contacts at initialisation to provide aggregation, and the portal API supports this by providing a method for loading all contacts in different modes (en-masse, in batches, etc).

I think it would be better do directly provide a aggregated contact list to the user inside the flatpak sandbox, because an aggregator inside the sandbox has no knowledge about the available low level data sources.

mjog:
The need for this portal could likely go away altogether for apps using Folks if the model I describe here is adopted: https://gitlab.gnome.org/GNOME/folks/-/issues/127#note_1006353

swick:
@mjog can this MR be closed then?

Mikenux:
Could someone explain to me in a simple way what is planned (features, general model of exposure and retrieval of contacts, what is used...)? Thanks!

PR #419, opened 2019-12-11
open-uri: Revamp the policy

matthiasclasen:
Collect all the code that handles the 'shall we open a dialog'
policy in one place. And tweak the rules we apply slightly, with
the goal of being less annoying by default, while still supporting
paranoid use cases.

By default, we skip the app chooser and directly launch the
default handler for http and inode/directory requests. We also
skip the app chooser if there is only one choice.

Lastly, we skip the app chooser if the last choice he been
selected often enough by the user (the threshold we use here
is 3).

The application can override these defaults by passing the
'ask' open with the OpenURI request. If that option is present,
we always show the app chooser.

The user can override everything by setting the threshold value
in the permission store to the special value 2147483647 (INT_MAX).
If the threshold has this value, we always show the app chooser.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -113,11 +113,13 @@ get_latest_choice_info (const char *app_id,
                         const char *content_type,
                         gchar **latest_id,
                         gint *latest_count,
-                        gint *latest_threshold)
+                        gint *latest_threshold,
+                        gboolean *always_ask)
 {
   char *choice_id = NULL;
-  gint choice_count = 0;
-  gint choice_threshold = DEFAULT_THRESHOLD;
+  int choice_count = 0;
+  int choice_threshold = DEFAULT_THRESHOLD;
+  gboolean ask = FALSE;
   g_autoptr(GError) error = NULL;
   g_autoptr(GVariant) out_perms = NULL;
   g_autoptr(GVariant) out_data = NULL;
@@ -136,6 +138,11 @@ get_latest_choice_info (const char *app_id,
         g_warning ("Unable to retrieve info for '%s' in the %s table of the permission store: %s",
                    content_type, PERMISSION_TABLE, error->message);
       g_clear_error (&error);
+    }
+
+  if (out_data != NULL)
+    {
+      g_variant_lookup (out_data, "always-ask", "b", &ask);
     }
 
   if (out_perms != NULL)
@@ -163,6 +170,7 @@ get_latest_choice_info (const char *app_id,
   *latest_id = choice_id;
   *latest_count = choice_count;
   *latest_threshold = choice_threshold;
+  *always_ask = ask;
 
   return (choice_id != NULL);
 }
@@ -253,8 +261,10 @@ update_permissions_store (const char *app_id,
   gint latest_count;
   gint latest_threshold;
   g_auto(GStrv) in_permissions = NULL;
+  gboolean ask;
 
-  if (get_latest_choice_info (app_id, content_type, &latest_id, &latest_count, &latest_threshold) &&
+  if (get_latest_choice_info (app_id, content_type,
+                              &latest_id, &latest_count, &latest_threshold, &ask) &&
       (g_strcmp0 (chosen_id, latest_id) == 0))
     {
       /* same app chosen once again: update the counter */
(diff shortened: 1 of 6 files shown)


matthiasclasen:
Note: when this is merged, https://github.com/flatpak/xdg-desktop-portal/wiki/The-Permission-Store needs some updates to reflect the new policy

TingPing:
From a user experience this all sounds acceptable.

I do worry that we are not considering the security aspect of this though. Just because you only have one handler doesn't mean its safe to pass arbitrary URIs to it.

TingPing:
I would also like the dialog to have a distinction between "always open with this application" and "only this time" option when selecting an application also.

matthiasclasen:
> I do worry that we are not considering the security aspect of this though. Just because you only have one handler doesn't mean its safe to pass arbitrary URIs to it.

But most likely the user has no idea about lurking vulnerabilities, so presenting him with the dialog does not really help that much, since he does not have the information to make that choice (and besides, the dialog will not offer a choice besides: cancel). And the app that is calling OpenURI doesn't know which handler is installed.

One thing we should do in any case is some basic sanity checks on the uri.

> I would also like the dialog to have a distinction between "always open with this application" and "only this time" option when selecting an application.

Will need some agreement from the design side.

>  Or maybe even allow the user to say always ask for this URI type in the future.

Thats what the 'paranoid' policy is about: set threshold to INT_MAX and we will always ask. Just need to convince the designers to export 'paranoid' switches in the control center, or put it in tweaks.

TingPing:
> Thats what the 'paranoid' policy is about: set threshold to INT_MAX and we will always ask. Just need to convince the designers to export 'paranoid' switches in the control center, or put it in tweaks.

Yes it would just be nice to be a visible option and make it possible to be per-uri type.

A workflow I personally use on Android is I leave it always asking for HTTP[S] and depending on the URL I send it to a private browser or to the normal one, I think it would be nice to allow that .

> presenting him with the dialog does not really help that much, since he does not have the information to make that choice 

You are right, I suppose opting into that paranoia is acceptable.

matthiasclasen:
I've added a better paranoid option now

I optionally read per-entry data to determine if we should unconditionally present the app
chooser. This allows the user to say 'always ask for http', regardless of what handler is
being used. The format of the data is an a{sv}, and we look for the 'always-ask' key, 
expecting a boolean.

src/open-uri.c
```diff
@@ -171,18 +163,33 @@ get_latest_choice_info (const char *app_id,
...
   return (choice_id != NULL);
 }
 
 static gboolean
 is_sandboxed (GDesktopAppInfo *info)
 {
-  g_autofree char *exec;
+  g_autofree char *flatpak = NULL;
+
+  flatpak = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (info), "X-Flatpak");
```
alexlarsson: X-Flatpak was added in 1.0, so I guess this ok at this point.

src/open-uri.c
```diff
@@ -171,18 +163,33 @@ get_latest_choice_info (const char *app_id,
...
+  desktop_id = g_app_info_get_id (info);
 
-  exec = g_desktop_app_info_get_string (info, G_KEY_FILE_DESKTOP_KEY_EXEC);
-  return strstr (exec, "flatpak run ") != NULL;
+  if (flatpak)
+    return g_steal_pointer (&flatpak);
+  else
+    return g_strndup (desktop_id, strlen (desktop_id) - strlen (".desktop"));
```
alexlarsson: This is not necessarily right. Apps are allowed to use "sub-ids" of the app id. And apps like libreoffice does this for its multiple apps.

alexlarsson: Oh, i see this is the fallback. Sorry.

alexlarsson: I wonder though if we should support such split desktop files to (optionally) specify their main id, so that the same permissions would be used for the flatpak and the host version of the app.

alexlarsson:
This looks good to me.

src/open-uri.c
nedrichards: It would have to be optional to support the 'private browser' use case that @TingPing mentioned above.

matthiasclasen:
For the 'private browser for certain urls' use case to be viable, the mime system has to learn to handle 'certain urls' first, imo. Then we can have handlers associated with such beasts, and store them in the permission store, etc.

src/open-uri.c
matthiasclasen: hmm, so to support application sub-ids here, I think I need to change course, and not normalize things to the flatpak ID. There can be multiple sub-ids, and they can be launched individually, so we want to let the app chooser present them and pass a sub-id back to the frontend to launch. And I guess that also implies we need to allow sub-ids in the permission store, so we can count how often the user has chosen the same sub-id.

PR #426, opened 2019-12-20
Add CI tests for xenial

alexlarsson:
This doesn't build libportal because that needs a new meson.


alexlarsson:
We need a backport of json_parser_steal_root().

alexlarsson:
I tried to get a new meson in to build libportal, but its seems quite hard as it needs new ninja, etc.

PR #430, opened 2019-12-22
tests: Skip Location tests if we have no D-Bus system bus

smcv:
Some CI systems do their builds in a container or chroot that does not
have a D-Bus system bus, and we can't expect Geoclue to work in that
situation. Ideally the location portal would report this as an error, but
at the moment the test just hangs if we cannot connect to the system bus.


smcv:
In particular, this test failed while I was trying to package 1.6.0 for Debian unstable (where we enable Geoclue, unlike the Github CI).

PR #436, opened 2020-01-14
PipeWire: update to 0.3 API

wtay:
This MR upates the API to the upcoming 3.0 API of PipeWire

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -97,7 +97,7 @@ AC_ARG_ENABLE(pipewire,
 	      [AS_HELP_STRING([--enable-pipewire],[Enable PipeWire support. Needed for screen cast portal])],
 	      enable_pipewire=$enableval, enable_pipewire=yes)
 if test x$enable_pipewire = xyes ; then
-	PKG_CHECK_MODULES(PIPEWIRE, [libpipewire-0.2 >= 0.2.6])
+	PKG_CHECK_MODULES(PIPEWIRE, [libpipewire-0.3 >= 0.2.90])
 	AC_DEFINE([HAVE_PIPEWIRE],[1], [Define to enable PipeWire support])
 fi
 AM_CONDITIONAL([HAVE_PIPEWIRE],[test "$enable_pipewire" = "yes"])
--- a/src/camera.c
+++ b/src/camera.c
@@ -141,7 +141,7 @@ open_pipewire_camera_remote (const char *app_id,
                              GError **error)
 {
   PipeWireRemote *remote;
-  struct spa_dict_item permission_items[1];
+  struct pw_permission permission_items[2];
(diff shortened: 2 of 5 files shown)


src/camera.c
```diff
@@ -158,12 +158,12 @@ open_pipewire_camera_remote (const char *app_id,
...
-  permission_items[0].value = "---";
+  permission_items[0] = PW_PERMISSION_INIT(PW_ID_CORE, PW_PERM_RWX);
+  permission_items[1] = PW_PERMISSION_INIT(PW_ID_ANY, 0)
 
-  pw_core_proxy_permissions (pw_remote_get_core_proxy (remote->remote),
-                             &SPA_DICT_INIT (permission_items,
-                                             G_N_ELEMENTS (permission_items)));
+  pw_client_update_permissions (pw_core_get_client(remote->core),
+		                G_N_ELEMENTS (permission_items),
+				permission_items);
```
TingPing: whitespace nitpick.

src/camera.c
```diff
@@ -158,12 +158,12 @@ open_pipewire_camera_remote (const char *app_id,
    * Hide all existing and future nodes by default. PipeWire will use the
    * permission store to set up permissions.
    */
-  permission_items[0].key = PW_CORE_PROXY_PERMISSIONS_DEFAULT;
-  permission_items[0].value = "---";
+  permission_items[0] = PW_PERMISSION_INIT(PW_ID_CORE, PW_PERM_RWX);
+  permission_items[1] = PW_PERMISSION_INIT(PW_ID_ANY, 0)
```
grulja: I wonder how this passed CI, there is a missing semicolon.

grulja:
Just for the info, I was testing xdg-desktop-portal with this patch and the new PipeWire version and it works.

configure.ac
```diff
@@ -97,7 +97,7 @@ AC_ARG_ENABLE(pipewire,
 	      [AS_HELP_STRING([--enable-pipewire],[Enable PipeWire support. Needed for screen cast portal])],
 	      enable_pipewire=$enableval, enable_pipewire=yes)
 if test x$enable_pipewire = xyes ; then
-	PKG_CHECK_MODULES(PIPEWIRE, [libpipewire-0.2 >= 0.2.6])
+	PKG_CHECK_MODULES(PIPEWIRE, [libpipewire-0.3 >= 0.2.9])
```
grulja: Maybe it should be >= 0.2.90? I don't know what are your plans with PW 0.2, but latest release is 0.2.7 so if you decide to do two more stable releases, then this check will pass and it won't build.

grulja:
@wtay can you update this PR based on the comments? I guess we will need this change to land soon and update Fedora rawhide with it, otherwise it will remain broken.

PR #443, opened 2020-03-05
Don't use AppArmor to detect snap confined clients

jhenstridge:
At the moment, we detect snaps by checking the AppArmor label, which works well on systems that have AppArmor loaded but fails elsewhere.  This branch attempts to fix that and also provide more information about the snap to xdg-desktop-portal.

The new strategy is as follows:

1. Perform a quick check to see whether the process is snap confined.  At present, this is checking whether the freezer cgroup membership is something snapd would place the process in.  This currently doesn't work for systems with a unified cgroup hierarchy, so will need some changes before landing.

2. If we think the process is a snap, shell out to `snap routine portal-info $pid` to identify the snap.  This will return metadata about the snap in keyfile format.  This means that the exact method of determining the metadata can change with snapd itself, and we can easily add more metadata as needed.

As with the old method, there is minimal overhead for systems not running snap confined processes, and no extra build dependencies.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -207,10 +207,27 @@ xdp_app_info_load_app_info (XdpAppInfo *app_info)
 
   g_return_val_if_fail (app_info != NULL, NULL);
 
-  if (app_info->id[0] == '\0')
-    return NULL;
+  switch (app_info->kind)
+    {
+    case XDP_APP_INFO_KIND_FLATPAK:
+      desktop_id = g_strconcat (app_info->id, ".desktop", NULL);
+      break;
 
-  desktop_id = g_strconcat (app_info->id, ".desktop", NULL);
+    case XDP_APP_INFO_KIND_SNAP:
+      desktop_id = g_key_file_get_string (app_info->u.snap.keyfile,
+                                          SNAP_METADATA_GROUP_INFO,
+                                          SNAP_METADATA_KEY_DESKTOP_FILE,
+                                          NULL);
+      break;
+
+    case XDP_APP_INFO_KIND_HOST:
+    default:
+      desktop_id = NULL;
+      break;
+    }
+
+  if (desktop_id == NULL)
+    return NULL;
 
   return G_APP_INFO (g_desktop_app_info_new (desktop_id));
 }
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -487,6 +487,46 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
   return g_steal_pointer (&app_info);
 }
 
+int
+_xdp_parse_cgroup_file (FILE *f, gboolean *is_snap)
+{
+  ssize_t n;
+  g_autofree char *id = NULL;
+  g_autofree char *controller = NULL;
+  g_autofree char *cgroup = NULL;
+  size_t id_len = 0, controller_len = 0, cgroup_len = 0;
+
+  g_return_val_if_fail(f != NULL, -1);
+  g_return_val_if_fail(is_snap != NULL, -1);
+
+  *is_snap = FALSE;
+  do
+    {
+      n = getdelim (&id, &id_len, ':', f);
+      if (n == -1) break;
+      n = getdelim (&controller, &controller_len, ':', f);
+      if (n == -1) break;
+      n = getdelim (&cgroup, &cgroup_len, '\n', f);
+      if (n == -1) break;
+
+      /* Only consider the freezer, systemd group or unified cgroup
+       * hierarchies */
+      if ((!strcmp (controller, "freezer:") != 0 ||
+           !strcmp (controller, "name=systemd:") != 0 ||
+           !strcmp (controller, ":") != 0) &&
+          strstr (cgroup, "/snap.") != NULL)
+        {
+          *is_snap = TRUE;
+          break;
+        }
+    }
+  while (n >= 0);
+
+  if (n < 0 && !feof(f)) return -1;
+
+  return 0;
+}
+
 static gboolean
 pid_is_snap (pid_t pid, GError **error)
 {
@@ -495,11 +535,6 @@ pid_is_snap (pid_t pid, GError **error)
   FILE *f = NULL;
   gboolean is_snap = FALSE;
   int err = 0;
-  ssize_t n;
-  g_autofree char *id = NULL;
-  g_autofree char *controller = NULL;
-  g_autofree char *cgroup = NULL;
-  size_t id_len = 0, controller_len = 0, cgroup_len = 0;
 
   g_return_val_if_fail(pid > 0, FALSE);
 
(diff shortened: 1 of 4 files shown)


alexlarsson:
What version of snap added support for "snap routine portal-info"?

alexlarsson:
If you're adding portal-info to snap, could you also implement the equivalent to flatpak info --file-access=$path $appid, as used by the document portal in app_has_file_access()?

src/xdp-utils.c
```diff
@@ -454,133 +453,91 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
...
-  /* Strip newline */
-  if (con[size - 1] == '\n')
+  cgroup_path = g_strdup_printf ("/proc/%u/cgroup", pid);
+  if (!g_file_get_contents (cgroup_path, &contents, NULL, &local_error))
     {
-      newline = &con[size - 1];
-      size--;
+      /* Don't propagate error if system does not support cgroups */
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        g_propagate_error (error, g_steal_pointer (&local_error));
```
alexlarsson: Its a bit weird to return NULL and not set error, maybe should comment this at the top of the function.

src/xdp-utils.c
```diff
@@ -454,133 +453,91 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
...
+      return NULL;
     }
 
-  if (parse_unconfined (con, size))
+  lines = g_strsplit(contents, "\n", -1);
```
alexlarsson: missing space

src/xdp-utils.c
```diff
@@ -454,133 +453,91 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
...
-
-      label = aa_splitcon (security_label_copy, NULL);
-      if (label && g_str_has_prefix (label, "snap."))
-        {
-          /* If the label belongs to a snap, it will be of the form
-           * snap.$PACKAGE.$APPLICATION.  We want to extract the package
-           * name */
+      return NULL;
+    }
+  if (!g_str_has_prefix(cgroup, "/snap."))
```
alexlarsson: missing space

alexlarsson: and some more below

src/xdp-utils.c
```diff
@@ -591,19 +548,27 @@ xdp_get_app_info_from_pid (pid_t pid,
...
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return NULL;
+        }
+    }
+
   if (app_info == NULL)
     app_info = xdp_app_info_new_host ();
 
-  return app_info;
+  return g_steal_pointer (&app_info);
```
alexlarsson: How did this even work before?

alexlarsson:
Overall this looks good to me, however do you have a plan for how it will work if you're on a unified cgroup? Also, maybe we should not fail if snap is old and doesn't support portal-info? Then we can just use the name?

alexlarsson:
@jhenstridge We want to do a release this week, and it would be nice if we could land this for the release.

alexlarsson:
I extracted the use-after-free fix to a separate PR

src/xdp-utils.c
jhenstridge: I don't think it ever did.  Previously it was only used by the `Device.AccessDevice()` portal API.  As I understand it, that API was called by an abandoned set of Pulse Audio patches.  It looks like it was being called by a Pipewire module in the past too, but that seems to have been removed.

Presumably it worked at some point in the past but was cleaned up without proper testing.

jhenstridge:
> What version of snap added support for "snap routine portal-info"?

At the moment, it looks like it will be in the snapd 2.44.1 point release.  I could push to get it into 2.44 if it is important.

> f you're adding portal-info to snap, could you also implement the equivalent to `flatpak info --file-access=$path $appid`, as used by the document portal in app_has_file_access()?

That's on my todo list.  Definitely something I want to get in place so we can get glib's g_app_info_launch_uris to export documents for snaps without exporting everything.

> Overall this looks good to me, however do you have a plan for how it will work if you're on a unified cgroup? Also, maybe we should not fail if snap is old and doesn't support portal-info? Then we can just use the name?

Proper support for handling unified cgroup systems [is close to landing][1], but is probably going to land in 2.45 at this point.  I had left this PR as draft status in the hope of updating it to properly handle unified cgroup systems, but it probably doesn't reduce the number of systems where snaps are detected given the current code relies on a kernel with AppArmor support.

The other thing I wanted to get implemented was exposing the desktop file ID for the snap so that icons and translated app names can be presented to the user.  This is complicated by the fact that some pieces of code assume that the desktop file can be derived from `app_id` by appending ".desktop", which doesn't hold here.  Perhaps that could be delayed to a future PR though, since it isn't a regression over the existing code.

[1]: https://github.com/snapcore/snapd/pull/7825

src/xdp-utils.c
jhenstridge: I was using the pattern of `parse_app_info_from_flatpak_info`, to separate out the case of an expected failure when the kernel feature is not available from the cases we expect to succeed but encounter an error.  I've added a comment to make this clear.

matthiasclasen:
So, should this wait until 2.44.1 is released, then? 

Whats the verdict?

alexlarsson:
Well, its also a bit problematic that the change breaks pre 2.44.1 versions of snap...

alexlarsson:
I guess best idea might be to do the release without this

matthiasclasen:
James, this is still marked as work in progress. Is it ready to be merged ?

jhenstridge:
@matthiasclasen: the branch is basically ready, but I thought you wanted to wait for the snapd side to be available in a release.  It missed the 2.44 release, but is included in 2.45 which is currently in the beta channel.

The last todo item for this PR was making sure this would work for unified cgroup systems.  The snapd side for that hasn't landed yet, but we now know what that support should look like.

jhenstridge:
I think this is ready for a round of review and merging.  The required snapd support was included in the 2.45 release, which is now on the stable channel.

I notice there is a CI failure, but it looks like the test suite has been timing out for every run since this run:

https://github.com/flatpak/xdg-desktop-portal/actions/runs/98023986

Erick555:
@alexlarsson @matthiasclasen @jhenstridge we are at snapd 2.47 now, shouldn't this get in?

src/xdp-utils.c
```diff
@@ -470,132 +469,125 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
...
-    return NULL;
-
-  /* Strip newline */
-  if (con[size - 1] == '\n')
+  f = fdopen (fd, "r");
+  if (f == NULL)
     {
-      newline = &con[size - 1];
-      size--;
+      err = errno;
```
alexlarsson, line 552: This seems to leak fd on error

alexlarsson:
This looks good to me (sans the least if fdopen fails, but that seems highly unlikely, maybe you can open a separate pr for that).
Sorry for the delay.

PR #447, opened 2020-03-09
New document-portal fuse backend with directory support

alexlarsson:
This reimplements the fuse backend in a much nicer way. It also adds a new flag to export and entire directory, bumping the portal version to 4. I also added a new testsuite for the fuse implementation.

This is not quite finished, outstanding issues are:
 * The code to re-use existing document ids is busted for directories.
 * I need to actually test this live with existing apps to make sure we don't regress.
 * There is a possibility the the new implementation run out of fd:s easier, I need to do some testing on this, especially on older distro versions with lower max-fd settings.
 * We're currently not doing *anything* about symlinks in the directory support, which means that they relative symlinks in the exported directories might break. This needs more consideration to see if we can solve it better.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -286,7 +286,7 @@ portal_delete (GDBusMethodInvocation *invocation,
 }
 
 static char *
-do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_existing, gboolean persistent)
+do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_existing, gboolean persistent, gboolean directory)
 {
   g_autoptr(GVariant) data = NULL;
   g_autoptr(PermissionDbEntry) entry = NULL;
@@ -294,12 +294,14 @@ do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_exis
   char *id = NULL;
   guint32 flags = 0;
 
-  g_debug ("Creating document at path '%s', resuse_existing: %d, persistent: %d", path, reuse_existing, persistent);
+  g_debug ("Creating document at path '%s', resuse_existing: %d, persistent: %d, directory: %d", path, reuse_existing, persistent, directory);
 
   if (!reuse_existing)
     flags |= DOCUMENT_ENTRY_FLAG_UNIQUE;
   if (!persistent)
     flags |= DOCUMENT_ENTRY_FLAG_TRANSIENT;
+  if (directory)
+    flags |= DOCUMENT_ENTRY_FLAG_DIRECTORY;
   data =
     g_variant_ref_sink (g_variant_new ("(^ayttu)",
                                        path,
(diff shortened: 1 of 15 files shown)


alexlarsson:
New version, and I think this is in pretty good shape.

I checked various old distros, and the hard ulimit (which we bump up to) is 4k fds, which I believe should be fine for normal use.

We now also handle lookup and reuse of directory document, as well as re-exporting files and dirs inside a directory document.

I did some simple testing of the file chooser portals with gedit and it seems to work fine.

For symlinks I don't really think we can do better. In general we try to avoid using pathnames in the fuse implementation and instead use O_PATH fds, as these are stable over mutations of the backing fd (i.e. renames/moves). But symlinks are inherently pathnames, and if they are relative is a bit of a pain handle them when we don't know the source pathname. I don't think allowing the user to create (potentially broken) symlinks that can point anywhere and passing them to others is security sensitive really though. I mean they can already do this in their regular directory in $HOME. What we risk is that relative symlinks going outside the directory don't work via the portal, but there is not really much we can do about that.

document-portal/document-portal.c
```diff
@@ -286,20 +286,22 @@ portal_delete (GDBusMethodInvocation *invocation,
 }
 
 static char *
-do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_existing, gboolean persistent)
+do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_existing, gboolean persistent, gboolean directory)
```
matthiasclasen: Should we consider introducing flags here ? The booleans start to pile up

alexlarsson: Possibly. However we have several flags with these value, but most also have some other values. It might be pretty confusing which set of flags go where too.

alexlarsson:
I added related support to flatpak document-export

hadess:
Are there tracker issues/MRs for what would be needed in terms of additional API, whether in xdg-desktop-portal{,-gtk} or GTK, to allow:
- applications to ask for access to the directory in which the file is, to open ancillary files in the same directory (eg. project files with external assets, subtitles with video files, cover art with music albums, etc.)
- full directory access (eg. adding non-stock directories as a library for music, video, etc.)

Also, are there any changes needed in g-c-c to list directories?

hadess:
> Also, are there any changes needed in g-c-c to list directories?

I thought it had a section of file access, but doesn't. I filed https://gitlab.gnome.org/GNOME/gnome-control-center/issues/913

hadess:
> full directory access (eg. adding non-stock directories as a library for music, video, etc.)

And filed https://github.com/flatpak/xdg-desktop-portal/issues/455 to handle this use case.

The other use case would be a bit harder.

hadess:
> applications to ask for access to the directory in which the file is, to open ancillary files in the same directory (eg. project files with external assets, subtitles with video files, cover art with music albums, etc.)

Filed as https://github.com/flatpak/xdg-desktop-portal-gtk/issues/285 because we'd probably want to figure out what the UI should look like first

Issue #455, opened 2020-03-13
Add "Select Folder" support to file chooser


hadess:
Requires:
- documenting and allowing "directory" option in xdg-desktop-portal: https://github.com/flatpak/xdg-desktop-portal/pull/456
- implementing "directory" option in xdg-desktop-portal-gtk: https://github.com/flatpak/xdg-desktop-portal-gtk/pull/284
- wiring up GTK's native file chooser: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1523

hadess:
I haven't finished the xdg-desktop-portal-gtk implementation, so I'll need to finish that up first and thoroughly test it.

matthiasclasen:
thanks for working on this!

hadess:
> I haven't finished the xdg-desktop-portal-gtk implementation, so I'll need to finish that up first and thoroughly test it.

It's done, but untested.

PR #456, opened 2020-03-13
file-chooser: Add "directory" option

hadess:
Add a new directory option to "OpenFile" to allow selecting directories
instead of files.

### DIFF
--- a/src/file-chooser.c
+++ b/src/file-chooser.c
@@ -439,6 +439,7 @@ static XdpOptionKey open_file_options[] = {
   { "accept_label", G_VARIANT_TYPE_STRING, NULL },
   { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
   { "multiple", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "directory", G_VARIANT_TYPE_BOOLEAN, NULL },
   { "filters", (const GVariantType *)"a(sa(us))", validate_filters },
   { "current_filter", (const GVariantType *)"(sa(us))", validate_current_filter },
   { "choices", (const GVariantType *)"a(ssa(ss)s)", validate_choices }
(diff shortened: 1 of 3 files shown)


hadess:
WIP for https://github.com/flatpak/xdg-desktop-portal/issues/455

src/file-chooser.c
```diff
@@ -439,6 +439,7 @@ static XdpOptionKey open_file_options[] = {
   { "accept_label", G_VARIANT_TYPE_STRING, NULL },
   { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
   { "multiple", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "directory", G_VARIANT_TYPE_BOOLEAN, NULL },
```
matthiasclasen: This means we'll pass the option to the backend, so we should probably update the backend interface description at the same time to also mention that option.

matthiasclasen:
One thing I wondered is: would it be nicer to make this a separate OpenDirectory method ?

hadess:
> One thing I wondered is: would it be nicer to make this a separate OpenDirectory method ?

I wondered that as well, but apart from the file filters, you'd probably want pretty much all the same options there. Really up to you, at the end of the day, it'd be hidden behind the GtkFileChooserNative API for app developers. Maybe @grulja strongly prefers a separate method?

grulja:
> > One thing I wondered is: would it be nicer to make this a separate OpenDirectory method ?
> 
> I wondered that as well, but apart from the file filters, you'd probably want pretty much all the same options there. Really up to you, at the end of the day, it'd be hidden behind the GtkFileChooserNative API for app developers. Maybe @grulja strongly prefers a separate method?

In QFileDialog it's basically same, it's same method, just a different flag so I don't have any preference. The QFileDialog has following flags:
`AnyFile, ExistingFile, Directory, ExistingFiles` 
Which means that even for the "multiple" option, it's just a matter of flag combination.

hadess:
Hopefully it works, because I still haven't had time to test it...

PR #459, opened 2020-03-23
FileChooser: document about 'writable' option for OpenFile method

grulja:


data/org.freedesktop.impl.portal.FileChooser.xml
```diff
@@ -97,6 +97,12 @@
               See org.freedesktop.portal.FileChooser.OpenFile() for details.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>writable b</term>
+            <listitem><para>
+              Whether the file is opened with write access. Default is no.
```
marcoscarpetta: Based on [this comment](https://github.com/flathub/flathub/issues/1410#issuecomment-602363947) and https://phabricator.kde.org/D28218, default is yes, isn't it?

grulja: That's default for the GTK file dialog, but if this option is not specified, then the opened file will be read-only.

Issue #460, opened 2020-03-24
Directory export should not change directory name


hadess:
When a directory is exported, it's exported to `/run/user/1000/doc/<id>/` where `<id>` is a computer-generated identifier. This means that a user wanting to access "My family photos" could end up with a directory called "b978c818" in the application's UI.

The document portal/fuse backend should instead export the directory as the sole child under that unique name, eg. `/run/user/1000/doc/b978c818/My family photos` instead of `/run/user/1000/doc/b978c818/`.

This is an behaviour break, but nothing uses the directory support in the document portal yet.

PR #461, opened 2020-03-24
Finish directory selection work

hadess:


hadess:
Still need to figure this https://github.com/flatpak/xdg-desktop-portal/issues/460 out, the "Support registering directories in documents portal helper" patch might need changes after that's fixed.

hadess:
This works now that https://github.com/flatpak/xdg-desktop-portal/pull/462 is merged.

PR #462, opened 2020-03-25
document-portal: Expose directory documents as a subdirectory

alexlarsson:
A path like /some/path/with/subdir now gets exposed as a document
directory '/run/user/1000/doc/1234abcd' with a 'subdir' child.

This means the actual name of the directory will be visible to
the app (and user) rather than some weird generated id.

Fixes https://github.com/flatpak/xdg-desktop-portal/issues/460


hadess:
That works exactly as expected, thanks!

Issue #464, opened 2020-03-28
O_FSYNC is undefined under musl


sirn:
In the new implementation of `document-portal-fuse.c` since 1.7.0, trying to compile xdg-desktop-portal under musl system will fail with:

```
document-portal/document-portal-fuse.c:215:15: error: 'O_FSYNC' undeclared (first use in this function); did you mean 'O_RSYNC'?
  215 |   if (flags & O_FSYNC)
      |               ^~~~~~~
      |               O_RSYNC
```

musl doesn't have `O_FSYNC` defined and [O_SYNC is used in place](https://git.musl-libc.org/cgit/musl/commit/?id=c1f4ed150137d793c9d07356305a89e8785e7e02). `O_SYNC` in glibc also returns the [same value](https://www.gnu.org/software/libc/manual/html_node/Operating-Modes.html) as `O_FSYNC`.

PR #465, opened 2020-03-28
document-portal: Fix compilation with musl

hadess:
Define O_FSYNC if it isn't already.

Closes: #464


sirn:
Thank you! I can confirm this has fixed compilation issue on musl.

hadess:
@alexlarsson if `O_SYNC` == `O_FSYNC`, does handling both flags still make sense?

PR #467, opened 2020-04-01
Try to use less file descriptors in document portal

alexlarsson:
This has some work that tries to keep down the nr of fds used in the document portal.
This was causing some issues when building on ubuntu, as it needed > 4096 fds in the fuse stress test.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -133,6 +133,9 @@ struct _XdpDomain {
   guint64 doc_dir_inode;
   guint32 doc_flags;
 
+  fuse_ino_t doc_parent_ino; /* ino nr of the parent of the document dir, used to invalidate entries */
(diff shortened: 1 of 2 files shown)


alexlarsson:
Hmm, testing this on ubuntu, and i got a super weird error:
```
463   writev(7</dev/fuse>, [{iov_base=")\0\0\0\3\0\0\0\0\0\0\0\0\0\0\0", iov_len=16}, {iov_base="\240\242\0\264\236\177\0\0\10\0\0\0\0\0\0\0", iov_len=16}, {iov_base="95b713f3\0", iov_len=9}], 3 <unfinished ...>
 <... writev resumed> )            = -1 ENOTDIR (Not a directory)
```

Disabling the entry invalidation seems to get rid of this, but I dunno how that could happen.

alexlarsson:
All right, with all these I was finally able to make an ubuntu deb that built, including the testsuite in my VM. Looks ok to me, but lemme do some testing of the results.

PR #481, opened 2020-05-01
Show app chooser when app to open doesn't exist

mcatanzaro:
If the app to open doesn't exist, we should go ahead and show the app chooser instead of spewing criticals and failing to open any app.

~~**Note that this will cause xdg-desktop-portal-gtk to crash.** I'll try to fix that next.~~

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -713,21 +713,26 @@ handle_open_in_thread_func (GTask *task,
           /* Launch the app directly */
           g_autoptr(GError) error = NULL;
 
-          g_debug ("Skipping app chooser");
-
           gboolean result = launch_application_with_uri (app, uri, parent_window, writable, &error);
-          if (request->exported)
+          if (result)
             {
-              if (!result)
-                g_debug ("Open request for '%s' failed: %s", uri, error->message);
-              g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
-              xdp_request_emit_response (XDP_REQUEST (request),
-                                         result ? XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS : XDG_DESKTOP_PORTAL_RESPONSE_OTHER,
-                                         g_variant_builder_end (&opts_builder));
-              request_unexport (request);
-            }
+              g_debug ("Skipped app chooser");
 
-          return;
+              if (request->exported)
+                {
+                  g_variant_builder_init (&opts_builder, G_VARIANT_TYPE_VARDICT);
+                  xdp_request_emit_response (XDP_REQUEST (request),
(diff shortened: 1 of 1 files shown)


mcatanzaro:
It's WIP only because of xdg-desktop-portal-gtk crash. This is the final version of my PR.

mcatanzaro:
> **Note that this will cause xdg-desktop-portal-gtk to crash.** I'll try to fix that next.

Fixed in https://github.com/flatpak/xdg-desktop-portal-gtk/pull/303

hadess:
> If the app to open doesn't exist, we should go ahead and show the app chooser instead of spewing criticals and failing to open any app.

I'd really like this behaviour to be rethought. I don't think it's working as well as it should be, and we're sorely lacking in any kinds of tests.

mcatanzaro:
Not opposed to rethinking how the app chooser dialog works.

That said, with the current design, falling back to displaying the dialog is surely the right thing to do.

src/open-uri.c
```diff
@@ -713,21 +713,26 @@ handle_open_in_thread_func (GTask *task,
           /* Launch the app directly */
           g_autoptr(GError) error = NULL;
 
-          g_debug ("Skipping app chooser");
-
           gboolean result = launch_application_with_uri (app, uri, parent_window, writable, &error);
-          if (request->exported)
+          if (result)
```
hadess, line 717: This means that you also won't skip the app chooser if we fail to export files. Which is probably not the intended behaviour.

How about:
```C
gboolean
app_exists (const char *app_id)
{
[...]
}
```
And around `if (app)`:
```C
gboolean app_exists;

app_exists = app_exists (app);
if (app && app_exists) {
[...]
} else if (!app_exists) {
  g_debug ("Would have used '%s' to skip app chooser, but does not exist", app);
}
```

Add add a g_warning and a return in `launch_application_with_uri()`.

mcatanzaro: > This means that you also won't skip the app chooser if we fail to export files. Which is probably not the intended behaviour.

No, we really do skip the app chooser when not exported. It's the next condition down. OK?

I added a g_debug() to warn when the application doesn't exist. I don't think it's worth g_warning() because it's expected that this will happen all the time.

hadess: I've reworked this, because I didn't like that we were waiting until the last minute to make those checks. I've kept the app info creation check, and I've added those changes to:
https://github.com/flatpak/xdg-desktop-portal/pull/473

hadess:
Obsoleted by https://github.com/flatpak/xdg-desktop-portal/pull/473

mcatanzaro:
Thanks!

PR #488, opened 2020-05-18
camera: add client-node factory to visible objects

wtay:
Or else a client will not be able to create the node to read from
the camera.


wtay:
As it turns out, we decided against this. The portal sets the 'portal.managed' client property and then the session manager is supposed to assign the right permissions. This used to be done in the 0.2 portal module, but in 0.3 this has to be done in the session manager.

PR #490, opened 2020-05-20
Multi camera

matthiasclasen:
Allow selecting a camera

### DIFF
--- a/src/camera.c
+++ b/src/camera.c
@@ -36,6 +36,20 @@ static XdpImplLockdown *lockdown;
 typedef struct _Camera Camera;
 typedef struct _CameraClass CameraClass;
 
+typedef struct {
+  char *name;
+  guint id;
+} CameraInfo;
+
+static void
+camera_info_free (gpointer data)
+{
+  CameraInfo *info = data;
+
+  g_free (info->name);
+  g_free (info);
+}
+
 struct _Camera
 {
   XdpCameraSkeleton parent_instance;
@@ -75,10 +89,45 @@ handle_access_camera_in_thread_func (GTask *task,
   Request *request = (Request *)task_data;
   const char *app_id;
   gboolean allowed;
+  char **id_array;
+  char **name_array;
+  char *choice;
 
   app_id = (const char *)g_object_get_data (G_OBJECT (request), "app-id");
 
-  allowed = device_query_permission_sync (app_id, "camera", request);
+  if (g_hash_table_size (camera->cameras) > 1)
+    {
+      GPtrArray *ids = g_ptr_array_new ();
+      GPtrArray *names = g_ptr_array_new ();
+      GHashTableIter iter;
+      gpointer key, value;
+
+      g_hash_table_iter_init (&iter, camera->cameras);
+      while (g_hash_table_iter_next (&iter, (gpointer *)&key, (gpointer *)&value))
+        {
+          CameraInfo *info = value;
+          char *id = g_strdup_printf ("%u", info->id);
+          g_ptr_array_add (ids, id);
+          g_ptr_array_add (names, g_strdup (info->name));
+        }
+      g_ptr_array_add (ids, NULL);
+      g_ptr_array_add (names, NULL);
+
+      id_array = (char **)g_ptr_array_free (ids, FALSE);
+      name_array = (char **)g_ptr_array_free (names, FALSE);
+    }
+  else
+    {
+      id_array = name_array = NULL;
+    }
+
+  choice = NULL;
+  allowed = device_query_permission_sync (app_id, "camera", request,
+                                          (const char **)id_array,
+                                          (const char **)name_array,
+                                          &choice);
+  g_strfreev (id_array);
+  g_strfreev (name_array);
 
   REQUEST_AUTOLOCK (request);
 
@@ -88,6 +137,18 @@ handle_access_camera_in_thread_func (GTask *task,
       guint32 response;
 
       g_variant_builder_init (&results, G_VARIANT_TYPE_VARDICT);
+
+      if (choice)
+        {
+          GVariantBuilder streaminfo;
+          GVariantBuilder streams;
+
+          g_variant_builder_init (&streaminfo, G_VARIANT_TYPE_VARDICT);
+          g_variant_builder_init (&streams, G_VARIANT_TYPE ("a(ua{sv})"));
+          g_variant_builder_add (&streams, "(ua{sv})", atoi (choice), &streaminfo);
+
+          g_variant_builder_add (&results, "{sv}", "streams", g_variant_builder_end (&streams));
+        }
 
       response = allowed ? XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS
                          : XDG_DESKTOP_PORTAL_RESPONSE_CANCELLED;
@@ -122,7 +183,6 @@ handle_access_camera (XdpCamera *object,
 
   app_id = xdp_app_info_get_id (request->app_info);
 
-
   g_object_set_data_full (G_OBJECT (request), "app-id", g_strdup (app_id), g_free);
 
   request_export (request, g_dbus_method_invocation_get_connection (invocation));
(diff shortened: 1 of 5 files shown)


jadahl:
I don't think we should send the node ids via the portal, they should just show up in PipeWire by PipeWire itself checking the permission store.

We also shouldn't use the node ID to identify the camera. Today node N will mean the built in web cam, but tomorrow it could just as well be the surveillance camera. Camera C might also consist of multiple nodes I suspect (see below), so it'd be misleading as well.

The reason for this is that the node IDs are volatile, the ID will change when you replug the device. It's different for screen casts, as they are by definition temporary, but camera access is remembered, so it doesn't make sense to communicate that here.

Looking at `pw-mon`'s output here, for the camera I get these two:

```
added:
	id: 33
	permissions: rwx
	type: PipeWire:Interface:Device (version 3)
 	params:
 	properties:
 		device.enum.api = "udev"
 		device.api = "v4l2"
 		media.class = "Video/Device"
 		api.v4l2.path = "/dev/video0"
 		device.plugged.usec = "23779664"
 		device.bus-path = "pci-0000:00:14.0-usb-0:8:1.0"
 		device.sysfs.path = "/sys/devices/pci0000:00/0000:00:14.0/usb1/1-8/1-8:1.0/video4linux/video0"
 		device.bus = "usb"
 		device.subsystem = "video4linux"
 		device.vendor.id = "04f2"
 		device.vendor.name = "SunplusIT Inc"
 		device.product.id = "b5ab"
 		device.product.name = "Integrated Camera: Integrated C"
 		device.serial = "SunplusIT_Inc_Integrated_Camera"
 		device.capabilities = ":capture:"
 		device.name = "v4l2_device.pci-0000:00:14.0-usb-0:8:1.0"
 		device.description = "Integrated Camera: Integrated C"
 		factory.id = "14"
 		client.id = "30"
 		object.id = "33"
 		object.path = "v4l2:/dev/video0"
 		api.v4l2.cap.driver = "uvcvideo"
 		api.v4l2.cap.card = "Integrated Camera: Integrated C"
 		api.v4l2.cap.bus_info = "usb-0000:00:14.0-8"
 		api.v4l2.cap.version = "5.6.10"
 		api.v4l2.cap.capabilities = "84a00001"
 		api.v4l2.cap.device-caps = "04200001"
added:
	id: 34
	permissions: rwx
	type: PipeWire:Interface:Device (version 3)
 	params:
 	properties:
 		device.enum.api = "udev"
 		device.api = "v4l2"
 		media.class = "Video/Device"
 		api.v4l2.path = "/dev/video1"
 		device.plugged.usec = "23779280"
 		device.bus-path = "pci-0000:00:14.0-usb-0:8:1.0"
 		device.sysfs.path = "/sys/devices/pci0000:00/0000:00:14.0/usb1/1-8/1-8:1.0/video4linux/video1"
 		device.bus = "usb"
 		device.subsystem = "video4linux"
 		device.vendor.id = "04f2"
 		device.vendor.name = "SunplusIT Inc"
 		device.product.id = "b5ab"
 		device.product.name = "Integrated Camera: Integrated C"
 		device.serial = "SunplusIT_Inc_Integrated_Camera"
 		device.capabilities = ":"
 		device.name = "v4l2_device.pci-0000:00:14.0-usb-0:8:1.0"
 		device.description = "Integrated Camera: Integrated C"
 		factory.id = "14"
 		client.id = "30"
 		object.id = "34"
 		object.path = "v4l2:/dev/video1"
 		api.v4l2.cap.driver = "uvcvideo"
 		api.v4l2.cap.card = "Integrated Camera: Integrated C"
 		api.v4l2.cap.bus_info = "usb-0000:00:14.0-8"
 		api.v4l2.cap.version = "5.6.10"
 		api.v4l2.cap.capabilities = "84a00001"
 		api.v4l2.cap.device-caps = "04a00000"
```

I don't know why there are two, my laptop only has one camera that I know of, but I suspect it's one of those situations where a single camera is built up of multiple hardware devices that can be used in combination some how. AFAIK cameras are getting more and more complex, where multiple "devices" are processed together in some way. I think it's still an open question whether things conceptually like [libcamera](https://libcamera.org/) should be dealt with before or after camera video passes through PipeWire. Either way, looking at that, it looks like we should use a subset of

```
 		device.vendor.id = "04f2"
 		device.vendor.name = "SunplusIT Inc"
 		device.product.id = "b5ab"
 		device.product.name = "Integrated Camera: Integrated C"
 		device.serial = "SunplusIT_Inc_Integrated_Camera"
```

as a "key", and not use node IDs for anything at all used for identification.

matthiasclasen:
If you look closely, the node IDs are not used for persistent device identification. We never store them (or any other device identifification) in the permission store.

This works exactly as  the screencast portal does: We look at the current pipewire graph, pull out a list of nodes that we consider cameras, and let the user choose one. Then we give that node ID to the application.

jadahl:
Pipewire still queries the permission store though, so whatever the user selects doesn't affect what cameras the application will get access to; it seems more like a outsourcing a piece of UI for selecting a camera, without any security benefits.

If it *would* set up static permissions, it'd instead break hotplugging, and we'd have to proxy that from PipeWire some how. I still don't think this is the right path to take at all; it doesn't allow implementing the mockups (as I saw them last time, it's a "Allow camera access" dialog).

I don't see the point in doing all this via the portal really. If there should be per camera device access control, I'd rather we have mockups for how the UI would look; I suspect it would not be selecting what camera every time, that'd make it impossible to implement basic camera applications as they behave on smart phones today. But for applications to see what cameras it has access to, it should just use PipeWire directly.

src/camera.c
```diff
@@ -159,7 +219,8 @@ open_pipewire_camera_remote (const char *app_id,
    * permission store to set up permissions.
    */
   permission_items[0] = PW_PERMISSION_INIT (PW_ID_CORE, PW_PERM_RWX);
-  permission_items[1] = PW_PERMISSION_INIT (PW_ID_ANY, 0);
+  permission_items[1] = PW_PERMISSION_INIT (remote->node_factory_id, PW_PERM_R);
```
jadahl, line 222: This is handled by `module-portal` in PipeWire. It was part of the PR as the `module-portal` had not been brought over to 0.3.

matthiasclasen: Mockups are just that: sketches that present an rough idea at a certain point in time. The ones we have for the camera portal are more than 3 years old now. I would not put too much weight on them.

> But for applications to see what cameras it has access to, it should just use PipeWire directly.

The point here was to give the user control over what cameras the application has access to. I don't see how you solve that without somebody enumerating cameras with in a user-presentable way.

And even if you give up on access control and say: let the app talk to pipewire directly, you just drop the "figure out user-presentable camera enumeration" problem onto the application.

matthiasclasen: > Pipewire still queries the permission store though, so whatever the user selects doesn't affect what cameras the application will get access to; it seems more like a outsourcing a piece of UI for selecting a camera, without any security benefits.

I was never entirely happy with this approach, and the recent "oops cameras don't work with pw 0.3" shows why: We treat pipewire as this magic black box. But it doesn't actually work, and we loose control and can't present a meaningful UI.

wtay: > I was never entirely happy with this approach, and the recent "oops cameras don't work with pw 0.3" shows why: We treat pipewire as this magic black box. But it doesn't actually work, and we loose control and can't present a meaningful UI.

This is supposed to be a temporary solution until we can merge a real multimedia session manager into GNOME. It all ties together; security, policy, device configuration, the apps that are running, the user, ...

Right now we have an example session manager that works indeed like a black box and that tries to get info from various places (portal, permission store) to do its thing.

In any case, even if there is a non-black box GNOME controlled session manager, it will probably still need to access other parts of the system (config, permission store, ..) to do all the things it needs to do.

A6GibKm:
> I don't think we should send the node ids via the portal, they should just show up in PipeWire by PipeWire itself checking the permission store.

To be able to use the file descriptor (correctly) one needs the node_id, otherwise one needs to open holes in the sandbox and add pipewire as a dependency. This is very unfriendly to developers, not to mention that at least for me it is impossible to build pipewire-rs in GNOME master.

With the node_id, you can just open a gstreamer pipewiresrc, set its node and fd properties and call it a day.

PR #492, opened 2020-05-21
Screencast: document source_type property in the response

grulja:
This is supposed to be a new addition for the clients to help them identify whether the user picked a screen or a window and what will be the content of the stream.

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -146,6 +146,14 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type u</term>
+            <listitem><para>
+              The type of the content which is being screen casted.
+              For available source types, see the AvailableSourceTypes property.
+              This property was added in version 3 of this interface.
+            </para></listitem>
+          </varlistentry>
         </variablelist>
     -->
     <method name="Start">
--- a/data/org.freedesktop.portal.ScreenCast.xml
+++ b/data/org.freedesktop.portal.ScreenCast.xml
@@ -192,6 +192,14 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type u</term>
+            <listitem><para>
+              The type of the content which is being screen casted.
+              For available source types, see the AvailableSourceTypes property.
+              This property was added in version 3 of this interface.
+            </para></listitem>
+          </varlistentry>
         </variablelist>
     -->
     <method name="Start">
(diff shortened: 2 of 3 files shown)


data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -192,6 +192,13 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type (u)</term>
```
jadahl, line 196: I don't think it should need to be a tuple. The other ones are because they contain multiple values.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -146,6 +146,13 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type (u)</term>
```
jadahl, line 150: Same here about not using a tuple.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -192,6 +192,13 @@
...
+            <listitem><para>
+              A type of the content which is being recorded.
```
jadahl, line 198: s/recorded/screen casted/ to not add new vocabulary.

data/org.freedesktop.portal.ScreenCast.xml
grulja: I didn't realize that (u) is a structure or a touple.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -192,6 +192,13 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type u</term>
+            <listitem><para>
+              A type of the content which is being screen casted.
+              For available source types, see the AvailableSourceTypes property.
```
jadahl, line 199: Should probably add a "This property was added in version 3 of this interface."

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -146,6 +146,14 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type u</term>
+            <listitem><para>
+              A type of the content which is being screen casted.
```
jadahl, line 152: *The type of content

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -192,6 +192,14 @@
               stream.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>source_type u</term>
+            <listitem><para>
+              A type of the content which is being screen casted.
```
jadahl, line 198: *The type of content

PR #493, opened 2020-05-23
FileChooser: Add a 'current_filter' return value

michaelweghorn:
Add a 'current_filter' element for the FileChooser response which is used to specify which filter
was selected in the file chooser.

This will allow making Gtk's 'gtk_file_chooser_get_filter' work for the portal native file chooser.

See https://gitlab.gnome.org/GNOME/gtk/-/issues/1820 for more details. I'll add a link to the corresponding gtk and xdg-desktop-portal-gtk merge/pull requests here once I have created them.

### DIFF
--- a/data/org.freedesktop.impl.portal.FileChooser.xml
+++ b/data/org.freedesktop.impl.portal.FileChooser.xml
@@ -110,6 +110,15 @@
             </para></listitem>
           </varlistentry>
           <varlistentry>
+            <term>current_filter (sa(us))</term>
+            <listitem>
+              <para>
+                The filter that was selected.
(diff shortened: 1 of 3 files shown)


michaelweghorn:
Corresponding changes in gtk and xdg-desktop-portal-gtk to make the whole use case work:

https://github.com/flatpak/xdg-desktop-portal-gtk/pull/311
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1959
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1960

matthiasclasen:
I don't particularly see why firefox needs to get the filter from the file chooser, but since you've done all the work, sure.

PR #510, opened 2020-07-09
Close fd in OpenURI portal

### LINKED ISSUE #509: xdg-open (OpenURI portal) is leaving files open
Minimal example using `flatpak-builder --run` for the call inside flatpak

```sh
flatpak $ xdg-open ~/test/dir
host $ lsof -f +D ~/test/
COMMAND    PID   USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME
xdg-deskt 4881 sophie   12u   DIR  253,1     4096 15082214 /home/sophie/test/dir
```

This also happens using glibs `g_app_info_launch_default_for_uri()` in a program started via `flatpak run`. The path is still open after exiting the flatpak.

I am using fusermount inside the flatpak and the showing the directory to the user via the OpenURI portal. However, I can't umount the fuse mount when shutting the program down ("Device or resource busy.") I would also be grateful if there is an idea for a workaround while this problem with the portal exists. Would it be possible to bypass the sandbox and run something like xdg-open on the host?

sophie-h:
Didn't managed to get portals to work with my own build, so this is untested.

Not sure if `xdp_impl_app_chooser_call_choose_application()` uses the fd since I have no idea where is function is defined.

Closes #509

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -622,6 +622,7 @@ handle_open_in_thread_func (GTask *task,
       gboolean fd_is_writable;
 
       path = xdp_app_info_get_path_for_fd (request->app_info, fd, 0, NULL, &fd_is_writable);
+      close (fd);
       if (path == NULL ||
           (writable && !fd_is_writable) ||
           (!xdp_app_info_is_host (request->app_info) && !writable && fd_is_writable))


matthiasclasen:
I think you've pointed out an issue, but we may need some more work to fix it properly.

One issue I see is with
```
fd = g_object_get_data (G_OBJECT (request), "fd");
```

if "fd" is not set on the request (as is the case if we're coming from the open-uri request), then this will set fd to 0, which is a valid fd that we __don't__ want to close.

I think we should explicitly set fd to -1 in that case, and make sure we check the fd before calling close().

We should assert that "fd" and "uri" are not set at the same time - if they were, we'd be leaking the fd in the if (uri) branch.

hadess:
> One issue I see is with
> 
> ```
> fd = g_object_get_data (G_OBJECT (request), "fd");
> ```
> 
> if "fd" is not set on the request (as is the case if we're coming from the open-uri request), then this will set fd to 0, which is a valid fd that we **don't** want to close.
> 
> I think we should explicitly set fd to -1 in that case, and make sure we check the fd before calling close().
> 
> We should assert that "fd" and "uri" are not s

Is https://github.com/flatpak/xdg-desktop-portal/pull/515 what you expected?

matthiasclasen:
looks done by hadess

Issue #487, opened 2020-05-18
fd leak prevents removable media from being ejected


ptomato:
I opened a file residing on removable media, through the portal, in the org.mozilla.Thunderbird app. Even after closing the app I couldn't eject the drive because it was "busy".

relevant entry from `lsof` (repeated many times):
```
xdg-docum  7925 14911 pool-/usr  ptomato   17u      REG                8,2 (file size redacted)       2282 /run/media/ptomato/(filename redacted)
```
Process 7925 is:
```
   7925 ?        Ssl    0:23 /usr/libexec/xdg-document-portal
```

sophie-h:
Very similar to #509. It seems like somehow nobody cared about closing fds returned by `g_unix_fd_list_append()`

matthiasclasen:
I will take a look at this soon

hadess:
https://github.com/flatpak/xdg-desktop-portal/pull/530 might be the fix (on top of the other fixes in https://github.com/flatpak/xdg-desktop-portal/pull/515).

orowith2os:
@ptomato can you reproduce this now?

ptomato:
I don't know if I recreated the exact circumstances, but this does not seem to be a problem any longer in xdg-desktop-portal 1.16.0.

GeorgesStavracas:
Thanks. Let's assume it was fixed then, but if you spot it again, please reopen

hfiguiere:
I thought this was #689 

Looks a lot like it.

PR #515, opened 2020-07-20
Fix wallpaper and open-uri fd leaks

hadess:
Should fix https://github.com/flatpak/xdg-desktop-portal/issues/509


sophie-h:
Can you also have a look at #487? The code and the problem looked very similar to me.

hadess:
> Can you also have a look at #487? The code and the problem looked very similar to me.

The best way to check would be to try it out in the same circumstances.

PR #520, opened 2020-08-05
tests: Divert service stdout to stderr

smcv:
These tests produce structured TAP output on stdout, so we cannot run
any subprocesses that would produce unstructured stdout. Unfortunately,
by default g_debug() writes to stdout, so running with G_MESSAGES_DEBUG
to get better logging from the various subprocesses breaks the test.
Use stderr instead.


smcv:
CI (still) fails due to #525. See #521 for a branch that includes this and has working CI.

PR #521, opened 2020-08-05
Improve installed-tests and run them during Github CI

smcv:
Includes #525, #520 and part of #503.

* All three commits from #525, see that PR for details

* xdp: Send messages to stderr, not stdout
    
    This is done for two reasons:
    
    1. Logging on stdout can interfere with machine-readable protocols.
    
    2. stdout is normally fully buffered, whereas stderr is normally
       line-buffered. It's a lot easier to debug the service when messages
       don't lag behind.

* tests: Divert service stdout to stderr, see PR #520 for details
 
* tests: use kill from PATH (from @jtojnar on #503)
    
    It is not guaranteed that /bin/kill exists and we are already using the bash builtin elsewhere in the script.

* Install session.conf.in to expected path for installed-tests

* tests: call test-document-fuse.py directly (from @jtojnar on #503)
    
    It has a shebang so no need to pass it to python3 in the tests.
    
    It also makes it easier for downstreams to wrap the python file for setting extra environment variables.
    
    [smcv: chmod the script +x]

* tests: Only run xdg-document-portal in foreground when uninstalled
    
    During build-time testing, we need to take special steps to run the
    just-built version of the portal. However, during "as-installed"
    testing we want to assert that it was installed correctly, and in
    particular that D-Bus .service file was installed correctly to make it
    activatable.

* test-document-fuse: Quote more defensively

* tests: Send SIGTERM to dbus-daemon
    
    There's no real need to kill it with SIGKILL, and in some configurations
    there is cleanup that is only performed when allowed to shut down
    gracefully.

* CI: Correct the name of the GLib package
    
    We want development headers, so libglib2.0-dev rather than libglib2.0-0.

* CI: Run installed-tests

---

Known omissions:

* We don't run the FUSE tests (but they'll probably fail or be skipped anyway in any CI environment that is based on containers)
* We don't have libcap2-bin (capsh) so some tests will be skipped
* We don't have Pipewire

I'm in the process of getting Pipewire 0.3 into Debian. When that has happened, the run that includes installed-tests could maybe be switched from Ubuntu 18.04 to Debian 11?


smcv:
This works for me on Debian unstable, using:

```
./configure --prefix="$HOME/tmp/usr" --disable-pipewire --disable-libportal --enable-installed-tests
make
make check
make install
XDG_DATA_DIRS="$HOME/tmp/usr/share:$XDG_DATA_DIRS" ginsttest-runner -d "$HOME/tmp/usr/share" xdg-desktop-portal
```

(I know that doesn't run *all* tests, but in particular it does cover the FUSE document-portal test.)

smcv:
With `TEST_IN_CI=1` and `--enable-libportal`, `test-portals.test` hangs after `/portal/openuri/exists`. I think that should probably be handled as a separate issue.

smcv:
I should drop the CI bits of this branch for now, since CI is consistently timing out anyway.

smcv:
Rebased on latest #525.

smcv:
Rebased to reflect #525 having been merged; no practical changes.

Issue #524, opened 2020-09-06
Regression in #473: test-portals now fails


smcv:
As noted on #521, test-portals now fails when running build-time tests with libportal enabled, which seems to be a regression caused by #473.

To reproduce:
* `./configure --prefix="$HOME/tmp/usr" --enable-pipewire --enable-libportal --enable-installed-tests`
* `make`
* `TEST_IN_CI=1 make check TESTS=test-portals`

With 69205f12 reverted, this passes.

With 5c4e4d22, the test hangs in `test_open_uri_http()` because `xdg-desktop-portal` crashes when `app_exists()` tries to look up `NULL` in a GHashTable:

```#0  g_str_hash (v=0x0) at ../../../glib/ghash.c:2333
#1  0x00007fca72e6eb0b in g_hash_table_lookup_node
    (hash_return=<synthetic pointer>, key=0x0, hash_table=0x557f84507000 = {...}) at ../../../glib/ghash.c:1511
#2  g_hash_table_lookup (hash_table=0x557f84507000 = {...}, key=key@entry=0x0) at ../../../glib/ghash.c:1511
#3  0x00007fca730828ca in desktop_file_dir_unindexed_get_app (desktop_id=0x0, dir=<optimized out>)
    at ../../../gio/gdesktopappinfo.c:1428
#4  desktop_file_dir_get_app (desktop_id=0x0, dir=<optimized out>) at ../../../gio/gdesktopappinfo.c:1428
#5  g_desktop_app_info_new (desktop_id=0x0) at ../../../gio/gdesktopappinfo.c:1983
#6  0x0000557f8282e97f in app_exists (app_id=<optimized out>) at src/open-uri.c:676
#7  handle_open_in_thread_func
    (task=<optimized out>, source_object=<optimized out>, task_data=0x557f8445dad0, cancellable=<optimized out>)
    at src/open-uri.c:676
#8  0x00007fca7306afae in g_task_thread_pool_thread (thread_data=0x557f84447ef0, pool_data=<optimized out>)
    at ../../../gio/gtask.c:1417
#9  0x00007fca72eab694 in g_thread_pool_thread_proxy (data=<optimized out>) at ../../../glib/gthreadpool.c:354
#10 0x00007fca72eaadad in g_thread_proxy (data=0x557f844ddcc0) at ../../../glib/gthread.c:819
#11 0x00007fca72e16ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
[...]

PR #525, opened 2020-09-06
Fix openuri test regression (#524)

smcv:
* open-uri: Don't crash if there is no default or latest app
    
    g_desktop_app_info_new() requires a non-NULL argument.
    
    Partially addresses #524.
    
    Fixes: 69205f12 "open-uri: Show app chooser when default app does not exist"

* open-uri: Assert that find_recommended_choices yields non-NULL
    
    choices is never NULL, but might be an array containing only NULL.

* tests: Make app chooser backend more debuggable
    
    Instead of crashing out when our expectations are not met, return an
    error that other components can log.

* open-uri: Use correct app ID to determine whether it exists
    
    x-d-p represents apps by their app ID *without* the .desktop suffix,
    but GDesktopAppInfo represents them by the app ID *with* the .desktop
    suffix.
    
    Fixes: 69205f12 "open-uri: Show app chooser when default app does not exist"  
    Resolves: #524

cc @hadess @mcatanzaro

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -554,10 +554,12 @@ static gboolean
 app_exists (const char *app_id)
 {
   g_autoptr(GDesktopAppInfo) info = NULL;
+  g_autofree gchar *with_desktop = NULL;
 
   g_return_val_if_fail (app_id != NULL, FALSE);
 
-  info = g_desktop_app_info_new (app_id);
+  with_desktop = g_strconcat (app_id, ".desktop", NULL);
+  info = g_desktop_app_info_new (with_desktop);
   return (info != NULL);
 }
 
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -669,6 +669,8 @@ handle_open_in_thread_func (GTask *task,
 
   /* collect all the information */
   find_recommended_choices (scheme, content_type, &default_app, &choices, &n_choices);
+  /* it's never NULL, but might be empty (only contain the NULL terminator) */
+  g_assert (choices != NULL);
   if (default_app != NULL && !app_exists (default_app))
     g_clear_pointer (&default_app, g_free);
   use_default_app = should_use_default_app (scheme, content_type);
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -555,6 +555,8 @@ app_exists (const char *app_id)
 {
   g_autoptr(GDesktopAppInfo) info = NULL;
 
+  g_return_val_if_fail (app_id != NULL, FALSE);
+
   info = g_desktop_app_info_new (app_id);
   return (info != NULL);
 }
@@ -667,13 +669,13 @@ handle_open_in_thread_func (GTask *task,
 
   /* collect all the information */
   find_recommended_choices (scheme, content_type, &default_app, &choices, &n_choices);
-  if (!app_exists (default_app))
+  if (default_app != NULL && !app_exists (default_app))
     g_clear_pointer (&default_app, g_free);
   use_default_app = should_use_default_app (scheme, content_type);
   get_latest_choice_info (app_id, content_type,
                           &latest_id, &latest_count, &latest_threshold,
                           &ask_for_content_type);
-  if (!app_exists (latest_id))
+  if (latest_id != NULL && !app_exists (latest_id))
     g_clear_pointer (&latest_id, g_free);
 
   skip_app_chooser = FALSE;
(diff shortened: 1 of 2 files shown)


smcv:
This made CI pass for the first time in a few months! \o/

src/open-uri.c
```diff
@@ -729,7 +733,7 @@ handle_open_in_thread_func (GTask *task,
         app = latest_id;
       else if (default_app != NULL)
         app = default_app;
-      else if (choices && app_exists (choices[0]))
+      else if (choices && choices[0] && app_exists (choices[0]))
```
hadess, line 736: Can you make find_recommended_choices() return early if the `infos` list is empty instead of returning an array with just the NULL terminator?

smcv: I could, but I'm not 100% sure I should. How strongly do you prefer this?

It has to set choices to *something*: either `NULL`, or a non-`NULL` pointer to an empty strv (just the `NULL` terminator). At the moment it looks as though it always sets choices to non-`NULL` (although possibly empty), which is the same behaviour as `g_variant_dup_strv()`.

I could make it early-return if you strongly prefer that, but that's actually more code and less consistent. We'd have to make sure it isn't passed to any of the GStrv functions that expect a non-`NULL` argument (like `g_strv_contains()`, `g_strv_equal()`, `g_strv_length()`), and that `xdp_impl_app_chooser_call_update_choices()` and `xdp_impl_app_chooser_call_choose_application()` will accept `NULL`.

If your objection is to the apparent redundancy here, I could sidestep it like this:

```suggestion
      else if (n_choices > 0 && app_exists (choices[0]))
```

Would you like that any better?

smcv: > and that `xdp_impl_app_chooser_call_update_choices()` and `xdp_impl_app_chooser_call_choose_application()` will accept `NULL`

I don't think they do: they end up in GVariant format string `^as`, which calls `g_variant_new_strv (ptr, -1)`, which wants a non-`NULL` *ptr* (`NULL` is only allowed if the length is explicitly 0, not if it's -1).

smcv: The diff would look something like this:

```diff
diff --git a/src/open-uri.c b/src/open-uri.c
index d503702..556fa71 100644
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -509,6 +509,14 @@ find_recommended_choices (const char *scheme,
   if (!infos)
     infos = g_app_info_get_all_for_type (content_type);
 
+  /* Don't return a non-NULL array that is empty */
+  if (!infos)
+    {
+      *choices = NULL;
+      *choices_len = 0;
+      return;
+    }
+
   n_choices = g_list_length (infos);
   result = g_new (char *, n_choices + 1);
   for (l = infos, i = 0; l; l = l->next)
@@ -542,6 +550,7 @@ app_info_changed (GAppInfoMonitor *monitor,
   content_type = (const char *)g_object_get_data (G_OBJECT (request), "content-type");
   find_recommended_choices (scheme, content_type, &default_app, &choices, &n_choices);
 
+  /* TODO: make sure this accepts NULL for choices */
   xdp_impl_app_chooser_call_update_choices (impl,
                                             request->id,
                                             (const char * const *)choices,
@@ -733,7 +742,7 @@ handle_open_in_thread_func (GTask *task,
         app = latest_id;
       else if (default_app != NULL)
         app = default_app;
-      else if (choices && choices[0] && app_exists (choices[0]))
+      else if (choices && app_exists (choices[0]))
         app = choices[0];
 
       if (app)
@@ -786,6 +795,7 @@ handle_open_in_thread_func (GTask *task,
 
   g_debug ("Opening app chooser");
 
+  /* TODO: make sure this accepts NULL for choices */
   xdp_impl_app_chooser_call_choose_application (impl,
                                                 request->id,
                                                 app_id,
```

hadess: I'm fine with either ways of fixing, but I think it should either always be non-NULL, with only a terminator, or it should be NULL when the array is empty. So your first suggestion is probably fine.

smcv: > I think it should either always be non-NULL, with only a terminator, or it should be NULL when the array is empty

OK, that makes sense!

smcv:
Should all be fixed in the latest version.

hadess:
Tip-top, thanks!

smcv:
Thanks! If you like CI coverage, #521 is also relevant, and also passing now.

hadess:
I'll let somebody who's more familiar with those to take a look. I'm the one that broke the stuff in this PR, so it's only fair I reviewed it...

PR #530, opened 2020-09-23
open-uri: Close fd after opening dir or file

hadess:
Use new XdpFd type with g_auto to make sure the file descriptor gets
closed when the function finishes.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -844,6 +844,7 @@ handle_open_uri (XdpOpenURI *object,
   if (!g_variant_lookup (arg_options, "ask", "b", &ask))
     ask = FALSE;
 
+  g_object_set_data (G_OBJECT (request), "fd", GINT_TO_POINTER (-1));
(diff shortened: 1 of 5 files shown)


src/open-uri.c
```diff
@@ -586,7 +586,7 @@ handle_open_in_thread_func (GTask *task,
   gboolean ask_for_content_type;
   GVariantBuilder opts_builder;
   gboolean skip_app_chooser = FALSE;
-  int fd;
+  g_auto(XdpFd) fd = -1;
```
jhenstridge, line 589: I was looking through the post-1.8.0 changes in the repo, and this change looked a bit suspect.

In particular, this function still has a `close(fd)` call buried in its body, so the destructor will sometimes call close() on a stale (or reallocated) file descriptor.  Might it be better to manage the lifetime by storing the fd with g_object_set_data_full() with an appropriate destroy_notify function (i.e. something that calls close() on values >= 0).

PR #532, opened 2020-09-28
FreeBSD support

arrowd:


arrowd:
Bump. Can this be merged in, please?

src/xdp-utils.c
```diff
@@ -472,6 +482,7 @@ parse_app_info_from_flatpak_info (int pid, GError **error)
 static gboolean
 aa_is_enabled (void)
 {
+#ifdef __Linux__
```
myfreeweb, line 485: isn't only lowercase `__linux__` available?

freebsd-stubs.h
```diff
@@ -0,0 +1,14 @@
...
+
+#define O_PATH 0
+#define O_DSYNC 0
+#define O_LARGEFILE 0
+#define O_NOATIME 0
+#define O_TMPFILE 0
+#define AT_EMPTY_PATH 0
+
+// stolen from /usr/include/c++/v1/errno.h
+#define ENODATA 9919
```
myfreeweb, line 12: How about `#define ENODATA ENOATTR` ?

freebsd-stubs.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _FREEBSD_STUBS_H
+#define _FREEBSD_STUBS_H
+
+#define O_PATH 0
```
myfreeweb, line 4: Interesting note: https://github.com/flatpak/xdg-desktop-portal/blob/cb54478e07f1b51537ee75179863c02b5ef97c01/src/xdp-utils.c#L937-L941 - seems like regular fds are preferred now.

So would be better to `#ifdef O_PATH` all the support for `O_PATH`

configure.ac
```diff
@@ -78,6 +80,10 @@ PKG_CHECK_MODULES(GLIB260, glib-2.0 >= 2.60,
                   [AC_DEFINE(GLIB_VERSION_MIN_REQUIRED, GLIB_VERSION_2_60, [Ignore massive GTimeVal deprecation warnings in 2.62])],
                   [true])
 
+PKG_CHECK_MODULES(EPOLL_SHIM, epoll-shim, [true], [true])
```
myfreeweb, line 83: This shouldn't be necessary at all, looks like the `sys/eventfd.h` include is dead code, eventfd is not actually used here

freebsd-stubs.h
```diff
@@ -0,0 +1,14 @@
...
+#define O_DSYNC 0
+#define O_LARGEFILE 0
+#define O_NOATIME 0
+#define O_TMPFILE 0
+#define AT_EMPTY_PATH 0
```
myfreeweb, line 9: The `AT_EMPTY_PATH` thing wouldn't work like that, really on FreeBSD the `fstatat` calls with this arg should be changed to regular old `fstat`.

IIUC, the point of this is that `fstatat(fd, "", &thing, AT_EMPTY_PATH|AT_SYMLINK_NOFOLLOW)` means "do fstat on the `O_PATH` symlink file descriptor itself" and we don't support `O_PATH`. So `fstatat` with the empty path would blow up on FreeBSD.

freebsd-stubs.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _FREEBSD_STUBS_H
+#define _FREEBSD_STUBS_H
+
+#define O_PATH 0
+#define O_DSYNC 0
+#define O_LARGEFILE 0
+#define O_NOATIME 0
+#define O_TMPFILE 0
```
myfreeweb, line 8: How about just

```diff
--- i/document-portal/document-portal-fuse.c
+++ w/document-portal/document-portal-fuse.c
@@ -286,8 +295,10 @@ open_flags_to_string (int flags)
     g_string_append (s, ",ASYNC");
   if (flags & O_FSYNC)
     g_string_append (s, ",FSYNC");
+#ifdef O_DSYNC
   if (flags & O_DSYNC)
     g_string_append (s, ",DSYNC");
+#endif
   if (flags & O_CREAT)
     g_string_append (s, ",CREAT");
   if (flags & O_TRUNC)
@@ -298,16 +309,22 @@ open_flags_to_string (int flags)
     g_string_append (s, ",CLOEXEC");
   if (flags & O_DIRECT)
     g_string_append (s, ",DIRECT");
+#ifdef O_LARGEFILE
   if (flags & O_LARGEFILE)
     g_string_append (s, ",LARGEFILE");
+#endif
+#ifdef O_NOATIME
   if (flags & O_NOATIME)
     g_string_append (s, ",NOATIME");
+#endif
   if (flags & O_NOCTTY)
     g_string_append (s, ",NOCTTY");
   if (flags & O_PATH)
     g_string_append (s, ",PATH");
+#ifdef O_TMPFILE
   if (flags & O_TMPFILE)
     g_string_append (s, ",TMPFILE");
+#endif

   return g_string_free (s, FALSE);
 }
```

and considering my other comments, we wouldn't need a `freebsd-stubs.h` at all

document-portal/document-portal-fuse.c
```diff
@@ -1965,7 +1978,11 @@ xdp_fuse_fallocate (fuse_req_t req,
 
   g_debug ("FALLOCATE %lx", ino);
 
+#ifdef __Linux__
```
myfreeweb, line 1981: lowercase `__linux__`  here too ?

myfreeweb:
Heh, I did some changes on my own without looking at PRs, now found this :)

I managed to get the document provider to add a file to the FUSE fs, since we don't have flatpak, this involved hacks:
- `xdp_app_info_new_host` returning non-empty string (to actually test the FS — originally for the host it just returns normal paths and doesn't use the document-provider)
- commenting out `parse_app_info_from_flatpak_info` calls
- in `xdp_app_info_get_path_for_fd` changing `/proc/self/fd` to `/compat/linux/dev/fd` where I mounted `fdescfs -o linrdlnk`

Also from `fusermount_argv` I needed to remove `auto_unmount`, it fails otherwise because that's not a recognized option.

But honestly the FUSE based document provider might not be the most useful thing for us…

<details>
<summary>
if anyone cares about actual FreeBSD sandboxing read here
</summary>

…unless someone wants to do something like bubblewrap based on jails, but why would you do that when we have some better ideas for sandboxing :) That is, I'd like to use the document portal with [capsicumizer](https://github.com/myfreeweb/capsicumizer) eventually, and for that I literally just need to pass the fd to the sandboxed application. Basically all I want is for `org.freedesktop.portal.FileChooser` to return fds in response to `OpenFile` etc. But I think I know how to fit this into the existing APIs, with a custom document provider and full control of what `open()` does in the sandboxed process, I can send the fds to the process from the provider, and return a special path that would be handled by the `open()` hook.

</details>

tl;dr maybe exclude all of `document-portal` from the build on FreeBSD at least for now, the patch would be really small then

freebsd-stubs.h
arrowd: There are a lot of places where it is used. For instance, `document-portal/document-portal-fuse.c:909`:

```
static int
xdp_nonphysical_document_inode_opendir (XdpInode *inode)
{
  XdpDomain *domain = inode->domain;
  xdp_autofd int dirfd = -1;
  int res;

  g_assert (domain->type == XDP_DOMAIN_DOCUMENT);
  g_assert (inode->physical == NULL);

  dirfd = open (domain->doc_path, O_PATH | O_DIRECTORY);
  if (dirfd < 0)
  ...
```

ifdef'ing the `open` call would obviously break this code. Or did I misunderstood you?

myfreeweb: Yes, in that file, we can only define `O_PATH` to 0, but in the linked `xdp-utils.c` we can disable a whole section handling it.

(again, not sure how useful the document portal would be really, maybe just drop it)

arrowd:
Thanks for your review. I hope I've addressed all your comments.

document-portal/document-portal-fuse.c
```diff
@@ -1351,7 +1379,11 @@ xdp_fuse_getattr (fuse_req_t req,
   g_assert (domain->type == XDP_DOMAIN_DOCUMENT);
 
   if (inode->physical)
+#ifndef __FreeBSD__
```
myfreeweb, line 1382: Maybe `#ifdef __linux__` these instead of `#ifndef __FreeBSD__`

myfreeweb:
Looks good!

jhenstridge:
This looks like it substantially changes the behaviour of the file system.  An `O_PATH` file descriptor is very different from a regular file descriptor, so doing `#define O_PATH 0` is almost certainly incorrect.

For example, consider this call:

```
  o_path_fd = openat (dirfd, name, O_PATH|O_NOFOLLOW, 0);
```

If the file path being opened is a symbolic link, on Linux the call will succeed returning an `O_PATH` fd representing the symlink itself (not what the symlink points to).  On FreeBSD  the call will fail because O_NOFOLLOW on its own says to reject paths containing symbolic links.

jhenstridge:
The example I gave was from the document portal code, where this PR defines `O_PATH` to zero:

https://github.com/flatpak/xdg-desktop-portal/blob/2fe4f30/document-portal/document-portal-fuse.c#L1597-L1599

The code might compile, but it now means something different now.

arrowd:
Would `openat (dirfd, name, 0, 0); ` work, then?

myfreeweb:
maybe just disable document-portal and don't port it

jhenstridge:
My point was that it changes the meaning.  For example, the `/var/run` path on my system is a symlink to `/run`.  Here's how the two file descriptor types look (using Python for convenience):

```python
>>> fd1 = os.open('/var/run', os.O_PATH|os.O_NOFOLLOW)
>>> stat.S_ISLNK(os.fstat(fd1).st_mode)
True
>>> fd2 = os.open('/var/run', os.O_RDONLY)
>>> stat.S_ISLNK(os.fstat(fd2).st_mode)
False
>>> stat.S_ISDIR(os.fstat(fd2).st_mode)
True
```

One file descriptor represents the symlink, and the other represents the symlink target: i.e. the difference between calling `lstat` and `stat` on a path.  And the reason the code is doing all this with file descriptors rather than paths is to avoid time-of-check/time-of-use race conditions.

arrowd:
I have removed `O_PATH` changes for now. Let's at least merge trivial build fixes.

arrowd:
Bump. Can this be merged in?

arrowd:
Another bump.

unrelentingtech:
An `O_PATH` implementation patch now exists https://reviews.freebsd.org/D29323

arrowd:
I'd still prefer to get this PR merged as is first.

arrowd:
It's been more than 6 months since this MR was submitted. Is there something wrong with it? Can we finally make it in?

arrowd:
Monthly bump.

matthiasclasen:
You said you removed the O_PATH parts, but there's still a bunch of #ifdef O_PATH in here.

I think jamesh's objection about that still stands.

arrowd:
> You said you removed the O_PATH parts, but there's still a bunch of #ifdef O_PATH in here.
> 
> I think jamesh's objection about that still stands.

Yep, I overlooked these. While there, I also removed conditionals around the code using `AT_EMPTY_PATH`.

arrowd:
Bump.

arrowd:
Yearly bump.

arrowd:
Rebased again.

emaste (commented):
<Intended to submit this on the unused header removal patch specifically>
Should be easily confirmed by CI or someone building this patch on Linux?

src/xdp-utils.c
```diff
@@ -815,7 +822,7 @@ xdp_filter_options (GVariant *options,
 
           continue;
         }
-         
+
```
emaste, line 825: Unrelated whitespace cleanup

emaste:
Overall looks fine to me, configure.ac change should move earlier in the series (i.e., before the related `HAVE_XXX` use)

arrowd:
Ping.

arrowd:
Bump.

arrowd:
@smcv I wonder if you can help us get this in?

GeorgesStavracas (changes requested):
I appreciate this could be frustrating since this PR is opened for 2 years now, but I gave it a quick look and I think there are a few things you could do to make it much easier to review and land, and also some things to update:

 * Update commit titles to `prefix: Title`, see the main branch for examples
 * Add some description to the commits. They're completely empty, and it's really hard (rude, even) to force reviewers to figure out what's going on.
 * There has been changes in glibc that could impact the code that you're changing - see https://github.com/ostreedev/ostree/issues/2685. Please double check if your changes are not impacted by that, and update the PR if they are.
 * Needs Meson support now

arrowd:
> Add some description to the commits. They're completely empty ...

Huh? What do you mean by that?

GeorgesStavracas:
I mean the commits have nothing beyond their titles. Usually a good commit message is organized as follows:

```
modified-file: Short title describing your change

Context of the bug/issue/change.

The problem, when happens and why we need the change.

Overview of the fix.

Closes <Github issue link>
```

arrowd:
Rebased onto tip, addressed comments.

meson.build
```diff
@@ -20,8 +20,10 @@ endif
 
 flatpak_intf_dir = get_option('flatpak-interfaces-dir')
 if flatpak_intf_dir == ''
-    flatpak_dep = dependency('flatpak', version: '>= 1.5.0')
-    flatpak_intf_dir = flatpak_dep.get_variable(pkgconfig: 'interfaces_dir')
+    flatpak_dep = dependency('flatpak', version: '>= 1.5.0', required : false)
```
GeorgesStavracas, line 23: I think it's better to require Flatpak if `host_machine.system()` is `linux`, something like:

```meson
flatpak_required = host_machine.system() in ['linux']
flatpak_dep = dependency('flatpak', version: '>= 1.5.0', required : flatpak_required)
```

src/xdp-utils.c
GeorgesStavracas: I'm okay with this unrelated cleanup :slightly_smiling_face:

document-portal/document-portal.c
```diff
@@ -21,8 +21,6 @@
 #include "file-transfer.h"
 #include "document-portal.h"
 
-#include <sys/eventfd.h>
```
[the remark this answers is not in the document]
GeorgesStavracas: Seems like CI is happy with this change.

GeorgesStavracas (changes requested):
Just a few more comments, looks mostly good otherwise

meson.build
arrowd: Why? As a FreeBSD developer I hate any sorts of `if (linux)` conditions, as they usually don't check for they are really intended for. Why should we use it here? Flatpak can be never porter to non-Linux systems?

smcv: > Flatpak can be never porter to non-Linux systems?

Correct. Flatpak relies on user namespaces, which are a Linux kernel feature and are not portable. It also makes various other Linux assumptions.

GeorgesStavracas: The Flatpak dependency is required for the build system to fetch the Flatpak interfaces directory, which is necessary for building the documentation pages for xdg-desktop-portal. I asked you to make this change because it currently is required on Linux already, and I don't want to change that behavior, and I was okay with opening an exception to FreeBSD given the amount of work you've put into it. But on a second look, I think a better approach would be to error out if the build system has `docbook-docs` enabled and `flatpak-interfaces-dir` is not passed and the `flatpak` dependency is not present.

> Flatpak can be never porter to non-Linux systems?

I'm afraid that's not a question I can answer, I assumed FreeBSD developers like yourself would know how feasible such a port it.

arrowd: > > Flatpak can be never porter to non-Linux systems?
> 
> Correct. Flatpak relies on user namespaces, which are a Linux kernel feature and are not portable. It also makes various other Linux assumptions.

FreeBSD has its own primitives for containers, so there is nothing fundamentally preventing Flatpak from being ported.

> But on a second look, I think a better approach would be to error out if the build system has docbook-docs enabled and flatpak-interfaces-dir is not passed and the flatpak dependency is not present.

This would disable building any docs, not only Flatpak-related.

smcv: > FreeBSD has its own primitives for containers, so there is nothing fundamentally preventing Flatpak from being ported

I'm sure it would be entirely possible to do something similar to Flatpak on FreeBSD, if people want to, but one of the design decisions in bubblewrap and Flatpak is that they're Linux-specific, don't aim to be portable, and don't attempt to restrict themselves to doing only portable things. x-d-p already has a small abstraction layer for multiple app frameworks, so if FreeBSD at some point has a Flatpak-like "app container" thing with the properties that x-d-p needs, then x-d-p could support that alongside Flatpak and Snap.

arrowd:
Bump.

arrowd:
Another bump.

meson.build
```diff
@@ -113,7 +113,7 @@ pipewire_dep = dependency('libpipewire-0.3',
                           required: get_option('pipewire'))
 libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 
-bwrap = find_program('bwrap')
+bwrap = find_program('bwrap', required: false)
```
jadahl, line 116: I wonder if we should make `bwrap` required on Linux, so that it isn't mistakenly leaved out if you compile yourself.

jadahl:
These changes seems mostly harmless to me, but I can't tell whether the lack of flags added in the document portal has any unwanted consequences.

It's also surprising the commits that just ifdef some header inclusion doesn't break compilation. Are those includes needed to begin with?

ngortheone:
@arrowd thanks for your work. Will this allow us to upgrade to `1.15` version?

arrowd:
> @arrowd thanks for your work. Will this allow us to upgrade to `1.15` version?

If this ever gets merged in. Speaking of which pinging again.

ngortheone:
@smcv maybe you could help too? :pray:

GeorgesStavracas:
It will be merged when it's considered ready. Please do not post comments asking for merging.

ngortheone:
@GeorgesStavracas sorry for reaching out to you in this manner. Gleb has been working hard and diligently addressing all comments for 2 years. Please understand that all this time we have to maintain a private fork/patchset to make xdg-desktop-portal usable on FreeBSD. And following upstream becomes more difficult.

Maybe instead of gatekeeping the change and trying to make it perfect we you could grant @arrowd maintainer access to the project? It would help FreeBSD greatly to have an ambassador in freedesktop related projects.  We are as interested in keeping this project working as you are.

Thanks

GeorgesStavracas:
Asking to grant someone without previous contributions to a project full access to the repository is absolutely unreasonable. "Gatekeeping" changes and "trying to make them perfect" is standard procedure in software development, we call it "code review", and having maintainer access to the project does imply permission to push without somebody else review it.

@ngortheone honestly, I appreciate you have the best of intentions here, but dealing with your misguided commentary has devoided me of more energy than another round of reviews. @arrowd has already done an outstanding job in applying review comments and got this pull request on the verge of being merged. Pinging random people you consider important won't make this go any faster, very likely will have the opposite effect. Please, disengage.

meson.build
```diff
@@ -113,7 +113,8 @@ pipewire_dep = dependency('libpipewire-0.3',
                           required: get_option('pipewire'))
 libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 
-bwrap = find_program('bwrap')
+bswap_required = host_machine.system() in ['linux']
```
GeorgesStavracas, line 116: `bswap` → `bwrap`

GeorgesStavracas (changes requested):
Only a minor nit now

GeorgesStavracas (approved):
We've briefly discussed this on the Flatpak Matrix channel today, and generally there are no objections to landing this given that the enablement work is mostly sprinkling ifdefs around. I'd like to make it clear that landing this does not imply any kind of "official" support, and the xdg-desktop-portal project still relies on BSD folks to actively monitor the project and send pull requests if the build breaks. We're not adding any BSD-based CI step due to that not being a core requirement of the project. These patches are appreciated, and I hope you continue to contribute to the project in the future.

arrowd:
Of course, we do plan to upstream any local patches that we come up with.

Is there any ETA for the next release date that would contain this change?

GeorgesStavracas:
The 1.16 milestone is set to December 1st

PR #549, opened 2020-11-24
xdp-utils: check if alternate doc portal path matches in path_for_fd()

### LINKED ISSUE #545: OpenURI.OpenFile() fails when used with document portal paths
We received a bug report with the Thunderbird snap related to launching files via xdg-desktop-portal.  The steps the user took were:

1. add an attachment to an email (picked via a `GtkFileChooserNative` dialog)
2. try to launch the attachment from the compose window (handled [via `gtk_show_uri`](https://dxr.mozilla.org/mozilla-central/source/toolkit/system/gnome/nsGIOService.cpp#120), which in turn calls `g_app_info_launch_default`).

I was able to reproduce this with a simplified test program:

https://gist.github.com/jhenstridge/34a1813b69df7a486e509db71675f09b

When run within snap confinement, I can reproduce the error.  Based on what I found, I'd assume the same would be true for Flatpak confinement.

The file name returned by the chooser will be something like `$XDG_RUNTIME_DIR/doc/$docid/$filename`.  The test application then opens this path to create a file descriptor to pass in the `OpenFile` D-Bus call.  Due to the document portal bind mount, the file opened is equivalent to `$XDG_RUNTIME_DIR/doc/by-app/$appid/$docid/$filename` outside of the sandbox.

When x-d-p receives the file descriptor, it uses `xdp_app_info_get_path_for_fd` to map it back to a host system file name.  This determines the path the confined app opened via `/proc`, then checks to see whether the file descriptor refers to the same file in the host file system by comparing stat information.

The problem arises because the `.../doc/$docid/$filename` and `.../doc/by-app/$appid/$docid/$filename` files provided by the document portal file system have different inode numbers, leading to the OpenFile request being rejected.

There are two possible solutions to this:

1. Have `xdp_app_info_get_path_for_fd` special case document portal paths some how.
2. Have `xdg-document-portal` use the same inode numbers for files exported through the by-app hierarchy.

jhenstridge:
The document portal uses different inode number when exposing a particular document in different parts of the file system.  As sandboxed apps only have a subtree of the document portal file system mounted, the "same file" checks in xdp_app_info_get_path_for_fd() would fail for document portal paths.

To fix this, we check to see whether the corresponding "by-app/$app_id" path matches the stat information of the file descriptor.

Fixes #545

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -884,6 +884,35 @@ verify_proc_self_fd (XdpAppInfo *app_info,
   return xdp_app_info_remap_path (app_info, path_buffer);
 }
 
+static char *documents_mountpoint = NULL;
+
+void
+xdp_set_documents_mountpoint (const char *path)
+{
+  g_clear_pointer (&documents_mountpoint, g_free);
+  documents_mountpoint = g_strdup (path);
+}
+
+/* alternate_document_path converts a file path  */
+char *
+xdp_get_alternate_document_path (const char *path, const char *app_id)
+{
+  int len;
+
+  /* If we don't know where the document portal is mounted, then there
+   * is no alternate path */
+  if (documents_mountpoint == NULL)
+    return NULL;
+
+  /* If the path is not within the document portal, then there is no
+   * alternative path */
+  len = strlen (documents_mountpoint);
+  if (!g_str_has_prefix (path, documents_mountpoint) || path[len] != '/')
+    return NULL;
+
+  return g_strconcat (documents_mountpoint, "/by-app/", app_id, &path[len], NULL);
+}
+
 char *
 xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
                               int fd,
@@ -981,8 +1010,22 @@ xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
       st_buf->st_dev != real_st_buf.st_dev ||
       st_buf->st_ino != real_st_buf.st_ino)
     {
-      /* Different files on the inside and the outside, reject the request */
-      return NULL;
+      /* If the path is provided by the document portal, the inode
+         number will not match, due to only a subtree being mounted in
+         the sandbox.  So we check to see if the equivalent path
+         within that subtree matches our file descriptor.
+
+         If the alternate path doesn't match either, then we treat it
+         as a failure.
+      */
+      g_autofree char *alt_path = NULL;
+      alt_path = xdp_get_alternate_document_path (path, xdp_app_info_get_id (app_info));
+
+      if (alt_path == NULL ||
+          stat (alt_path, &real_st_buf) < 0 ||
+          st_buf->st_dev != real_st_buf.st_dev ||
+          st_buf->st_ino != real_st_buf.st_ino)
+        return NULL;
     }
 
   if (writable_out)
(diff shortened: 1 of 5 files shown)


jhenstridge:
With this patch applied, the test program I referenced in the bug functions correctly when under snap confinement.  I'd assume the same is true for Flatpak confinement.

src/xdp-utils.c
```diff
@@ -981,8 +1011,22 @@ xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
...
-      return NULL;
+      /* If the path is provided by the document portal, the inode
+         number will not match, due to only a subtree being mounted in
+         the sandbox.  So we check to see if the equivalent path
+         within that subtree matches our file descriptor.
+
+         If the alternate path doesn't match either, then we treat it
+         as a failure.
+      */
+      g_autofree char *alt_path = alternate_document_path
```
hadess, line 1022: Splitting this in two lines would probably help with the readability.

src/xdp-utils.c
```diff
@@ -884,6 +884,36 @@ verify_proc_self_fd (XdpAppInfo *app_info,
   return xdp_app_info_remap_path (app_info, path_buffer);
 }
 
+static char *documents_mountpoint = NULL;
+
+void
+xdp_set_documents_mountpoint (const char *path)
+{
+  g_clear_pointer (&documents_mountpoint, g_free);
```
hadess, line 892: Why do we need to allow this variable to be set multiple times?

src/xdp-utils.c
```diff
@@ -981,8 +1011,22 @@ xdp_app_info_get_path_for_fd (XdpAppInfo *app_info,
...
+         within that subtree matches our file descriptor.
+
+         If the alternate path doesn't match either, then we treat it
+         as a failure.
+      */
+      g_autofree char *alt_path = alternate_document_path
+        (path, xdp_app_info_get_id (app_info));
+
+      if (alt_path == NULL ||
+          stat(alt_path, &real_st_buf) < 0 ||
```
hadess, line 1026: Space before `(`

src/xdp-utils.c
```diff
@@ -884,6 +884,36 @@ verify_proc_self_fd (XdpAppInfo *app_info,
...
+xdp_set_documents_mountpoint (const char *path)
+{
+  g_clear_pointer (&documents_mountpoint, g_free);
+  documents_mountpoint = g_strdup (path);
+}
+
+
+/* alternate_document_path converts a file path  */
+char *
+alternate_document_path(const char *path, const char *app_id)
```
hadess, line 899: Might be useful to have a simple unit test to test this codepath. Don't forget to add a space before the parenthesis.

hadess (changes requested):

src/xdp-utils.c
jhenstridge: It isn't necessary for xdg-desktop-portal or xdg-document-portal.  It turned out to be useful in writing the test, so I've left it as is.

src/xdp-utils.c
jhenstridge: I've added a test to `test-xdp-utils.c`.

hadess:
Looks correct, if @matthiasclasen or @alexlarsson want to chime in.

alexlarsson:
LGTM, thanks @jhenstridge !

Issue #363, opened 2019-09-16
file handles


sonnyp:
Here is my use case; I basically made a browser, I use `GtkFileChooserNative` to allow users to set the download destination. Once the file is downloaded I trigger a notification with 2 actions "Open file" and "Open folder" that use "GioAppInfo.launch_default_for_uri".

See https://github.com/sonnyp/Tangram/pull/54

Unfortunately with flatpak that won't work because `GtkFileChooserNative.get_uri` returns something like `file:///run/user/1000/doc/90e34d70/MyFile.txt`.

I know this description is very Gtk oriented but I couldn't find any way to achieve that in the spec anyway.

Basically what I need is some kind of handle I could pass to various APIs that open trusted processes ?

matthiasclasen:
I'm not entirely clear what you are trying to achieve, and what the problem is.

Your app is a flatpak that downloads files ?

And it tries to open the files using other flatpak'ed apps ?

sonnyp:
> Your app is a flatpak that downloads files ?

Yes.

> And it tries to open the files using other flatpak'ed apps ?

Exactly. 

And I'd like to do so without any filesystem permission, just using portals.

matthiasclasen:
If you use the file chooser portal to save a file outside the sandbox, then yes,the location you get will be something like file:///run/user/1000/doc/90e34d70/MyFile.txt, because we use the document portal to make the user-chosen location available for writing to your sandboxed app.

I'm not sure I understand where the problem is.

sonnyp:
> If you use the file chooser portal to save a file outside the sandbox, then yes,the location you get will be something like file:///run/user/1000/doc/90e34d70/MyFile.txt, because we use the document portal to make the user-chosen location available for writing to your sandboxed app.

Yep makes sense.

> The canonical way to implement 'Open file' inside a flatpak app would be to use the OpenURI portal.

Yep but that doesn't work with a "sandbox" `file://` URI. Right?

> I'm not sure I understand where the problem is.

Unless I'm missing/misunderstanding something I guess it's more of a feature request than a problem.

I would like to be able to save a file **and** open it using portals. Which implies that

* my app doesn't know the real path of the file
* OpenURI portal (or equivalent) would somehow be able to map that "sandbox" `file:` URI to the real host path URI.

matthiasclasen:
> > The canonical way to implement 'Open file' inside a flatpak app would be to use the OpenURI portal.
> 
[...]

PR #552, opened 2020-12-16
Use real filesystem path for OpenDirectory

YaLTeR:
Before this change the file manager would open inside the document portal directory like `/run/user/1000/doc/10288756` which is not very user-friendly. With this change it opens in the real filesystem path. Now actions like "Show in Files" can be properly implemented.

cc https://github.com/flatpak/xdg-desktop-portal/issues/363

### DIFF
--- a/src/documents.c
+++ b/src/documents.c
@@ -200,3 +200,28 @@ register_document (const char *uri,
   doc_path = g_build_filename (documents_mountpoint, doc_id, basename, NULL);
   return g_filename_to_uri (doc_path, NULL, NULL);
 }
+
+char *
+get_real_path_for_doc_path (const char *path,
+                            const char *app_id)
+{
+  g_autofree char *doc_id = NULL;
+  gboolean ret = FALSE;
+  char *real_path = NULL;
+
+  if (app_id == NULL || *app_id == '\0')
+    return g_strdup (path);
+
+  ret = xdp_documents_call_lookup_sync (documents, path, &doc_id, NULL, NULL);
+  if (!ret)
+    return g_strdup (path);
+
+  if (!g_strcmp0 (doc_id, ""))
+    return g_strdup (path);
+
+  ret = xdp_documents_call_info_sync (documents, doc_id, &real_path, NULL, NULL, NULL);
+  if (!ret)
+    return g_strdup (path);
+
+  return real_path;
+}
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -668,8 +668,10 @@ handle_open_in_thread_func (GTask *task,
 
       if (open_dir)
         {
-          char *dir = g_path_get_dirname (path);
+          char *real_path = get_real_path_for_doc_path (path, app_id);
           g_free (path);
+          char *dir = g_path_get_dirname (real_path);
+          g_free (real_path);
           path = dir;
         }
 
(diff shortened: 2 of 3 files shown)


src/documents.c
```diff
@@ -200,3 +200,29 @@ register_document (const char *uri,
...
+char *
+get_real_path (const char *path,
+               const char *app_id,
+               GError **error)
+{
+  g_autofree char *doc_id = NULL;
+  gboolean ret = FALSE;
+  char *real_path = NULL;
+
+  if (app_id == NULL || *app_id == 0)
```
hadess, line 213: This should probably be `*app_id == '\0'` right?

src/open-uri.c
```diff
@@ -668,8 +668,11 @@ handle_open_in_thread_func (GTask *task,
 
       if (open_dir)
         {
-          char *dir = g_path_get_dirname (path);
+          // Get the real path so the file manager opens the real dir and not the document portal dir.
```
hadess, line 671: We don't use C++ style comments. This comment shouldn't be necessary if the name of the function was clearer.

src/documents.c
```diff
@@ -200,3 +200,29 @@ register_document (const char *uri,
...
+    return g_strdup (path);
```
hadess, line 214: Maybe the function should return `NULL` if there's an error?

src/documents.c
```diff
@@ -200,3 +200,29 @@ register_document (const char *uri,
   doc_path = g_build_filename (documents_mountpoint, doc_id, basename, NULL);
   return g_filename_to_uri (doc_path, NULL, NULL);
 }
+
+char *
+get_real_path (const char *path,
+               const char *app_id,
+               GError **error)
```
hadess, line 207: You can remove the error argument if it's never going to be used.

src/documents.c
```diff
@@ -200,3 +200,29 @@ register_document (const char *uri,
   doc_path = g_build_filename (documents_mountpoint, doc_id, basename, NULL);
   return g_filename_to_uri (doc_path, NULL, NULL);
 }
+
+char *
+get_real_path (const char *path,
```
hadess, line 205: `get_path_for_doc_path()`?

hadess (changes requested):

src/documents.c
YaLTeR: If the path is not a document path then it can be returned as-is because it's already "good".

YaLTeR:
> Any chance you could make a change that would exercise this new code path in the test app in https://github.com/flatpak/libportal ?

Done: https://github.com/flatpak/libportal/pull/35

YaLTeR:
Anything else I should do?

src/documents.c
```diff
@@ -200,3 +200,28 @@ register_document (const char *uri,
...
+  char *real_path = NULL;
+
+  if (app_id == NULL || *app_id == '\0')
+    return g_strdup (path);
+
+  ret = xdp_documents_call_lookup_sync (documents, path, &doc_id, NULL, NULL);
+  if (!ret)
+    return g_strdup (path);
+
+  if (!g_strcmp0 (doc_id, ""))
```
matthiasclasen, line 219: I prefer g_strcmp0 (doc_id, "") == 0, but not a big deal

PR #561, opened 2021-02-03
Settings: do not put failed implementations into the list

grulja:
With settings portal ability to load multiple backend implementations, there is a possibility that one of them will not load correctly and be placed into the list anyway. This results into failed implementation being used, we will call Read/ReadAll on it and it will fail and crash. To avoid that, we will create a temporary list, check real number of working implementations and put them into the final list we will use later.

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -45,7 +45,7 @@ struct _SettingsClass
 };
 
 static XdpImplSettings **impls;
-static int n_impls;
+static int n_impls = 0;
 
 GType settings_get_type (void) G_GNUC_CONST;
 static void settings_iface_init (XdpSettingsIface *iface);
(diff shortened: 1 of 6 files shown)


grulja:
Alternative and much simpler solution (thought not that correct), would be just add:
```
if (impls[i] == NULL)
  continue;
```
into both for cycles in s**ettings_handle_read()** and **settings_handle_read_all()**.

grulja:
The test now fails, because when there is no backend implementation loaded, I will make it fail. Looks I will need to extend tests with some backend implementation.

src/settings.c
```diff
@@ -176,29 +176,52 @@ settings_create (GDBusConnection *connection,
...
+  impls = g_new (XdpImplSettings *, n_impls);
+
+  for (i = 0, j = 0; i < n_impls_tmp && j < n_impls; i++, j++)
+    {
+      if (impls_tmp[i] != NULL)
+        impls[j] = impls_tmp[i];
     }
 
+  g_free(impls_tmp);
+
```
matthiasclasen, line 225: I think a much simpler approach is sufficient here. No need to loop twice. Just allocate impls to the full length, fill it up with as many non-NULL objects as we get, and set n_impls accordingly. Nevermind that the array has some extra room in the end.


```
impls = g_new (XdpImplSettings, n);
for (i = 0; j = 0; i < n; i++)
  {
      impl = xdp_impl_settings_proxy_new_sync (...);
      if (impl)
        impls[j++] = impl;
  }
n_impls = j;
```

matthiasclasen: Thanks for adding a test!

PR #563, opened 2021-02-21
Regenerate docs for version 1.8.1

takluyver:
By doing:

```
git checkout 1.8.1
./autogen.sh --enable-docbook-docs
make
```

& taking the HTML file from the doc/ folder.


takluyver:
I see the docs are now updated by a CI workflow (#678), so this is no longer needed.

PR #566, opened 2021-03-04
openuri: add an activation_token option

emersion:
This allows the original client to give an activation token to the
spawned client.

See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/50


data/org.freedesktop.portal.OpenURI.xml
```diff
@@ -71,6 +71,14 @@
...
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem><para>
+              A token that can be used to activate the chosen application.
+            </para><para>
+              The activation_token option was introduced in version 4 of the interface.
+            </para></listitem>
+          </varlistentry>
```
jadahl, line 81: Probably want this as part of the app chooser spec too.

emersion: Done

jadahl:
I guess this also needs some plumbing to the app chooser dialog, but documenting is a good start.

data/org.freedesktop.impl.portal.AppChooser.xml
```diff
@@ -62,6 +64,14 @@
             <listitem><para>The filename to choose an application for. Note that this
                is just a basename, without a path.</para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem><para>
+              A token that can be used to activate the file chooser.
```
matthiasclasen, line 70: Is it a typo that this talks about a file chooser ?

emersion: Ah, indeed, "file chooser" is wrong here. But as far as I understand, since this is the backend API, this D-Bus call will just present the file chooser to the user, and won't actually start it.

I've updated this option to refer to "the application chooser" instead.

And I've added an activation token in results as well, which was missing. This is necessary in case choosing an application has required user interaction: the portal impl should generate a new activation token from that interaction.

emersion:
Pushed a new commit to add the necessary glue code to xdg-desktop-portal itself.

emersion:
Gentle ping

emersion:
Thanks for merging!

PR #568, opened 2021-03-11
Add a security policy

mwleeds:


mwleeds:
As far as I know x-d-p doesn't have a stable series that's officially supported. cc @matthiasclasen ?

smcv:
I thought x-d-p used odd/even versioning, like flatpak, GLib and dbus?

If that's the case, then the supported branch is 1.8.x (and maybe an older branch if there's someone supporting it), and there's currently no development branch?

mwleeds:
> I thought x-d-p used odd/even versioning, like flatpak, GLib and dbus?
> 
> If that's the case, then the supported branch is 1.8.x (and maybe an older branch if there's someone supporting it), and there's currently no development branch?

Looks like you're right based on the release history. I added that info to the doc here but I don't have permissions to merge it.

Issue #579, opened 2021-04-29
Application ID is empty when an host app uses portals


carlocastoldi:
Like the title says, I found myself in need (and will) to use the `Inhibit` portal from a not yet flatpak-_able_ application.
Unlike many other portal methods `Inhibit` doesn't take an `app_id` as parameter of the call, but the backend API presents it as a parameter **without** even specifying anywhere in the docs (I couldn't find it) that `app_id` is an empty string when `Inhibit` is used.
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/xdp-utils.c#L150

This clashes with [xdg-desktop-portal-gtk](https://github.com/flatpak/xdg-desktop-portal-gtk) because Gtk requires a not-empty Application ID in order to grant the inhibition permission, making the usage of the portal practically unusable on GNOME by an host app.

From what @matthiasclasen told me, the portal relies on the unfakeable ID gotten from flatpak (or snap?) and not from the app-id specified in the program. The reason for this. as far as i know, appears to be that sometimes app id is used (by whom?) as a key for storing things and passing it outside of a sandbox could be a security threat.
* Why is it a problem to pass the ID as input to portals such as `Inhibit`, `Screenshot` or `ScreenCast`, while for some others such as `Documents`, `FileChooser`, `AppChooser` or `Print` it isn't?
* If I wanted to make a pull request to change the above-mentioned line in `xdp-utils.c` without changing Inhibit's interface, would I be able to retrieve the app ID from the window's handle? If yes, how? I would be ok to start by doing it for Wayland's application windows only. 
* If you think it is _really_ impossible and an hassle to handle this case, should I try to make a pull request for [xdg-desktop-portal-gtk](https://github.com/flatpak/xdg-desktop-portal-gtk) so that it retrieves the ID from the handle or—if it is not possible—it uses a placeholder ID (like in [`gnome-session-inhibit`](https://gitlab.gnome.org/GNOME/gnome-session/blob/master/doc/man/gnome-session-inhibit.xml#L62))? I know the latter wouldn't be a beautiful solution, but it's better then denying it altogether.
* Can you update the [docs ](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html) so that it is clear that `app_id` is empty for some portal backend API? I have spent many hours investigating this problem and I would like if others didn't do my same mistake in the future.

### Why am I troubling so much using portals? I could _just use XYZ_!
[...]

PR #581, opened 2021-05-03
[WIP] Retrieve a not-empty Application ID for un-sendboxed clients that use cgroups

carlocastoldi:
Following #579, I present a simple but yet effective solution to retrieve the Application ID from PID's cgroup.
The usage of cgroups to manage processes by DEs is an innovative approach pushed forward by both KDE & GNOME, and uses the application's ID to distinguish the various resources.

I thus intend to check if the cgroup of the client is compliant to the [standard](https://systemd.io/DESKTOP_ENVIRONMENTS/) and, if it is, use the contained ID. Otherwise proceed with how we have always done: assigning `""` as `app_id`.

You can test this patch by running [my demo](https://gist.github.com/carlocastoldi/b52a3434c36b8bf185afe4a545ff90d8) with an appropriate `.desktop` file or with one of the two following command:
```
systemd-run --user --scope -u app-com.myapp.InhibitorTest-`uuidgen | sed s/-//g`.scope   ./xdg-portal-inhibitor-test-gtk4
```
```
systemd-run --user -u app-com.myapp.InhibitorTest@`uuidgen | sed s/-//g`.service   ./xdg-portal-inhibitor-test-gtk4
```

### Currently not working
An application run without  cgroups isn't recognised as such and an un-meaningful `app_id` is passed down.


swick:
This is a good change IMO. However some code doesn't use XdpAppInfo's XDP_APP_INFO_KIND_HOST to determine that the client is non-sandboxed but instead just checks the app_id for NULL. Non-sandboxed clients for which the app id can be determined then get the same treatment as sandboxed clients.

carlocastoldi:
> However some code doesn't use XdpAppInfo's XDP_APP_INFO_KIND_HOST to determine that the client is non-sandboxed but instead just checks the app_id for NULL.

In the current implementation when would the `app_id` be NULL? As of now shouldn't it be set to `""` for host applications?

To avoid changing the code without being sure, I post here the potential tests that perhaps could look at `XdpAppInfo`'s `kind`:https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-store.c#L64
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L331
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L369
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L532 (doubt)
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L230 (doubt. There is a multitude of similar tests)
https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/wallpaper.c#L184 (doubt)

Additionally there are many checks on `app_id` in `document-portal/document-portal.c`, but I am not sure if they are actually testing if the client is an host or just that if the app_id is empty/NULL (similarly to `src/wallpaper.c` above).

carlocastoldi:
> Non-sandboxed clients for which the app id can be determined then get the same treatment as sandboxed clients.

Where could this not be desirable?

Erick555:
I guess in case of document portal trying unsandboxed apps to use it isn't desirable although the portal will fail anyway as it won't be able to determine app file permissions which is unfortunately also true for sandboxed but non-flatpak/snap apps.

src/xdp-utils.c
```diff
@@ -143,11 +144,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+get_app_id_from_pid (pid_t pid, XdpAppInfo *app_info)
+{
+  char *unit, **splits, **app_id;
+  int res, len;
+
+  res = sd_pid_get_user_unit(pid, &unit);
+  if (res == -ENODATA || res < 0 || !g_str_has_prefix (unit, "app-"))
+    // the session might not be managed by systemd
+    // or there could be error fetching own systemd units
+    // or the unit might not be started by the the desktop environment (e.g. it's a script run from terminal)
```
matthiasclasen, line 181: Please use /* */ and put the comment before the if

src/xdp-utils.c
```diff
@@ -143,11 +144,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+    {
+      app_info->id = g_strdup ("");
+      return;
+    }
+  splits = g_strsplit (unit, "-", -1);
+  len = g_strv_length (splits);
+  if (g_str_has_suffix (unit, ".service"))
+    // format: app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
+    app_id = &g_strsplit (splits[len-1], "@", 2)[0];
```
matthiasclasen, line 195: this leaks

src/xdp-utils.c
```diff
@@ -143,11 +144,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  res = sd_pid_get_user_unit(pid, &unit);
+  if (res == -ENODATA || res < 0 || !g_str_has_prefix (unit, "app-"))
+    // the session might not be managed by systemd
+    // or there could be error fetching own systemd units
+    // or the unit might not be started by the the desktop environment (e.g. it's a script run from terminal)
+    {
+      app_info->id = g_strdup ("");
+      return;
+    }
+  splits = g_strsplit (unit, "-", -1);
```
matthiasclasen, line 186: splits leaks

src/xdp-utils.c
```diff
@@ -143,11 +144,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  len = g_strv_length (splits);
+  if (g_str_has_suffix (unit, ".service"))
+    // format: app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
```
matthiasclasen, line 194: here too: please use /* */ and put the comment before the if

src/xdp-utils.c
```diff
@@ -143,11 +144,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  app_info->id = g_strdup (*app_id);
+}
+
 static XdpAppInfo *
-xdp_app_info_new_host (void)
+xdp_app_info_new_host (pid_t pid)
 {
   XdpAppInfo *app_info = xdp_app_info_new (XDP_APP_INFO_KIND_HOST);
-  app_info->id = g_strdup ("");
+  get_app_id_from_pid(pid, app_info);
```
matthiasclasen, line 204: please put a space before (

matthiasclasen:
> > However some code doesn't use XdpAppInfo's XDP_APP_INFO_KIND_HOST to determine that the client is non-sandboxed but instead just checks the app_id for NULL.
> 
> In the current implementation when would the `app_id` be NULL? As of now shouldn't it be set to `""` for host applications?
> 
> To avoid changing the code without being sure, I post here the potential tests that perhaps could look at `XdpAppInfo`'s `kind`:
> 
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-store.c#L64

We should pass down the appinfo here.

> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L331
> 
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L369
> 
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/location.c#L532
> (doubt)

Here too.

> 
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L230
> (doubt. There is a multitude of similar tests)
>
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/src/wallpaper.c#L184
> (doubt)

Not sure it makes a difference here, but we have the app_info right there.

carlocastoldi:
> @matthiasclasen: _"We should pass down the appinfo here."_

I was trying to do so, bu is there a simple way to retrieve XdgAppInfo from XdgDomain or from `xdp_fuse_lookup()`? If yes, can I change the domain struct so that it has an appinfo instead of the `app_id`, keeping the same semantic (i.e., `NULL for root, by-app, non-app id`)?

Problem is that I don't think I am able to retrieve the AppInfo from there. I can see where the app_id is coming from, but atm I think I lack the knowledge in order to get something else.
I think I need to modify the XdpDomain because, by trying to pass down the appinfo, I followed the following path:
* `document-store.c/document_entry_get_permissions()`
  * `document-store.c/document_entry_has_permissions()`
    * `document-portal-fuse.c/app_can_see_doc()`
      * `ensure_doc_inode()`
        * `xdp_fuse_lookup()`
          * https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L3026

P.S. similarly here: https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L1727

Makefile.am
```diff
@@ -27,6 +27,7 @@ CLEANFILES += $(dbus_service_DATA)
 CLEANFILES += $(systemduserunit_DATA)
 EXTRA_DIST += $(service_in_files)
 
+CFLAGS = -lsystemd
```
smcv, line 30: This should become a check for libsystemd in the build system (Flatpak has a suitable check in its `configure.ac`), keeping libsystemd optional. If compiled without libsystemd support, the answer should always be "I don't know what app this is".

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+decode_hex_escapes (char *str[])
+{
+  char hex[] = "0xhh";
+
+  for (int i=0; (*str)[i]!='\0' && (*str)[i+1]!='\0'; ++i)
+  {
+    if ((*str)[i] == '\\' && (*str)[i+1] == 'x')
+      {
+        hex[2] = (*str)[i+2];
+        hex[3] = (*str)[i+3];
```
smcv, line 158: If `hex[2]` is `\0` then this will be out of bounds.

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+        char unescaped = (char) strtol (hex, NULL, 16);
+        (*str)[i+2] = '\\';
+        (*str)[i+3] = unescaped;
+        i += 3;
+      }
+  }
+  char **splits = g_strsplit(*str, "\\x\\", -1);
+  g_free (*str);
+  *str = g_strjoinv ("", splits);
+  g_strfreev (splits);
```
smcv, line 168: After going to heroic lengths to do the decoding in-place with no extra allocations, this does a lot of malloc/free cycles at the end...

I think a better approach to this would be to take a `const char *` argument and iterate through it, appending characters to a `GString` buffer, returning a newly allocated string with `g_string_free (gstr, FALSE)`.

Makefile.am
GeorgesStavracas, line 30: Perhaps make it a compile-time optional, enabled by default?

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  return;
+}
+
+void
+get_app_id_from_pid (pid_t pid, XdpAppInfo *app_info)
+{
+  char *unit, **dash_splits, **at_splits, **app_id;
+  int res, len;
+
+  res = sd_pid_get_user_unit(pid, &unit);
```
GeorgesStavracas, line 178: Style nit: missing space before `(`

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  /*
+   * the session might not be managed by systemd
+   * or there could be error fetching own systemd units
+   * or the unit might not be started by the the desktop environment (e.g. it's a script run from terminal)
+   */
```
GeorgesStavracas, line 183: I don't think this comment is necessary - probably better to mention these cases in the commit message?

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  if (res == -ENODATA || res < 0 || !g_str_has_prefix (unit, "app-"))
```
GeorgesStavracas, line 184: `g_str_has_prefix` is not NULL-safe, so `!unit || !g_str_has_prefix (...)`

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+    app_id = &dash_splits[len-2];
+  /* format: app[-<launcher>]-<ApplicationID>[@<RANDOM>].service */
+  else {
+    at_splits = g_strsplit (dash_splits[len-1], "@", 2);
+    app_id = &at_splits[0];
+  }
+  decode_hex_escapes (app_id);
+  app_info->id = g_strdup (*app_id);
+  g_strfreev(at_splits);
+  g_strfreev (dash_splits);
```
GeorgesStavracas, line 206: You can remove these calls by declaring these variables as `g_auto(GStrv) at_splits = NULL` and `g_auto(GStrv) dash_splits = NULL`

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+    {
+      app_info->id = g_strdup ("");
+      return;
+    }
+  dash_splits = g_strsplit (unit, "-", -1);
+  len = g_strv_length (dash_splits);
+  /*
+   * format:
+   * app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
+   * app[-<launcher>]-<ApplicationID>-autostart.service -> no longer true since systemd v248
```
smcv, line 194: I think instead of relying on the part after the app-ID to not contain any additional `-` (e.g. `app-gnome-com.example.Foobar-123-456.service` would break this logic), it might make more sense to assume that the `<launcher>` does not contain dots (in practice it's something like `gnome`, `flatpak` or `steam`, which doesn't), so that the logic could be:

* remove trailing `.service`, `.slice` or `.scope`
* remove an `@` and anything after it, if present
* split into dash-separated components
* if second component contains at least two dots, it's the app ID
* else if third component contains at least two dots, it's the app ID

src/xdp-utils.c
```diff
@@ -144,11 +145,72 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+   */
+  if (g_str_has_suffix (unit, ".scope") || g_str_has_suffix (unit, "-autostart.service"))
+    app_id = &dash_splits[len-2];
+  /* format: app[-<launcher>]-<ApplicationID>[@<RANDOM>].service */
+  else {
+    at_splits = g_strsplit (dash_splits[len-1], "@", 2);
+    app_id = &at_splits[0];
+  }
```
GeorgesStavracas, line 202: It's better to check `if (g_str_has_suffix (unit, ".service"))` here, and let the `else` condition set the app id to an empty string and return

GeorgesStavracas (changes requested):

src/xdp-utils.c
smcv, line 178: `unit` is leaked, at the moment.

smcv (commented):
I agree with @matthiasclasen that using `XdpAppInfo` everywhere is a prerequisite for doing this.

I think if someone wants to move this forward (@GeorgesStavracas? @mwleeds?) then the way to do it would be:

* first, go through all the places that use an app ID for decisions, and make sure they get an `XdpAppInfo` instead, with `kind == XDP_APP_INFO_KIND_HOST` instead of `g_str_equal (app_id, "")` as the way to tell that the caller is un-sandboxed;
* then, reapply changes similar to these on top of that refactoring

mwleeds:
> > > However some code doesn't use XdpAppInfo's XDP_APP_INFO_KIND_HOST to determine that the client is non-sandboxed but instead just checks the app_id for NULL.
> > 
> > 
> > In the current implementation when would the `app_id` be NULL? As of now shouldn't it be set to `""` for host applications?
> > To avoid changing the code without being sure, I post here the potential tests that perhaps could look at `XdpAppInfo`'s `kind`:
> > https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-store.c#L64
> 
> We should pass down the appinfo here.
> 

Seems to me we need two variants of `document_entry_get_permissions()`, one that operates on a `XdpAppInfo` and one on an app id, because we only have the app id available for the app being granted permissions in `portal_grant_permissions()` (see `target_app_id`) not an `XdpAppInfo`. Conceptually though I'm not sure when one sandboxed app would be granting another sandboxed app permissions to a document (or maybe it's intended for use by unsandboxed processes).

mwleeds:
> > @matthiasclasen: _"We should pass down the appinfo here."_
> 
> I was trying to do so, bu is there a simple way to retrieve XdgAppInfo from XdgDomain or from `xdp_fuse_lookup()`? If yes, can I change the domain struct so that it has an appinfo instead of the `app_id`, keeping the same semantic (i.e., `NULL for root, by-app, non-app id`)?
> 
> Problem is that I don't think I am able to retrieve the AppInfo from there. I can see where the app_id is coming from, but atm I think I lack the knowledge in order to get something else. I think I need to modify the XdpDomain because, by trying to pass down the appinfo, I followed the following path:
> 
> * `document-store.c/document_entry_get_permissions()`
>   
>   * `document-store.c/document_entry_has_permissions()`
>     
>     * `document-portal-fuse.c/app_can_see_doc()`
>       
>       * `ensure_doc_inode()`
>         
>         * `xdp_fuse_lookup()`
>           
>           * https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L3026
> 
> P.S. similarly here:
> 
> https://github.com/flatpak/xdg-desktop-portal/blob/89d2197002f164d02d891c530dcaa2808f27f593/document-portal/document-portal-fuse.c#L1727

Looks like there is no need to change `document-portal-fuse.c` to use `XdpAppInfo` because it's not making any decisions based on `xdp_app_info_get_id` returning `NULL` or `""`.  It is only ever working with valid non-empty app IDs for `XDP_DOMAIN_BY_APP` and `XDP_DOMAIN_APP`, and for `XDP_DOMAIN_ROOT` the access is not dependent on app ID.

But we can test the document portal with these patches (when they're ready) just to be sure.

Erick555:
`it would be a bit pointless since the document portal doesn't use the fuse filesystem to expose files to unsandboxed callers, and passes them the host path directly instead.`

There may be sandboxed but not recognized as such callers though. I wonder if there could be a opt-in switch to force document portal to use fuse filesystem for those.

mwleeds:
Preparatory PR for this is here: #718

mwleeds:
> > it would be a bit pointless since the document portal doesn't use the fuse filesystem to expose files to unsandboxed callers, and passes them the host path directly instead.
> 
> There may be sandboxed but not recognized as such callers though. I wonder if there could be a opt-in switch to force document portal to use fuse filesystem for those.

In a sense there already is a switch in that you can create `/.flatpak-info` in the filesystem of a process and it will be treated as sandboxed, which WebKitGTK already does, but it's a bit of a hack. Alternatively the document portal could check if the calling process has access to a file. It already does that for sandboxed callers using `/proc/self/fd`. Probably best to open a separate issue to discuss this if you want, since it would be an independent change from the ones here.

Erick555:
The `/.flatpak-info` or `GTK_USE_PORTAL=1` won't cut it for document portal because the portal assumes `sandbox==flatpak` and consults flatpak metadata for file permissions instead of checking if actual file exist in sandbox.

https://github.com/flatpak/xdg-desktop-portal/blob/main/document-portal/document-portal.c#L648

https://github.com/flatpak/xdg-desktop-portal/blob/main/document-portal/document-portal.c#L569

Obviously the former will always fail for non-flatpak sandbox and portal will pass the host path even if it doesn't exist in sandbox.

The issue for that exist: https://github.com/flatpak/xdg-desktop-portal/issues/680

src/xdp-utils.c
mwleeds: Seems like a helpful comment to me

src/xdp-utils.c
mwleeds: I came up with a regex that seems to work for this: `"^app-(?:[[:alnum:]]+\\-)?(.+?\\..+?\\..+?)(?:[@\\-][0-9]*)?(?:-autostart)?(?:\\.scope|\\.service|\\.slice)$"`

But unfortunately the unit returned by systemd for gnome-software (which runs as `/usr/bin/gnome-software --gapplication-service`) is `app-gnome-gnome\x2dsoftware\x2dservice-9502.scope` (where 9502 is the pid in my case) so parsing that in any way is not going to get us the correct app ID of `org.gnome.Software`.

mwleeds:
Closing this in favor of #719 which is in better shape but also not ready to be merged.

Issue #592, opened 2021-06-15
Flatpak: file paths in /var/data, etc. not handled by OpenURI portal


leamas:
Our application OpenCPN (www.opencpn.org) uses plugins which are dl-loaded flatpak extensions. Some of these have a need to display a local file like Help.html using the host browser. This should work using the sandboxed xdg-open if I have understood this, but it doesn't.

A test case: after loading our Shipdriver plugin we have a file test.html available in the sandbox as  /var/data/opencpn/plugins/ShipDriver_pi/data/test.html. If this file is fed to xdg-open in the sandbox it silently fails to display the file.

OTOH, displaying a file from the main application like /app/share/opencpn/authors.html  works just fine.

---
OS: Fedora 33
Flatpak: 1.10.2

Steps to reproduce (requires a x86_64 box):
  - Start org.opencpn.OpenCPN using `--command=bash`
  - Start application: `$ /app/bin/opencpn &`
  - In the application, click the top-left Options icon, the one under the hamburger.
  - Select the Plugins tab
  - Push the Import Plugin button
  - Import the attached plugin
  - In the list above the Import button, scroll down to the Shipdriver plugin
  - Push the Install button.
  - Check the "enabled" checkbox.
  - In the sandbox shell, try to open the file test.html:
    `$  xdg-open /var/data/opencpn/plugins/ShipDriver_pi/data/test.html`
  - Watch nothing happen

Expected results:
  - The test file should be displayed
  
Actual results:
   - Nothing displayed, return code zero, no error messages. 

[ShipDriver-2.6.3.0-+2106151524.f666917_flatpak-18.08-x86_64.tar.gz](https://github.com/flatpak/xdg-desktop-portal/files/6657134/ShipDriver-2.6.3.0-%2B2106151524.f666917_flatpak-18.08-x86_64.tar.gz)

smcv:
> we have a file test.html available in the sandbox as `/var/data/opencpn/plugins/ShipDriver_pi/data/test.html`

Oh, I might have misunderstood you on the mailing list. Are these plugins downloaded at runtime by the app itself, rather than being Flatpak extensions that are downloaded out-of-band by Flatpak?

I think this might be a problem that is specific to using `/var/data`. If you access the file via `$XDG_DATA_HOME` (which expands to something like `/home/you/.var/org.opencpn.OpenCPN/data` and is a separate mount point for the same directory as `/var/data`), does it work?

[...]

PR #593, opened 2021-06-15
utils: Remap Flatpak /var/config, /var/data

smcv:
In a Flatpak app, /var/config and /var/data are parallel mount points for
    the same directories as $XDG_CONFIG_HOME and $XDG_DATA_HOME, which are
    canonically under ~/.var/app/$FLATPAK_ID. /var/config and /var/data
    don't exist as conventional FHS paths, so it seems safe to assume that
    if they appear inside a Flatpak app, it's these aliases that are meant.

/var/cache and /var/tmp are not handled here, because those are
    ambiguous and will need a different code structure to handle: they're
    usually aliases for $XDG_CACHE_HOME and $XDG_CACHE_HOME/tmp respectively,
    but it would be reasonable for a Flatpak app to have
    --filesystem=/var/tmp, and not impossible for it to have
    --filesystem=/var/cache/something.
    
Resolves: #592

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -334,6 +334,14 @@ xdp_app_info_remap_path (XdpAppInfo *app_info,
       else if (g_str_has_prefix (path, "/run/flatpak/app/"))
         return g_build_filename (g_get_user_runtime_dir (), "app",
                                  path + strlen ("/run/flatpak/app/"), NULL);
+      else if (g_str_has_prefix (path, "/var/config/"))
+        return g_build_filename (g_get_home_dir (), ".var", "app",
+                                 app_info->id, "config",
+                                 path + strlen ("/var/config/"), NULL);
+      else if (g_str_has_prefix (path, "/var/data/"))
+        return g_build_filename (g_get_home_dir (), ".var", "app",
+                                 app_info->id, "data",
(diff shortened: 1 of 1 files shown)


smcv:
Because these paths are not 100% Flatpak-specific and someone might conceivably use `--filesystem=/var/cache` or `--filesystem=/var/tmp`, we might want to hold off on doing this for those two paths until someone (perhaps me) has refactored this code so that instead of having a single canonical translation for a path found in `/proc`, it can use a sequence of possibilities, stat each of them in turn, and use whichever one matches the (device, inode) pair from the container. We'll need that anyway if we want to cope gracefully with `--persist`.

matthiasclasen:
Should we merge this as-is, or are you still working on that refactoring?

smcv:
I think don't merge this right now and I'll try to improve it. If I can't get the refactoring done soon, then merging a version that only handles `/var/config` and `/var/data` (but not `/var/cache` and `/var/tmp`) would be a good fallback plan.

smcv:
> If I can't get the refactoring done soon, then merging a version that only handles /var/config and /var/data (but not /var/cache and /var/tmp) would be a good fallback plan.

I haven't done that yet, so a partial solution now seems better than waiting for a more complete version that'll need restructuring.

PR #595, opened 2021-06-17
ScreenCast: Add 'VIRTUAL' source type

jadahl:
This is a source type that by the portal implementation backend should
be implemented as a per screen cast stream created virtual monitor. The
dimension of the stream for these types of screen cast streams are
typically determined during the pipewire stream negotiation.

---

Use case for this is e.g. gnome-network-displays like applications; anything that wants to provide a virtual monitor like functionality.

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -175,6 +175,7 @@
         <simplelist>
           <member>1: MONITOR</member>
           <member>2: WINDOW</member>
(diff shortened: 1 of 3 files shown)


matthiasclasen:
Wonder if we need to bump the version property for this?

jadahl:
Maybe; it doesn't really change any behavior other than starting to maybe advertise support for "virtual" sources.

Another question is whethere there should be ways to add hints e.g. how many should be opened; not sure. In the current implementation it's just a GtkSwitch that opens one or zero.

PR #598, opened 2021-06-19
permission-store: Add GetPermission

### LINKED ISSUE #583: [PermissionStore] consider GetPermission inclusion?
We currently have `SetPermission` and `DeletePermission` which operate with specific apps, but we don't have a `GetPermission` that does the same. Instead, we rely on `LookUp` which brings all `(table, id, app)` tuples.

From what I have seen, in GNOME Control Center, and what I needed to do for Flatseal, we call `LookUp`, check if the app is present among all tuples, to then access the permission data (e.g., ["yes"], etc).

As more UI tools ala-flatseal appear, I believe a `GetPermission` would make more sense for such tools as these operate per-app basis. Another reason is that the permission store data tends to grow, which means that `LookUp` will keep bringing more and more unnecessary data for the case I am describing here (but I haven't really measure the impact of this).

Turning this question around, Is there a specific reason why there ins't `GetPermssion`?

tchx84:
We currently have SetPermission and DeletePermission which operate
with specific apps, but we didn't have a GetPermission method that
does the same. Instead, we rely on LookUp which brings all (table,
id, app) tuples.
   
From what I have seen,in GNOME Control Center and what I needed
to do for Flatseal, we call LookUp, check if the app is present
among the retrieved tuples,  to then access the permission data
we are looking for (e.g, ["yes"], etc).

As more UI tools ala-flatseal appear, a GetPermission would make
more sense for such tools as these operate per-app basis.

Another reason is that the permission store data tends to grow,
which means that  LookUp  will keep bringing more and more data
unnecessarily for the case described.

Closes #583

### DIFF
--- a/data/org.freedesktop.impl.portal.PermissionStore.xml
+++ b/data/org.freedesktop.impl.portal.PermissionStore.xml
@@ -146,6 +146,23 @@
     </method>
 
     <!--
+        GetPermission:
+        @table: the name of the table to use
+        @id: the resource ID to modify
+        @app: the application ID to modify
+        @permissions: permissions to get
+
+        Gets the entry for an application and a resource
+        in the given table.
+    -->
+    <method name="GetPermission">
+      <arg name='table' type='s' direction='in'/>
+      <arg name='id' type='s' direction='in'/>
+      <arg name='app' type='s' direction='in'/>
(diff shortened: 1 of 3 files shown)


matthiasclasen:
I don't see any problem with adding such an api.

Thanks for adding tests!

PR #599, opened 2021-06-20
Improve error handling when passing files by fd

smcv:
* verify_proc_self_fd: Raise a GError
    
    This lets us log the failure reason as a debug message on failure.

* xdp_app_info_get_path_for_fd: Factor out check_same_file
    
    As with the previous commit, this lets us log a more detailed debug
    message when we don't get a match.

* utils: Add fallback definition for G_DBUS_METHOD_INVOCATION_HANDLED
    
    When implementing GDBus methods, it's not always obvious that
    "return TRUE" on error is correct - but in fact it is, because the
    boolean result is handled/unhandled rather than the usual success/error.
    GLib 2.68 introduced macros to make this more obvious. Backport them
    here for older GLib versions.
    
    I'm not systematically converting existing code right now, but the new
    macros can be used in new code.

* email: Handle error from xdp_app_info_get_path_for_fd
    
    Otherwise we would append NULL to the GVariantBuilder, which is
    considered to be a programming error, resulting in a runtime warning
    and not actually attaching the intended attachment.

* open-uri: Emit debug messages even if request was not exported yet

* trash: Log debug messages on failure

* xdp_app_info_get_path_for_fd: Raise a GError
    
    This allows better context-sensitive logging.

---

This is mostly just adding more detailed `g_debug()` when a file cannot be passed to the portal. We log this via `g_debug()` rather than sending information back to the calling app in order to avoid leaking information about the host filesystem into the sandbox of a Flatpak (or Snap) app.

I'm in two minds about "email: Handle error from xdp_app_info_get_path_for_fd": do we want trying to attach a file that cannot be shared with the host to cause an error, or do we want to log a message and gracefully degrade to sending the email without the attachment? Marked as draft until this is resolved.

I originally thought the right answer was an error (as implemented here), but while testing this I found that the `xdg-email --attach` in flatpak-xdg-utils has, effectively, silently ignored `--attach` since 2017 (flatpak/flatpak-xdg-utils#50, fix proposed in flatpak/flatpak-xdg-utils#51), and it still doesn't implement multiple `--attach`; so clearly nobody has found that feature to be particularly important?

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -362,10 +362,15 @@ validate_fd (int fd,
   g_autofree char *name = NULL;
   xdp_autofd int dir_fd = -1;
   struct stat real_st_buf;
+  g_autoptr(GError) local_error = NULL;
 
-  path = xdp_app_info_get_path_for_fd (app_info, fd, 0, st_buf, writable_out);
+  path = xdp_app_info_get_path_for_fd (app_info, fd, 0, st_buf, writable_out, &local_error);
+
   if (path == NULL)
-    goto errout;
+    {
+      g_debug ("Invalid fd passed: %s", local_error->message);
+      goto errout;
+    }
 
   if ((ensure_type == VALIDATE_FD_FILE_TYPE_REGULAR || ensure_type == VALIDATE_FD_FILE_TYPE_ANY) && S_ISREG (st_buf->st_mode))
     {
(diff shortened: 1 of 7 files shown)


matthiasclasen:
I merged the `xdg-email --attach fix`, so imo we can go with what you have implemented here.

smcv:
> I merged the xdg-email --attach fix, so imo we can go with what you have implemented here.

OK, if you're happy with these semantics then I think it's ready.

PR #600, opened 2021-06-30
DRAFT: Add Realtime portal

### LINKED ISSUE #355: Add a RealtimeKit Portal
On behalf of the [PulseEffects](https://github.com/wwmm/pulseeffects) project I would like to request a portal interface for accessing the [RealtimeKit](https://directory.fsf.org/wiki/RealtimeKit) scheduling service.

PulseEffects needs this in order to optionally (off by default) set process priorities to prevent audio stutter.

The main problem is that RealtimeKit wants a host namespace pid, but since a Flatpak app runs in its own pid namespace it will send an unusable pid. Because of this its not possible to temporarily workaround the issue by granting talk access to **org.freedesktop.RealtimeKit1** on the system bus.

This particular issue could probably be solved the same way as in [org.freedesktop.portal.GameMode](https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.portal.GameMode.xml).

Not sure what else would be required in order for this to work well. Feedback is much appreciated.

*Note that I am not a developer and will unfortunately not be able to contribute this myself.*

PulseEffects upstream issue thread:
https://github.com/wwmm/pulseeffects/issues/445

TingPing:
This adds an initial portal that proxies to RealtimeKit on the host.

WIP as some further testing is required.

Closes #355


data/org.freedesktop.portal.Realtime.xml
```diff
@@ -0,0 +1,71 @@
...
+      <method name="MakeThreadHighPriorityWithPID">
+        <arg type="t" name="process" direction="in"/>
+        <arg type="t" name="thread" direction="in"/>
+        <arg type="i" name="priority" direction="in"/>
+      </method>
+
+      <!--
+          GetProperty:
+          @property: Property name. One of "MaxRealtimePriority", "MinNiceLevel", "RTTimeUSecMax".
+          @result: The property value
```
matthiasclasen, line 62: Why not expose these as actual DBus properties ?

src/realtime.c
```diff
@@ -0,0 +1,273 @@
...
+
+  g_dbus_proxy_call (G_DBUS_PROXY (realtime->rtkit_proxy),
+                     "MakeThreadRealtimeWithPID",
+                     g_variant_new ("(ttu)", pids[0], thread, priority),
+                     G_DBUS_CALL_FLAGS_NONE,
+                     -1,
+                     NULL,
+                     on_call_ready,
+                     g_object_ref (invocation));
+
```
matthiasclasen, line 113: I guess we do not expect to ever make this interactive?

I still think it should propably still have a permission to let users deny this request.

src/realtime.c
```diff
@@ -0,0 +1,273 @@
...
+                     g_variant_new ("(ss)", "org.freedesktop.RealtimeKit1", property_name),
+                     G_DBUS_CALL_FLAGS_NONE,
+                     -1,
+                     NULL,
+                     on_get_property_ready,
+                     g_object_ref (invocation));
+
+  return TRUE;
+}
+
```
matthiasclasen, line 204: This is a little odd, to add your own GetProperty method, instead of implementing DBus properties.

data/org.freedesktop.portal.Realtime.xml
```diff
@@ -0,0 +1,71 @@
...
+
+      <!--
+          MakeThreadHighPriorityWithPID:
+          @process: Process id
+          @thread: Thread id
+          @priority: Priority
+      -->
+      <method name="MakeThreadHighPriorityWithPID">
+        <arg type="t" name="process" direction="in"/>
+        <arg type="t" name="thread" direction="in"/>
```
matthiasclasen, line 55: As has been pointed out in the issue: it is a little odd to let the app specify its own pid - the portal can just get that from peer creds.

src/realtime.c
TingPing: Whats the proper way to implement non-blocking getters with dbusproxy?

data/org.freedesktop.portal.Realtime.xml
TingPing: The pid can be another process, webkit setting it on a subprocess for example.

src/realtime.c
TingPing: Its just such a low level detail i didnt think anything would want to expose a permission for it.

data/org.freedesktop.portal.Realtime.xml
```diff
@@ -0,0 +1,71 @@
...
+      <!--
+          MakeThreadRealtimeWithPID:
+          @process: Process id
+          @thread: Thread id
+          @priority: Priority
+      -->
+      <method name="MakeThreadRealtimeWithPID">
+        <arg type="t" name="process" direction="in"/>
+        <arg type="t" name="thread" direction="in"/>
+        <arg type="u" name="priority" direction="in"/>
```
jadahl, line 44: Any reason why these don't use the regular Request objects?

TingPing: What is gained by that? I thought they were mostly for user interactive APIs.

jadahl: So this is meant to be always non-interative and just a "filter portal" towards the real rt kit one? I was assuming there would be some backend involved that sets the right permission in the permission store, or at least theoretically making it possible. That would make this interactive as a side effect.

TingPing: > So this is meant to be always non-interative and just a "filter portal" towards the real rt kit one? 

That was my intention yes. RTKit already does some sanity checks to ensure the process is set up to handle being real time and won't lock up your machine. I don't know why a person would want to prevent this.

matthiasclasen: permissions are checked in the frontend, so no need for a backend for that.

jadahl: And the user-querying would work how?

src/realtime.c
TingPing: Ok so I tried to use `g_dbus_proxy_get_cached_property()` originally but it didn't work.

It turns out RealtimeKit does not implement `GetAll` or `PropertyChanged`, so its not possible for us to cache properties.

I think this ugly method is the best way to do this?

data/org.freedesktop.portal.Realtime.xml
TingPing: See https://github.com/flatpak/xdg-desktop-portal/pull/600#discussion_r684665453

src/realtime.c
TingPing: OK I ended up manually caching the values. There is a potential for them to become outdated but I'm not sure the value in RealtimeKit ever changes anyway.

hfiguiere:
excellent!

PR #601, opened 2021-07-05
openuri: Use FileManager1 in OpenDirectory

### LINKED ISSUE #564: Select the file when opening file manager with OpenDirectory
OpenDirectory method of the OpenURI portal accepts a fd of a file, so it theoretically can use https://www.freedesktop.org/wiki/Specifications/file-manager-interface/ to open a directory with file selected. There are sadly no analogue of the FileManager service for sandboxed apps and this makes "show file in folder" feature of applications less convenient :(

refi64:
This will result in the file entry being highlighted upon opening the directory. Fixes #564.

(Also includes an fd double-close fix I found while writing this.)

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -682,6 +682,7 @@ handle_open_in_thread_func (GTask *task,
       uri = g_filename_to_uri (path, NULL, NULL);
       g_object_set_data_full (G_OBJECT (request), "uri", g_strdup (uri), g_free);
       close (fd);
+      fd = -1;
       g_object_set_data (G_OBJECT (request), "fd", GINT_TO_POINTER (-1));
     }
 
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -43,6 +43,12 @@
 #include "permissions.h"
 #include "documents.h"
 
+#define FILE_MANAGER_DBUS_NAME "org.freedesktop.FileManager1"
+#define FILE_MANAGER_DBUS_IFACE "org.freedesktop.FileManager1"
+#define FILE_MANAGER_DBUS_PATH "/org/freedesktop/FileManager1"
+
+#define FILE_MANAGER_SHOW_ITEMS "ShowItems"
+
 #define PERMISSION_TABLE "desktop-used-apps"
 
 #define DEFAULT_THRESHOLD 3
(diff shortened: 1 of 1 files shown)


src/open-uri.c
```diff
@@ -1005,6 +1049,20 @@ open_uri_create (GDBusConnection *connection,
...
+                                                 G_DBUS_PROXY_FLAGS_NONE,
+                                                 NULL,
+                                                 FILE_MANAGER_DBUS_NAME,
+                                                 FILE_MANAGER_DBUS_PATH,
+                                                 FILE_MANAGER_DBUS_IFACE,
+                                                 NULL,
+                                                 &error);
+  if (!open_uri->file_manager)
+    {
+      g_warning ("Failed to create FileManager proxy: %s", error->message);
```
matthiasclasen, line 1062: I would not warn about this. It is an expected condition.

src/open-uri.c
```diff
@@ -668,10 +676,45 @@ handle_open_in_thread_func (GTask *task,
...
+                  xdp_request_emit_response (XDP_REQUEST (request),
+                                             XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS,
+                                             g_variant_builder_end (&opts_builder));
+                  request_unexport (request);
+                  return;
+                }
+            }
+
           g_free (path);
           char *dir = g_path_get_dirname (real_path);
```
matthiasclasen, line 717: Could avoid the local variable here, and just say

```
path = g_path_get_dirname (real_path);
```

src/open-uri.c
```diff
@@ -1005,6 +1049,20 @@ open_uri_create (GDBusConnection *connection,
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (impl), G_MAXINT);
 
   open_uri = g_object_new (open_uri_get_type (), NULL);
+  open_uri->file_manager = g_dbus_proxy_new_sync(connection,
```
matthiasclasen, line 1052: Coding style nit: space before (

refi64:
@matthiasclasen done.

Forgot to mention this before, but for the sake of context: I had a bit of an internal debate as to whether or not this should maybe be a new implementation method, or just doing this. In the end, I figured:

- most implementations would probably end up doing this anyway
- even if more were done later, this would be a useful fallback to have

PR #605, opened 2021-07-23
location: Update accuracy translation

fmuellner:
Currently the translation uses the geoclue enum values rather than
the documented values, which breaks consumers like geoclue's client
library that translates accuracy values according to the documentation.


matthiasclasen:
As discussed on irc: +1

PR #608, opened 2021-07-27
remote-desktop: better error messages on device access failures

bilelmoussaoui:
Uses a more meaningful error messages in case a device access wasn't provided after
starting the session


matthiasclasen:
Thanks, looks good to me.

PR #609, opened 2021-07-27
use correct type for the session object path

bilelmoussaoui:
The session id is defined as the object path of the session.
We should use a type of 'o' instead of 's'.

Note that the docs mentions the correct type https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-method-org-freedesktop-portal-ScreenCast.CreateSession.

### DIFF
--- a/src/inhibit.c
+++ b/src/inhibit.c
@@ -366,7 +366,7 @@ create_monitor_done (GObject *source_object,
     }
 
   g_variant_builder_add (&results_builder, "{sv}",
-                         "session_handle", g_variant_new ("s", session->id));
+                         "session_handle", g_variant_new ("o", session->id));
 
 out:
   if (request->exported)
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -221,7 +221,7 @@ create_session_done (GObject *source_object,
     }
 
   g_variant_builder_add (&results_builder, "{sv}",
-                         "session_handle", g_variant_new ("s", session->id));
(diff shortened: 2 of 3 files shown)


GeorgesStavracas (approved):

GeorgesStavracas (approved):
This could use a better commit message title, but the code LGTM

grulja:
It's unfortunate that this will break existing clients, for example it will take weeks (or months) until fixed WebRTC (Chrome) gets to users and until then they won't be able to share their screen.

[1] - https://webrtc-review.googlesource.com/c/src/+/232329

grulja:
Fedora 35 already updated to the new version and that effectively broke screensharing in Chrome.

@matthiasclasen can we revert this change at least in Fedora until we get this fixed in Chrome? I can see OBS and Firefox fixed it the way it doesn't matter what type we use so it should be safe to revert it for a while so things don't break and people are happy.

matthiasclasen:
Oh, grr. Lets see what we can do about this

grulja:
> Oh, grr. Lets see what we can do about this

Thank you. I believe we can re-revert this change back in ~1-2 months once we are sure it doesn't break anything.

Conan-Kudo:
I cherry-picked a fix for obs-studio to deal with this too... https://pkgs.rpmfusion.org/cgit/free/obs-studio.git/commit/?id=311816c1758129bb843a0def17a3b394b2642da3

PR #613, opened 2021-08-10
Implement portal for GPowerProfileMonitor

hadess:


hadess:
The GLib API landed:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2222

and its installed tests exercise this MR, so merging this now.

PR #621, opened 2021-08-30
Add a portal for Emulated Input

whot:
*Note: filing this as Draft to get some early review please. I can't really say I fully know what I'm doing here, it was a lot of copy/paste and fixing, so I may have missed something obvious.*

This PR adds a new portal: `org.freedesktop.portal.EmulatedInput` together with its `impl` equivalent.
Two calls:
- `CreateSession` to determine if a client may emulate input
- `ConnectToEIS` which connects to an [EIS implementation](https://gitlab.freedesktop.org/libinput/libei/) and passes the socket back to the application. After that is done, the application talks directly to the EIS implementation, similar to pipewire.

Note that this is just a gateway portal, there is no actual events being sent through the portal, this is all handled through libeis/libei.

We need an `impl.portal` here because there may be multiple EIS implementations running, so we rely on the implementation to know which one we actually need to connect to. I'm planning to add this to mutter.

### DIFF
--- a/Makefile.am.inc
+++ b/Makefile.am.inc
@@ -5,6 +5,7 @@ PORTAL_IFACE_FILES =\
 	$(top_srcdir)/data/org.freedesktop.portal.Device.xml \
 	$(top_srcdir)/data/org.freedesktop.portal.Documents.xml \
 	$(top_srcdir)/data/org.freedesktop.portal.Email.xml \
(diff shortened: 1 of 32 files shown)


whot:
Closing this, no point keeping it open. We'll be going through the RemoteDesktop API instead, see #762

PR #625, opened 2021-09-05
pipewire: Ensure roundtrips have a timeout

refi64:
This ensures the portal itself can still start successfully (albeit with
reduced abilities) if PipeWire itself is non-functional.

Note that this is not just hypothetical, as I have seen various reports
of this over time.

### DIFF
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -25,6 +25,8 @@
 
 #include "pipewire.h"
 
+#define ROUNDTRIP_TIMEOUT_SECS 10
+
 typedef struct _PipeWireSource
 {
   GSource base;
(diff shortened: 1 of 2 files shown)


hadess:
> Note that this is not just hypothetical, as I have seen various reports
> of this over time.

Would be great if you could add a few references in the commit message then.

refi64:
@hadess sadly none officially reported onto the issue tracker, the most recent case that caused me to look into opening this PR was when a friend updated to F35 and then had pipewire break due to old configs lying around. The other times were generally surrounded by various life scenarios where I didn't get to open an issue at the moment but had identified this as the cause.

Now if just adding the above anecdotes is what you mean, I can definitely do that, but otherwise there unfortunately isn't much to link to.

hadess:
It would be best to add the anecdote and remove the mention of reports as those don't exist. Did you manage to test the changes another way?

refi64:
@hadess commit message is updated, I was able to essentially test it by sending them the compiled binary (since we're both using Fedora), and they were able to confirm it "fixed" the problem for them (of course PW was still not working, but there were no longer any issues with xdg-desktop-portal's startup).

In addition, I *did* try simply removing the registry listener so that no events would be sent in order to locally test that the timeout was functioning. From what I can tell, the situation should be identical from the portal's perspective (open socket but the registry event is not being received).

hadess:
I asked Wim to check whether this was the right approach and I got told "I guess so". So that's good enough for me. Only question is whether the spa_source needs destroyed in finalize (`spa_loop_utils_destroy_source`?)

refi64:
@hadess ah whoops, the source should be destroyed now

Issue #629, opened 2021-09-09
Provide a standard dark style preference key in the settings portal


Exalm:
Background: https://blog.elementary.io/the-need-for-a-freedesktop-dark-style-preference/

That's now live in elementary OS 6, and I'm currently implementing [the same thing](https://gitlab.gnome.org/GNOME/libadwaita/-/merge_requests/224) in GNOME, and I think it would be a good time to actually standardize it instead of having separate incompatible preferences.

The main reason to do it here and not just have a gsettings key in `org.gnome.desktop.interface` and pass that through the portal as is would be apps like Firefox or Telegram - right now they don't really have any way to get the system preference on elementary, or the upcoming one on GNOME, or the similar looking preference in KDE, etc.

Settings portal currently doesn't define any keys and leaves it to implementation. What if we actually defined a key that the implementation would proxy where they need it?

My prototype implementation uses the following:

* Namespace: `org.freedesktop.appearance`
* Key: `color-scheme`
* Possible values: strings `default`, `prefer-light`, `prefer-dark`.

The elementary setting is https://github.com/elementary/default-settings/blob/master/accountsservice/io.elementary.pantheon.AccountsService.xml#L43-L52

CC @danrabbit @CassidyJames from the elementary side

CC @aleixpol from the KDE side - I'm not sure who would be the relevant person here

aleixpol:
Hey, it does make a lot of sense.

It won't be entirely trivial to implement in our end, since our users select a colour scheme to use rather than light vs dark but seeing how varied toolkits are nowadays, standardising colour schemes would definitely be out of reach.

+1 in general from the our side. I've shared it in some KDE channels in case someone has any ideas.
Thanks!

danrabbit:
We currently abstract this away a bit for developers in Granite so it would be fairly simple to adopt whatever is agreed upon here without breaking API for developers. Very +1 from the elementary camp. Thanks :)

Exalm:
> It won't be entirely trivial to implement in our end, since our users select a colour scheme to use rather than light vs dark but seeing how varied toolkits are nowadays, standardising colour schemes would definitely be out of reach.

Yeah... I suppose it would need to be a separate thing from the general KDE color schemes. In theory it should be possible to do it based on whether the window foreground color is lighter or darker than background.

DarkTrick:
Thank you for the effort on implementing it on GNOME.

I would suggest a file for storing the setting, though. This way it's the most compatible and long-lived format.

[...]

PR #633, opened 2021-09-19
settings: Add color-scheme key

Exalm:
Specify a key for getting the system's preferred color scheme.

Fixes https://github.com/flatpak/xdg-desktop-portal/issues/629

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -27,9 +27,27 @@
     of host settings required for toolkits similar to XSettings.
     It is not for general purpose settings.
 
-    Currently the valid keys are entirely implementation details
-    that are undocumented. If you are a toolkit and want to use
-    this please open an issue.
+    Currently the interface provides the following keys:
+
+    <variablelist>
+      <varlistentry>
+        <term>org.freedesktop.appearance color-scheme u</term>
+        <listitem><para>
+          Indicates the system's preferred color scheme.
+          Supported values are:
+          <simplelist>
+            <member>0: No preference</member>
+            <member>1: Prefer dark appearance</member>
+            <member>2: Prefer light appearance</member>
+          </simplelist>
+          Unknown values should be treated as 0 (no preference).
+        </para></listitem>
+      </varlistentry>
+    </variablelist>
+
+    Implementations can provide other keys; they are entirely
+    implementation details that are undocumented. If you are a
+    toolkit and want to use this please open an issue.
 
     Note that the Settings portal can operate without this backend,
     implementing it is optional.
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -27,9 +27,27 @@
     of host settings required for toolkits similar to XSettings.
     It is not for general purpose settings.
 
-    Currently the valid keys are entirely implementation details
-    that are undocumented. If you are a toolkit and want to use
-    this please open an issue.
+    Currently the interface provides the following keys:
+
+    <variablelist>
+      <varlistentry>
+        <term>org.freedesktop.appearance color-scheme u</term>
+        <listitem><para>
+          Indicates the system's preferred color scheme.
+          Supported values are:
+          <simplelist>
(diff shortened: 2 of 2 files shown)


data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -27,9 +27,27 @@
...
+        <term>org.freedesktop.appearance color-scheme u</term>
+        <listitem><para>
+          Indicates the system's preferred color scheme.
+          Supported values are:
+          <simplelist>
+            <member>0: No preference</member>
+            <member>1: Prefer dark appearance</member>
+            <member>2: Prefer light appearance</member>
+          </simplelist>
+          Unknown values should be considered same as 0.
```
GeorgesStavracas, line 43: `Unknown values should be treated as 0 (no preference).`

data/org.freedesktop.portal.Settings.xml
```diff
@@ -27,9 +27,27 @@
...
+        <term>org.freedesktop.appearance color-scheme u</term>
+        <listitem><para>
+          Indicates the system's preferred color scheme.
+          Supported values are:
+          <simplelist>
+            <member>0: No preference</member>
+            <member>1: Prefer dark appearance</member>
+            <member>2: Prefer light appearance</member>
+          </simplelist>
+          Unknown values should be considered same as 0.
```
GeorgesStavracas, line 43: `Unknown values should be treated as 0 (no preference).`

TingPing:
Looks fine to merge once implementations exist.

Exalm:
https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/12

Do we need an equivalent `xdg-desktop-portal-gtk` change? On one hand it's fairly GNOME-specific, but `-gtk` already has plenty of GNOME things in it, since `xdg-desktop-portal-gnome` exists now, not sure what's the policy regarding that.

Exalm:
elementary implementation: https://github.com/elementary/settings-daemon/pull/38

TingPing:
I'll leave the -gtk question to @GeorgesStavracas

GeorgesStavracas:
> Do we need an equivalent xdg-desktop-portal-gtk change?

I would say, yes.

Exalm:
> I would say, yes.

Ok - https://github.com/flatpak/xdg-desktop-portal-gtk/pull/354

lilianmoraru:
@Exalm, @GeorgesStavracas, @TingPing
Could you please also add support for High Contrast?
Microsoft does it very nicely: https://support.microsoft.com/en-us/windows/use-high-contrast-mode-in-windows-10-fedc744c-90ac-69df-aed5-c8a90125e696 , allowing users that maybe have problems seeing a particular color to switch to a different one.
Here they describe the event that apps will receive and how to read the user's desired colors, in order to avoid hardcoding them: https://docs.microsoft.com/en-us/windows/win32/w8cookbook/high-contrast-mode#solution

Exalm:
We have it in GNOME, but I consider it a separate issue when it comes to having a standardized one - for example, elementary doesn't have it.

lilianmoraru:
GNOME has a separate accessibility toggle with no dedicated D-Bus connection to listen on, for apps to handle it(and missing documentation).
It needs a more official/portable way of doing it(IMO). I thought that once you are here, this could be handled as well.

Exalm:
I mean until recently (still is in 41) it was a theme - not even a flag. One thing at a time :p

WhyNotHugo:
I'm trying to read this setting with non-flatpak'd apps (in my case, d-feet), but I get:

`` `
('g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: '
 'Portal operation not allowed: Unable to open /proc/147042/root (9)')
```

Are native applications supposed to use a different mechanism to retrieve this value from the portal?

swick:
IMO no, if it's not working for some reason that's a bug.

rusty-snake:
Screenshot_2022-01-13_15-12-49-fs8

WhyNotHugo:
@rusty-snake Reported in #737

PR #635, opened 2021-09-22
Move xdg-permission-store to Background slice

unusual-thoughts:
Follow-up to https://github.com/flatpak/xdg-desktop-portal/pull/614

I think `xdg-permission-store` fits the description from `man:system.special(7)`
>background.slice
> All services running low-priority background tasks should use this slice. This permits resources to be preferentially assigned to the other slices. Examples include non-interactive tasks like file indexing or backup operations where latency is not important.

### DIFF
--- a/document-portal/xdg-document-portal.service.in
+++ b/document-portal/xdg-document-portal.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=flatpak document portal service
+PartOf=graphical-session.target
 
 [Service]
 BusName=org.freedesktop.portal.Documents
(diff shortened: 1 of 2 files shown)


GeorgesStavracas (approved):
LGTM

unusual-thoughts:
I also added `PartOf=graphical-session.target` so that the services are automatically shut down on logout

smcv:
> All services running low-priority background tasks should use this slice. This permits resources to be preferentially assigned to the other slices

Are you sure this is the right slice, and not `session.slice`?

xdg-permission-store doesn't do heavy I/O or computation, so giving it too high a priority level seems unlikely to have an impact on interactive responsiveness. However, when you use portal interfaces, if the latency of querying the permission store is bad, it could delay use of the portal interface and make it less responsive.

PR #638, opened 2021-09-29
Screencast restore

GeorgesStavracas:
This is my first working version of a working way to restore screencast streams. First, I'd like to remind what goals I had in mind when working on it:

 * Users must be able to control whether or not any application will be able to restore streams. Users must be able to revoke these permissions at any point in time. This is non-negotiable. ¹
 * Avoid a broad "allow all" permission.
 * The fallback is always to drop permissions and query the user again.

That being said, let's jump right into the implementation.

## Implementation

There are 2 important workflows happening in parallel: acquiring a token, and using it.

### 1. Acquiring a restore token

 1. Applications that want to be able to restore screencast streams pass the `persist_mode` option to the `SelectSources()` method. I'll explain persist modes below.
 2. `xdg-desktop-portal` pass the `persist_mode` option to the portal implementation.
 3. The portal implementation queries the user (usually via a dialog) which monitor or window to share, and whether they allow it to be restored. (It should default to not allow.)
 4. If the screencast session should be restored, the portal implementation hands `xdg-desktop-portal` the implementation-specific data to be restored. This is passed via the `restore_data` option, at the response of the `Start()` call.
 5. `xdg-desktop-portal` stores this data somewhere with an unique identifier that I called `restore_token`.
 6. `xdg-desktop-portal` pass `restore_token` to the application.

The application then should save this restore token wherever it thinks it's appropriate.

### 2. Using the restore token

Now that the application has a restore token, it can restore a previous stream. This is how it happens:

 1. Applications that want to restore a screencast stream pass the `restore_token` option to the `SelectSources()` method.
 2. `xdg-desktop-portal` receives the token, and performs a lookup for the implementation-specific data associated with this token.
 3. `xdg-desktop-portal` consumes this token and removes the permission.
 4. `xdg-desktop-portal` sends this restore data to the portal implementation as the `restore_data` option to the `SelectSources()` method.
 5. The portal implementation uses this restore data to match windows and monitors without presenting a dialog. If it fails to match (due to, for example, any monitors or windows not being available), it fallbacks to presenting the dialog.

### Token lifecycle 

Tokens can only be used once to restore streams. The permissions they hold are revoked after use. That's what xdg-desktop-portal does at step 2.3. Applications must not assume anything about tokens. Portal implementations never have access to the restore tokens.

### Persist modes

In this first implementation, I've added two modes of persistence:

 1. **Transient**: the restore token is valid as long as the application is alive. It's stored in memory and revoked when the application closes its D-Bus connection. The use-case I had in mind with this mode is browsers.
 2. **Persistent**: the restore token is stored in disk, using the permissions store, and is valid until the user manually revokes it.

-----

See https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/14 for a reference implementation.


GeorgesStavracas:
@smcv here's the screencast restore thing I've mentioned. I'd love to hear your thoughts on this. I'm not confident that this is a good approach, much less a secure one.

smcv:
> The restore token can be stolen by any host system app via monitoring the session bus

Host system apps can also arbitrarily impersonate each other, e.g. if a host system app suspects that you might give Steam permission to record your screen (for Remote Play), it can tell xdg-desktop-portal "hello, I'm Steam". Or it can impersonate a web browser, or a standalone video-streaming client, or anything else, in a way that is indistinguishable from a Flatpak app. Anything Flatpak can do (not a Flatpak app, but Flatpak itself!), a host system app can do too.

In general, a host system app can also ptrace xdg-desktop-portal, or append `export LD_PRELOAD="$HOME/.malicious-code-goes-here"` to `~/.profile` and use that to load its own arbitrary code into xdg-desktop-portal and host system apps. We are not going to be able to avoid any of this without making use of LSMs (SELinux/AppArmor/Smack/etc.), which are non-portable, but provide the only mechanism currently available for Unix sockets (and hence D-Bus) to distinguish between "applications" running as the same uid in the face of potentially malicious host-system applications.

The solution is to avoid running host system apps that you don't completely trust: they are in the trusted computing base (TCB), on the trusted side of the [airtight hatchway](https://devblogs.microsoft.com/oldnewthing/20060508-22/?p=31283).

Of course, this is far from ideal - if we were designing a modern app framework from scratch, without backward-compatibility concerns, we'd want literally every app to be identifiable as part of the app framework, as they are in Android or iOS, with a minimum of code in the TCB. However, users of Unix OSs are used to being able to add arbitrary programs to the TCB, with the tradition being that host system apps are the majority and sandboxed apps are a minority; and the price we pay for that freedom is that the programs added to the TCB are ... in the TCB.

smcv:
A host system app could also do the equivalent of `xdg-desktop-portal --replace` and *become* the portal.

If we assume an attacker that somehow only has the ability to monitor D-Bus passively (Monitoring and/or legacy eavesdropping, the same mechanisms used in new and old dbus-monitor respectively), then we could craft a way to avoid the token being visible on the bus. The easiest way would be for the portal's client (browser, VNC server, Steam, etc.) to fd-pass the write end of a pipe to the portal, and the portal could write the token into the pipe, then close the pipe; the client could read into a buffer until EOF, and that's their token. An eavesdropper can't read from the write end of a pipe, so it would be useless to them.

However, that's 100% security theatre, because if an attacker has the unfiltered access to the session bus that is necessary to carry out Monitoring or legacy eavesdropping, then they can also carry out active attacks. Perhaps the simplest active attack would be to ask the bus nicely for the `org.freedesktop.portal.Desktop` name, *become* the portal, and reimplement anything it can do.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -119,6 +119,37 @@
...
+
+              <simplelist>
+                <member>0: Do not persist (default)</member>
+                <member>1: Permissions persist as long as the application is running</member>
+                <member>2: Permissions persist forever</member>
+              </simplelist>
+
+              If the permission for the session to persist is granted, a restore token will
+              be returned via the #org.freedesktop.portal.Request::Response signal of the
+              #org.freedesktop.portal.Screencast.Start method.
```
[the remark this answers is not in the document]
jadahl: For example, if an application wants such a token to restore, let the user either click `[ Share and remember ]` or `[ Share once ]` maybe. Or a pre-selected `[x] Remember for this application` or something. Whether the UI should include this should be a design decision, but the API could be prepared to make it possible by communicating the decision to the application.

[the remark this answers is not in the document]
GeorgesStavracas: Done

GeorgesStavracas:
This is ready for review now. The missing bits (removing transient permissions when a client disconnects, and apply review comments) are implemented.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -81,6 +81,37 @@
               Whether to allow selecting multiple sources. Default is no.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>restore_data as</term>
+            <listitem><para>
+              The data to restore from a previous session.
```
jadahl, line 87: This should probably contain a tip for what to pass here, e.g. the same array of strings that was received in a previous response signal.

Also, why is it an array? Does the order of elements matter? Do they have meaning or is it just implementation specific blobs?

GeorgesStavracas: This is just the easiest way to deal with the permission store. I'm not entirely happy with how the permission store is used in this branch, but I just couldn't meaningfully use the "value" field of the permission store without overcomplicating the code. Let me go through it once again...

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -81,6 +81,37 @@
...
+          </varlistentry>
+          <varlistentry>
+            <term>persist_mode u</term>
+            <listitem><para>
+              How this session should persist. Default is 0. Accepted values are:
+
+              <simplelist>
+                <member>0: Do not persist (default)</member>
+                <member>1: Permissions persist as long as the application is running</member>
+                <member>2: Permissions persist forever</member>
```
matthiasclasen, line 105: I think the last one should maybe say: Permissions persist until explicitly revoked

data/org.freedesktop.impl.portal.ScreenCast.xml
GeorgesStavracas: Switch to a more sophisticated usage of the permission store. Data is now stored as:

```
Table      Object                               App                       Permissions                    Data
screencast 1be22dff-xxx-yyyy-zzzz-4ffdf1602084 com.obsproject.Studio     yes                            ['monitor::AUO:0x313d:0x00000000']

```

data/org.freedesktop.impl.portal.ScreenCast.xml
GeorgesStavracas: Makes sense, updated.

src/screen-cast.c
```diff
@@ -624,12 +915,98 @@ collect_screen_cast_stream_data (GVariantIter *streams_iter)
...
+      g_clear_pointer (&screen_cast_session->restore_token, g_free);
+      g_clear_pointer (&screen_cast_session->restore_data, g_variant_unref);
+      return;
+    }
+
+  switch (screen_cast_session->persist_mode)
+    {
+    case PERSIST_MODE_NONE:
+      g_clear_pointer (&screen_cast_session->restore_token, g_free);
+      g_clear_pointer (&screen_cast_session->restore_data, g_variant_unref);
```
jadahl, line 934: Should the transient/persintent permission be cleared here too, if there existed a token?

Same question applies to transient when setting persistent, and persistent when setting transient.

src/screen-cast.c
```diff
@@ -100,6 +114,146 @@ GType screen_cast_session_get_type (void);
 
 G_DEFINE_TYPE (ScreenCastSession, screen_cast_session, session_get_type ())
 
+static void
+set_persistent_permissions (const char *app_id,
+                            const char *restore_token,
+                            GVariant   *restore_data)
```
jadahl, line 120: sneaked in some fancy variable name alignment I see ;P

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -169,6 +200,14 @@
               'multiple' was set to 'true' as part of the SelectSources method.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>restore_token s</term>
+            <listitem><para>
+              The restore token.
+
```
jadahl, line 207: This should probably mention that the token in a single-use token, and it will only be allowed to be used to restore a session once. Maybe e.g.

```
This token is a single use token that can later be used to restore
a session. See  org.freedesktop.portal.ScreenCast.SelectSources
for details.
```

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -119,6 +119,37 @@
...
+          <varlistentry>
+            <term>restore_token s</term>
+            <listitem><para>
+              The token to restore a previous session.
+
+              If the stored session cannot be restored, this value is ignored
+              and the user will be prompted normally. This may happen when, for
+              example, the session contains a monitor or a window that is not
+              available anymore, or when the stored permissions are withdrawn.
+
```
jadahl, line 131: As with the documentation for the token itself, maybe add something about the inability to reuse the same token again after restoring once. E.g.

```
After having used the restore token once, it will no longer be valid.
To restore the same session again, use the new restore token sent
in response to starting this session.
```

or something.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -123,6 +154,37 @@
...
+              The data to be restored. This is the data that a portal
+              implementation sent in reaction to transient or persistent
+              modes.
+
+              Data generated by a portal implementation must be consumable by
+              the same portal implementation, but not necessarily by others.
+              Thus, it is important to properly handle 'restore_data' not being
+              in a valid format. This may happen when, e.g., the user switched
+              to another desktop environment, or is using a different portal
+              implementation.
```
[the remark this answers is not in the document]
GeorgesStavracas: I don't think the permission store has any particular requirements for the variant. I opted for  a simple `v` here because that's as generic as it can be, and can cover pretty much everything any portal implementation need.

src/screen-cast.c
```diff
@@ -373,12 +527,134 @@ validate_cursor_mode (const char *key,
...
+
+      if (g_strcmp0 (screen_cast_select_sources_options[i].key, "restore_token") == 0)
+        {
+          g_autoptr(GVariant) restore_data = NULL;
+          g_autofree char *restore_token = NULL;
+
+          if (is_remote_desktop_session (session))
+            {
+              g_set_error (error, XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                           "Remote desktop sessions cannot be restored");
```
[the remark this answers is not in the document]
GeorgesStavracas: I assumed the fact that this is not documented in `org.freedesktop.portal.RemoteDesktop` would be enough :slightly_smiling_face: 

Do you have any particular suggestions as to where to document this?

src/screen-cast.c
GeorgesStavracas: A mistake

src/screen-cast.c
GeorgesStavracas: Yup. Done

data/org.freedesktop.portal.ScreenCast.xml
GeorgesStavracas: Added

data/org.freedesktop.portal.ScreenCast.xml
GeorgesStavracas: Added too

GeorgesStavracas:
See https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/14 for a working, complete implementation of this feature.

GeorgesStavracas:
While working on the corresponding API in libportal, a question arised: should "restoring a stream" also restore extra information like the cursor mode, output types, etc?

To expand a bit, libportal has this API:

```c
void xdp_portal_create_screencast_session (XdpPortal           *portal,
                                           XdpOutputType        outputs,
                                           XdpScreencastFlags   flags,
                                           XdpCursorMode        cursor_mode,
                                           GCancellable        *cancellable,
                                           GAsyncReadyCallback  callback,
                                           gpointer             data);
```

Inspired by that, I thought about adding a sibling function that would look like this:

```c
void xdp_portal_restore_screencast_session (XdpPortal                *portal,
                                            XdpScreencastPersistMode  persist_mode,
                                            const char               *restore_token,
                                            GCancellable             *cancellable,
                                            GAsyncReadyCallback       callback,
                                            gpointer                  data);
```

Which seems quite natural, almost obvious of an API. Which is great.

I'm still pondering the pros and cons of making stream restore also account for cursor mode, output types, etc.

 **Pros**:
 * Easier to write an elegant corresponding libportal API
 * Portal implementations don't need to check if the to-be-restored stuff matches the output types (e.g. trying to restore a window stream, but the session is configured to monitor streams only)

**Cons**:
 * The more state we store, the harder it gets to add introduce new features or change existing ones
 * It might be annoying for apps to be "stuck" with their previous screencast state if they want to restore. Specially when some of these options can affect performance (e.g. embedded vs metadata cursors), it might be better to not restrict the API in this way.

I'm currently leaning towards the current pull request, which does not store any additional information. I'd love to hear your thoughts on this

GeorgesStavracas:
I've been talking to various design-oriented folks about this feature, and what would an ideal interaction would look like, and apparently this proposal is flexible enough to allow portal implementations to do pretty sophisticated things (like, waiting for a window or monitor to be plugged when restoring a previous stream, etc). It doesn't seem necessary to add more properties to SelectSources() to be able to do that. So I guess this API design is playing out nicely in practice!

jadahl:
> ...
> I'm currently leaning towards the current pull request, which does not store any additional information. I'd love to hear your thoughts on this

Another reason to not attempt to restore cursor mode etc is that the session might not be available to be restored, meaning the application still needs to provide those parameters in case a new session will be created.

GeorgesStavracas:
> Another reason to not attempt to restore cursor mode etc is that the session might not be available to be restored, meaning the application still needs to provide those parameters in case a new session will be created.

Great point... scratch that idea then :)

src/screen-cast.c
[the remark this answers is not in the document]
GeorgesStavracas: What I mean is that this code cannot be reached by using `org.freedesktop.portal.RemoteDesktop` - it will bail out earlier, when parsing the options and noticing that `persist_mode` is there but not allowed. I added this check in the code merely as a security measure, I don't expect that this code is ever reached in practice.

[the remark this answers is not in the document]
jadahl: This function is called from `org.freedesktop.ScreenCast.SelectSources`, which can be called on a remote desktop session too. Or am I missing somewhere where we already return an error before getting here?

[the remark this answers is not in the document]
GeorgesStavracas: Oh, I understand what you mean now! Sorry for this confusion. I've update this pull request to (1) fail earlier if it's a remote desktop session and has persist_mode != 0, and (2) document this behavior in the XML file.

data/org.freedesktop.impl.portal.ScreenCast.xml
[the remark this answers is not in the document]
GeorgesStavracas: Thanks for the suggestion. I've added a paragraph with (a modified version of) it.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -123,6 +154,44 @@
...
+              SelectSources() method call. The portal backend should store enough
+              information in 'restore_data' to be able to restore the session
+              later. The portal backend can deal with unavailable sources however
+              they see fit.
+
+              Data generated by a portal implementation must be consumable by
+              the same portal implementation, but not necessarily by others.
+              Thus, it is important to properly handle 'restore_data' not being
+              in a valid format. This may happen when, e.g., the user switched
+              to another desktop environment, or is using a different portal
```
columbarius, line 189: Does this mean, that changing portal implementation would drop a stored permission an implementation doesn't understand, or would this result in multiple entries, which are send sequentially until the implementation does find a valid one?
Would it be possible to register the namespace of the portal implementation with it?

GeorgesStavracas: > Does this mean, that changing portal implementation would drop a stored permission an implementation doesn't understand

This.

>  Would it be possible to register the namespace of the portal implementation with it?

I don't think forcing any particular format here would be helpful. Giving portal frontends the flexibility to store whatever data they want is the key part of this; it allows frontends to implement whatever interaction they see fit. Each portal frontend should check if the variant they're receiving is in a format they recognize, then parse this variant and check if the values are expected.

[For the GNOME implementation](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/14), we went with `(suxxa(uus))` as the variant format, because we need to store: `("GNOME", version, creation time, last used time, array of (stream id, stream type, stream recovery string))`. The "creation time" and "last used time" fields will be used by GNOME Settings to show information about this permission; "stream recovery string" is Mutter-specific.

columbarius: > I don't think forcing any particular format here would be helpful. Giving portal frontends the flexibility to store whatever data they want is the key part of this;

I don't want to force a full spec on what can be stored, but I usually like structures with a fixed format and one entry for arbitrary data. I have no experience with dbus and as such i can't really offer a working solution, but would it possible to save the path of the portal `org.freedesktop.impl.portal.{gnome,wlr,...}` in some specified entry of the permission store and the permission store will only send entries matching to the token and the impl? Another idea would be to specify that the first entry in the permission store has to be the portal name, such that all data would start with "GNOME", "WLROOTS", ... and this would ease parsing if a valid format is present.

DE independent permissions aren't possible at all with the current state, but it would be nice for people using multiple DE's to have the permanent settings survive a switch. 

This probably would take a "keep token" response to the client and managing multiple tokens and would make the whole thing much more complicated... 

So tldr: A standardized first entry in the variant format would be nice to avoid collitions, everything else is just nice to have if possible in a reasonable way.

GeorgesStavracas: > DE independent permissions aren't possible at all with the current state

This is currently possible, it's just discouraged. I'd prefer that each portal implementation focuses on itself and how to provide the best experience they can. But any portal implementation can check the variant type and see if it matches any other implementation's format and load it. You don't need an extra string for that.

columbarius: > This is currently possible, it's just discouraged.

How with non-standardized restore_data? 

Another idea: Should the data be only used by the portal, or should third applications (like flatseal) be allowed to edit those entries? In this case the application has to extinguish between datas from different portals.

GeorgesStavracas: > How with non-standardized restore_data?

```c
if (g_variant_check_format_string (data, "(suxxa(uus))", FALSE))
  {
      /* restore GNOME session */
  }
else if (g_variant_check_format_string (data, "(foo{bar})", FALSE))
  {
      /* restore someone else's session */
  }
```

columbarius: Ok, sry if thats now to pendantic:
```
if (g_variant_check_format_string (data, "(suxxa(uus))", FALSE))
``` 
should be 
```
if (g_variant_check_format_string (data, "(suxxa(uus))", FALSE) && first s of variant == "GNOME")
```
since two portals could use the same variant format. And restoring someone else's session would most likely be just not restoring anything and querying the user again or adding support to parse another portal implementations?

So my two concerns are:
* Collisions in the format_variant (Can be mitigated by checking the format and every portal impl individually placing a magic token somewhere, therefore maybe standardizing on the position of this magic token on the first place)
* Loosing the state by switching DEs (Setting up a scene in obs in gnome, switching to kde, scene is started , kde can't parse the permissions and drops them, user sets up new permissions, switches back to gnome, all permissions get dropped ...) 

Since the application has to store the token and doesn't know about the used portal implementation the second point seems not solvable, except an application might check sth. like `XDG_CURRENT_DESKTOP` and store multiple tokens for each DE that way.

Please correct me if I got sth. wrong or am missunderstanding sth.

GeorgesStavracas: To be clear, adding an extra string is necessarily a bad idea. The problem is where to draw the line. Is a string alone sufficient? Should it also be versioned? Who should set these strings? Are portal implementations allowed to "lie" about who their vendor strings? Should xdg-desktop-portal itself derive this vendor string? If so, where from?

These are all API questions that only exist if we try to add more data to this, and I'm reasonably sure that there is no decision that would make everyone happy. Ultimately, that's why I opted for a plain `v` variant there - it's none of x-d-p's business, and backends can still try to restore the data they want, if so they wish.

GeorgesStavracas: If you're changing your entire desktop environment, restoring streams is the smallest of your problems. If you're on KDE and switch to GNOME, or to Sway, or to any other environment, there is very little to be carried over. This API is made for restoring streams, optimizing this API for this case is going to overcomplicate things with essentially no gains.

columbarius: > To be clear, adding an extra string is necessarily a bad idea. The problem is where to draw the line. Is a string alone sufficient? Should it also be versioned? Who should set these strings? Are portal implementations allowed to "lie" about their vendor strings? Should xdg-desktop-portal itself derive this vendor string? If so, where from?

Sry now I get the point of your missunderstanding. Against xdg-desktop-portal an arbitrary variant is really fine! What I meant was more an agreement between portal implementations to start the content of their variant with a "vendor name".

columbarius: > This API is made for restoring streams, optimizing this API for this case is going to overcomplicate things with essentially no gains.

And sry again for being out of the scope of this API discussion: We can keep the API simple as is and if applications want/have a need to work on mutliple DE's they won't be able to carry over their settings (I didn't want to imply that, since the DE's might be too different for that), but they can detect the running DE and only use tokens obtained while running with this DE and as such keep all permissions wrt. the running DE.

GeorgesStavracas: >  What I meant was more an agreement between portal implementations to start the content of their variant with a "vendor name".

Oh I see. We can absolutely agree on a format with a at least few common fields that can be interpreted by every implementation. What I was trying to say is that I don't think this should be enforced in *here* - this API provides the mechanism, not the policy, to restore data.

aleixpol (approved):

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -81,6 +81,37 @@
               Whether to allow selecting multiple sources. Default is no.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>restore_data as</term>
```
smcv, line 85: Is this meant to be the same thing as the `restore_data v` in the `Response`?

How sure are we that an array of UTF-8 strings is the right type?

data/org.freedesktop.impl.portal.ScreenCast.xml
smcv: > We can absolutely agree on a format with a at least few common fields that can be interpreted by every implementation. What I was trying to say is that I don't think this should be enforced in here - this API provides the mechanism, not the policy, to restore data.

I think it's OK for the implementation to not try to validate (motto: "mechanism, not policy"), but it would still be useful to mention the convention in the specification.

It might be best to establish a convention that the `restore_data` is a struct starting with something like `GNOME-0` so that if a future `x-d-p-gnome` changes the information that it needs to store, it can use `GNOME-1`, ... for incompatible struct payloads? Or were you intending to use an append-only struct similar to the convention used in Linux syscalls?

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -123,6 +154,44 @@
...
+              This option was added in version 4 of this interface.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>restore_data v</term>
+            <listitem><para>
+              The data to be restored. This is the data that a portal
+              implementation sent in reaction to transient or persistent
+              modes.
+
```
smcv, line 177: Do you want a

```suggestion
             </para><para>
```

here and elsewhere?

src/screen-cast.c
```diff
@@ -100,6 +114,142 @@ GType screen_cast_session_get_type (void);
...
+
+  locker = g_mutex_locker_new (&transient_permissions_lock);
+
+  if (!transient_permissions)
+    return;
+
+  g_hash_table_iter_init (&iter, transient_permissions);
+  while (g_hash_table_iter_next (&iter, (gpointer *) &key, NULL))
+    {
+      g_auto(GStrv) split = g_strsplit (key, ":::", 2);
```
smcv, line 246: If the keys in this hash table start with D-Bus bus names, then I'd suggest using a delimiter that is a character that cannot appear in a syntactically valid D-Bus bus name at all (`:` can appear; `:::` cannot, but you have to read the spec more carefully to know that).

`/` would seem like an obvious choice, resulting in something like `:1.42/blah-blah-blah` splitting into sender `:1.42` and transient permissions `blah-blah-blah`.

smcv (commented):
I haven't reviewed the implementation in detail.

GeorgesStavracas:
I have changed the `restore_data` format to `(suv)` and added the corresponding type checks in the code. This is roughly what @columbarius suggested, so rejoice, this is in now :)

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -81,6 +81,47 @@
...
+              example, the session contains a monitor or a window that is not
+              available anymore, or when the stored permissions are withdrawn.
+            </para>
+            <para>
+              The restore data is composed of the vendor name (e.g. "GNOME" or
+              "KDE"), the version of the implementation-specific private data,
+              and the implementation-specific private data itself.
+            </para>
+            <para>
+              This option was added in version 4 of this interface.
```
mwleeds, line 101: We need to update the note at the top to say it's version 4? And similarly for `org.freedesktop.portal.Screencast.xml`

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -169,10 +207,29 @@
...
+            </para></listitem>
+          </varlistentry>
         </variablelist>
 
         Stream properties include:
         <variablelist>
+          <varlistentry>
+            <term>id s</term>
+            <listitem><para>
+              Opaque identifier. Will be unique for this stream for and local
```
mwleeds, line 227:
```suggestion
              Opaque identifier. Will be unique for this stream and local
```

mwleeds, line 227:
```suggestion
              Opaque identifier. Will be unique for this stream and local
```

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -123,6 +164,47 @@
...
+          <varlistentry>
+            <term>restore_data (suv)</term>
+            <listitem><para>
+              The data to be restored. This is the data that a portal
+              implementation sent in reaction to transient or persistent
+              modes.
+            </para>
+            <para>
+              This data will be stored in the permission store if the effective
+              persist mode is 2, and may be passed in a future as part of the
```
mwleeds, line 190:
```suggestion
              persist mode is 2, and may be passed in the future as part of the
```

data/org.freedesktop.impl.portal.ScreenCast.xml
GeorgesStavracas: Good catch. Updated.

GeorgesStavracas:
libportal counterpart available at https://github.com/flatpak/libportal/pull/59

mwleeds:
I didn't finish reviewing this today but I can pick it back up tomorrow.

src/screen-cast.c
```diff
@@ -624,12 +929,123 @@ collect_screen_cast_stream_data (GVariantIter *streams_iter)
...
+          delete_transient_permissions (session->sender, screen_cast_session->restore_token);
+        }
+
+      g_clear_pointer (&screen_cast_session->restore_token, g_free);
+      g_clear_pointer (&screen_cast_session->restore_data, g_variant_unref);
+      break;
+
+    case PERSIST_MODE_TRANSIENT:
+      if (screen_cast_session->restore_token == NULL)
+        screen_cast_session->restore_token = g_uuid_string_random ();
```
mwleeds, line 964: The docs for `g_uuid_string_random()` say it has the same randomness guarantees as `GRand`, which can fallback to using the current time as a seed if `/dev/urandom` isn't available. I guess we're not worried since it's an unlikely attack vector?

mwleeds:
I can't say I understand every bit of this but I reviewed it fairly well and I think we can merge once coming to a conclusion on the above comment.

src/screen-cast.c
GeorgesStavracas: I am not completely sure this can't be abused by a malicious app, but there are a few points I had in mind when writing this code:

 * As Simon said in the first comments of this PR, there is nothing we can do against malicious apps running unsandboxed on the host. They can do literally anything to these permissions.
 * `xdg-desktop-portal` checks the app id of the app that is requesting the permission, and stores the tuple `(restore-token, app-id)`. For a malicious app to breach any security using this feature, it'd have to lie about it's app id, which either requires not being sandboxed, or tricking users with a fake app id.
 *  Even if a malicious app attempts to brute-force every possible restore token, they still wouldn't be able to predict *what* that restore token will restore (this feature guarantees that apps never know what is being shared with them; they can only restore previous sessions, but they don't know what these sessions had.)

All3xJ:
In which portal version will this be implemented? I'm looking forward this "remember the choice" feature to not have always these prompts.

aleixpol:
> In which portal version will this be implemented?

I'm pretty sure GNOME already has it, for Plasma it should be available for 5.25 https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/79/

All3xJ:
> > In which portal version will this be implemented?
> 
> I'm pretty sure GNOME already has it, for Plasma it should be available for 5.25 https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/79/

which version of GNOME? I have 41.5 and don't have it

mwleeds:
> > > In which portal version will this be implemented?
> > 
> > 
> > I'm pretty sure GNOME already has it, for Plasma it should be available for 5.25 https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/79/
> 
> which version of GNOME? I have 41.5 and don't have it

42 according to https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/blob/main/NEWS

All3xJ:
> > > > In which portal version will this be implemented?
> > > 
> > > 
> > > I'm pretty sure GNOME already has it, for Plasma it should be available for 5.25 https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/79/
> > 
> > 
> > which version of GNOME? I have 41.5 and don't have it
> 
> 42 according to https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/blob/main/NEWS

I have updated my archlinux to GNOME 42, but portal window still has not the option to remember the choice

jadahl:
> I have updated my archlinux to GNOME 42, but portal window still has not the option to remember the choice

Does the application make use of the new feature? Are you running it in a flatpak?

All3xJ:
> > I have updated my archlinux to GNOME 42, but portal window still has not the option to remember the choice
> Are you running it in a flatpak?
no.
>Does the application make use of the new feature?
I thought that this new update would automatically add the option "remember the choice". So the application devs have to manually implement it? I can make a feature request in flameshot issues

jadahl:
> no.

If not via flatpak (or snap) then they need to be run in a cgroup compatible with https://github.com/flatpak/xdg-desktop-portal/pull/719

> I thought that this new update would automatically add the option "remember the choice".

No, the application needs to keep track of a token that it can use to restore a previous session.

All3xJ:
Ok thanks. But is this only for screencast, or can it be implemented on Flameshot (that uses Screenshot portal)?

All3xJ:
Is there a way to readapt the screencast restore code by @GeorgesStavracas to Screenshot portal?

marler8997:
Is there any reason this wasn't also added to remote desktop sessions?

P.S. the documentation in the PR description is great, but, would be very helpful to include some of those details in the documentation (i.e. details about the Token lifecycle).

jadahl:
> Is there any reason this couldn't also be added to remote desktop sessions?

See https://github.com/flatpak/xdg-desktop-portal/issues/850.

Issue #636, opened 2021-09-24
Slow startup due to deadlock if o.fd.FileManager1 is implemented by GtkApplication


minhduc0711:
As a result, several apps have to wait while the service was starting.

This starts happening when I updated to 1.10.1. Notice the time between the `Starting Portal service` and `Started Portal service`:
```
● xdg-desktop-portal.service - Portal service
     Loaded: loaded (/usr/lib/systemd/user/xdg-desktop-portal.service; static)
     Active: active (running) since Fri 2021-09-24 10:24:12 +07; 46s ago
   Main PID: 31907 (xdg-desktop-por)
      Tasks: 10 (limit: 23842)
     Memory: 3.7M
        CPU: 392ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/xdg-desktop-portal.service
             └─31907 /usr/lib/xdg-desktop-portal

Sep 24 10:23:47 acer-aspire-a5 systemd[924]: Starting Portal service...
Sep 24 10:24:12 acer-aspire-a5 systemd[924]: Started Portal service.
Sep 24 10:24:42 acer-aspire-a5 xdg-desktop-por[31907]: Failed to get application states: GDBus.Error:org.freedesktop.portal.Error.Failed: Could not get window list
```

The service started immediately in 1.8.1:
```
● xdg-desktop-portal.service - Portal service
     Loaded: loaded (/usr/lib/systemd/user/xdg-desktop-portal.service; static)
     Active: active (running) since Fri 2021-09-24 10:26:46 +07; 1s ago
   Main PID: 34405 (xdg-desktop-por)
      Tasks: 8 (limit: 23842)
     Memory: 3.1M
        CPU: 101ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/xdg-desktop-portal.service
             └─34405 /usr/lib/xdg-desktop-portal

Sep 24 10:26:46 acer-aspire-a5 systemd[924]: Starting Portal service...
Sep 24 10:26:46 acer-aspire-a5 systemd[924]: Started Portal service.                                                    
```

FYI I also use sway, so there might be some configuration problems.
Edit: I checked https://github.com/flatpak/xdg-desktop-portal/issues/607, but my xdg-desktop-portal-wlr seems to start up properly.

AidanGG:
[...]

PR #639, opened 2021-09-30
open-uri: Avoid calling into o.fd.FileManager1 during startup

smcv:
Making a synchronous call into a GTK implementation of o.fd.FileManager1
can cause a deadlock, because GtkApplication makes a synchronous call
into the Inhibit portal when started in a non-GNOME, non-XFCE environment.
In this situation, each service does not take its bus name (and hence
complete service-activation) until it gets a reply from the other, which
obviously can't work.

We don't actually need a GDBusProxy here: we don't need to subscribe to
any signals, watch any properties or hold any state, so it's just as
straightforward to use g_dbus_connection_call_sync() on a just-in-time
basis.

Fixes: 69961f3f "openuri: Use FileManager1 in OpenDirectory"  
Resolves: https://github.com/flatpak/xdg-desktop-portal/issues/636

---

Not thoroughly tested yet: because the issue mentioned Thunar, I tried to reproduce the issue in an XFCE environment, but unfortunately XFCE is one of the few environments where this *won't* happen.

Another option would be to create the FileManager proxy asynchronously, but that would require thread synchronization between the thread that creates the proxy and the thread that handles the request, and life's too short for that.

Ideally we'd fix this deadlock at both ends: in xdg-desktop-portal, *and* in GTK.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -60,8 +60,6 @@ typedef struct _OpenURIClass OpenURIClass;
 struct _OpenURI
 {
   XdpOpenURISkeleton parent_instance;
-
-  GDBusProxy* file_manager;
 };
 
 struct _OpenURIClass
@@ -689,7 +687,6 @@ handle_open_in_thread_func (GTask *task,
         {
           g_autofree char *real_path = get_real_path_for_doc_path (path, app_id);
 
-          if (open_uri->file_manager != NULL)
             {
               /* Try opening the directory via the file manager interface, then
                  fall back to a plain URI open */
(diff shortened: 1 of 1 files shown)


AidanGG:
I tested this change on Arch + Sway + Thunar + xdg-desktop-portal-gtk and it seems to have resolved the dead locking issue for me.

ErikReider:
This solves the issue for me too :)

src/open-uri.c
```diff
@@ -689,29 +687,43 @@ handle_open_in_thread_func (GTask *task,
...
+                                                      G_DBUS_CALL_FLAGS_NONE,
+                                                      -1,
+                                                      NULL,
+                                                      &local_error);
+
               if (result == NULL)
                 {
-                  g_warning ("Failed to call " FILE_MANAGER_SHOW_ITEMS ": %s",
-                             local_error->message);
+                  if (g_error_matches (local_error, G_DBUS_ERROR,
```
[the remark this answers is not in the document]
smcv: I wondered about that myself, but I think that would actually be wrong behaviour: if we can't contact the FileManager1 the first time, there's no reason it can't succeed when we try again later.

For example, if o.fd.FileManager1 isn't service-activatable, the first time we call it there might not be an implementation running, but by the time we ping it for a second time, the user might have started an implementation that grabbed the name.

At the moment, o.fd.FileManager1 implementations usually *are* service-activatable, but in some ways that's actually undesirable too, because D-Bus has no concept of a desktop-specific implementation of a bus name. So for example, if a GNOME user and an XFCE user share a computer, it's arbitrary (readdir() order, i.e. basically random) whether the session bus will choose GNOME's Nautilus or XFCE's Thunar to be today's implementation of o.fd.FileManager1.

At some point I'd like to resolve that by querying the `g_app_info_get_default_for_type()` for `inode/directory`, and then checking whether it's a `GDesktopAppInfo` that is `DBusActivatable`. If it is, we can try invoking *that* name via D-Bus as a FileManager1, before falling back to attempting service-activation - which will make us more likely to launch the user's preferred file manager. That's out of scope for solving the deadlock, though.

The other reason I didn't do this is that the OpenURI implementation runs in a worker thread, and there isn't any guarantee that we don't have two calls to OpenURI running concurrently; so we would need thread-synchronization between them.

src/open-uri.c
```diff
@@ -689,29 +687,43 @@ handle_open_in_thread_func (GTask *task,
         {
           g_autofree char *real_path = get_real_path_for_doc_path (path, app_id);
 
-          if (open_uri->file_manager != NULL)
```
refi64, line 692: I think you want to remove the immediately following brace pair and dedent all the code within?

refi64 (approved):
LGTM other than a minor formatting issue

src/open-uri.c
smcv: I've added a second commit which does that.

PR #641, opened 2021-10-01
build: Fix cross-compilation by adjusting pkg-config use

smcv:
Patch from Helmut Grohne, forwarded from a Debian bug report:

xdg-desktop-portal fails to cross build from source, because configure.ac
hard codes the build architecture pkg-config on one occasion, rather than
taking the host architecture pkg-config from $PKG_CONFIG. The relevant
occasion is unnecessary as the version check can be fused in an earlier
access of the relevant package with the right pkg-config.

[Commit message added by Simon McVittie, based on the text of the
Debian bug report.]

Resolves: https://bugs.debian.org/985411


GeorgesStavracas (approved):
LGTM

PR #645, opened 2021-10-05
Do not destroy the registry object during creation of PipeWireRemote

msizanoen1:
Currently, the PipeWire registry object is destroyed when
discover_node_factory_sync finishes it's functionality. This causes
registry events to not be delivered to the global_removed_cb and
global_added_cb callbacks after discover_node_factory_sync, breaking
the functionality of the
org.freedesktop.portal.Camera.IsCameraPresent property in case of
camera device hot(un)plug.

This changes the discover_node_factory_sync function to store the
registry in the PipeWireRemote structure and destroy it in
pipewire_remote_destroy.

### DIFF
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -126,17 +126,13 @@ static gboolean
 discover_node_factory_sync (PipeWireRemote *remote,
                             GError **error)
 {
-  struct pw_registry *registry;
-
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
Could you please rebase against latest master? Also, please change the commit title to `pipewire: Destroy registry object with remote`

GeorgesStavracas (approved):
LGTM

PR #659, opened 2021-11-11
device: Add OpenPipeWireRemote() method

GeorgesStavracas:
It works very similarly to the same-named method of the Camera portal, except it receives a list of devices to access.

This new method returns an error if an unsupported device is requested, or if the app doesn't have permission to access any of the requested devices.

### DIFF
--- a/data/org.freedesktop.portal.Device.xml
+++ b/data/org.freedesktop.portal.Device.xml
@@ -28,7 +28,7 @@
       Not a portal in the strict sense, since the API is not directly
       accessible to applications inside the sandbox.
 
-      This documentation describes version 1 of this interface.
+      This documentation describes version 2 of this interface.
     -->
   <interface name="org.freedesktop.portal.Device">
     <!--
@@ -57,6 +57,39 @@
       <arg type="as" name="devices" direction="in"/>
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        OpenPipeWireRemote:
+        @options: Vardict with optional further information
+        @fd: File descriptor of an open PipeWire remote.
+
+        Open a file descriptor to the PipeWire remote where the microphone,
+        speakers, and camera nodes are available. The file descriptor should be
+        used to create a <classname>pw_core</classname> object, by using
+        <function>pw_context_connect_fd</function>.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>devices as</term>
+            <listitem><para>
+              The devices to access. Supported devices for this call are 'camera',
+              'microphone', and 'speakers'.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+        This method will only succeed if the application already has permission
+        to access any one of the devices requested.
+
+        This option was added in version 2 of this interface.
+    -->
+    <method name="OpenPipeWireRemote">
+      <annotation name="org.gtk.GDBus.C.Name" value="open_pipewire_remote"/>
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="h" name="fd" direction="out"/>
     </method>
     <property name="version" type="u" access="read"/>
   </interface>
--- a/data/org.freedesktop.portal.Device.xml
+++ b/data/org.freedesktop.portal.Device.xml
@@ -91,6 +91,33 @@
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="h" name="fd" direction="out"/>
     </method>
+
+    <!--
+        IsCameraPresent:
+
+        A boolean stating whether there is any camera available.
+
+        This property was added in version 2 of this interface.
+    -->
+    <property name="IsCameraPresent" type="b" access="read"/>
+
+    <!--
+        IsSpeakerPresent:
+
+        A boolean stating whether there is any speaker available.
+
+        This property was added in version 2 of this interface.
+    -->
+    <property name="IsSpeakerPresent" type="b" access="read"/>
+    <!--
+        IsMicrophonePresent:
+
+        A boolean stating whether there is any microphone available.
+
+        This property was added in version 2 of this interface.
+    -->
+    <property name="IsMicrophonePresent" type="b" access="read"/>
+
     <property name="version" type="u" access="read"/>
   </interface>
 </node>
--- a/src/device.c
+++ b/src/device.c
@@ -30,6 +30,7 @@
 
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <gio/gunixfdlist.h>
 #include <gio/gdesktopappinfo.h>
 
 #include "device.h"
@@ -38,6 +39,10 @@
 #include "xdp-dbus.h"
 #include "xdp-impl-dbus.h"
 #include "xdp-utils.h"
+
+#ifdef HAVE_PIPEWIRE
+#include "pipewire.h"
+#endif
 
 #define PERMISSION_TABLE "devices"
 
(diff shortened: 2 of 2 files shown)


GeorgesStavracas:
I couldn't add the tests since they rely on libportal implementing this, but it's not there yet. I'll add the corresponding code as a separate step after working on libportal's implementation.

GeorgesStavracas:
@wtay hey, i've opened the corresponding media-session MR relative to this portal: https://gitlab.freedesktop.org/pipewire/media-session/-/merge_requests/25

I'd love to hear your thoughts on this portal

data/org.freedesktop.portal.Audio.xml
```diff
@@ -0,0 +1,74 @@
...
+        </variablelist>
+
+        Following the #org.freedesktop.portal.Request::Response signal, if
+        granted, #org.freedesktop.org.Audio.OpenPipeWireRemote can be used to
+        open a PipeWire remote.
+    -->
+    <method name="AccessAudio">
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
```
matthiasclasen, line 52: Is there no way for the app to express that it is only interested in input, or only in output ?

data/org.freedesktop.portal.Audio.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2021 Endles OS Foundation, LLC
```
ebassi, line 3:
```suggestion
 Copyright (C) 2021 Endless OS Foundation, LLC
```

data/org.freedesktop.portal.Audio.xml
GeorgesStavracas: I thought a bit about that, and decided to not add until requested by any reviewed :slightly_smiling_face: 

I'll add that

GeorgesStavracas: Added

GeorgesStavracas:
Following IRC discussions with Wim, I'm changing this PR to add a new method to the Devices portal instead. The reasoning behind this shift in direction is that it is better to open one single PipeWire remote with different media roles, instead of one per media role. Things like GStreamer will benefit from it.

The Camera portal is made somewhat useless with this change, but I didn't touch it for now. At some point, it'll be interesting for us to retire the Camera portal.

jadahl:
> Following IRC discussions with Wim, I'm changing this PR to add a new method to the Devices portal instead. The reasoning behind this shift in direction is that it is better to open one single PipeWire remote with different media roles, instead of one per media role. Things like GStreamer will benefit from it.

I've been also thinking it's a good idea to explore a "common" portal for opening PipeWire remotes to not have to use multiple PipeWire remotes for different. I'm not sure about placing it inside the `Device` portal though, it's something that has been used by e.g. pulseaudio to check permissions from outside the sandbox IIRC. Might be something we can eventually remove, so I think adding a `org.freedesktop.portal.Media` or something is better way.

> The Camera portal is made somewhat useless with this change, but I didn't touch it for now. At some point, it'll be interesting for us to retire the Camera portal.

Yes, makes sense to deprecate the camera portal if we add something that can handle both camera and other media roles.

GeorgesStavracas:
> I'm not sure about placing it inside the Device portal though, it's something that has been used by e.g. pulseaudio to check permissions from outside the sandbox

This pull request doesn't interfere with PulseAudio in any way, applications that need device access should use the `AccessDevice()` method exactly like they used to. This pull request merely adds a way to grab a filtered PipeWire fd for cameras, speakers, and / or microphones.

jadahl:
Hmm. I see what you mean, the client would take over "ask for access" action from the camera portal/pulseaudio, by doing it "manually" via AccessDevice. Yea, maybe this is the best way.

jadahl:
Maybe we should add properties for whether a certain device class exists on the system, as currently handled by the camera portal. Quoting the Camera pull request:

> In order to allow an application to only request camera permission would there be any cameras available, there is a IsCameraPresent boolean property that will expose whether there are any cameras available at all.

GeorgesStavracas:
Good idea. That would translate to adding `IsCameraPresent`, `IsSpeakerPresent`, and `IsMicrophonePresent` here. I can add that. Should we change these property names?

jadahl:
> Should we change these property names?

They seem fine to me already.

GeorgesStavracas:
Added these properties! Most of this code was adapted from the Camera portal

msizanoen1:
How will monitor devices be handled in this proposal? IMO the capability to output to an audio device and to capture output from an audio device should be separate.

jadahl:
> How will monitor nodes be handled in this proposal? IMO the capability to output to an audio device and to capture output from an audio device should be separate.

This portal (with this MR) sets the PipeWire "media role" to either "Speaker", "Microphone", or "Camera". It is then up to PipeWire to determine whether a moniter output device falls into any of these categories. I would assume it shouldn't. @wtay is this the case?

Whether we should add monitor devices here, I'm not sure. Some how we need a portal to allow "arbitrary" PipeWire routing e.g. for DAW:s, which the "monitor" capture somewhat falls into, but not sure how that permission model should look like.

msizanoen1:
Apparently in PipeWire output and monitor are two port on the same node/device. It should be possible to separate monitor ports into a separate category and change PipeWire and the media session controller perform access control on port-level granularity instead of device/node-level. (Currently the function responsible for checking permission to create a link between two ports works on the device/node-granularity level and is [a stub that always return success](https://github.com/PipeWire/pipewire/blob/c387e83c65b7a55ea4fbf784c46d7b20d5640fc6/src/pipewire/impl-link.c#L1112-L1119)).

wtay:
> Currently the function responsible for checking permission to create a link between two ports works on the device/node-granularity level and is a stub that always return success)

It's not really a problem, the port permissions are first checked when a client wants to create a link and the client just can make the link. 

It is usually the session manager that makes the links for clients, and it usually has all permissions anyway. It is the session manager that should deny creating links between a monitor port and the client stream when it has no permission.

data/org.freedesktop.portal.Device.xml
```diff
@@ -58,6 +58,66 @@
...
+            <term>devices as</term>
+            <listitem><para>
+              The devices to access. Supported devices for this call are 'camera',
+              'microphone', and 'speakers'.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+        This method will only succeed if the application already has permission
+        to access any one of the devices requested.
```
devurandom, line 84: "any one" or "all"?

GeorgesStavracas: "any one" is the accurate wording

Naheem: It might be useful to change "one" to a different word to avoid ambiguity of English language. maybe "single"?

rmader:
@GeorgesStavracas @jadahl just a reminder to look into this one soon if possible :)

data/org.freedesktop.portal.Device.xml
```diff
@@ -58,6 +58,66 @@
...
+        <function>pw_context_connect_fd</function>.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>devices as</term>
+            <listitem><para>
+              The devices to access. Supported devices for this call are 'camera',
+              'microphone', and 'speakers'.
+            </para></listitem>
```
msizanoen1, line 79: Note: the `speakers` permission should control the application's ability to capture output sound from the output device. The application shouldn't need any special permissions to play sound, so as to be consistent with other platforms' permission model.

msizanoen1 (changes requested):

data/org.freedesktop.portal.Device.xml
smcv: Mechanically, this is not true: applications in a Flatpak sandbox have no way to play sound unless their permissions allow for it, either statically (`--socket=pulseaudio`, `--socket=pipewire` or maybe `--device=all`), or dynamically (something like this interface).

I agree it would make sense for this interface to have a way to grant audio output permission to apps statically, rather than prompting the user - but that would have to be implemented by the app asking xdg-desktop-portal for access to the speakers, and xdg-desktop-portal replying "yes, here is your file descriptor" without asking the user's permission first. This is analogous to the way the OpenUri portal will open `http:` and `https:` URLs in a web browser without asking permission (again, to match app-author expectations from other platforms), but the app still has to ask the portal to do it.

If capturing the mixed output of all applications (a "monitor" in PulseAudio terms, I don't know what Pipewire calls this) is something that sandboxed apps will want to do, then I think that would be a fourth device in this interface.

GeorgesStavracas:
Converting to draft until the following is done:

 * Reconsider permissions naming; specifically, "speakers"
 * Add support for "monitor" permission type

Be-ing:
I've been looking in to the camera portal API and I was a bit surprised by the all-or-nothing character of a security feature and that seems to be replicated here. I expected the permissions to be granted on a per device basis rather than per category of device.

> That would translate to adding IsCameraPresent, IsSpeakerPresent, and IsMicrophonePresent here. I can add that. Should we change these property names?

How about using more general names, "audio input" and "audio output"? The portal has no idea what devices are actually connected to the audio interface. Does it make sense to ask permission to use speakers when only headphones are connected? For all we know the user could be setting up an audio effects loop with a DAW and a guitar pedal without anybody hearing the audio.

Be-ing:
> Add support for "monitor" permission type

The word "monitor" without further context is ambiguous. A computer screen is a "monitor", which could be confused with the ScreenCast portal. Studio speakers are also called "monitors". But I don't think either of these meanings are what you're talking about.

wtay:
> I've been looking in to the camera portal API and I was a bit surprised by the all-or-nothing character of a security feature and that seems to be replicated here. I expected the permissions to be granted on a per device basis rather than per category of device.

The portal does not handle permissions, It only signals to the session manager to enforce permissions on 'camera' devices. Using the permission store you get per device permissions (and new devices that pop up later also get the right permissions).

vchernin:
Should there be a fallback mechanism for Flatpaks to access the PipeWire socket in case this method isn't available? Even after this API is introduced, apps would have to keep `filesystem=xdg-run/pipewire-0` in their manifests for older xdg-desktop-portal versions.

I imagine a similar fallback to `wayland` and `fallback-x11` which allows Wayland supporting apps to have safer permissions on Wayland desktops, while still working on X11. 

Or is fallback unnecessary given most distros using PipeWire would typically have newer xdg-desktop-portal? Even if `filesystem=xdg-run/pipewire-0` should be kept around for some time, it's not the worst sandbox hole.

jadahl:
After some IRC discussion, we'll need wireplumber to set up "roles" or something for the missing device classes. A starting point for this is https://gitlab.freedesktop.org/pipewire/wireplumber/-/merge_requests/455.

data/org.freedesktop.portal.Device.xml
```diff
@@ -58,6 +58,66 @@
...
+        used to create a <classname>pw_core</classname> object, by using
+        <function>pw_context_connect_fd</function>.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>devices as</term>
+            <listitem><para>
+              The devices to access. Supported devices for this call are 'camera',
+              'microphone', and 'speakers'.
```
orowith2os, line 78: I was poking around the OBS Studio PRs regarding things like virtual cameras, and I wonder if adding the ability for the portal to ask the user to let an app create virtual webcams would be a good idea?

jadahl:
Something that came to mind during a related discussion: might want to create a new interface entry point, e.g. `org.freedesktop.portal.Devices` to avoid confusion about the existing method that's useless from sandboxed apps.

GeorgesStavracas:
Closing as this stalled, and needs a fresh reimagination

PR #660, opened 2021-11-17
CI: Use a privileged docker container to support FUSE tests

3v1n0:
In order to run various tests, we need a working /dev/fuse, and this is
currently only possible in a privileged container [1].
Github actions workflow doesn't suppport it natively, so we need to
initialize a new one ourself, we just keep it running and execute the
commands when we need them, not to depend on a static image, so that the
github workflow can still be followed as atomic operations.

Also ensuring now that the installed tests are installed and that they're actually running.

/cc @TingPing 

[1] https://github.com/docker/for-linux/issues/321

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -20,6 +20,7 @@ jobs:
       CFLAGS: -fsanitize=undefined -fsanitize-undefined-trap-on-error -fsanitize=address -Wp,-D_FORTIFY_SOURCE=2
       BUILD_CONTAINER: ${{ matrix.compiler }}-build-container
       RUN_CMD: docker exec -t -w /src -e TEST_IN_CI -e ASAN_OPTIONS -e G_MESSAGES_DEBUG -e XDG_DATA_DIRS ${{ matrix.compiler }}-build-container
+      AS_USER: runuser -u tester --
 
     steps:
     - name: Prepare container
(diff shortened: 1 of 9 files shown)


.github/workflows/check.yml
```diff
@@ -8,52 +8,65 @@ env:
...
+      run: |
+        docker run --name $BUILD_CONTAINER \
+          --tty --device /dev/fuse --cap-add SYS_ADMIN \
+          --security-opt apparmor:unconfined \
+          -v $(pwd):/src \
+          -e DEBIAN_FRONTEND \
+          -e DEBCONF_NONINTERACTIVE_SEEN=true \
+          -e TERM=dumb \
+          -e CC -e CFLAGS \
+          -d ubuntu:rolling sleep infinity
```
TingPing, line 53: Why does this need to be rolling, I'd prefer staying on versioned releases.

3v1n0: Sure, [that's just an alias](https://hub.docker.com/_/ubuntu) for the latest-released release (so 21.10 at the date).

I know the name is misleading, but that's refers to the latest stable, otherwise it would have been `devel` (that is `jammy` 22.04 right now).

TingPing: I prefer just setting 21.10 as it avoids a chance of waking up with it suddenly broken when 22.04 comes out, instead we can choose when to dedicate time to updating it.

3v1n0: Ok, fair enough... I've changed it.

3v1n0:
> If you could rebase this on master manually I'll merge this.

Oh, sorry I thought I did already, apparently I used the wrong remote for pulling.

Thanks!

3v1n0:
Ah, should I maybe open an issue, but it would be nice to probably use github packages to just build an image every now and then so that we don't have to do it all the times.

TingPing:
> Ah, should I maybe open an issue, but it would be nice to probably use github packages to just build an image every now and then so that we don't have to do it all the times.

Yeah go ahead and open an issue. I'm up for it but haven't had time to do it.

PR #670, opened 2021-12-01
Sort implementations based on the `UseIn` key

Marukesu:
this fixes an possible issue in  `org.freedesktop.portal.Settings` that they would call an fallback backend first instead of the supported one,  possibly breaking the `org.freedesktop.appearance.color-scheme` support, if both implements it.

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -103,12 +103,48 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
   return TRUE;
 }
 
+static gboolean
+g_strv_case_contains (const gchar * const *strv,
+                      const gchar         *str)
+{
+  for (; strv && *strv != NULL; strv++)
+    {
+      if (g_ascii_strcasecmp (str, *strv) == 0)
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
 static gint
-sort_impl_by_name (gconstpointer a,
-                   gconstpointer b)
+sort_impl_by_use_in_and_name (gconstpointer a,
+                              gconstpointer b)
 {
   const PortalImplementation *pa = a;
   const PortalImplementation *pb = b;
+  const char *desktops_str = NULL;
+  g_auto(GStrv) desktops = NULL;
+  int i;
+
+  desktops_str = g_getenv ("XDG_CURRENT_DESKTOP");
+
+  if (desktops_str == NULL)
+    desktops_str = "";
+
+  desktops = g_strsplit (desktops_str, ":", -1);
+
+  for (i = 0; desktops[i] != NULL; i++)
+    {
+      gboolean use_a = g_strv_case_contains ((const char **)pa->use_in, desktops[i]);
+      gboolean use_b = g_strv_case_contains ((const char **)pb->use_in, desktops[i]);
+
+      if (use_a != use_b)
+        return use_b - use_a;
+      else if (use_a)
+        break;
+      else
+        continue;
+    }
 
   return strcmp (pa->source, pb->source);
 }
@@ -159,20 +195,7 @@ load_installed_portals (gboolean opt_verbose)
         }
     }
 
-  implementations = g_list_sort (implementations, sort_impl_by_name);
-}
-
-static gboolean
-g_strv_case_contains (const gchar * const *strv,
-                      const gchar         *str)
-{
-  for (; *strv != NULL; strv++)
-    {
-      if (g_ascii_strcasecmp (str, *strv) == 0)
-        return TRUE;
-    }
-
-  return FALSE;
+  implementations = g_list_sort (implementations, sort_impl_by_use_in_and_name);
 }
 
 PortalImplementation *


src/portal-impl.c
```diff
@@ -103,13 +103,51 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
   return TRUE;
 }
 
+static gboolean
+g_strv_case_contains (const gchar * const *strv,
+                      const gchar         *str)
+{
+  for (; *strv != NULL; strv++)
```
GeorgesStavracas, line 110: For extra paranoia:

```suggestion
  for (; strv && *strv != NULL; strv++)
```

src/portal-impl.c
```diff
@@ -103,13 +103,51 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
...
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
 static gint
-sort_impl_by_name (gconstpointer a,
-                   gconstpointer b)
+sort_impl_by_use_in (gconstpointer a,
```
GeorgesStavracas, line 120: Since you're still comparing by name in the fallback case, `sort_impl_by_use_in_and_name` seems more appropriate

src/portal-impl.c
```diff
@@ -103,13 +103,51 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
...
+                     gconstpointer b)
 {
   const PortalImplementation *pa = a;
   const PortalImplementation *pb = b;
 
+  const char *desktops_str = g_getenv ("XDG_CURRENT_DESKTOP");
+  g_auto(GStrv) desktops = NULL;
+  int i;
```
GeorgesStavracas, line 128: Please move this up in the function, and separate declaration and assignment:

```c
  const PortalImplementation *pa = a;
  const PortalImplementation *pb = b;
  const char *desktops_str = NULL;
  g_auto(GStrv) desktops = NULL;
  int i;

  desktop_str = g_getenv ("XDG_CURRENT_DESKTOP");
  ...
```

src/portal-impl.c
```diff
@@ -103,13 +103,51 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
...
+
+      if (use_a == use_b)
+        {
+          if (use_a)
+             break;
+          else
+            continue;
+        }
+
+      return use_a ? -1 : 1;
```
GeorgesStavracas, line 148: I think it looks a bit easier to read if written like this:

```suggestion
      if (use_a != use_b)
        return use_a ? -1 : 1;
      else if (use_a)
        break;
      else
        continue;
```

GeorgesStavracas (changes requested):
Some comments, thanks for working on it

src/portal-impl.c
```diff
@@ -103,12 +103,48 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
...
+
+  desktops = g_strsplit (desktops_str, ":", -1);
+
+  for (i = 0; desktops[i] != NULL; i++)
+    {
+      gboolean use_a = g_strv_case_contains ((const char **)pa->use_in, desktops[i]);
+      gboolean use_b = g_strv_case_contains ((const char **)pb->use_in, desktops[i]);
+
+      if (use_a != use_b)
+        return use_a ? -1 : 1;
```
GeorgesStavracas, line 142: Ugly nitpick: maybe exploit the fact that TRUE = 1 and FALSE = 0, and do:

```suggestion
        return use_b - use_a;
```

GeorgesStavracas (changes requested):
Please squash the commits together, and either apply or reject my nitpicky suggestion below. LGTM when you do that.

GeorgesStavracas:
The title of the first commit should be something like `impl: Sort by UseIn and name`

Please add a description of your changes, what are they fixing, and how it is done, in the commit message. Please keep in mind that commit message lines are limited to 72 characters in length.

Marukesu:
it should be all good now.

src/portal-impl.c
```diff
@@ -103,12 +103,48 @@ register_portal (const char *path, gboolean opt_verbose, GError **error)
...
+
+  desktops = g_strsplit (desktops_str, ":", -1);
+
+  for (i = 0; desktops[i] != NULL; i++)
+    {
+      gboolean use_a = g_strv_case_contains ((const char **)pa->use_in, desktops[i]);
+      gboolean use_b = g_strv_case_contains ((const char **)pb->use_in, desktops[i]);
+
+      if (use_a != use_b)
+        return use_a - use_b;
```
GeorgesStavracas, line 142: Hm, isn't it supposed to be the opposite? We want to return -1 when `use_a` is 1, and that only happen with `use_b - use_a`

Marukesu: You're right, I don't inverted those, should be right now.

GeorgesStavracas (approved):

GeorgesStavracas:
Thanks, this looks good to me now

PR #673, opened 2021-12-05
Update Spanish translation

oscfdezdz:


GeorgesStavracas:
@fitojb what do you think?

oscfdezdz:
GNOME Spanish translation team has reviewed and approved the translation. I should have mentioned it before, my bad 😅
https://l10n.gnome.org/vertimus/xdg-desktop-portal/master/po/es/

PR #675, opened 2021-12-10
Make sure to properly close sessions when invoked from portal backend implementation

grulja:
We don't seem to close our XdpSession when XdpImplSession::Closed() signal is called and we don't emit XdpSession::Closed() signal when it actually happens. Better description is in each commit.


grulja:
> _No description provided._

GeorgesStavracas (approved):
Both changes make sense to me, thanks!

PR #678, opened 2021-12-11
CI: build & deploy the docs

bilelmoussaoui:
cc @TingPing

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,43 @@
+name: Documentation
+
+on: [push, pull_request]
+
+jobs:
+  docs:
+    runs-on: ubuntu-20.04
+    container:
+      image: ubuntu:22.04
+    steps:
+      - name: Install dependencies
+        env:
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          apt-get update
+          apt-get upgrade -y
+          apt-get install -y --no-install-recommends \
+          make automake autoconf libtool gettext autopoint gcc \
+          gtk-doc-tools shared-mime-info desktop-file-utils gnome-desktop-testing xmlto fuse libflatpak-dev \
+          libglib2.0-dev libgeoclue-2-dev libjson-glib-dev libfontconfig1-dev libfuse-dev libportal-dev libpipewire-0.3-dev
+
+      - name: Check out xdg-desktop-portal
+        uses: actions/checkout@v2
+
+      - name: Build docs
+        run: |
+          ./autogen.sh --enable-docbook-docs
+          make
+
+      - name: Prepare docs
+        working-directory: doc
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Build docs
+        run: |
+          ./autogen.sh --enable-docbook-docs
+          make
+
+      - name: Prepare docs
+        working-directory: doc
+        run: |
+          mkdir published-docs
+          mv portal-docs.html ./published-docs
```
bilelmoussaoui, line 34:
```suggestion
          mv portal-docs.html ./published-docs
```
shouldn't this file be installed as `index.html` instead? :) we could create a file that redirects to the new url from the old one to avoid any breakages

TingPing: Sure, go ahead.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+          apt-get upgrade -y
+          apt-get install -y --no-install-recommends \
+          make automake autoconf libtool gettext autopoint gcc \
+          gtk-doc-tools shared-mime-info desktop-file-utils gnome-desktop-testing xmlto fuse libflatpak-dev \
+          libglib2.0-dev libgeoclue-2-dev libjson-glib-dev libfontconfig1-dev libfuse-dev libportal-dev libpipewire-0.3-dev
+
+      - name: Check out xdg-desktop-portal
+        uses: actions/checkout@v2
+
+      - name: Build docs
```
TingPing, line 25: I changed the main build workflow to build docs itself. Does it make sense to just add the deploy step to one of those instead of having its own? Fewer duplicated places to maintain dependencies and whatnot.

TingPing: Or we could keep this and disable docs in the main workflow so for contributors it has a more clear result.

bilelmoussaoui: I moved the docs part out of the checks jobs. Ideally, there would be some containers built that contain all the dependencies needed for the CI to run, but that's out of the scope of this PR & it seems to be fast enough, so it's fine I guess?

TingPing: Yeah, I'm ok with that.

TingPing:
@bilelmoussaoui After merging the deploy step has failed: https://github.com/flatpak/xdg-desktop-portal/runs/4499571313?check_suite_focus=true

TingPing:
Attempted fix: https://github.com/flatpak/xdg-desktop-portal/commit/9c5bf0c3e6bcf4d09df04e0ca84e000a256b642a

TingPing:
Which failed: https://github.com/flatpak/xdg-desktop-portal/runs/4499584217?check_suite_focus=true

PR #690, opened 2022-01-13
Add max-width

TheEvilSkeleton:
It makes the page easier to read.

This is how it looks like on a 1080p display:
image


TheEvilSkeleton:
I added an extra newline by accident so I removed it. It should be good to go.

GeorgesStavracas (approved):

Issue #681, opened 2021-12-21
Dynamic Launcher Portal design questions


mwleeds:
Context: https://gitlab.gnome.org/Teams/Design/os-mockups/-/issues/144

This issue is for discussing the technical architecture of a new portal that would allow sandboxed applications to install `.desktop` files and icons used by those desktop files to make launcher entries in the desktop environment's application launcher. The primary use case for the portal is to allow Epiphany or Chromium to install web apps when running under Flatpak (or Snap). However it could have other use cases.

cc @matthiasclasen 

## Requirements
- Sandboxed applications should be able to request installation of a launcher, resulting in a dialog where the user can see the name and icon, and in the case of web apps the URL. Depending on the options used by the calling application, the user can edit the name (Chromium would not allow this). The sandboxed application can then use the name returned by the portal and other information to create a `.desktop` file which it would pass to the portal. The portal would rewrite the `Exec` line to prepend it with `/usr/bin/flatpak run com.example.App` using the app ID of the calling app. This ensures that (a) the user is consenting to the sandboxed application installing the launcher and (b) the launcher is only able to run binaries in the same sandbox as the app. The portal would then install the `.desktop` file and icon file to locations in $XDG_DATA_DIRS which are not writable to the sandboxed app (unless that permission was specifically granted).
- The portal could also be used by unsandboxed applications. This makes sense so that Epiphany can use the same code path when installing web apps rather than maintaining two code paths that do the same thing.
- For the initial implementation the portal does not need to allow the launcher to be edited after being installed. Epiphany currently allows editing the name or icon of installed web apps, and apps may also want to update their own name or icon for example if their logo changes, but this could be left out of the MVP.
- There also needs to be a mode where the portal allows the installation of a `.desktop` launcher by a sandboxed application without any user interaction, but only if a token (a UUID string) is provided which was previously obtained by an unsandboxed trusted system component. This is necessary so that GNOME Software can install web apps without requiring user interaction beyond clicking the "Install" button. GNOME Software would talk to Epiphany over D-Bus and provide the token acquired from the portal, so that Epiphany can stay in charge of the on-disk format for its web apps.
- The portal also needs to allow deleting launchers. For sandboxed applications we could only allow deletion of launchers installed by that application so that the portal doesn't need to do any user interaction.

## Open questions
- Is the interface in the right shape?
[...]

Issue #326, opened 2019-05-17
Desktop file portal


lionirdeadman:
Certain programs (Lutris, Epiphany, menulibre, etc.) may want to create desktop files, we should give a portal to do so.

We probably want a way to track what files each flatpak create to remove them when the flatpak is deleted.

TingPing:
This was just briefly discussed by Epiphany and some thoughts brought up:

Not only does `flatpak` need to have knowledge of these new launchers but also stores like `gnome-software` will want appdata and the ability to delete them itself most likely via a new `flatpak` API.

sonnyp:
[ Dynamic Launcher Portal design questions #681](https://github.com/flatpak/xdg-desktop-portal/issues/681)

mwleeds:
Released in 1.14.0

PR #696, opened 2022-01-21
Add dynamic launcher portal

mwleeds:
This is not ready to be merged since I haven't implemented the gnome backend or the libportal side of it, and I haven't tested it. But it can be reviewed.

For context see #681 and #326

### DIFF
--- a/data/org.freedesktop.portal.DynamicLauncher.xml
+++ b/data/org.freedesktop.portal.DynamicLauncher.xml
@@ -1,0 +1,332 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2021 Matthew Leeds
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Matthew Leeds <mwleeds@protonmail.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.DynamicLauncher:
+      @short_description: Portal for installing application launchers onto the
+        desktop
+
+      The DynamicLauncher portal allows sandboxed (or unsandboxed) applications
+      to install launchers (.desktop files) which have an icon associated with them
+      and which execute a command in the application. The desktop environment
+      would display the launcher to the user in its menu of installed applications.
+      For example this can be used by a sandboxed browser to install web app
+      launchers. The portal also allows apps to uninstall the launchers, launch
+      them, and read the desktop file and icon data for them.
+
+      The standard way to install a launcher is to use the PrepareInstall() method
+      which results in a dialog being presented to the user so they can confirm
+      they want to install the launcher. Then, the token returned by PrepareInstall()
+      would be passed to the Install() method to complete the installation.
+
+      However, in the rare circumstance that an unsandboxed process such as a
+      system component needs to install a launcher without user interaction, this
+      can be accomplished by using the RequestInstallToken() method and passing
+      the acquired token to Install().
+
+      This documentation describes version 1 of this interface.
+  -->
+  <interface name="org.freedesktop.portal.DynamicLauncher">
+    <!--
+        Install:
+        @token: Token proving authorization of the installation
+        @desktop_file_id: The .desktop file name to be used
+        @desktop_entry: The text of the Desktop Entry file to be installed, see below
+        @options: Vardict with optional further information
+
+        Installs a .desktop launcher and icon into appropriate directories to
+        allow the desktop environment to find them. Please note that this method
+        overwrites any existing launcher with the same id. If you want to
+        present the user with a confirmation dialog in that case, you can check
+        for it using the GetDesktopEntry() method and clean up any state from
+        the previous launcher if you want.
+
+        @token must be a token that was returned by a previous
+        org.freedesktop.portal.DynamicLauncher.PrepareInstall() or
+        org.freedesktop.portal.DynamicLauncher.RequestInstallToken() call.
+        The token can only be used once and is valid for up to five minutes.
+
+        The icon and name used for the launcher will be the ones from the previous
+        org.freedesktop.portal.DynamicLauncher.PrepareInstall() or
+        org.freedesktop.portal.DynamicLauncher.RequestInstallToken() call.
+
+        The @desktop_file_id must have ".desktop" as a suffix. Except in the
+        special case when the calling process has no associated app ID,
+        @desktop_file_id must have the app ID followed by a period as a prefix,
+        regardless of whether the calling process is sandboxed or unsandboxed.
+
+        The @desktop_entry should be a valid desktop entry file beginning with
+        "[Desktop Entry]", except it should not include Name= or Icon= entries
+        (if present, these will be overwritten by the portal implementation).
+        The Exec= entry will be rewritten to call the sandboxed application e.g.
+        via "flatpak run", if the application is sandboxed.
+
+        It is recommended to include a TryExec= line with either a binary name
+        or an absolute path. The launcher will be deleted if the TryExec binary
+        cannot be found on session start.
+
+        The @options vardict currently has no supported entries.
+    -->
+    <method name="Install">
+      <arg type="s" name="token" direction="in"/>
+      <arg type="s" name="desktop_file_id" direction="in"/>
+      <arg type="s" name="desktop_entry" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+    <!--
+        PrepareInstall:
+        @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
+        @name: The default name for the launcher
+        @icon_v: A #GBytesIcon icon as returned by g_icon_serialize(). Must be
+          a png or jpeg no larger than 512x512, or an svg
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Presents a dialog to the user to allow them to see the icon, potentially
+        change the name, and confirm installation of the launcher.
+
+        Supported keys in the @options vardict:
+        <variablelist>
+          <varlistentry>
+            <term>handle_token s</term>
+            <listitem><para>
+              A string that will be used as the last element of the @handle. Must be a valid
+              object path element. See the #org.freedesktop.portal.Request documentation for
+              more information about the @handle.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>modal b</term>
+            <listitem><para>
+              Whether to make the dialog modal. Defaults to yes.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>launcher_type u</term>
+            <listitem><para>
+              The type of launcher being created. For supported values see the
+              SupportedLauncherTypes property. Defaults to "Application".
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>target s</term>
+            <listitem><para>
+              For a launcher of type "Webapp", this is the URL of the web app
+              being installed. This is displayed in the user-facing dialog.
+              For other launcher types, this is not needed.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>editable_name b</term>
+            <listitem><para>
+              If true, the user will be able to edit the name of the launcher.
+              Defaults to true.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>editable_icon b</term>
+            <listitem><para>
+              If true, the user will be able to edit the icon of the launcher,
+              if the implementation supports this. Defaults to false.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>name s</term>
+            <listitem><para>
+              The name chosen by the user for the launcher.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>token s</term>
+            <listitem><para>
+              Token that can be passed to a subsequent org.freedesktop.portal.DynamicLauncher.Install() call to
+              complete the installation without another dialog.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="PrepareInstall">
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="s" name="name" direction="in"/>
+      <arg type="v" name="icon_v" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+    <!--
+        RequestInstallToken:
+        @name: The name that will be used in the desktop file
+        @icon_v: A #GBytesIcon icon as returned by g_icon_serialize(). Must be
+          a png or jpeg no larger than 512x512, or an svg
+        @options: Vardict with optional further information
+        @token: the token to be used with the Install() method
+
+        This method is intended for use only by specific components that
+        have their application ID allowlisted in the portal backend (e.g. GNOME
+        Software and KDE Discover). It is otherwise not guaranteed to work.
+
+        The token returned by this method can be used to avoid the need for a
+        confirmation dialog; the token can be passed to the Install() method
+        just as if it were acquired via the PrepareInstall() method.
+
+        The @options vardict currently has no supported entries.
+    -->
+    <method name="RequestInstallToken">
+      <arg type="s" name="name" direction="in"/>
+      <arg type="v" name="icon_v" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="s" name="token" direction="out"/>
+    </method>
+    <!--
+        Uninstall:
+        @desktop_file_id: The .desktop file name
+        @options: Vardict with optional further information
+
+        This method deletes the desktop file and corresponding icon from the
+        appropriate directories to remove the launcher referred to by
+        @desktop_file_id.
+
+        The @desktop_file_id must have ".desktop" as a suffix. Except in the
+        special case when the calling process has no associated app ID,
+        @desktop_file_id must have the app ID followed by a period as a prefix,
+        regardless of whether the calling process is sandboxed or unsandboxed.
+
+        For example, Epiphany, which has the app ID "org.gnome.Epiphany"
+        in stable builds, might use a @desktop_file_id like
+        "org.gnome.Epiphany.WebApp_e9d0e1e4b0a10856aa3b38d9eb4375de4070d043.desktop"
+        In that example the desktop file would exist at the path
+        "~/.local/share/xdg-desktop-portal/applications/org.gnome.Epiphany.WebApp_e9d0e1e4b0a10856aa3b38d9eb4375de4070d043.desktop"
+        with a sym link in "~/.local/share/applications/".
+        The checksum at the end of the file name is an implementation detail in
+        Epiphany and not required by the portal.
+
+        This method is intended to be called by the application that created the
+        launcher, e.g. a web browser, so it can clean up associated data as part
+        of the uninstallation. Consequently, the proper way for a software center
+        to remove a launcher is by using the APIs provided by the application
+        that installed it. For example, for GNOME Software to remove web
+        launchers created by Epiphany, it would use the
+        org.gnome.Epiphany.WebAppProvider D-Bus interface.
+
+        Please note that this method call will fail if the specified launcher
+        already does not exist.
+
+        The @options vardict currently has no supported entries.
+    -->
+    <method name="Uninstall">
+      <arg type="s" name="desktop_file_id" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+    <!--
+        GetDesktopEntry:
+        @desktop_file_id: The .desktop file name
+        @contents: the contents of the named .desktop file
+
+        This function returns the contents of a desktop file with the name
+        @desktop_file_id in @contents.
+
+        The @desktop_file_id must have ".desktop" as a suffix. Except in the
+        special case when the calling process has no associated app ID,
+        @desktop_file_id must have the app ID followed by a period as a prefix.
+
+        This method only works for desktop files that were created by the
+        dynamic launcher portal.
+    -->
+    <method name="GetDesktopEntry">
+      <arg type="s" name="desktop_file_id" direction="in"/>
+      <arg type="s" name="contents" direction="out"/>
+    </method>
+    <!--
+        GetIcon:
+        @desktop_file_id: The .desktop file name
+        @icon_v: the icon as a serialized #GBytesIcon
+        @icon_format: one of "png", "jpeg", "svg"
+        @icon_size: the width and height in pixels of the icon
+
+        This function returns the contents of the icon specified in the "Icon"
+        key of the desktop file with the name @desktop_file_id in @icon_v. The
+        icon #GVariant can be passed to g_icon_deserialize() to reconstruct the
+        #GIcon.
+
+        The @desktop_file_id must have ".desktop" as a suffix. Except in the
+        special case when the calling process has no associated app ID,
+        @desktop_file_id must have the app ID followed by a period as a prefix.
+
+        The format and size of the icon are returned in @icon_format and
+        @icon_size. For svg icons, @icon_size is currently always set to 4096,
+        but don't depend on that as it may change in the future.
+
+        This method only works for desktop files that were created by the
+        dynamic launcher portal.
+    -->
+    <method name="GetIcon">
+      <arg type="s" name="desktop_file_id" direction="in"/>
+      <arg type="v" name="icon_v" direction="out"/>
+      <arg type="s" name="icon_format" direction="out"/>
+      <arg type="u" name="icon_size" direction="out"/>
+    </method>
+    <!--
+        Launch:
+        @desktop_file_id: The .desktop file name
+        @options: Vardict with optional further onformation
+
+        This function launches the app specified by @desktop_file_id.
+
+        The @desktop_file_id must have ".desktop" as a suffix. Except in the
+        special case when the calling process has no associated app ID,
+        @desktop_file_id must have the app ID followed by a period as a prefix.
+
+        This method only works for desktop files that were created by the
+        dynamic launcher portal.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem><para>
+              A token that can be used to activate the chosen application.
+            </para><para>
+              The activation_token option was introduced in version 1 of the interface.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="Launch">
+      <arg type="s" name="desktop_file_id" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+    <!--
+        SupportedLauncherTypes:
+
+        A bitmask of available launcher types. Currently defined types are:
+
+        <simplelist>
+          <member>1: Application. A launcher that represents an application.</member>
+          <member>2: Webapp. A launcher that represents a web app.</member>
+        </simplelist>
+    -->
+    <property name="SupportedLauncherTypes" type="u" access="read"/>
+    <property name="version" type="u" access="read"/>
+  </interface>
+</node>
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -20,6 +20,8 @@
 
 #include "config.h"
 
+#ifdef HAVE_GLIB_2_66
+
 #include <locale.h>
 #include <stdio.h>
 #include <string.h>
@@ -1248,3 +1250,5 @@ dynamic_launcher_create (GDBusConnection *connection,
 
   return G_DBUS_INTERFACE_SKELETON (dynamic_launcher);
 }
+
+#endif /* HAVE_GLIB_2_66 */
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -253,9 +253,22 @@ save_icon_and_get_desktop_entry (const char  *desktop_file_id,
   if (exec == NULL)
     return NULL;
 
-  exec_strv = g_strsplit (exec, " ", -1);
+  if (!g_shell_parse_argv (exec, NULL, &exec_strv, error))
+    return NULL;
+
+  /* Don't let the app give itself access to host files */
+  if (xdp_app_info_get_kind (xdp_app_info) == XDP_APP_INFO_KIND_FLATPAK &&
+      g_strv_contains ((const char * const *)exec_strv, "--file-forwarding"))
+    {
+      g_set_error (error,
+                   XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   _("Desktop entry given to Install() must not use --file-forwarding"));
+      return NULL;
+    }
+
   prefixed_exec_strv = xdp_app_info_rewrite_commandline (xdp_app_info,
-                                                         (const char * const *)exec_strv);
+                                                         (const char * const *)exec_strv,
+                                                         TRUE /* quote escape */);
   if (prefixed_exec_strv == NULL)
     {
       g_set_error (error,
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -1,0 +1,1250 @@
+/*
+ * Copyright © 2022 Matthew Leeds
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Matthew Leeds <mwleeds@protonmail.com>
+ */
+
+#include "config.h"
+
+#include <locale.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+#include <gio/gunixinputstream.h>
+#include <gio/gdesktopappinfo.h>
+#include <glib/gi18n.h>
+
+#include "dynamic-launcher.h"
+#include "request.h"
+#include "call.h"
+#include "xdp-dbus.h"
+#include "xdp-impl-dbus.h"
+#include "xdp-utils.h"
+
+#define MAX_DESKTOP_SIZE_BYTES 1048576
+
+typedef struct _DynamicLauncher DynamicLauncher;
+typedef struct _DynamicLauncherClass DynamicLauncherClass;
+
+struct _DynamicLauncher
+{
+  XdpDynamicLauncherSkeleton parent_instance;
+};
+
+struct _DynamicLauncherClass
+{
+  XdpDynamicLauncherSkeletonClass parent_class;
+};
+
+static XdpImplDynamicLauncher *impl;
+static DynamicLauncher *dynamic_launcher;
+
+static GMutex transient_permissions_lock;
+static GHashTable *transient_permissions;
+
+GType dynamic_launcher_get_type (void) G_GNUC_CONST;
+static void dynamic_launcher_iface_init (XdpDynamicLauncherIface *iface);
+
+G_DEFINE_TYPE_WITH_CODE (DynamicLauncher, dynamic_launcher, XDP_TYPE_DYNAMIC_LAUNCHER_SKELETON,
+                         G_IMPLEMENT_INTERFACE (XDP_TYPE_DYNAMIC_LAUNCHER, dynamic_launcher_iface_init));
+
+typedef enum {
+  DYNAMIC_LAUNCHER_TYPE_APPLICATION = 1,
+  DYNAMIC_LAUNCHER_TYPE_WEBAPP = 2,
+} DynamicLauncherType;
+
+static GVariant *
+get_launcher_data_and_revoke_token (const char *token)
+{
+  g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&transient_permissions_lock);
+  GVariant *launcher_data_wrapped;
+
+  if (!transient_permissions)
+    return NULL;
+
+  if (!g_uuid_string_is_valid (token))
+    return NULL;
+
+  launcher_data_wrapped = g_hash_table_lookup (transient_permissions, token);
+  if (launcher_data_wrapped)
+    {
+      g_autoptr(GVariant) launcher_data = NULL;
+      guint timeout_id;
+
+      g_variant_get (launcher_data_wrapped, "(vu)", &launcher_data, &timeout_id);
+
+      g_source_remove (timeout_id);
+      g_hash_table_remove (transient_permissions, token);
+
+      return g_steal_pointer (&launcher_data);
+    }
+
+  return NULL;
+}
+
+static gboolean
+validate_desktop_file_id (const char  *app_id,
+                          const char  *desktop_file_id,
+                          GError     **error)
+{
+  const char *after_app_id;
+
+  if (!g_str_has_suffix (desktop_file_id, ".desktop"))
+    {
+      g_set_error (error,
+                   XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   _("Desktop file id missing .desktop suffix: %s"), desktop_file_id);
+      return FALSE;
+    }
+
+  if (app_id == NULL || *app_id == '\0')
+    return TRUE;
+
+  after_app_id = desktop_file_id + strlen (app_id);
+  if (!g_str_has_prefix (desktop_file_id, app_id) || *after_app_id != '.')
+    {
+      g_set_error (error,
+                   XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   _("Desktop file id missing app id prefix '%s.': %s"),
+                   app_id, desktop_file_id);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+write_icon_to_disk (GVariant    *icon_v,
+                    const char  *icon_subdir,
+                    const char  *icon_path,
+                    GError     **error)
+{
+  g_autoptr(GIcon) icon = NULL;
+  g_autoptr(GFile) icon_file = NULL;
+  g_autoptr(GFileOutputStream) icon_stream = NULL;
+  GBytes *icon_bytes;
+  gconstpointer bytes_data;
+  gsize bytes_len;
+
+  icon = g_icon_deserialize (icon_v);
+  g_assert (G_IS_BYTES_ICON (icon));
+  icon_bytes = g_bytes_icon_get_bytes (G_BYTES_ICON (icon));
+
+  g_mkdir_with_parents (icon_subdir, 0700);
+  icon_file = g_file_new_for_path (icon_path);
+  icon_stream = g_file_replace (icon_file, NULL, FALSE, G_FILE_CREATE_REPLACE_DESTINATION, NULL, error);
+  if (icon_stream == NULL)
+    return FALSE;
+
+  /* Use write_all() instead of write_bytes() so we don't have to worry about
+   * partial writes (https://gitlab.gnome.org/GNOME/glib/-/issues/570).
+   */
+  bytes_data = g_bytes_get_data (icon_bytes, &bytes_len);
+  if (!g_output_stream_write_all (G_OUTPUT_STREAM (icon_stream),
+                                  bytes_data, bytes_len,
+                                  NULL, NULL, error) ||
+      !g_output_stream_close (G_OUTPUT_STREAM (icon_stream), NULL, error))
+    return FALSE;
+
+  return TRUE;
+}
+
+static GKeyFile *
+save_icon_and_get_desktop_entry (const char  *desktop_file_id,
+                                 const char  *desktop_entry,
+                                 GVariant    *launcher_data,
+                                 XdpAppInfo  *xdp_app_info,
+                                 char       **out_icon_path,
+                                 GError     **error)
+{
+  g_autoptr(GVariant) icon_v = NULL;
+  g_autoptr(GDesktopAppInfo) desktop_app_info = NULL;
+  g_autoptr(GKeyFile) key_file = g_key_file_new ();
+  g_autofree char *exec = NULL;
+  g_auto(GStrv) exec_strv = NULL;
+  g_auto(GStrv) prefixed_exec_strv = NULL;
+  g_auto(GStrv) groups = NULL;
+  g_autofree char *prefixed_exec = NULL;
+  g_autofree char *tryexec_path = NULL;
+  g_autofree char *icon_path = NULL;
+  g_autofree char *icon_subdir = NULL;
+  const char *name, *icon_extension, *icon_size;
+  const char *app_id;
+
+  g_variant_get (launcher_data, "(&sv&s&s)", &name, &icon_v, &icon_extension, &icon_size);
+  g_assert (name != NULL && name[0] != '\0');
+  g_assert (icon_v);
+  g_assert (icon_extension != NULL && icon_extension[0] != '\0');
+  g_assert (icon_size != NULL && icon_size[0] != '\0');
+
+  app_id = xdp_app_info_get_id (xdp_app_info);
+
+  if (!g_key_file_load_from_data (key_file, desktop_entry, -1,
+                                  G_KEY_FILE_KEEP_COMMENTS | G_KEY_FILE_KEEP_TRANSLATIONS,
+                                  error))
+    {
+      g_set_error (error,
+                   XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   _("Desktop entry given to Install() not a valid key file"));
+      return NULL;
+    }
+
+  /* The desktop entry spec supports more than one group but we don't in case
+   * there's a security risk.
+   */
+  groups = g_key_file_get_groups (key_file, NULL);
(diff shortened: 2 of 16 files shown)


mwleeds:
Do we really still care about Ubuntu 16.04 support? It is EOL except for their [ESM](https://ubuntu.com/security/esm) program.

mwleeds:
> Do we really still care about Ubuntu 16.04 support? It is EOL except for their [ESM](https://ubuntu.com/security/esm) program.

I made the new portal only compile when we have GLib 2.66, so we can still maintain compatibility for the existing code with old GLib for the sake of security patches.

mwleeds:
I realized that the implementation currently doesn't allow the `editable-icon` option to be implemented by the backend, even though the option is there. I'm going to work on getting this tested without editable icons for now; that may need to be a v2 feature.

mwleeds:
I also notice that the notification portal takes an icon as a GVariant (made with `g_icon_serialize()`) whereas here we take it as a file descriptor. I made it an fd because I thought it's not best practice to send an image in a D-Bus message due to message size limits and performance concerns. I still think this is a good design, but just noting that it's not in line with some existing code we have.

ramcq:
> I also notice that the notification portal takes an icon as a GVariant (made with `g_icon_serialize()`) whereas here we take it as a file descriptor. I made it an fd because I thought it's not best practice to send an image in a D-Bus message due to message size limits and performance concerns. I still think this is a good design, but just noting that it's not in line with some existing code we have.

I'm not sure it's worth it...? My recollection is that a D-Bus message can be 10MB in size, and the icons are expected to be significantly smaller. Requiring a fd means that if you are composing / sizing / etc the icon dynamically you are forced to place it on disk or set up a memory-backed fd (/dev/shm I guess?) which is a bit of a pain. @matthiasclasen may have an opinion on this but for a relatively rare process of adding an app, requiring an fd might be overkill. If we were streaming video, sure... :)

mwleeds:
> > I also notice that the notification portal takes an icon as a GVariant (made with `g_icon_serialize()`) whereas here we take it as a file descriptor. I made it an fd because I thought it's not best practice to send an image in a D-Bus message due to message size limits and performance concerns. I still think this is a good design, but just noting that it's not in line with some existing code we have.
> 
> I'm not sure it's worth it...? My recollection is that a D-Bus message can be 10MB in size, and the icons are expected to be significantly smaller. Requiring a fd means that if you are composing / sizing / etc the icon dynamically you are forced to place it on disk or set up a memory-backed fd (/dev/shm I guess?) which is a bit of a pain. @matthiasclasen may have an opinion on this but for a relatively rare process of adding an app, requiring an fd might be overkill. If we were streaming video, sure... :)

Yeah the way it's currently implemented:

- Software doesn't have to do any extra work since it already caches remote icons to disk so we just pass that path to libportal which gets an fd for it (but maybe Software could start modifying the icons like Epiphany does).
- Epiphany downloads icons to disk (except on the fallback path that uses favicons), then reads it, makes it have rounded edges and be a certain size, and then writes it back out to a tmp file to be passed to libportal.

So if message size limits aren't an issue it might make sense to pass a GVariant.

mwleeds:
Separately, there is the question of how to handle installation of a launcher that would overwrite an existing one. Currently the portal just does the overwrite without confirmation, but Epiphany checks if the app already exists before calling the install method and presents a confirmation dialog if so. It could make sense to have some sort of overwrite confirmation in the portal implementation, but having it controlled by the application has the advantage that it can clear any other state associated with the launcher being overwritten. However it does mean that the confirmation dialog is separate rather than a child of the main portal dialog.

mwleeds:
> So if message size limits aren't an issue it might make sense to pass a GVariant.

I checked 18 PWAs and found one icon that was 275 kB but all the rest were an order of magnitude smaller.

matthiasclasen:
Message size would be an issue for photos. I don't see it being a problem for icons

mwleeds:
> Message size would be an issue for photos. I don't see it being a problem for icons

Okay, I can change the implementation to use a GVariant instead for the icon.

In other news, I just was able to successfully test this with unsandboxed Epiphany. Still need to test sandboxed Epiphany but it seems to be working so far.

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,223 @@
...
+
+        The @options vardict currently has no supported entries.
+    -->
+    <method name="RequestInstallToken">
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="s" name="name" direction="in"/>
+      <arg type="h" name="icon_fd" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="s" name="token" direction="out"/>
+    </method>
```
GeorgesStavracas, line 184: I find this separation of `PrepareInstall()` and `RequestInstallToken()` not ideal. The portal daemon can detect if an app is sandboxed or not, and therefore can decide when not to ask implementations to show a dialog. If the application is unsandboxed, it would just accept the all the data passed through `PrepareInstall()` unmodified.

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,223 @@
...
+        org.freedesktop.portal.DynamicLauncher.PrepareInstall() or
+        org.freedesktop.portal.DynamicLauncher.RequestInstallToken() call.
+        The token is valid for up to five minutes.
+
+        The icon and name used for the launcher will be the ones from the previous
+        org.freedesktop.portal.DynamicLauncher.PrepareInstall() or
+        org.freedesktop.portal.DynamicLauncher.RequestInstallToken() call.
+
+        The @desktop_entry should be a valid desktop entry file beginning with
+        "[Desktop Entry]", except it should not include Name= or Icon= entries
```
GeorgesStavracas, line 70: Should not, or must not?

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,223 @@
...
+        Epiphany and not required by the portal.
+
+        If the calling process is unsandboxed, @desktop_file_id need only have
+        a ".desktop" suffix.
+
+        This method is intended to be called by the application that created the
+        launcher, e.g. Epiphany, so it can clean up associated data such as the
+        profile directory as part of the uninstallation. The proper way for a
+        software center to remove an Epiphany-installed web app is using the
+        org.gnome.Epiphany.WebAppProvider D-Bus interface.
```
GeorgesStavracas, line 212: This reads a bit too GNOMEy to me. Portals are DE-agnostic, so any mention to DEs must be clearly only as examples. Allow me to suggest the following:

```
        This method is intended to be called by the application that created the
        launcher, e.g. a web browser, so it can clean up associated data such as
        the profile directory as part of the uninstallation. Consequently, the
        proper way for a software center to remove a launcher is by using the APIs
        provided by the application that installed it. For example, for GNOME
        Software to remove web launchers created by Epiphany, it would use the
        org.gnome.Epiphany.WebAppProvider D-Bus interface.
```

src/dynamic-launcher.c
```diff
@@ -0,0 +1,794 @@
...
+                   "Desktop file id missing .desktop suffix: %s", desktop_file_id);
+      return FALSE;
+    }
+
+  if (app_id == NULL || *app_id == '\0')
+    return TRUE;
+
+  after_app_id = desktop_file_id + strlen (app_id);
+  if (!g_str_has_prefix (desktop_file_id, app_id) ||
+      *after_app_id != '.')
```
GeorgesStavracas, line 114: Nitpick: these could be in the same line

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,223 @@
...
+        @options: Vardict with optional further information
+        @result: 0 to indicate success, a nonzero number to indicate failure
+
+        Installs a .desktop launcher and icon into appropriate directories to
+        allow the desktop environment to find them.
+
+        @token must be a token that was returned by a previous
+        org.freedesktop.portal.DynamicLauncher.PrepareInstall() or
+        org.freedesktop.portal.DynamicLauncher.RequestInstallToken() call.
+        The token is valid for up to five minutes.
```
GeorgesStavracas, line 63: How many times can this token be used during these 5 minutes?

src/dynamic-launcher.c
```diff
@@ -0,0 +1,794 @@
...
+  g_autoptr(GKeyFile) desktop_keyfile = NULL;
+  g_autofree char *icon_path = NULL;
+  g_autofree char *desktop_dir = NULL;
+  g_autofree char *desktop_path = NULL;
+  g_autofree char *link_path = NULL;
+  g_autofree char *relative_path = NULL;
+  g_autoptr(GFile) link_file = NULL;
+  gint result = -1;
+
+  launcher_data = get_launcher_data_and_revoke_token (arg_token);
```
GeorgesStavracas, line 334: I think it would be useful to make `get_launcher_data_and_revoke_token()` validate `arg_token` with `g_uuid_string_is_valid()`

src/dynamic-launcher.c
```diff
@@ -0,0 +1,794 @@
...
+    }
+
+  g_hash_table_insert (transient_permissions,
+                       g_strdup (token),
+                       g_variant_ref (launcher_data));
+
+  /* Revoke the token if it hasn't been used after 5 minutes, in case of
+   * client bugs. This is what the gnome print portal implementation does.
+   */
+  g_timeout_add_seconds (300, install_token_timeout, g_strdup (token));
```
GeorgesStavracas, line 439: You need to store the source id somewhere, and remove the timeout when the token is used

GeorgesStavracas (changes requested):
One common pattern in portals that can provide a variety of options is to expose the availability of these options using D-Bus properties. In this case, I think it needs a `SupportedLauncherTypes` property (uint32):

```xml
SupportedLauncherTypes:

A bitmask of available launcher types. Currently defined types are:

<simplelist>
  <member>1: Application. A launcher that represents an application.</member>
  <member>2: Link. A launcher that represents a webapp.</member>
</simplelist>
```

`PrepareInstall()` would then receive a `launcher_type` (uint32) as part of the `options` vardict. Knowing ahead of type what kind of launcher it is will allow portal implementations to provider more specialized interfaces. Despite being technically a bitmask, `PrepareInstall()` must only accept one type. ¹

By providing a `launcher_type`, the naming of the `url` option doesn't feel natural to me - perhaps rename to `target`? Ideas welcomed.

---
 
 ¹ -  This can be validated with `__builtin_popcount(launcher_types) == 1`.

data/org.freedesktop.portal.DynamicLauncher.xml
mwleeds: I don't think that pattern works because I want to use the portal UI in both sandboxed and unsandboxed Epiphany, so we don't have to maintain two code paths there. So the split between `PrepareInstall` and `RequestInstallToken` lets the app choose whether it wants UI to be presented, and the UI also provides editing the name (and potentially the icon eventually) so it is useful.

mwleeds:
> 



> One common pattern in portals that can provide a variety of options is to expose the availability of these options using D-Bus properties. In this case, I think it needs a `SupportedLauncherTypes` property (uint32):
> 
> ```
> SupportedLauncherTypes:
> 
> A bitmask of available launcher types. Currently defined types are:
> 
> <simplelist>
>   <member>1: Application. A launcher that represents an application.</member>
>   <member>2: Link. A launcher that represents a webapp.</member>
> </simplelist>
> ```
> 
> `PrepareInstall()` would then receive a `launcher_type` (uint32) as part of the `options` vardict. Knowing ahead of type what kind of launcher it is will allow portal implementations to provider more specialized interfaces. Despite being technically a bitmask, `PrepareInstall()` must only accept one type. ¹
> 
> By providing a `launcher_type`, the naming of the `url` option doesn't feel natural to me - perhaps rename to `target`? Ideas welcomed.
> 
> ¹ - This can be validated with `__builtin_popcount(launcher_types) == 1`.

That makes sense to make the launcher types more explicit. But I guess application launchers wouldn't have a target so it would be NULL in that case. Still seems fine to name it `target` in case there's another use for it in the future. The implementation currently only uses it to display the URL in small text at the bottom of the [dialog](https://gitlab.gnome.org/Teams/Design/os-mockups/uploads/a6378d68610b96d8d4b881a8000db240/epiphany-install-app-dialog.png).

data/org.freedesktop.portal.DynamicLauncher.xml
ramcq: My instinct is not to bother here, just say that the portal implementation will replace these unconditionally. If someone is adding flatpak support to an application that otherwise writes out templated .desktop files, you're forcing more work on them - either different templates, or adding a wrapper that parses/modifies them, etc. Better to be forgiving in what is accepted here and just enforce what we need.

mwleeds:
Currently we only add a `TryExec=` entry to the desktop file for Flatpaks, since we know how to find the wrapper script for them in e.g. `~/.local/share/flatpak/exports/bin/`. We then use that in `migrate_renamed_app_launchers()` which gets run at startup to rename launchers for apps that have been renamed and delete launchers for apps that have been uninstalled. Presumably we want the latter part, deleting no-longer-needed launchers, to work for non-Flatpaks as well, so we could:

- Try to determine the absolute path to use as `TryExec=` by looking at `/proc/$(pid of caller)/exe`, but this isn't perfect. It doesn't work for something executed in a toolbox container for example, or
- Use the first string in the `Exec=` line, since `TryExec=` doesn't have to be an absolute path according to the spec, or
- Require the caller add a `TryExec=` line.

For Epiphany I couldn't find an easy way to get the absolute path to the binary so I had it add `TryExec=epiphany` ([link to MR](https://gitlab.gnome.org/GNOME/epiphany/-/merge_requests/1067)).

data/org.freedesktop.portal.DynamicLauncher.xml
mwleeds: once

mwleeds:
> > Message size would be an issue for photos. I don't see it being a problem for icons
> 
> Okay, I can change the implementation to use a GVariant instead for the icon.
> 

Looks like there's no easy way to tell the file format of a serialized `#GBytesIcon`, so we could either (a) gain a dependency on gdk and use gdk-pixbuf to write it out as a png regardless of what it originally was, or (b) add a parameter so the caller can specify the file extension, with the allowed values the same as Flatpak (".png", ".jpeg", ".svg"). The second option seems simpler so I'll go with that.

mwleeds:
This is now working using serialized icons rather than file descriptors. I tested it with the -gtk backend and epiphany.

gasinvein:
Sorry to interject, but this supports only one icon per desktop id, right? May support for multiple icons for the same desktop id (i.e. for different icon sizes) be added in the future, or it's out of scope of this portal?

mwleeds:
> Sorry to interject, but this supports only one icon per desktop id, right? May support for multiple icons for the same desktop id (i.e. for different icon sizes) be added in the future, or it's out of scope of this portal?

The Desktop Entry [spec](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html) only supports one `Icon=` key in desktop files. It can be either an absolute path or the name of an icon provided by a theme, but this portal only supports absolute paths (if it becomes clear at some point that there's a valid use case for themed icons that support could be added).

I think the reasoning is that an icon of a size such as 192x192 (512x512 is the max) can be used for all the places it's needed in the desktop environment. Do you have a use case for needing icons of different sizes?

gasinvein:
> I think the reasoning is that an icon of a size such as 192x192 (512x512 is the max) can be used for all the places it's needed in the desktop environment. Do you have a use case for needing icons of different sizes?

Steam games ship desktop icons in different sizes. Sure just the largest one (or closest to 192x192) can be used, but I thought it would be better to install all sizes available to avoid resizing.

mwleeds:
Looks like we actually need one more method for this portal to allow Epiphany to reach feature parity with non-Flatpak Epiphany, because Epiphany lets you launch just-installed web apps by clicking on a shell notification. I suppose we need something like `Launch()` which would take a desktop file name, presumably check that it's prefixed with the caller's app ID as we do in the other methods, and call `g_app_info_launch()`, perhaps with some sanitization of the environment variables.

Even though we have the `GetDesktopEntry()` method to get the .desktop file contents, and `g_desktop_app_info_new_from_keyfile()`, that GLib API doesn't produce a valid `GAppInfo` when called from within Flatpak, so passing it to `g_app_info_launch()` doesn't work. It makes sense for GLib to assume that the launching is happening on the host, so I think more portal API is the solution.

However installing web apps from Flatpak Epiphany using this portal does work.

mwleeds:
> 



> Looks like we actually need one more method for this portal to allow Epiphany to reach feature parity with non-Flatpak Epiphany, because Epiphany lets you launch just-installed web apps by clicking on a shell notification. I suppose we need something like `Launch()` which would take a desktop file name, presumably check that it's prefixed with the caller's app ID as we do in the other methods, and call `g_app_info_launch()`, perhaps with some sanitization of the environment variables.
> 
> Even though we have the `GetDesktopEntry()` method to get the .desktop file contents, and `g_desktop_app_info_new_from_keyfile()`, that GLib API doesn't produce a valid `GAppInfo` when called from within Flatpak, so passing it to `g_app_info_launch()` doesn't work. It makes sense for GLib to assume that the launching is happening on the host, so I think more portal API is the solution.
> 

Alternatively I guess it would be possible for the sandboxed app to spawn the command used in the `Exec=` of the desktop file, just without the `flatpak run ...` part of it, but it seems better to use the standard GLib API via the portal.

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,287 @@
...
+
+        If the calling process is unsandboxed, @desktop_file_id need only have
+        a ".desktop" suffix.
+
+        This method only works for desktop files that were created by the
+        dynamic launcher portal.
+    -->
+    <method name="Launch">
+      <arg type="s" name="desktop_file_id" direction="in"/>
+    </method>
```
mwleeds, line 273: This API is missing two things that might be considered important:

- a timestamp, which per the docs of gdk_app_launch_context_set_timestamp():

> The timestamp should ideally be taken from the event that triggered the launch.
> 
> Window managers can use this information to avoid moving the focus to the newly launched application when the user is busy typing in another window. This is also known as ‘focus stealing prevention’.

- and a way to set what screen the launch happens on, which is not needed by Epiphany but which could be needed in some contexts. However, on Wayland it is [not possible](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4434/diffs) to control the workspace or screen a window is launched on (according to ebassi). And on X11 it is only possible to control the workspace (via `gdk_app_launch_context_set_desktop()` not the physical monitor.

So perhaps a way to control the workspace is not worth the trouble, and just a timestamp is enough. But it does come at the cost of requiring each backend to implement another method.

mwleeds: On closer inspection it looks like there's already a portal that uses `g_app_info_launch`, the open-uri portal. And there we also have support for xdg activation tokens (https://github.com/flatpak/xdg-desktop-portal/pull/566) which appear to be a mechanism for focus stealing prevention that is intended to replace timestamps ([docs](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/50/diffs#diff-content-7b1a427b51eea37983e26cee3b13856c7731eb35)). So we should follow suit in this portal.

mwleeds:
cc @jadahl @aleixpol in case you want to check the use of xdg activation tokens here to see if it looks right?

mwleeds:
> Currently we only add a `TryExec=` entry to the desktop file for Flatpaks, since we know how to find the wrapper script for them in e.g. `~/.local/share/flatpak/exports/bin/`. We then use that in `migrate_renamed_app_launchers()` which gets run at startup to rename launchers for apps that have been renamed and delete launchers for apps that have been uninstalled. Presumably we want the latter part, deleting no-longer-needed launchers, to work for non-Flatpaks as well, so we could:
> 
> * Try to determine the absolute path to use as `TryExec=` by looking at `/proc/$(pid of caller)/exe`, but this isn't perfect. It doesn't work for something executed in a toolbox container for example, or
> * Use the first string in the `Exec=` line, since `TryExec=` doesn't have to be an absolute path according to the spec, or
> * Require the caller add a `TryExec=` line.
> 
> For Epiphany I couldn't find an easy way to get the absolute path to the binary so I had it add `TryExec=epiphany` ([link to MR](https://gitlab.gnome.org/GNOME/epiphany/-/merge_requests/1067)).

I went with not requiring `TryExec` but if present we use it to garbage collect the launcher when its parent app is uninstalled. Requiring it would mean more work for apps as discussed [above](https://github.com/flatpak/xdg-desktop-portal/pull/696#discussion_r794659074), and I don't know how to handle it for snaps.

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -0,0 +1,301 @@
...
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem><para>
+              A token that can be used to activate the chosen application.
+            </para><para>
+              The activation_token option was introduced in version 4 of the interface.
+            </para></listitem>
```
GeorgesStavracas, line 284: This new portal certainly won't hit the ground running like this :slightly_smiling_face:

data/org.freedesktop.portal.DynamicLauncher.xml
GeorgesStavracas: Sorry, I'm still not sold on this approach. The reason I don't think it's a good idea is because portal design generally avoids delegating this kind of decision to apps. x-d-p knows if an app is sandboxed or not, and can make decisions based on that; apps, on the other hand, could call the wrong D-Bus method for a variety of reasons (broken sandbox detection, evil intents, etc.)

I would like to suggest an alternative approach: pass a hint to whether the requesting app wants a dialog or not, as part of the `options` vardict. I imagine it would be something like this:

```
interaction_mode → uint32

The suggested interaction mode. Supported values are:

  * 0: Default. Portal implementations decide how to present this request. This is the default.
  * 1: Interactive. Present a dialog even when not necessary.
```

In this case ^^^, I don't think we need a `SupportedInteractionModes` property, because portal implementations *must* implement all modes.

(Naming TBD; if you can think of better terminology, please let me know.)

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+static gboolean
+validate_url (const char  *key,
+              GVariant    *value,
+              GVariant    *options,
+              GError     **error)
+{
+  const char *url = g_variant_get_string (value, NULL);
+  g_autoptr(GError) local_error = NULL;
+
+  if (!g_uri_is_valid (url, G_URI_FLAGS_NONE, &local_error))
```
GeorgesStavracas, line 461: Is this correct for the non-webapp case?

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+  GVariant *launcher_data_wrapped;
+
+  if (!transient_permissions)
+    {
+      transient_permissions = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                                     g_free, (GDestroyNotify)g_variant_unref);
+    }
+
+  /* Revoke the token if it hasn't been used after 5 minutes, in case of
+   * client bugs. This is what the gnome print portal implementation does.
```
GeorgesStavracas, line 370: Nitpick: `GNOME`

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+   */
+  timeout_id = g_timeout_add_seconds_full (G_PRIORITY_DEFAULT, 300, install_token_timeout,
+                                           g_strdup (token), g_free);
+  launcher_data_wrapped = g_variant_new ("(vu)", launcher_data, timeout_id);
```
GeorgesStavracas, line 374: Interesting approach, but I'd rather use a custom structure here, what do you think? Was there any specific reason to avoid adding a new structure to the code?

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+
+  if (!g_file_delete (desktop_file, NULL, &error))
+    goto error;
+
+  if (icon_path && g_str_has_prefix (icon_path, icon_dir))
+    {
+      icon_file = g_file_new_for_path (icon_path);
+      if (!g_file_delete (icon_file, NULL, &error))
+        goto error;
+    }
```
GeorgesStavracas, line 691: I wonder if this particular function should be a bit more lenient with errors. For example, here, if deleting the desktop file errs, the icon won't be deleted. I think it should try and delete all files, even when one file errs on deletion.

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+  if (!validate_desktop_file_id (app_id, arg_desktop_file_id, &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
+  icon_dir = g_build_filename (g_get_user_data_dir (), "xdg-desktop-portal", "icons", NULL);
+  desktop_dir = g_build_filename (g_get_user_data_dir (), "xdg-desktop-portal", "applications", NULL);
+  g_mkdir_with_parents (desktop_dir, 0700);
+  g_mkdir_with_parents (icon_dir, 0700);
```
GeorgesStavracas, line 668: Feels a bit odd to create directories in the Uninstall() method.

src/dynamic-launcher.c
```diff
@@ -0,0 +1,868 @@
...
+  g_autofree char *desktop_path = NULL;
+  const char *activation_token = NULL;
+  g_autoptr(GAppLaunchContext) launch_context = NULL;
+  g_autoptr(GDesktopAppInfo) app_info = NULL;
+
+  if (!validate_desktop_file_id (app_id, arg_desktop_file_id, &error))
+    goto error;
+
+  desktop_dir = g_build_filename (g_get_user_data_dir (), "xdg-desktop-portal", "applications", NULL);
+  g_mkdir_with_parents (desktop_dir, 0700);
```
GeorgesStavracas, line 758: Here too. I don't think Launch() should create folders?

GeorgesStavracas (changes requested):
A few more comment, I was ready to do some nitpicking but there's not much to nitpick here, the code itself looks good

src/dynamic-launcher.c
mwleeds: You're right, we need to only do this validation in the webapp case.

src/dynamic-launcher.c
mwleeds: I think it makes sense because `launch_data` is already a `GVariant` so we're working with them anyway here. I think a `struct` would be a bit "heavyweight" here but honestly don't see much of a good argument in either direction.

data/org.freedesktop.portal.DynamicLauncher.xml
mwleeds: > Sorry, I'm still not sold on this approach. The reason I don't think it's a good idea is because portal design generally avoids delegating this kind of decision to apps. x-d-p knows if an app is sandboxed or not, and can make decisions based on that; apps, on the other hand, could call the wrong D-Bus method for a variety of reasons (broken sandbox detection, evil intents, etc.)

Sorry I don't see how this is a reason we should move the functionality of `RequestInstallToken()` into `PrepareInstall()`. It is great for x-d-p to make decisions based on whether an app is sandboxed or not, but in this case we do need a way for the app to either install with a dialog or install without, and I think these flows are different enough that it makes sense for them to have different methods. A malicious app that's sandboxed is going to get a dialog regardless of which choice we make here, so I don't understand the point of bringing that up.

> I would like to suggest an alternative approach: pass a hint to whether the requesting app wants a dialog or not, as part of the options vardict. I imagine it would be something like this:
> 
> interaction_mode → uint32
> 
> The suggested interaction mode. Supported values are:
> 
>   * 0: Default. Portal implementations decide how to present this request. This is the default.
>   * 1: Interactive. Present a dialog even when not necessary.

I don't think this makes sense, because we need to support all three of (1) sandboxed Ephy install w/ dialog, (2) unsandboxed Ephy install w/ dialog, and (3) Software install w/o dialog. In your scheme for (3) Software would not be able to specify that there needs to be no interaction (since it has its own UI) and would have to rely on the portal deciding not to show UI, which seems more fragile. It could possibly work as a tri-state, where there's a "Force no interaction" option, but this just seems like over-engineering when we already have a working solution.

Here are the reasons I think the split between the two methods makes sense:

- Having `PrepareInstall` default to not showing a dialog if the app is unsandboxed and the proposed "Force interaction" option isn't set (I think that's what you're suggesting) doesn't make a lot of sense since the user wouldn't have a chance to change the name of the launcher before it gets created (if the `editable-name` option is set). For what it's worth Chrome shows a confirmation dialog when installing a PWA even without letting the user change the launcher name, so maybe that's more evidence that it's the right UX.
- The `RequestInstallToken` method implementation has a check at the top to error out if the caller is sandboxed, and it's easy to see that this is correct when reading the code, whereas a similar check in `PrepareInstall` would have to be conditional on the `Force no interaction` option being set and therefore a bit harder to check for correctness.
- There's a lot of complexity in `PrepareInstall` that is not relevant when the caller does not want a dialog: the `parent_window` argument, the use of the `org.freedesktop.portal.Request` pattern, and all of the keys in the `options` vardict. I don't see how it makes sense to force clients to deal with all that when they don't need it (and to force x-d-p handle those things being omitted correctly).
- For what it's worth there's already precedent for methods that are only for unsandboxed callers: e.g. `org.freedesktop.portal.Device.AccessDevice()`

I do appreciate your time reviewing this and I don't intend to be combative.

mwleeds: If you'd like to take this to a court of appeals we could ask a third party for an opinion :laughing:

mwleeds:
It was [pointed out](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1182#note_1374036) in the review of the epiphany Software plugin that Software wants to know the dimensions of the icon that's specified in the .desktop file for these dynamic launchers, because it can sort icons by size and use different ones in different contexts when there's more than one available for one app. And getting the width and height by reading the icon file when we're creating a `GsApp` for a web app is a bit expensive. Icons specified in appstream data usually have width and height attributes, and themed icons are stored in per-size subdirectories (except `scalable` and `symbolic`), so it might make sense to follow suit here and make the size information easily available by putting the icons in subdirectories, e.g. `~/.local/share/xdg-desktop-portal/icons/192x192/` and `~/.local/share/xdg-desktop-portal/icons/scalable/`.

To get the size information we could take the icon-validator from Flatpak and change it to output the size and format rather than just reading in those things (or split off icon-validator into its own repo so we can share the code). Once we have it as part of x-d-p we would be able to make its use non-optional in this portal and the notification portal, both of which currently don't validate icons if `LIBEXECDIR/flatpak-validate-icon` isn't present. This would mean adding a dependency on `bwrap` but that seems pretty reasonable. It would also mean we can check the format of files (png/svg/jpg) rather than trusting what the caller of the portal says.

Alternatively we could put the icons in `~/.local/share/icons` which is specified by the Icon Theme Specification and use just the name in the desktop file `Icon` key rather than an absolute path. But since these icons are not actually part of a theme I think absolute paths make more sense.

smcv:
> To get the size information we could take the icon-validator from Flatpak and change it to output the size and format rather than just reading in those things (or split off icon-validator into its own repo so we can share the code). Once we have it as part of x-d-p

... like the way I've been patching it back into x-d-p in Debian ever since it was removed, to avoid having a circular dependency between x-d-p and Flatpak? :-)

mwleeds:
> > To get the size information we could take the icon-validator from Flatpak and change it to output the size and format rather than just reading in those things (or split off icon-validator into its own repo so we can share the code). Once we have it as part of x-d-p
> 
> ... like the way I've been patching it back into x-d-p in Debian ever since it was removed, to avoid having a circular dependency between x-d-p and Flatpak? :-)

Oh interesting. Didn't know the history there. The rationale in ea7357758652c1697790f18e6a9556c963b3b8cd for dropping the copied code is "We don't want to duplicate the code". It is only a couple hundred lines so the duplication doesn't seem that concerning, but we could make a separate repo for it. The way of finding the right `bwrap` binary would be a bit complicated since it's only in `$(libexecdir)` not `$PATH`.` flatpak-validate-icon` tries the `FLATPAK_BWRAP` env var and then falls back to a value defined at build time, and this lets the unit tests use the `builddir` copy of `bwrap` and outside the unit tests it uses either a value specified at build time or a fallback `$(libexecdir)/flatpak-bwrap`. So if we keep the env var and the build time option, does splitting `flatpak-validate-icon` into a separate repo sound like the way to go?

src/dynamic-launcher.c
mwleeds: Yeah. My thinking was that it would be easier to reason about the error conditions if you only had to worry about the file not existing rather than its parent directories, but I suppose it's `G_IO_ERROR_NOT_FOUND` either way. I can remove these.

src/dynamic-launcher.c
mwleeds: Makes sense

smcv:
> The way of finding the right bwrap binary would be a bit complicated since it's only in $(libexecdir) not $PATH.

For Debian's purposes it doesn't matter, because Debian's packages for all these tools use a system copy of bubblewrap in `/usr/bin`, but I agree that's more problematic for upstream.

I wonder whether it's time to require (or at least strongly recommend) a system copy of bubblewrap?

smcv:
> we could make a separate repo for it

I know this might seem strange for a Debian maintainer to say, because distributions like Debian are usually the most opposed to duplication, but for an amount of code this small, packaging it separately seems higher-cost than just duplicating the code. The packaging would be as large as the code itself.

I suppose it could be a submodule-only project like libglnx if people really want it to have separate git history, but the code-review story for submodules isn't great (`git log` in flatpak cannot tell you much about the history of variant-schema-compiler or libglnx) and neither is the ability to do bugfix-only backports without also introducing new features (the only operation that is particularly straightforward for submodules is "update to latest").

mwleeds:
> I wonder whether it's time to require (or at least strongly recommend) a system copy of bubblewrap?

The only scenario I can think of when it's helpful to have the ability to use `bwrap` as a submodule is when the system version is too old for what you need. That seems like more of an issue for Flatpak than `x-d-p` since the portals are unlikely to need a specific version, so maybe we should just use system `bwrap` for `x-d-p`'s copy of the icon validator.

mwleeds:
> Alternatively we could put the icons in `~/.local/share/icons` which is specified by the Icon Theme Specification and use just the name in the desktop file `Icon` key rather than an absolute path. But since these icons are not actually part of a theme I think absolute paths make more sense.

Looking into it a bit more it seems there are more reasons not to use themed icons here. We want to be able to determine the file path of the icon in the portal code, both so we can clean up after ourselves in the `Uninstall()` method, and so we can provide the raw icon data in the `GetIcon()` method (not implemented yet but I will soon). Finding the file path for a `GThemedIcon` is [complicated](https://gitlab.gnome.org/GNOME/gnome-software/-/blob/a49908878e998171bf806175c6b7a092af5f7c1b/lib/gs-app.c#L4350) and requires `gtk`. Also conceptually since we only expect these apps to use the icon provided through this portal, they should not need the infrastructure icon themes provide of being able to install icons at different sizes, being able to override icons in inherited themes, etc.

So I think I should:

- Change the D-Bus interface to no longer require `icon_format` on the `PrepareInstall()` and `RequestInstallToken()` methods
- Change the portal code to determine the format and size using a custom copy of `flatpak-validate-icon`
- Put icons in per-size subdirectories of `~/.local/share/xdg-desktop-portal/icons/` except `svg` icons which go in `scalable/`
- Make `gnome-software` infer the size of icons from their file path, and treat scalable ones as preferable in its size-sorted list of icons, perhaps by setting `4096` as their size (`gs_icon_set_width()`).
- Have epiphany's migration code for existing web apps follow the pattern too, putting icons in `192x192/`

data/org.freedesktop.portal.DynamicLauncher.xml
GeorgesStavracas: I'd like to have third and fourth party opinions here, for sure :slightly_smiling_face:

mwleeds: Maybe @pwithnall since he's been looking at the Software side of this already?

mwleeds:
> So I think I should:
> 
> * Change the D-Bus interface to no longer require `icon_format` on the `PrepareInstall()` and `RequestInstallToken()` methods
> * Change the portal code to determine the format and size using a custom copy of `flatpak-validate-icon`
> * Put icons in per-size subdirectories of `~/.local/share/xdg-desktop-portal/icons/` except `svg` icons which go in `scalable/`

This is implemented now. I will test it after the dust settles on the various related PRs so I can test it all at once.

data/org.freedesktop.portal.DynamicLauncher.xml
mwleeds: For what it's worth the `PrepareInstall()` method here is modeled after the `PreparePrint()` method in the Print portal, which creates a dialog regardless of whether or not the caller is sandboxed.

matthiasclasen:
I may not be in the right mindset to comment here. I have more questions than answers, and dislike many of the things I see in the suggested portal api. Some questions/comments:

- Why is there a "Launch" ? I would say launching applications is a user task and unrelated to installing launchers
- I don't like that the api docs seem to imply that callers need to use the api differently depending on whether they are sandboxed or not. That seems bad

mwleeds:
> I may not be in the right mindset to comment here. I have more questions than answers, and dislike many of the things I see in the suggested portal api. Some questions/comments:
> 

Okay, I'll try to answer your specific concerns but I don't think I can do much about unspecified concerns.

> * Why is there a "Launch" ? I would say launching applications is a user task and unrelated to installing launchers

Epiphany has a feature where the user can click on the notification that appears when a web app is installed to launch it. Since the desktop file is not directly accessible to Epiphany, launching it has to happen on the portal side, so in order to support clicking on the aforementioned notification we need a portal method. There may be other cases where applications want to launch their dynamic launchers, for example in Steam when you want to play a game or in Bottles when you want to launch a Windows app. We could decide that launching should be part of a separate portal interface, but I'm not sure what restrictions we'd want to put on sandboxed apps. As written in this PR, sandboxed apps can only launch their own dynamic launchers.

> * I don't like that the api docs seem to imply that callers need to use the api differently depending on whether they are sandboxed or not. That seems bad

I'm not sure if you're referring to the restrictions on `.desktop` file names or to the `RequestInstallToken()` method?

For the `.desktop` file names, we don't have an app ID available for unsandboxed apps, so I don't think we can put any restrictions on them.

For the `RequestInstallToken()` method, I think we need some way for Software to install web apps without any portal dialog, since Software has its own UI for the install operation and it needs to let Epiphany finish the installation even if sandboxed (to avoid re-creating the previous buggy implementation where Software tried to imitate Epiphany's desktop files and create them on its own). So I don't really see a way around it aside from moving that functionality into `PrepareInstall()` as Georges suggested. If that's what you are suggesting I'd appreciate if one or both of you could respond to the [concerns](https://github.com/flatpak/xdg-desktop-portal/pull/696#discussion_r799143821) I have about that.

Thanks for taking a look at this

matthiasclasen:
> I'm not sure if you're referring to the restrictions on `.desktop` file names or to the `RequestInstallToken()` method?

I was referring to language like:

     When initiated by a sandboxed application, the PrepareInstall() method would...
  
      By contrast, an unsandboxed process such as a system component can choose...

This seems very much against the spirit of flatpak. An app shouldn't have to care if it is sandboxed
or not, the apis should work (or deny service) just the same.

mwleeds:
> > I'm not sure if you're referring to the restrictions on `.desktop` file names or to the `RequestInstallToken()` method?
> 
> I was referring to language like:
> 
> ```
>  When initiated by a sandboxed application, the PrepareInstall() method would...
> 
>   By contrast, an unsandboxed process such as a system component can choose...
> ```
> 
> This seems very much against the spirit of flatpak. An app shouldn't have to care if it is sandboxed or not, the apis should work (or deny service) just the same.

Some of this might just be bad wording because `PrepareInstall()` does work exactly the same for sandboxed and unsandboxed applications. See if the wording I just pushed seems any better.

swick:
> Today I learned there are https://github.com/flatpak/xdg-desktop-portal/pull/581 to set the app ID properly for unsandboxed applications (processes)

That one has been stalled for some time now. If someone could take a look at it that would be great.

mwleeds:
Georges and I talked this morning and decided on this as the way forward:

- Make the `RequestInstallToken` method work for both sandboxed and unsandboxed processes
- Have `RequestInstallToken` defer to backends on the decision of whether to allow non-interactive launcher installation
- Have backends use the caller's app ID to decide whether to allow, and hard-code the app IDs for GNOME Software and KDE Discover in the respective back-ends (I haven't started a kde backend though and probably don't have time).
- Get #581 merged before this PR so we can depend on the app ID being set for unsandboxed processes.

I'm glad we could find a good solution and get this moving along!

data/org.freedesktop.portal.DynamicLauncher.xml
mwleeds: As noted [below](https://github.com/flatpak/xdg-desktop-portal/pull/696#issuecomment-1051090694), we came to a resolution on this discussion.

mwleeds:
> * Get [Retrieve a not-empty Application ID for un-sandboxed clients that use cgroups #581](https://github.com/flatpak/xdg-desktop-portal/pull/581) merged before this PR so we can depend on the app ID being set for unsandboxed processes.
> 

Unfortunately this turned out to be very complicated. #719 successfully implements getting the app ID from the systemd scope unit attached to the calling process, but it doesn't work if you run the app from the command line, which is a pretty significant issue for development workflows. In the interest of getting this portal merged in the near future, I suggest that we only enforce the app ID allowlist on sandboxed apps and allow any unsandboxed app to use `RequestInstallToken()`. In general such apps would be able to install the desktop launcher on their own without the help of the portal so it should not entail privilege escalation.

Longer term, I can think of a couple solutions to #579:

- Provide a portal-wide `SetAppID()` method and only allow unsandboxed processes to use it. It would be called before other portal methods and would let us keep those method signatures as they are.
- Defer to back-ends on the mapping from pids to app IDs, and implement it in each one. The gnome backend would talk to gnome-shell at a currently-nonexistent D-Bus interface, kde would do something similar, and so on.

swick:
> In the interest of getting this portal merged in the near future, I suggest that we only enforce the app ID allowlist on sandboxed apps and allow any unsandboxed app to use RequestInstallToken().

That sounds reasonable to me.

> Provide a portal-wide SetAppID() method and only allow unsandboxed processes to use it. It would be called before other portal methods and would let us keep those method signatures as they are.

Also sounds reasonable.

> Defer to back-ends on the mapping from pids to app IDs, and implement it in each one. The gnome backend would talk to gnome-shell at a currently-nonexistent D-Bus interface, kde would do something similar, and so on.

Even gnome-shell is best-effort so you still won't be able to get the correct appid in all cases and you still have to handle empty appid's. All of that complexity would be only for slightly improving the best-effort matching.

mwleeds:
I updated this branch so we only skip the app ID check in `RequestInstallToken()` if the app is both unsandboxed *and* its app ID couldn't be determined, and updated the docs to warn people not to depend on the method succeeding unless their app ID is in the allowlist, so we have the freedom to later make it a strict allowlist if desired.

Also pushed some other minor changes; I believe this is ready for another round of review when someone has time.

GeorgesStavracas:
I think it's worth splitting e6fd8abbebe486d75959c8aa5d797f6d64267076 into a separate pull request, and land independently of this one.

GeorgesStavracas:
@mwleeds as a preparation for another round of reviews, could you please squash the fixups and rebase on top of master?

mwleeds:
> @mwleeds as a preparation for another round of reviews, could you please squash the fixups and rebase on top of master?

Sure, done

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+      g_set_error (&error,
+                   XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_FAILED,
+                   _("Desktop file '%s' icon at unrecognized path"), arg_desktop_file_id);
+      goto error;
+    }
+
+  icon_file = g_file_new_for_path (icon_path);
+  icon = g_file_icon_new (icon_file);
+  stream = g_loadable_icon_load (G_LOADABLE_ICON (icon), 0, NULL, NULL, NULL);
+  /* Icons are usually smaller than 1 MiB. Set a 10 MiB
```
GeorgesStavracas, line 886: Nitpick: Add a newline above the comment

GeorgesStavracas (changes requested):
Another round

src/validate-icon.c
```diff
@@ -85,6 +85,17 @@ validate_icon (const char *arg_width,
...
+  if (width != height)
+    {
+      g_printerr ("Expected a square icon but got: %dx%d\n", width, height);
+      return 1;
+    }
+
+  /* Print the format and size for consumption by (at least) the dynamic
+   * launcher portal.
+   */
+  g_print ("%s:%d", name, width);
```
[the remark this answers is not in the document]
mwleeds: Reading the file twice in two processes would be pretty bad for performance.... However we could possibly upstream this change to Flatpak and Flatpak would just ignore the output.

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+    {
+      transient_permissions = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                                     g_free, (GDestroyNotify)g_variant_unref);
+    }
+
+  /* Revoke the token if it hasn't been used after 5 minutes, in case of
+   * client bugs. This is what the GNOME print portal implementation does.
+   */
+  timeout_id = g_timeout_add_seconds_full (G_PRIORITY_DEFAULT, 300, install_token_timeout,
+                                           g_strdup (token), g_free);
```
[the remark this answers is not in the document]
mwleeds: I don't really see how that's possible. The frontend doesn't have any knowledge of the token, and if we were going to change the architecture so that it does we would need a good reason to do so.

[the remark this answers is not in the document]
GeorgesStavracas: Ah right, alright then.

src/validate-icon.c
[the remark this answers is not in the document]
mwleeds: Here's that PR: https://github.com/flatpak/flatpak/pull/4803

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+
+  launcher_data_wrapped = g_hash_table_lookup (transient_permissions, token);
+  if (launcher_data_wrapped)
+    {
+      guint timeout_id;
+      g_autoptr(GVariant) launcher_data = NULL;
+      g_variant_get (launcher_data_wrapped, "(vu)", &launcher_data, &timeout_id);
+      g_source_remove (timeout_id);
+      g_hash_table_remove (transient_permissions, token);
+      return g_steal_pointer (&launcher_data);
```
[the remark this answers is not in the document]
mwleeds: Sure that looks better

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+  g_autofree char *icon_path = NULL;
+  g_autofree char *icon_subdir = NULL;
+  const char *name, *icon_extension, *icon_size;
+  const char *app_id;
+
+  g_variant_get (launcher_data, "(&sv&s&s)", &name, &icon_v, &icon_extension, &icon_size);
+  g_assert (name);
+  g_assert (icon_v);
+  g_assert (icon_extension);
+  g_assert (icon_size);
```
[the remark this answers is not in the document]
mwleeds: Makes sense

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+
+static gboolean
+validate_launcher_type (const char  *key,
+                        GVariant    *value,
+                        GVariant    *options,
+                        GError     **error)
+{
+  guint32 launcher_type = g_variant_get_uint32 (value);
+  g_autoptr(GError) local_error = NULL;
+  guint32 supported_launcher_types = xdp_dynamic_launcher_get_supported_launcher_types (XDP_DYNAMIC_LAUNCHER (dynamic_launcher));
```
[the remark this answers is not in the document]
mwleeds: :+1:

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+  if (xdp_app_info_is_host (call->app_info) && g_str_equal (app_id, ""))
+    {
+      response = 0;
+    }
+  else if (!xdp_impl_dynamic_launcher_call_request_install_token_sync (impl,
+                                                                       app_id,
+                                                                       arg_options,
+                                                                       &response,
+                                                                       NULL, /* cancellable */
+                                                                       &error))
```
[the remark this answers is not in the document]
mwleeds: Hmm, that's not actually equivalent though, because in your implementation an unsandboxed app that has its app ID set would not have the app ID checked by the backend.

src/request.c
```diff
@@ -310,6 +310,13 @@ get_token (GDBusMethodInvocation *invocation)
             options = g_variant_get_child_value (parameters, 1);
           }
     }
+  else if (strcmp (interface, "org.freedesktop.portal.DynamicLauncher") == 0)
+    {
+        if (strcmp (method, "PrepareInstall") == 0 )
+          {
+            options = g_variant_get_child_value (parameters, 3);
+          }
```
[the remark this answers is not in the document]
mwleeds: I think that would be fine if we want to change the whole rest of the function, but as it is it's consistent with the rest of the function.

src/xdp-utils.c
```diff
@@ -345,6 +348,50 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
...
+      g_autofree char *app_path = g_key_file_get_string (app_info->u.flatpak.keyfile,
+                                                         FLATPAK_METADATA_GROUP_INSTANCE,
+                                                         FLATPAK_METADATA_KEY_APP_PATH, NULL);
+      char *path = original_app_path ? original_app_path : app_path;
+      g_autofree char *app_slash = g_strconcat ("app/", app_info->id, NULL);
+      char *app_slash_pointer;
+      g_autofree char *tryexec_path = NULL;
+
+      if (path == NULL || *path == '\0')
+        return NULL;
```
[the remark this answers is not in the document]
mwleeds: Sure that looks nice

src/dynamic-launcher.c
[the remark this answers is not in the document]
GeorgesStavracas: The important part would be moving `response = 0` to the `else` block, I didn't pay much attention to the if conditions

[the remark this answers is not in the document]
mwleeds: > Ok, so you're proposing:
> 
> ```
>   if ((!xdp_app_info_is_host (call->app_info) ||
>        !g_str_equal (app_id, "")) &&
>       !xdp_impl_dynamic_launcher_call_request_install_token_sync (...
> ```
> 

Oops, just realized that would also not be correct, since it would overwrite the `response` value to `0` in the `else` block in case `xdp_impl_dynamic_launcher_call_request_install_token_sync()` returns `TRUE`, rather than accepting the `response` value from the backend.

src/dynamic-launcher.c
```diff
@@ -0,0 +1,1221 @@
...
+    if (g_strcmp0 (icon_extension, "svg") == 0)
+      subdir = g_strdup ("scalable");
+    else
+      subdir = g_strdup_printf ("%sx%s", icon_size, icon_size);
+
+    icon_subdir = g_build_filename (g_get_user_data_dir (), "xdg-desktop-portal", "icons", subdir, NULL);
+    icon_path = g_build_filename (icon_subdir, icon_name, NULL);
+
+    g_key_file_set_string (key_file, G_KEY_FILE_DESKTOP_GROUP, "Icon", icon_path);
+  }
```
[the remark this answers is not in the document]
GeorgesStavracas: Okay

src/dynamic-launcher.c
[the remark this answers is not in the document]
GeorgesStavracas: Nevermind then

GeorgesStavracas (approved):
Looks good. We shall land this now. Thanks!

GeorgesStavracas:
@mwleeds let's focus on the -gnome and -gtk portal implementations now. I've filed https://bugs.kde.org/show_bug.cgi?id=451510 for KDE.

mwleeds:
> @mwleeds let's focus on the -gnome and -gtk portal implementations now. I've filed https://bugs.kde.org/show_bug.cgi?id=451510 for KDE.

Sounds good thanks for filing that. Of course there are other back-ends as well but I don't think we have an obliglation to file a ticket for each one.

bilelmoussaoui:
Is it possible to add the docs for the new portal?

mwleeds:
> Is it possible to add the docs for the new portal?

Yup, thanks for the reminder: https://github.com/flatpak/flatpak-docs/pull/325

PR #711, opened 2022-02-16
Add a global shortcut portal

### LINKED ISSUE #624: Global keyboard shortcut portal
We have received several requests from elementary AppCenter app developers for a way to set (or prompt a user to set) a global keyboard shortcut to launch their app or a specific feature of their app (e.g. an [Application action](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s11.html)).

Pre-sandbox, these developers would set the GSettings for the desktop itself to add their app to the global keyboard shortcuts. Obviously this is non-ideal and only reasonably worked because we human reviewed every app’s source upon submission and update.

Today in Flatpaks, we recommend developers direct users to System Settings > Keyboard > Shortcuts and tell users how to add a custom shortcut manually.

Ideally, however, apps could use a Portal to request a system-wide shortcut (along with a description/rationale of the feature), and then we could provide a UI to display the request, let the user pick a shortcut, avoid conflicts, etc.

aleixpol:
It's designed so that applications can register actions that can be
triggered globally (i.e. regarless of the system's state, like focus).

WIP because:
* It's untested
* There's no implementation for it
* Needs reviews

Fixes #624


data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,134 @@
...
+      org.freedesktop.portal.GlobalShortcut:
+      @short_description: GlobalShortcut portal backend interface
+
+      This this portal lets applications register global shortcuts so they can
+      act regardless of the system state upon an input event.
+
+      To use this, an app will first create a session under which its actions
+      will be collected. Then shortcuts can be registered, listed and eventually
+      unregistered. It's up to the portal implementation to perform the registration
+      and tracking of that shortcut throughout the session's lifespan.
```
whot, line 32: fwiw, I found this description to be confusing because afaict you cannot register [1] a shortcut here, you can merely *bind* [2] to existing shortcuts. So aiui you can set up a "zoom in" shortcut in the compositor which will then show up in the list below. The application can "bind" to that shortcut but only because it exists.

An application cannot say "here's a new shortcut 'rotate left'" with this interface.

[1] "register" being interpreted as "create a new thing and add it to a lookup table"
[2] "bind" being interpreted as "make known that I want to use this thing"

src/globalshortcuts.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Authors:
+ *       Aleix Pol Gonzalez <aleixpol@kde.org>
+ */
+
+#pragma once
+
+#include <gio/gio.h>
+
+GDBusInterfaceSkeleton * globalshrtcuts_create (GDBusConnection *connection,
```
whot, line 25: /me frantically throws an 'o' at this function

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,156 @@
...
+        Deactivated:
+        Notifies that a shortcut is not active anymore.
+
+        @session_handle: Session that requested the shortcut
+        @shortcut_id: the application-provided ID for the notification
+    -->
+    <signal name="Deactivated">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="shortcut_id" direction="in"/>
+    </signal>
```
whot, line 142: activated/deactivated indicates that applications care about the state, equivalent to press/release. Which also suggests that applications *may* care about the duration of that state for things like key repeat. Is that the intention here? If not, a single signal `Triggered` may be sufficient?

whot (commented):
fwiw, personally I could really use a few examples here to better understand what the intention is of this portal. I get that it would address the simple "ctrl+c" shortcut type but it seems overspecced for that simple case. And I think it may be underspecced for the more complex cases that require key repeats. So I'm wondering: what are the specific use-cases you are trying to solve here?

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,156 @@
...
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @results: Vardict with the results of the call
+
+        List the shortcuts registered in the @p handle session.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>names a{sv}</term>
```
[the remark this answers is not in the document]
aleixpol: Correct, addressed.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,156 @@
...
+            <term>shortcut_ids as</term>
+            <listitem><para>
+              The ids of the supported shortcuts
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>shortcut_descriptions as</term>
+            <listitem><para>
+              User-readable description of what the shortcut does
+            </para></listitem>
```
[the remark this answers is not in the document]
aleixpol: Fair enough, addressed.

src/globalshortcuts.h
aleixpol: Thanks!

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: It's important to support use-cases like Push-to-Talk

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: I've fixed this description, it was a left-over from the discussion that happened in the originating ticket. Please tell me if it makes sense to you now.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,134 @@
...
+
+      Upon a shortcut trigger, Activated will be emitted until when it's not
+      triggered anymore, then Deactivated will be issued.
+  -->
+  <interface name="org.freedesktop.portal.GlobalShortcuts">
+        <!--
+        CreateSession:
+        @options: Unused so far
```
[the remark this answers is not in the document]
aleixpol: Not sure what you mean. This is what the CreateSession method looks like in ScreenCast:

```
    <method name="CreateSession">
    </method>
```

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,151 @@
...
+
+        List the shortcuts registered in the @p handle session.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>names a{sv}</term>
+            <listitem><para>
+              The different shortcut ids that have been registered mapped to user-readable descriptions for the app to present that describes the shortcut.
+            </para></listitem>
```
whot, line 77: as a further thought to future-proof this: maybe `aa{sv}`, so an array of dicts, resolving to something like:
```yaml
- shortcut_id: foo
  description: blah
  enabled: true
- shortcut_id: bar
  ...
```

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: > If so, don't you need a description to go from application to portal too? Something like Bind("zoom-in", "Zoom into the canvas") so the portal can use a human-readable version to display what the user should create a mapping for

When we CreateSession we already pass a text explaining what the shortcut is for. The portal can use that.

> And if you want to future-proof this, make it an a{sv} passed in via options, so you can later add things like enabled.

We can always have more methods to interact with the shortcut. We could have a setEnabled(shortcut-id) kind of call.

> I wonder how that'll work with translations btw but i18n is not my forte. Or switching keyboard layouts after the shortcut was bound, because "Press Ctrl+A" might as well now be an effective "Ctrl+Q". This part is largely compositor-dependent though.

The portal will offer strings in the user's language. And yes, since it's for the user it can be either "Ctrl+A", "Control+A", "Press Control and without letting go, press A" and any variation. Obviously, conciseness is advisable. FWIW, other more esoteric options would be possible like "Say Banana to your toaster", again responsibility of your portals implementation.

smcv:
> I get that it would address the simple "ctrl+c" shortcut type but it seems overspecced for that simple case

My understanding is that this is for global shortcuts that are always active, even while an unrelated application has keyboard focus. Instead of "copy" or "zoom in", think more like "start recording".

While an application has focus, it can receive Ctrl+C (or similar) in the usual X11 or Wayland way, without needing portals.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: There is also the risk of overcomplicating everything just in case.

jadahl:
Lets say we have an application that can control the sun shade of a physical window, and that it wants to bind the actions "roll up sun shade" and "roll down sun shade" using the global shortcut portal to some two key bindings.

Could you describe step by step how the application does this, when a portal dialog will be presented, and how the application does the second time it launches?

aleixpol:
> Could you describe step by step how the application does this, when a portal dialog will be presented, and how the application does the second time it launches?

app:
```
void activatedCallback(name)
{
    if (name == "shade-up")
        shade->up();
    else if (name == "shade-down")
        shade->down();
}

request = CreateSession({
    "shade-up", i18n("Rolls the sun shades up"),
    "shade-down", i18n("Rolls the sun shades down")
})

request.exec(); // Wait for it to finish...
session = request.result()["session_handle"];

connect(session.activated, activatedCallback);

if (we really want to make sure the global shortcuts are active) {
    request = ListShortcuts(session);
    request.exec(); // Wait for it to finish...
    shortcuts = request.result(); // [2]
    
    if (shortcuts["shade-up"] == null) {
        request = Bind("shade-up"); // The portal might prompt here [1]
        request.exec(); // Wait for it to finish...
        request.makeSureItDidNotFail();
    }
    if (shortcuts["shade-down"] == null) {
        request = Bind("shade-down"); // The portal might prompt here [1]
        request.exec(); // Wait for it to finish...
        request.makeSureItDidNotFail();
    }
}

// We update the UI for the user to know how to trigger it
{
    request = ListShortcuts(session); // [3]
    request.exec(); // Wait for it to finish...
    shortcuts = request.result();
    
    ui.shadeUpLabel.text = shortcuts["shade-up"]
    ui.shadeDownLabel.text = shortcuts["shade-down"]
}
```

1. The portal will always be notified and asked to bind so the application can call this again if the binding needs changing.
2. on the second executions the portal should restore the bindings that were initially registered. We could consider including the ListShortcuts output to the CreateSession's request output
3. We call ListShortcuts twice for readability, Bind() does offer this information in the request's output

data/org.freedesktop.portal.GlobalShortcuts.xml
whot: > When we CreateSession we already pass a text explaining what the shortcut is for. The portal can use that.

indeed, I read over that. for the documentation, maybe use different terms like "description" for what the shortcut is supposed to do and "definition" for what the shortcut is set to trigger by? or whatever else the thesaurus suggests. Makes things easier for denser people like me :)

GeorgesStavracas (changes requested):
I have a few comments about the design of the API. None of the C code was reviews, I'll review the implementation once we figure out the D-Bus bits.

Something I'd like us to think through is the security model of this API. The current iteration implies that applications would need to request the same shortcuts on startup, and no permissions are stored anywhere. This can be an annoying behavior. On the other hand, allowing any app to request global shortcuts without any extra check is not ideal either. So the question is: how would we implement apps only requesting permissions once?

jadahl:
I agree with @GeorgesStavracas that it'd be good to be able to batch multiple Bind requests into a single `BindShortcuts` as described above.

Thinking more about the permission model would indeed be good. What needs to be ensured is that: applications should not be allowed to arbitrarily bind shortcuts without user consent; at least the API must make sure that a backend can implement such semantics. It'd also be good to allow an application to be able to request new bindings, or change existing ones, thus the initial set of bindings being set up at session construction is a bit awkward.

Given this, maybe we can take inspiration from the screen cast session restore functionality, by making the following changes:

1. Replace the `shortcuts` parameter in `CreateSession` with a `restore_token` string, 
2. Add a `restore_token` to the data passed via the response signal emitted after `CreateSession`
3. Replace `Bind` with `BindShortcuts` to allow for binding in batches (thus setting up multiple bindings in the same dialog)
4. Pass internationalized description that was previously part of the `shortcuts` parameter in `CreateSession` together with `BindShortcuts`

With this, the flow would be altered slightly, but I believe it should be possible to implement the same user interface as you described in the issue:
```
void activatedCallback(name)
{
    if (name == "shade-up")
        shade->up();
    else if (name == "shade-down")
        shade->down();
}

options = {};
if (has_restore_token)
    options["restore_token"] = get_saved_restore_token();

request = CreateSession(options);

response = request.exec(); // Wait for it to finish...
session = request.result()["session_handle"];

response_restore_token = response["restore_token"];

needs_to_bind = False;
existing_bindings
if (response_restore_token is null)
    needs_to_bind = True;
else {
    existing_bindings = ListShortcuts(session);
    if (existing_bindings.length < wanted_bindings &&
        should_nag_user_about_bindings())
}
    

connect(session.activated, activatedCallback);

if (we really want to make sure the global shortcuts are active) {
    request = ListShortcuts(session);
    request.exec(); // Wait for it to finish...
    shortcuts = request.result(); // [2]
    
    if (shortcuts["shade-up"] == null) {
        request = Bind("shade-up"); // The portal might prompt here [1]
        request.exec(); // Wait for it to finish...
        request.makeSureItDidNotFail();
    }
    if (shortcuts["shade-down"] == null) {
        request = Bind("shade-down"); // The portal might prompt here [1]
        request.exec(); // Wait for it to finish...
        request.makeSureItDidNotFail();
    }
}

// We update the UI for the user to know how to trigger it
{
    request = ListShortcuts(session); // [3]
    request.exec(); // Wait for it to finish...
    shortcuts = request.result();
    
    ui.shadeUpLabel.text = shortcuts["shade-up"]
    ui.shadeDownLabel.text = shortcuts["shade-down"]
}
```

GeorgesStavracas:
Another question that just popped up: should this protocol support binding the same shortcut to multiple key combinations?

jadahl:
> Another question that just popped up: should this protocol support binding the same shortcut to multiple key combinations?

Sounds like something an impl backend can do without involving any APIs.

swick:
> Sounds like something an impl backend can do without involving any APIs.

If we want to give apps access to a string which describes what the shortcut is bound to this might not be so straightforward.

GeorgesStavracas:
Converted to a Draft so no one (read: me) accidentally merges it

aleixpol:
> Thinking more about the permission model would indeed be good. What needs to be ensured is that: applications should not be allowed to arbitrarily bind shortcuts without user consent; at least the API must make sure that a backend can implement such semantics.

There's no way for an app to register here. We have not even specified how that would work. I'm afraid we might be miscommunicating.

> It'd also be good to allow an application to be able to request new bindings, or change existing ones, thus the initial set of bindings being set up at session construction is a bit awkward.

It could be nice to show some UI to let the portal handle the UX for all global shortcuts. I wouldn't expect this to be how every app does it though, as I expect apps to want to show their own UI for listing these shortcuts.

> Given this, maybe we can take inspiration from the screen cast session restore functionality, by making the following changes:
> 
>     1. Replace the `shortcuts` parameter in `CreateSession` with a `restore_token` string,
>     2. Add a `restore_token` to the data passed via the response signal emitted after `CreateSession`
>     3. Replace `Bind` with `BindShortcuts` to allow for binding in batches (thus setting up multiple bindings in the same dialog)
>     4. Pass internationalized description that was previously part of the `shortcuts` parameter in `CreateSession` together with `BindShortcuts`

My understanding is that we should keep the shortcuts an app has requested by its name, no need to keep a token. I don't know in which cases we'd consider the application a different one.

In practice, we can consider shortcuts to be static in a specific version of an app and could change between versions but it should only change slightly.

I do want us to be able to offer a UX that doesn't _necessarily_ have to show UI from the portals, as we've seen this lead us to rather awkward workflows with cross-platform apps.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,165 @@
...
+        ListShortcuts:
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @options: Unused so far
+
+        List all the shortcuts with their invocation descriptions.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>shortcut_ids a{sv}</term>
```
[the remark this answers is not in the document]
aleixpol: Why do we want to receive back the description of the shortcut?

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,165 @@
...
+
+    <!--
+        ShortcutsChanged:
+        @session_handle: Session that requested the shortcut
+
+        Indicates that the information associated with some of the shortcuts has changed.
+    -->
+    <signal name="ShortcutsChanged">
+      <arg type="o" name="session_handle" direction="out"/>
+    </signal>
```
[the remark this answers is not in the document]
aleixpol: We could include the information of ListShortcuts(), as this is what it would generally trigger, sure.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,165 @@
...
+              The user-readable description of the registered binding. This is intended for the app to be able to render the shortcut into the UI, allowing the user to learn the shortcut
+            </para></listitem>
+          </varlistentry>
+      </variablelist>
+    -->
+    <method name="Bind">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="shortcut_id" direction="in"/>
+      <arg type="o" name="request_handle" direction="out"/>
+    </method>
```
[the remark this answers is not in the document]
aleixpol: Added. Followed the semantics we've used so far elsewhere for consistency, not that they can't change still.

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
GeorgesStavracas: I don't think we want, this was just an example. Though, I'm considering the possibility of making `trigger` be an array of strings (and rename it to `triggers` of course), so that this portal can eventually deal with multiple bindings.

aleixpol:
> Another question that just popped up: should this protocol support binding the same shortcut to multiple key combinations?

You can also give from gnome's portal a description that is `Ctrl+A or Meta+Z`, this would allow us to make the supported feature set be less tied to the spec.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,188 @@
...
+          </varlistentry>
+      </variablelist>
+    -->
+    <method name="Bind">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="shortcut_id" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QVariantMap"/>
+    </method>
```
GeorgesStavracas, line 117: With the introduction of `BindShortcuts`, I think this method is unecessary now. People can use `BindShortcuts` for all cases, including individual shortcuts.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,188 @@
...
+              The different shortcut ids that have been registered mapped to user-readable descriptions for the app to present that describes how the shortcut can be invoked.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="BindShortcuts">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="as" name="shortcuts" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
```
GeorgesStavracas, line 142: Do you have any particular thoughts about documenting a `replace` boolean option? My idea is that it would be sent as part of the `options` vardict, and would control whether the array of shortcuts should be merged or replace the currently active shortcuts.

GeorgesStavracas (commented):
I appreciate Apol is still working on the previous round of review, but I had a few comments to flush

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: Yes, I'll have to agree. In my PoC this proved to be harder to implement than I had hoped.

aleixpol:
Did a PoC implementation of the portal.
https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/80

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
jadahl: How do you mean the information should be listed if the type is `a{sv}`? If the key is the `id`, then what`s the `v`?

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,163 @@
...
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>session_id s</term>
+            <listitem><para>
+              The session id. A string representing the created global shortcuts session.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>shortcuts a{sv}</term>
```
jadahl, line 49: Also here about how the type is supposed te represent the data passed. Is the "key" the `id`, but what's in the `v`?

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,163 @@
...
+            <listitem><para>
+              The different shortcut ids that have been registered mapped to user-readable descriptions for the app to
+              present that describes the shortcut.
```
jadahl, line 52: What will the portal do with these shortcut IDs? I see no details about whether they are intended to be bound or not, or if they are just for preparing a future `BindShortcuts` call.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,163 @@
...
+      <arg type="a{sv}" name="results" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out0" value="QVariantMap"/>
+    </method>
+
+    <!--
+        BindShortcuts:
+        @handle: Object path for the #org.freedesktop.impl.portal.Request object representing this call
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
+        @shortcuts: The identifier of the shortcuts we intend to register, empty for all shortcuts
```
jadahl, line 97: What happens if one tries to bind shortcuts not part of the `CreateSession` call?

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,134 @@
...
+        ListShortcuts:
+        @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
+        @options: Unused so far
+
+        Create a global shortcuts session.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>shortcut_ids  a{sv}</term>
```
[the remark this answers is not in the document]
aleixpol: Good catch. I forgot to update it when I changed the .impl one.

Do you think it would be more clear if it was specified `a{ss}`? I guess so since it's strings.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: It should error out. I am not sure how this is generally communicated, do you know what would be a good example?

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
jadahl: `a(ss)` for a list of the tuple pair of two strings, but I don't think it's a good idea, but would rather see `a(sa{sv})` as suggested elsewhere, i.e. so that now it can be

```
[
   'water-plants': {
     'description': 'Water the plants',
   }
]
```

but later extended to 
```
[
   'water-plants': {
     'description': 'Water the plants',
     'preferred-key-binding': '<Super>+W',
   }
]
```

if such a hint would ever be added.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: It can close the session and start a new one with the previous + the added one. This is how I imagined it at least.

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
jadahl: > Any reason why you prefer a(sa{sv}) to a{a{ss}}?

`a{sv}` for the per binding properties so that they can remain "type safe", so that e.g. an int can be an int, a list a list, so that we don't need no encode such things into a string.

> WRT the preferred-key-binding, we'll first need to standardise these strings.

Yea, I think as long as we prepare the API to make it possible to extend, it makes sense to leave that for a follow up version.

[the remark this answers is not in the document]
jadahl: Or you mean `aa{sv}` instead of `a(sa{sv})`? I.e. have the `id` in the vararg as its own property. That would work too.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
jadahl: I see. Re-creating the session is racy; if one triggers a binding right at the time the app decides to recreate the session, it can be lost. Maybe not a big deal since it's very unlikely to happen very often. Then again, could just as well be one session per binding couldn't it?

aleixpol: > Re-creating the session is racy; if one triggers a binding right at the time the app decides to recreate the session, it can be lost. Maybe not a big deal since it's very unlikely to happen very often. 

Correct. It shouldn't happen though during general runtime of the app.

> Then again, could just as well be one session per binding couldn't it?

Yes.

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
aleixpol: > Or you mean aa{sv} instead of a(sa{sv})?

Yes, sorry. I'm not all that fluent in this types languages yet.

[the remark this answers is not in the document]
aleixpol: I've changed all dictionaries to be a(sa{sv}) for consistency. Please review the latest changes and see if it's what you had in mind.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: This has been addressed since

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
aleixpol: This is somewhat the case now, please re-review

src/globalshortcuts.c
```diff
@@ -0,0 +1,556 @@
...
+  object_class->finalize = global_shortcuts_session_finalize;
+
+  session_class = (SessionClass *)klass;
+  session_class->close = global_shortcuts_session_close;
+}
+
+static GlobalShortcutsSession *
+global_shortcuts_session_new (GVariant *options,
+                         Request *request,
+                         GError **error)
```
GeorgesStavracas, line 109: Please align the function arguments so that they start after the parenthesis column. Do this here, throughout this entire file, and also the header file. It should look like this:

```c
static GlobalShortcutsSession *
global_shortcuts_session_new (GVariant *options,
                              Request *request,
                              GError **error)
```

src/globalshortcuts.c
```diff
@@ -0,0 +1,556 @@
...
+                  gpointer data)
+{
+  g_autoptr(Session) session = lookup_session (session_id);
+  GlobalShortcutsSession *global_shortcuts_session = (GlobalShortcutsSession *)session;
+
+  g_debug ("Received ShortcutsChanged %s", session_id);
+
+  if (global_shortcuts_session && !global_shortcuts_session->closed) {
+    xdp_dbus_global_shortcuts_emit_shortcuts_changed(data, session_id);
+  }
```
GeorgesStavracas, line 527: Single line `if`s don't need curly braces

src/globalshortcuts.c
```diff
@@ -0,0 +1,556 @@
...
+                  const char *session_id,
+                  gpointer data)
+{
+  g_autoptr(Session) session = lookup_session (session_id);
+  GlobalShortcutsSession *global_shortcuts_session = (GlobalShortcutsSession *)session;
+
+  g_debug ("Received ShortcutsChanged %s", session_id);
+
+  if (global_shortcuts_session && !global_shortcuts_session->closed) {
+    xdp_dbus_global_shortcuts_emit_shortcuts_changed(data, session_id);
```
GeorgesStavracas, line 526: All function calls should have a space before the parenthesis (e.g. `xdp_dbus_global_shortcuts_emit_shortcuts_changed (`)

src/globalshortcuts.c
```diff
@@ -0,0 +1,556 @@
...
+  if (!session)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  g_variant_builder_init (&options_builder, G_VARIANT_TYPE_VARDICT);
+  options = g_variant_builder_end (&options_builder);
+
+  g_warning("seeeeeeeeeet %x %x %x", request, quark_request_session, session);
```
GeorgesStavracas, line 245: Uh oh

src/globalshortcuts.c
```diff
@@ -0,0 +1,556 @@
...
+  GlobalShortcutsSession *global_shortcuts_session = (GlobalShortcutsSession *)session;
+
+  global_shortcuts_session->closed = TRUE;
+}
+
+static void
+global_shortcuts_session_finalize (GObject *object)
+{
+  G_OBJECT_CLASS (global_shortcuts_session_parent_class)->finalize (object);
+}
```
GeorgesStavracas, line 86: This override can be removed, since it's empty

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,184 @@
...
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In3" value="QVariantMap"/>
+    </method>
+
+    <!--
+        Activated:
+        Notifies about a shortcut becoming active.
+
+        @session_handle: Session that requested the shortcut
+        @shortcut_id: the application-provided ID for the notification
+        @timestamp: When the event happened in ms
```
GeorgesStavracas, line 150: Does it make any sense for timestamps to have a unit like this? I always thought generic timestamps like this one were just opaque numbers that don't really represent any specific time or range, and should be used strictly for comparison.

GeorgesStavracas (changes requested):
Some more comments, this time I started to look at the C code since the D-Bus API seems to be reaching some level of maturity

src/globalshortcuts.c
aleixpol: I'll do my best. Is there a clang-format config for it?

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: I've seen it backfire. When I was implementing the recording of pipewire streams into ffmpeg I found the lack of units a big problem when reading documentation. I understand it's not the same case though, still I think it's good to say what we are talking about.

We are not providing a frame of reference, so it's still for comparison anyway.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,156 @@
...
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>description</term>
+            <listitem><para>
+              The user-readable description of the registered binding. This is intended for the app to be able to render the shortcut into the UI, allowing the user to learn the shortcut
+            </para></listitem>
+          </varlistentry>
+      </variablelist>
+    -->
```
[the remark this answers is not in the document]
aleixpol: There is no bind anymore, so closing for now.

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
aleixpol: Added.

data/org.freedesktop.portal.GlobalShortcuts.xml
jadahl: Specifying unit seems good, but maybe not the origin (e.g. what clock other than it's not something affected by timezone changes or real time clock changes). Maybe make it microseconds instead though? It's the source of most timestamps; milliseconds feels more like an X11 relic.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,188 @@
...
+        Shows a dialog with all shortcuts to configure and offers them back to the app when done.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>shortcuts a(sa{sv})</term>
+            <listitem><para>
+              The different shortcut ids that have been registered. They are mapped to a list of pairs including:
+                - description: User-readable text describing what the shortcut does
+                - trigger: User-readable text describing how to trigger the shortcut
```
jadahl, line 129: What's the purpose of repeating the same information that was provided when creating the session?

If the assumption is that one creates a session consisting of a predefined set of shortcuts, and a application wanting to request that a subset of them to be bound, it should be enough to just provide a list of shortcut IDs; the rest is already there.

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: Changing to µs

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: The idea was that we'd get the triggers that were just bound. I agree that getting the descriptions again is not that useful, I just added it so that we get the shortcuts dictionary every time rather than several very similar ones.

After implementing, it seems to me that it might be too synchronous an API.

aleixpol (commented):
The implementation I did for Plasma is working as expected now.

aleixpol:
Ping?

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,170 @@
...
+
+ Author: Aleix Pol Gonzalez <aleixpol@kde.org>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.GlobalShortcut:
+      @short_description: GlobalShortcut portal backend interface
+
+      This this portal lets applications register global shortcuts so they can
```
PolyMeilex, line 26: Is "This this" a typo?

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,195 @@
...
+
+ Author: Aleix Pol Gonzalez <aleixpol@kde.org>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.GlobalShortcut:
+      @short_description: GlobalShortcut portal backend interface
+
+      This this portal lets applications register global shortcuts. These shortcuts can be
```
PolyMeilex, line 26: Same here, is double this a typo?

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: Yes, they were typos. Thanks for pointing it out!

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,195 @@
...
+      When a shortcut triggers, Activated is emitted. When it is not triggered anymore, then
+      Deactivated is issued.
+  -->
+  <interface name="org.freedesktop.portal.GlobalShortcuts">
+      <!--
+        CreateSession:
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Creates a global shortcuts session.
```
jadahl, line 43: This should describe what happens to the passed shortcuts; e.g. whether they are potentially restored from previous sessions. I assume they can't possibly be automatically bound the first time this is called for an app?

data/org.freedesktop.portal.GlobalShortcuts.xml
jadahl: Maybe it should be documented in `ListShortcuts` and just referred to from here?

jadahl:
> but does this allow for a shortcut to be visible to both the focused application and the application(s) that have registered it?

This would make a globally registered shortcut capture the input events that was bound, so that they would not reach the focused application, but the application that registered the global shortcut, in the form of a "activated"/"deactivated" signal.

It'd be visible only to the application that registered the shortcut, not any other application.

jadahl:
> How do you think it could be addressed?

What exactly is it that needs to be addressed? What's the use case / behavior you're after? After all, "global shortcuts" is about making an app key binding (or equivalent) "global" i.e. be active no matter the focus.

Frogging101:
> What's the use case / behavior you're after?

Essentially, if I bind my Ctrl key as push-to-talk in a voice chat application, I still want to be able to use it elsewhere.

jadahl:
The global keybinding for push-to-talk would only be active when the push-to-talk application is running and actively asking for it (by having an active global shortcuts session).

Frogging101:
Yes, but that would mean that whenever the application is running and I'm engaged in voice chat, I cannot use my Ctrl key (or whichever key is bound) for anything else. That would be a problem.

jadahl:
> so I assume it's up to the implementation whether the input event is "eaten" by the key binding. Is that right?

It's currently not clearly defined, but should probably be.

PolyMeilex:
This protocol does not define any string to keyboard shortcut standard, right? So apps can't provide their "default"/"preferred" keybindings to the portal. All descriptions/hints are in human-readable form.

If so, my question is: is this left for potential future update? Or is this something that we don't want to do at all?
Requiring users to bind every key manually without any hints/defaults seems way too DIY for a DE. In my opinion, portal popup should be used to accept default shortcuts and solve conflicts or change stuff as user see fit, rather than "define everything yourself" popup.

aleixpol:
> I don't see anything in the protocol that would mandate this one way or another, so I assume it's up to the implementation whether the input event is "eaten" by the key binding. Is that right?

Not specifying it means that the implementation can decide what to do with it. I am not sure of what we'd win by specifying it but there could be something I am missing.

Frogging101:
> I'm not sure what we'd win by specifying it
It would ensure consistent behaviour across implementations.

brunoais:
First, I agree with Brodie's opinions. If you want to know his opinions, you can see them here:
https://odysee.com/@BrodieRobertson:5/xdg-desktop-portals-fix-waylands-biggest:1

Second, I have a few opinions here and I'll put into different comments:

Related to [@aleixpol's answer, specially the code block](https://github.com/flatpak/xdg-desktop-portal/pull/711#issuecomment-1063114892):

I don't see how the shortcuts are being defined there. Are they defined in a separate file? If so, it would only be useful to a very limited amount of users.

Related to [@GeorgesStavracas's comment](https://github.com/flatpak/xdg-desktop-portal/pull/711#pullrequestreview-910903645) where he mentions:
> On the other hand, allowing any app to request global shortcuts without any extra check is not ideal either. 

Exactly.

> So the question is: how would we implement apps only requesting permissions once?

For the UI, I would take inspiration from Android in this regard. In the popup that appears requesting for access to the key binding: 
1. Add a checkbox allowing the user to confirm whether the app will have exclusive access to the shortcut (the app can set the default value of this one or we can allow the app to suggest but it's up to the user to tick the checkbox; depends on how secure it should be)
2. Add a checkbox the user can tick to tell flatpak that there's no need to ask again for that key combination. 
3. Mention a list of other apps that are already using the combination and whether they are currently running.
Then the user can finish the confirmation by clicking either allow or deny.

For the implementation,

1. The app can request on whether a global shortcut is temporary or not. By default they are not.
2. The app can hint on whether a global shortcut is exclusive or not. By default they are not.
3. The app must define whether the bind is to a character or to a key on the keyboard (I'd do this either transparent to the user or with just some small mention in the prompt. For the large majority of users, it doesn't matter).

For the 2nd element, the user not allowing the shortcut to be the way the application wants does not prevent it from working. I.e. If the shortcut is requested to be exclusive but the user doesn't allow it, the shortcut works as non-exclusive. 

For the 3rd element, when requesting a bind, I'd make the API so the program can provide either the key name (for character bind), usa standard QWERTY key name (this one is inspired by the web standards [explained @ MDN](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values) (for key bind) or the scan code. 

When the checkbox is ticked, I'd store that status somewhere where no program running in flatpak sandbox (regardless of permissions provided) can edit (except if through a file picker portal).

I think for the storage of the information something equivalent to the existing storage of permissions (`org.freedesktop.impl.portal.PermissionStore`) would work nicely.

[About @aleixpol's comment](https://github.com/flatpak/xdg-desktop-portal/pull/711#issuecomment-1073934576):
> My understanding is that we should keep the shortcuts an app has requested by its name, no need to keep a token. I don't know in which cases we'd consider the application a different one.

I would consider per packaged bundle (such as flatpak package). If permission is granted to a the bundle, it's granted to everything running in that bundle environment regardless of which program is effectively running (default, `bash`, `python`, etc...). If there's a single individual executable, then the bundle it the individual executable. If one executable requests, it applies to all programs in the bundle.
That also allows packaging separate programs for keybinding for cases of need or even for convenience of the programmer.

> In practice, we can consider shortcuts to be static in a specific version of an app and could change between versions but it should only change slightly.

I advocate for the shortcuts to be dynamic so the user can set his own for his convenience.


[From @GeorgesStavracas suggestion:](https://github.com/flatpak/xdg-desktop-portal/pull/711#issuecomment-1073981260)

I object. The details are above in this comment. Related to the implementation. How more options should exist.

[From @swick's suggestion:](https://github.com/flatpak/xdg-desktop-portal/pull/711#issuecomment-1073997290)

> For example if you want to delete a shortcut and the app doesn't delete it but deletes it in its UI you suddenly have a shortcut bound to a trigger which you are not aware of. How does it work if you want to bind multiple triggers? How does it work if you bind triggers which require rich UI? How do you handle conflicts and potential conflicts?

That's an app problem. It's up for the app to question the portal which key bindings has the user removed himself. Possibly, every time the user focuses on the keybinding window (when it appears or when focus is regained). I wouldn't solve this in the portal itself. If there's something for the portal to do, would be to send an event to the app letting it know that the user made the change.

[From @Frogging101's comment comment:](https://github.com/flatpak/xdg-desktop-portal/pull/711#issuecomment-1111448398)

> I don't see anything in the protocol that would mandate this one way or another, so I assume it's up to the implementation whether the input event is "eaten" by the key binding. Is that right?

I think it should be mandated and it should require approval by the user (see above in my comment).


Let me know if I missed something. As a developer and as a user, I'm quite interested in getting this working in a useful way.

GeorgesStavracas:
Tracking issue for GNOME implementation: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/47

@aleixpol please add a reference implementation of this portal to https://github.com/flatpak/libportal too

GeorgesStavracas:
I think it's safe to assume https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/56 is landed and move on with the rest of the work

aleixpol:
> @aleixpol please add a reference implementation of this portal to https://github.com/flatpak/libportal too

If you could help there it would be much appreciated, I am clearly not great at writing glib-y code.

I'll be updating this PR soon with the default shortcuts.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,170 @@
...
+        Create a global shortcuts session.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>shortcuts a(sa{sv})</term>
+            <listitem><para>
+              The different shortcut ids that have been registered. They are mapped to user-readable descriptions of the trigger. This can be used by the app to describe the shortcut to the user.
+
+              By default these shorcuts are not bound. To set them, BindShortcuts needs to be called for the portal
```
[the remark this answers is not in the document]
aleixpol: Addressed, it turns out I'd done this same typo several times... ^^'

orowith2os:
As an end user, I hope nobody minds if I put out my thoughts on how this should be implemented:

I, personally, like how Android handles permissions. Ask once, and it has the permission until you revoke it. However, maybe it would be possible to add an "Only for this session", and/or an "Allow this keybinding for (specified amount of time)" option so that, when the window closes/time is up, the permission is reset (this may be handled by the `xdg-desktop-portal-*`, I'm not sure exactly how the different all the `xdg-desktop-portal-*`s are)

Maybe, it would be possible to have all the permissions managed by the `xdg-desktop-portal` program itself instead of `xdg-desktop-portal-*`, instead using `xdg-desktop-portal-*` to ask for permissions and how long they should be given. This, I think, would allow for more standard keybindings for those users that switch between multiple desktop environments, instead of needing to re-set them every time they enter an environment, along with letting programs hook into `xdg-desktop-portal` instead of `xdg-desktop-portal-*` if it's a GUI manager for keybinding permissions.

orowith2os:
As an end user, I hope nobody minds if I put out my thoughts on how this should be implemented:

I, personally, like how Android handles permissions. Ask once, and it has the permission until you revoke it. However, maybe it would be possible to add an "Only for this session", and/or an "Allow this keybinding for (specified amount of time)" option so that, when the window closes/time is up, the permission is reset.

Maybe, it would be possible to have all the permissions managed by the `xdg-desktop-portal` program itself instead of `xdg-desktop-portal-*`, instead using `xdg-desktop-portal-*` to ask for permissions and how long they should be given. This, I think, would allow for more standard keybindings for those users that switch between multiple desktop environments, instead of needing to re-set them every time they enter an environment.

aleixpol:
Can  somebody please either accept this or raise a concern? It's been 1+ month without progress now.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,196 @@
...
+                - description: User-readable text describing what the shortcut does
+                - trigger: User-readable text describing how to trigger the shortcut
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="BindShortcuts">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="as" name="shortcuts" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
```
swick, line 138: Not documented

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,195 @@
...
+      <arg type="a{sv}" name="options" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In4" value="QVariantMap"/>
+    </signal>
+
+    <!--
+        ShortcutsChanged:
+        @session_handle: Session that requested the shortcut
+        @shorcuts: The different shortcut ids that have been registered. They are mapped to a list of pairs including:
+          - description: User-readable text describing what the shortcut does
+          - trigger: User-readable text describing how to trigger the shortcut
```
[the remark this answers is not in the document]
swick: I tend to agree.

swick:
@aleixpol There is still a few unaddressed comments from jadahl that I find very reasonable.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,184 @@
...
+        Notifies about a shortcut becoming active.
+
+        @session_handle: Session that requested the shortcut
+        @shortcut_id: the application-provided ID for the notification
+        @timestamp: When the event happened in ms
+    -->
+    <signal name="Activated">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="shortcut_id" direction="in"/>
+      <arg type="t" name="timestamp" direction="in"/>
```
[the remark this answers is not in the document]
jadahl: Should we specify what clock the timestamp is in? Without it, it's rather useless other than comparing to other `Activated` signal emissions.

mildsunrise:
I'm wondering how easy / proper would it be to allow apps to provide metadata with every shortcut, or to extend the protocol in the future. for things like icon, localization and so on

jadahl:
> I'm wondering how easy / proper would it be to allow apps to provide metadata with every shortcut, or to extend the protocol in the future. for things like icon, localization and so on

The API uses "varargs" for these types of things, i.e. key value sets, so adding more metadata is relatively easy, as it doesn't require changing the API other than describing the new keys.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
aleixpol: At the moment we have a comment in CreateSession specifying:

```
By default these shortcuts are not bound. To set them, BindShortcuts needs to be called for the portal
              to register the different triggers with the user.
```

This should be enough, AFAIU.

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: Yes, it was indeed a bit weird. Massaged the wording across the document to make it all consistent.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,195 @@
...
+
+        Creates a global shortcuts session.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>shortcuts a(sa{sv})</term>
+            <listitem><para>
+              The different shortcut ids that have been registered. They are mapped to a list of pairs including:
+                - description: User-readable text describing what the shortcut does
```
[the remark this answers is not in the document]
aleixpol: Done.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -0,0 +1,184 @@
...
+        List all the shortcuts with their invocation descriptions.
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>shortcuts a(sa{sv})</term>
+            <listitem><para>
+              The different shortcut ids that have been registered. They are mapped to a list of pairs including:
+                - description: User-readable text describing what the shortcut does
+                - trigger: User-readable text describing how to trigger the shortcut
```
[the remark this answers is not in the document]
aleixpol: I think it makes sense to provide the information always in the same format, so I'd say so. But we can leave it out if you say so.

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: Referring to ListShortcuts now.

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
aleixpol: The unused options argument is on now.

Changed the timestamp to map the wording in the location portal `@timestamp: The timestamp, as seconds and microseconds since the Unix epoch.`

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
aleixpol: I'm not sure what you mean, what would that look like?

aleixpol:
@jadahl sorry there were some comments you included that I'd lost in the immensity of GitHub's UI. I think it's all addressed or commented on further now.

GeorgesStavracas:
@aleixpol i think this PR now needs to update the Meson files too. Shouldn't be hard to add, and won't impact review too much, but I'd appreciate if you could add it too since next xdg-desktop-portal release will be done with Meson :slightly_smiling_face:

aleixpol:
> @aleixpol i think this PR now needs to update the Meson files too. 

That explains the CI exploding, should be addressed

data/org.freedesktop.portal.GlobalShortcuts.xml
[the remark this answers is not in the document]
jadahl: The idea would be to have a <vararg> list in some generic top level like documentation section, and everywhere where the same format is expected to be passed, the documentation will say "See ... for details about the shortcut vararg." or something, so that when we e.g. add a new entry to the dictionary, it'll only be added to one place.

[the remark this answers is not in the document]
aleixpol: Ah right, makes sense. I've put it on the class definition and pointed there. Not sure if there's a better way of doing this.

GeorgesStavracas (approved):
There still are a few things to correct here, but I think it'll be more productive to do that myself :slightly_smiling_face:

aleixpol:
Glad to see it's in, I'd have expected a PR into this branch but I guess that will work anyway. Thanks, please poke me for the polishes.

aleixpol:
> I apologise for being a bit obtuse. Will a shortcut exclusivity option be addressed in x-d-p, or will this be left up to the backend implementations? With shortcut exclusivity I refer to sending the keys pressed to the focused application as well as the input event for the activated shortcut.

This protocol is only for system-wide shortcuts. As a foscussed application you can always do as you wish with the input you are given.

> If there's any public IRC channel to catch up to internal communications that I am unaware of please let me know.

No, all the communication to put this together in this Pull Request.

RichardFevrier:
Would it be possible with this portal for a third party app (KDE Settings/GNOME Settings/other...) to be able to list and manage all global shortcuts from every app using them?

Issue #713, opened 2022-02-23
Question: why don't Session objects support extra interfaces


whot:
This is a design question, not a bug.

The portals that create session follow the same basic approach: a `Foo.CreateSession()` method that returns an object path (via `Request`) and a bunch of methods in the form `Foo.DoSomething(session_handle, args)`.

I'm wondering: why don't the sessions themselves implement the required interfaces instead? So the returned session object implements `org.freedesktop.portal.Session` but also the `Foo`-specific `org.freedesktop.portal.FooSession` which has the `DoSomething(args)` handle.

As a more specific example, the [`ScreenCast` interface](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.ScreenCast) would then have the same `ScreenCast.CreateSession` function, but the returned object would also implement `org.freedesktop.ScreenCastSession`.
And that interface would have 
```
org.freedesktop.portal.ScreenCastSession.SelectSources(IN  a{sv} options,
                                                       OUT o     handle);
```

I understand existing portals won't change, I'm just wondering what the reasoning behind this is (for the implementation of future portals).

jadahl:
I can't think of a reason other than too much C thinking when creating the API. In a very early version, the session was only a ID, not its own object even, accessed from multiple portals (`RemoteDesktop` and `ScreenCast`), but turning it into a generic object didn't go all the way.

Looks as if `Location` and `Inhibit` works more or less the same way. I wonder whether we should just follow this trend "just because" or if it makes sense to break it?

whot:
> if it makes sense to break it?

I think it does because it makes the API more flexible and less likely to be mis-used. You can't pass the wrong session handle into a function because - well, it's implicit in the session object itself. In the case of multiple simultaneous sessions this would also allow for session-specific property values and signals to be sent to one session only. Right now neither of those are easily possible.

whot:
One comment @jadahl mentioned elsewhere - another reason for the session being a generic interface is that in the case of RemoteDesktop and ScreenCast, the session isn't assigned to either type until e.g `SelectSources` is called.

PR #714, opened 2022-02-24
Add a InputCapture portal

whot:
This portal is aimed at allowing applications to capture input from
physical devices, the prime use case is the server component of a
Barrier/Synergy setup where local devices are used to control a remote
(Barrier/Synergy client) display.

The general flow for a `InputCapture` session is:
- application queries the "zones" available (usually representing the desktop)
- application sets up "pointer barriers" at the edges
- application requests a libei-compatible handle for the actual input
  events
- when a cursor moves against/across a pointer barrier, the compositor
  notifies the client and re-routes input events from libinput to libei
  instead of controlling the local cursor

Notable: the actual input events are not part of the portal here,
they're handled by the direct libei connection between compositor and
application.

The compositor is in charge of everything - it can restrict some screens
from showing up in the zones, it can deny pointer barriers at
locations, it decides when a pointer barrier is crossed, it can stop
capturing (or filter events) at any time, etc.

This requires active/passive libei contexts, see
https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/80

- libportal PR: https://github.com/flatpak/libportal/pull/96

cc @jadahl, @ofourdan, @p12tic


data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+    <!--
+        Enable:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @options: Vardict with optional further information
+
+        Enable input capturing. This does not immediately trigger capture, it
+        merely enables the capturing to be triggered at some future point
+        (e.g. by the cursor moving across a barrier). If and when that happens,
+        the Request object triggers the #org.freedesktop.portal.Request::Response
+        signal.
```
jadahl, line 252: Do we really want to communicate the `enabled` signal via a one shot request object? That would make moving the pointer back and forth across a trigger barrier racy, as one might try to attempt to move the pointer back into the intended barrier before the new `Enable()` was called.

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+        that of the 2k display. A cursor movement on the 2k screen covers N
+        pixels but that same physical cursor movement on the 4k display covers
+        2*N pixels. Likewise, a pointer barrier extending to the height of the
+        4k screen needs to be half as tall (in pixels) as that on the 2k screen
+        despite being the same physical size.
+    -->
+    <method name="GetRegions">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a(uuiid)" name="regions" direction="out"/>
```
jadahl, line 156: One idea to add as an out argument is a `serial` that should be passed to `SetPointerBarriers` to make sure the used region source is the most recent one.

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+
+        Applications must ignore unknown capabilities.
+
+        Currently defined types are:
+
+        <simplelist>
+          <member>1: RELATIVE_POINTER</member>
+          <member>2: ABSOLUTE_POINTER</member>
+          <member>4: KEYBOARD</member>
+          <member>8: TOUCH</member>
```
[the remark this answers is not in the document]
whot: For triggers, we have DBus methods (currently `SetPointerBarrier` but I can envision at least a `ActivateNow` type too). Those have return values, so I think it may be good enough to just error out from those and let the application deal with it. If in the future we notice that applications need to know ahead of time whether they can set up a barrier, we can add those, separate to the input capabilities. But right now I think they're not needed - doubly so because the capturing triggers will be very few, so it's likely all implementation will support all of them.

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="capabilities" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        GetRegions:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @options: Vardict with optional further information
+        @regions: An array of 4 integers, each specifying a region's width, height, x/y offset and physical scale
```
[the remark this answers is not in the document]
whot: as a general rule, the devices exposed in the CaptureInput portal will more closely match the **physical** input devices, not the logical ones. And those don't have that physical scale, that one only exists on screen.

And since we expect the pointer barriers to be set up on screen, I think we need the physical scale as part of the region.

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+        and the application should re-request the current regions to update its internal state.
+
+        Note that regions are session-specific, there is no guarantee that two
+        applications see the same screen regions. An empty region list implies
+        that no pointer barriers can be set.
+
+        The physical scale by default is 1.0, or the proportion of
+        logical:physical size of the screen. This is required on systems that
+        do not have support for high-resolution displays but employ a mix of
+        high- and low-resolution displays.
```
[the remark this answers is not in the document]
whot: I'll reword to something like "if unsure what to pass, use 1.0" which was the intended meaning

data/org.freedesktop.portal.CaptureInput.xml
whot: What I *want* to do is simply make this a signal on the session object like #713 :smile: And everything else, come to think of it. Failing that, keeping the request open until it triggered seemed like the best approach given the current infrastructure. More documentation is needed in that cancelling that request is akin to `Disable()` but otherwise it works all fine in my head.

I'm not sure on the raciness:
- application calls `Enable()` with the `Request` object
- pointer moves across the barrier, compositor starts re-routing events
- events start flowing on libei | `Request` object succeeds (in either order, can't do much about that bit)
- application calls `Disable()` | compositor decides its enough (in either order)
- compositor stops capturing, `Disabled` signal

Because it's input capture, there's no "moving back" once the compositor triggered, the input events are re-routed and the cursor doesn't move. This is a bit hairier when you're just capturing keyboard events (or a subset of pointer devices) based on pointer motion but even then the above works - once capturing starts, the event flow starts.

whot:
I'm aware of #711, so far my best idea is to have a specific method `SetKeyboardShortcut(shortcut_id)`, with the `shortcut_id` being the identifier to bind the two portals together. I'm *not* planning to re-implement keyboard shortcuts in CaptureInput, I don't think that would work well.

data/org.freedesktop.portal.CaptureInput.xml
jadahl: > What I want to do is simply make this a signal on the session object like https://github.com/flatpak/xdg-desktop-portal/issues/713 smile

The signal could have the session handle as a parameter.

Anyhow, what I was thinking of is that when the capturing is disabled, there will be an amount of time where a "recapture" isn't possible because it hasn't been explicitly re-enabled. This can be avoided by adding a `Enabled` signal, and making `Enable()` create a session that can be enabled/disabled any amount of times. Maybe it'd be more clear what I was thinking if one renames things:
```
<method name="Enable">...</method>
<method name="Disable">...</method>
<method name="Release">...<arg name="position" type="(ii)"/></method>
<signal name="Activated">...</signal>
<signal name="Deactivated">...</signal>
```
So the flow would be

1. CreatePointerConstraints
1. Enable
...
1. Activated (events coming)
1. Release(x,y)
1. Deactivated (no more events)
...
1. Activated (events coming)
1. Deactivated (no more events)

Looking at that, the API becomes bigger, so maybe that's not good. What raises my eye brows is that I'm not sure requests have been used in this way before; they tend to be used to relatively short lived and represent the user making a decision, not a async "enable" request that might be responded to 2 hours later. As in, the request to enable enable the capture was already granted more or less immediately, it's just that it hasn't triggered yet, which makes it feel a bit odd.

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
jadahl: Shouldn't the those properties (physical vs logical movement) be handled by libei events instead? Or do you think it's better to leave out any meaning of logical pixels out of the libei API? (what would need to be added would be a `get_scale()` or something to `ei_region`.

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
jadahl: I'm thinking of when the portal backend should present a dialog to the user. At this point, it should be known what the application wants, both in terms of possilbe triggers and the type of input devices. Doing it at `Enable()` is probably too late, since `ConnectToEIS()` would be called prior to that, which wouldn't be possible since it's not known what kind of permissions there will be. Doing it at `SetPointerBarrier()` wouldn't be good either, since that means adding another delayed trigger would require yet another dialog. My point is that at some point before any `ConnectToEIS()` or `Enable()`, the whole intention of the application must be known, so that a complete picture of its needs can be presented to the user.

jadahl:
> I'm aware of #711, so far my best idea is to have a specific method `SetKeyboardShortcut(shortcut_id)`, with the `shortcut_id` being the identifier to bind the two portals together. I'm _not_ planning to re-implement keyboard shortcuts in CaptureInput, I don't think that would work well.

I think the question is more about input routing, not about triggering. When input is captured by an application, does global shortcuts registered by #711 gets triggered or not? Should this be possible to tweak with some property?

For the intended primary use case (Synergy/Barrier/whatever the new one is called), I'd say captured input should not trigger keyboard shortcuts, as input being captured really should be seen as being designated for another computer, but are there use cases where this isn't desirable?

whot:
> but are there use cases where this isn't desirable?

yes, if we (theoretically) switch qemu to use libei instead of wl_relative_pointer that would be a prime use-case for this: you'd still want ctrl+alt to escape the pointer capture. I haven't added to that effect yet beyond the documentation that the compositor can filter any event because without more specific use-cases it's hard to get that API right. And much of it relies on the individual use-cases anyway, e.g. you *may* want to not forward volume keys in a synergy setup because the music is playing on the server, not the client. Or maybe it's playing on the client and you *do* want to forward them.

With the `shortcut_id` from #711 it would be possible to negotiate a set of either explicitly prevented or explicitly forwarded shortcuts. My current thought though is that may be best to not provide any specific solution for this and wait for real-world use-cases to run into those issues, so we have real problems to solve.

data/org.freedesktop.portal.CaptureInput.xml
whot: I wondered if a "oneshot" or "re-arm" arguments could help here. But no matter how I type it out, it comes down to the same thing: we need `Enable()` and `Disable()` and a `Disabled` signal. And we do need the `Activated/Deactivated` signals, I can't find a way to make the API smaller. So despite several attempts, I'm coming up with the same flow as you did above.

If (mis)using the Request like I did above is not ideal, I'm happy to change it to the "normal" flow that relies on signals.

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
whot: Yep, we don't disagree. I'm just suggesting to leave this off for now and work it into the various `options a{sv}` and `result a{sv}` if we end up having a specific use-case for it. Probably at `CreateSession()` time, on the assumption that if we allow an application to set up pointer barrier and it ends up setting up the wrong one that's a bug rather than a sandbox violation.

As for "when to present the dialog", I'd say at `CreateSession` time. That currently already has the capabilities field so "An application wants to capture pointer input" is possible there. `SetPointerBarriers` uses a Request and has a return value for failed barriers, so an extra dialog could be inserted here, if that's really needed.

I think (and that needs to be spelled out in the docs) that the EIS implementation needs integrate with the `impl.portal` correctly anyway, because `ConnectToEis()` is only called once but if regions change and/or the client set up invalid pointer barriers later, the fd is still there even though the client must no longer receive events.

jadahl:
There would be two levels of breaking the capture in the qemu/virt-manager case wouldn't there? As in a application level and compositor level. A compositor level escape hatch would always be possible (as it is with the "inhibit keyboard shortcuts" Wayland protocol), but the virt-manager level escape hatch is controlled by virt-manager, which would see all the events before they reach qemu I suppose?

data/org.freedesktop.portal.CaptureInput.xml
jadahl: My gut feeling is that `Activated/Deactivated` is better than a very long lived request object. It'd make this API less "unique" which might be a good thing.

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
jadahl: For the sake of pointer barriers alone, I still don't see how any scale is relevant. The only thing that matters is the screen edges in a logical pixel grid that represents the layout the compositor uses.

Sometimes the logical pixel grid matches the physical in some sub-regions, sometimes it doesn't; sometimes the region is "HiDPI", but it is so somewhat independently of the scale here, as in, a scale == 1.0 if the scale is to represent the logical - physical pixel grid relationship doesn't mean it isn't a HiDPI monitor and it doesn't mean raw input isn't in some way transformed.

What we really are after, isn't that always in what way *relative* input events are transformed? For example lets say we have two identical HiDPI setups:

  * A) 4K monitor, represented by a 2K region; logical - physical pixel scale is `2`, clients draw with scale `2`
  * B) 4K monitor, represented by a 4K region; logical - physical pixel scale is `1`, clients draw with scale `2`

In both A) and B) the displayed output is completely identical, both show a HiDPI client exactly the same way, pixel by pixel, but the way input events are represented in libei would differ. The same applies to physically moving a pointer devices a given distance - it will result in the same distance traveled by the pointer sprite on screen.

In A) moving a relative pointer device `(dx, dy)` units after pointer acceleration etc moves the pointer `(dx, dy)` logical pixels in the grid. For something looking at absolute events it'll see `(x, y)` and then `(x', y')`, and to get the original device delta, it just needs to do `dx = x' - x` and `dy = y' - y`.

For A) moving a relative pointer device `(dx, dy)` units after pointer acceleration etc moves the pointer `(dx * 2, dy * 2)` logical pixels in the grid, since the logical - physical pixel scale is `2`. Thus when looking at libei events, a relative pointer should still send `(dx, dy)`, but absolute events would see first `(x, y)` then `(x + dx * 2, y + dy * 2)`. If something that looks only at absolute events sees `(x, y)` then `(x', y')`, in order to get the original input event it will need to know the scale the compositor used to boost the pointer movement (lets call it `s`), thus to get the original `(dx, dy)` it will need to do `dx = (x' - x) / s` and `dy = (y' - y) / s`.

To come up with some kind of conclusion of what I'm thinking:

* For pointer barriers alone, I don't think scale is needed anywhere at all.

* In order to be able to get the original post-processed relative input event from two absolute events, the scale used to scale input events is needed - not the one related to the pixel grid.

* The only time a physical - logical scale is needed is to know how far a relative event traveled in physical space, but I'm not sure I understand when this is actually needed when capturing input?

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
whot: obsolete because we're removing the scale

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+
+        For example, two 24 inch monitors, one 2k and one 4k are the same
+        physical size but the resolution on the 4k display is double
+        that of the 2k display. A cursor movement on the 2k screen covers N
+        pixels but that same physical cursor movement on the 4k display covers
+        2*N pixels. Likewise, a pointer barrier extending to the height of the
+        4k screen needs to be half as tall (in pixels) as that on the 2k screen
+        despite being the same physical size.
```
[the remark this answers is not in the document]
whot: obsolete because we're removing the scale

data/org.freedesktop.portal.CaptureInput.xml
whot: done locally, will be in the next push

data/org.freedesktop.portal.CaptureInput.xml
whot: done locally now, will be in the next push

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
whot: First - agreed, scale for pointer barriers appears to not be necessary.

For the events, there are missing bits, I tried to type out the four cases that apply:

## Setup
Server-side: 
- we have a 4k screen with logical-physical `2` on synergy server
  - technically the client-scale doesn't matter from an input perspective so we can leave it out here
- we have a touchscreen and a mouse
- synergy server connects to portal and receives `CaptureInput` region `(2560x1440@0,0)`, scale `1.0`                                                                                                                                                                             
  - scale here can be `1` because the client-scale is completely transparent to anyone but the compositor, those pixels might as well not exist and we effectively have a 2k screen
- synergy server sets up a pointer barrier at `x=2560, y=[0-1440]`
- compositor creates EIS device capability `rel` with **no input region**
- compositor creates EIS device capability `abs` with an input region of `(500x312@0,0)` in **mmm**                                                                                                                                                                             
  - **Missing:** there is no connection between the portal region and the EIS region

And on the synergy-client side:
- we have a 4k screen with logical-physical `1`                                                                                                                                                                           
- synergy-client has an EIS region of `(1920x1080@0,0)` with scale `2.6`
  - contrived example to make this explanation more complex. 
 
## rel → rel event
- mouse/libinput generate a relative motion of `(dx, dy)`                                                                                                                                                                                                                       
- compositor converts this to `(2dx, 2dy)` *physical* pixels movement on-screen                                                                                                                                                                                                 
- input capture activates                                                                                                                                                                                                                                                       
- mouse/libinput generate a relative motion event of `(dx, dy)`                                                                                                                                                                                                                 
- compositor passes this as `(dx, dy)` to the libei context on synergy-server                                                                                                                                                                                                   
- synergy-server sends `(dx,dy)` as-is to synergy-client                                                                                                                                                                                                                     
- synergy-client converts to local delta `(dx/scale, dy/scale)` (where `scale == 2.6`) and passes that to EIS                                                                                                                                                                                           
- remote compositor moves cursor by that delta

## abs → abs event                                                                                                                                                                                                                                                             
- touchscreen generates absolute motion to `(x, y)`                                                                                                                                                                                                                             
- compositor converts this to the matching screen position using `libinput_event_touch_get_x_transformed()` in the 4k range                                                                                                                                                     
- input capture activates                                                                                                                                                                                                                                                       
- touchscreen generates absolute motion event to `(x, y)`
- compositor passes these as `(xmm, ymm)` in `mm` to the libei context on synergy-server                                                                                                                                                                     
- synergy-server converts this to abs on the client by converting from `500x312` in mm to the `1920x1080` in logical pixels range and sends `(ax, ay)` to the client                                                                                                                                                                                                                                            
- synergy-client libei sends abs `(ax, ay)` to the EIS implementation                                                                                                                                                                                                           
- remote compositor converts `(ax, ay)` (in the 1920 range) to the right position in the 4k range
 
## rel → abs event                                                                                                                                                                                                                                                             
- mouse/libinput generate a relative motion of `(dx, dy)`                                                                                                                                                                                                                       
- compositor converts this to `(2dx, 2dy)` *physical* pixels movement on-screen                                                                                                                                                                                                 
- input capture activates                                                                                                                                                                                                                                                       
- mouse/libinput generate a relative motion event of `(dx, dy)`                                                                                                                                                                                                               
- compositor passes this as `(dx, dy)` to the libei context on synergy-server                                                                                                                                                                                                   
- synergy-server converts sends `(dx,dy)` on as-is to synergy-client
  - technically synergy-server does everything here too but that's irrelevant to this example
- synergy-client applies local scale and converts the delta `(dx/scale, dy/scale)` to absolute position `(ax, ay)` in that region                                                                                                                                                                       
- synergy-client libei sends abs `(ax, ay)` to the EIS implementation                                                                                                                                                                                                           
- compositor converts `(ax, ay)` to the right position on the 4k screen  (mapping 1920x1080 to the 4k actual range)  

## abs → rel event
- touchscreen generates absolute motion to `(x, y)`                                                                                                                                                                                                                             
- compositor converts this to the matching screen position using `libinput_event_touch_get_x_transformed()` in the 4k range                                                                                                                                                     
- input capture activates                                                                                                                                                                                                                                                       
- touchscreen generates absolute motion event to `(x, y)`, then `(x', y')`                                                                                                                                                                                                      
- compositor passes these as `(xmm, ymm)` and `(x'mm, y'mm)` in `mm` to the libei context on synergy-server                                                                                                                                                                     
- synergy-server calculates delta as `(dxmm, dymm)`                                                                                                                                                                                                                             
  - **Missing**: we cannot reliably convert abs motion in mm to rel motion in logical pixels with the current data.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
- synergy-server converts this to delta on the client by calculating factor `f` to convert from `500x312` to the `1920x1080` range and applying that conversion to `(dxmm * f, dymm * f)` and sending that as relative event to synergy-client                                                                                                                                                                                                
- synergy-client applies local scale `2.6` on the incoming delta and passes it on to EIS                                                                                                                                                                                       
- compositor moves cursor by that delta

## Summary

So, the missing bits in the above are:
We cannot convert from abs motion in mm to rel motion in logical pixels, and the reason we can't is that there is no association between a physical region and the logical region it may be mapped to. Or like in the example above: there's no hint that the 500x312mm touchscreen corresponds to the 2k region on the host.

This **could** be a problem, like in the `abs → rel` case. If we had an association between the two, we could map the mm to pixels.

The other bit that is missing is that the size of pixels differs. All the above treats logical pixels as "reachable pixel", but in the case of a 4k screen and a VGA screen, the same pixel movement will be significantly different physical speed. Maybe a scale < 1.0 could help there, haven't thought through that yet.

whot:
Summary after a meeting I had with @jadahl last night, putting it here so it doesn't end up hidden in a resolved conversation. This was about https://github.com/flatpak/xdg-desktop-portal/pull/714#discussion_r823371424 and which other bits we need. In conclusion:
- we are going to rename the portal "Regions" into "Zones" to avoid any confusion with the libei Region. They look too similar despite being independent that we want to disambiguate them properly. Zones isn't a great name but that's what the thesaurus suggested.
- libei is going to add the API to check whether a device is a physical device. If so, it will have a size in mm (width + height but no offset). 
  - as a general rule, an active libei context (xdotool, synergy-client, any emulating application) uses logical devices, a passive libei context (synergy-server, any capturing application) uses physical devices
- libei will update so that the input event data it sends is either in logical pixels (for logical devices) or mm (for physical devices)
- libei logical devices will continue to have regions, they will not have a size. A libei physical device will not have any ei Regions.
- if an absolute input device is mapped to a specific screen and its input is to be captured, there is no hint in the portal or libei about the mapping - the data is simply in mm relative to the device's top-left corner. The mapping is left to the application to figure out.

whot:
Renamed to `InputCapture` because that sounds less awkward. Suggested by @jadahl (blame him, not me! he even *likes* pitchforks)

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,547 @@
...
+          <varlistentry>
+            <term>direction i</term>
+            <listitem><para>
+                The directional movement to trigger the pointer barrier. A
+                negative integer implies the pointer barrier triggers on negative
+                movement across the pointer barrier (e.g. "up"
+                for a horizontal barrier), a positive integer implies the
+                pointer barrier triggers on positive movement across
+                the pointer barrier (e.g. "right" for a vertical barrier).
+                A direction of zero triggers the barrier in either direction.
```
jadahl, line 260: Should we make this a "enum" i.e. a `u` where we define `positive` and `negative`? Alt. limit it to `-1` and `1` so that `-42` and `123` aren't valid directions.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,547 @@
...
+        Note that unlike the X11 XFixes PointerBarriers, these pointer barriers do not
+        block movement of the cursor.
+
+        Pointer barriers are situated on the top or left of their respective pixels and a
+        pointer barrier must be within or at the extent of the available
+        zone. For example, a zone of 1920x1080 at offset 0,0 allows
+        a pointer barrier at x1=1920, y1=0, x2=1920, y2=1080.
+        A pointer barrier at the extent of a zone is considered triggered
+        when the pointer would logically move off that zone, even if the
+        actual cusor movement is clipped to the zone.
```
jadahl, line 210: Should we limit these barriers on the outer bounds of all the zones? Or should it be possible to create a barrier in between two adjecent zones?

Also what about barriers right in the middle? Sounds like something that shouldn't be possible, since the application cannot possibly know what is there.

Also, should it be allowed that one barrier can "touch" multiple zones? I.e. if you have

```
+---+---+
| 1 | 2 |
+---+---+
A   B   C
```
Can a barrier be created between A and C, or should it be two barriers, one between A and B, and another between B and C?

whot: Agree with the outer-bound zones-only, that's (at least for version 1) a sensible approach.

And I think requiring A-B and B-C separately makes sense. There's a use-case for having them separate anyway (synergy may have two different clients "above" 1 and 2) and it makes the checks a bit simpler when 1 and 2 are of different sizes/offsets.

data/org.freedesktop.portal.InputCapture.xml
whot: good point, will amend

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,547 @@
...
+                around. Implementations of this portal must to increase the
+                serial number by an sensible amount to allow for wrapping detection.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>cursor_position ii</term>
+            <listitem><para>
+              The current cursor position. Note that this position may be
+              outside the Zones available to this session - this indicates
+              movement larger than a single pixel.
```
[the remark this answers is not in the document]
whot: The latter, so you may get 1924x1100 on a 1920x1080 region if you moved "out" of the bottom right corner. Will amend, thanks.

data/org.freedesktop.portal.InputCapture.xml
jadahl: On the other hand, with barriers only being allowed on the outer edges of all the zones, the direction isn't needed as only one direction makes any sense.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,547 @@
...
+              It is best to view this set as a negative confirmation - a
+              capability that was requested but is missing is an indication that
+              this application may not capture events of that capability.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <method name="CreateSession">
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
```
jadahl, line 123: We're expecting to show the dialog in response to this method being called, right? Since When we have the session, the app will be able to call things like GetZones() etc without any extra nagging.

With that assumption, we need a `parent_window` input argument, i.e. a
```
        @parent_window: Identifier for the application window, see <link linkend="parent_window">Common Conventions</link>
```

and 

```xml
```

data/org.freedesktop.portal.InputCapture.xml
whot: done in the latest updates, direction was removed as well

data/org.freedesktop.portal.InputCapture.xml
whot: done

data/org.freedesktop.portal.CaptureInput.xml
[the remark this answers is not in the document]
whot: documentation updated

data/org.freedesktop.portal.InputCapture.xml
whot: done!

data/org.freedesktop.portal.CaptureInput.xml
```diff
@@ -0,0 +1,406 @@
...
+        </variablelist>
+
+        The following results get returned via the #org.freedesktop.portal.Request::Response signal:
+        <variablelist>
+          <varlistentry>
+            <term>serial u</term>
+            <listitem><para>
+                A serial number that can be used to synchronize with any
+                side-channels. This serial number has no intrinsic meaning but
+                is guaranteed to increase by an unspecified amount on each call.
```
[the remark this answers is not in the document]
whot: i've added comments for this now

whot:
We now have an implementation, and it seems to work too! (according to the pytest test suite that is)

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+        @options: Vardict with optional further information
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Release any ongoing input capture. No
+        #org.freedesktop.portal.InputCapture::Deactivated signal will be emitted.
+
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>cursor_position ii</term>
```
jadahl, line 350: Maybe make this `(dd)` to not loose precision.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+                This allows an application to have a synchronization point and
+                attribute an event sequence to the portal interaction.
+
+                Applications must be able to handle the serial number wrapping
+                around. Implementations of this portal must to increase the
+                serial number by an sensible amount to allow for wrapping detection.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>cursor_position ii</term>
```
jadahl, line 441: Maybe make this (dd) to not loose precision.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+        A pointer barrier is considered triggered when the pointer would
+        logically move off that zone, even if the actual cusor movement is
+        clipped to the zone.
+
+        A zero-sized array of pointer barriers removes all existing pointer barriers
+        for this session. Setting pointer barriers immediately suspends the
+        current session and the application must call Enable() after this method.
+
+        The @serial must be equivalent to the last returned serial number of the
+        org.freedesktop.portal.CpatureInput.GetZones() method. A serial mismatch
```
jadahl, line 232: CpatureInput -> CaptureInput

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+        Set up zero or more pointer barriers. Pointer barriers are horizontal
+        or vertical lines that should trigger the start of input capture when the cursor moves
+        across the pointer barrier. In the future,
+        when the compositor has deemed the pointer barrier to be crossed, it may
+        complete the org.freedesktop.portal.InputCapture.Enable() request and input events
+        (from compositor-determined input devices) are sent to the application
+        via the transport layer.
+
+        Note that unlike the X11 XFixes PointerBarriers, these pointer barriers do not
+        block movement of the cursor.
```
jadahl, line 204: This sentence is no longer relevant, since barriers are always on the edge of the screen.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+        @options: Vardict with optional further information
+        @barriers: An array of vardicts, each specifying one barrier
+        @serial: A unique serial number referring to the zone  set
+        @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+        Set up zero or more pointer barriers. Pointer barriers are horizontal
+        or vertical lines that should trigger the start of input capture when the cursor moves
+        across the pointer barrier. In the future,
+        when the compositor has deemed the pointer barrier to be crossed, it may
+        complete the org.freedesktop.portal.InputCapture.Enable() request and input events
```
jadahl, line 199: We're using normal signals instead of long-lived method calls now.

data/org.freedesktop.portal.InputCapture.xml
whot: done (for all occurances)

data/org.freedesktop.portal.InputCapture.xml
whot: fixed

data/org.freedesktop.portal.InputCapture.xml
whot: Updated

jadahl:
Draft backend implementation in xdg-desktop-portal-gnome: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/33

It lacks libei integration, but can create barriers, get zones etc.

garnacho:
Looking at the DBus interface and implementation this makes sense to me. A slight concern I have is about the fine grained capabilities that are captured in `CreateSession`.

On one hand, the behavior of the compositor with non-captured capabilities is unspecified (and probably could be called "implementation detail" here), but leaving it like that is subject to many possible interpretations and grades of complexity in handling it.

On the other hand, there are also combinations that are more directly awkward to handle, like the behavior of pointers when they enter a barrier but pointers are not in the captured capability set.

From the Mutter perspective I've been looking to reduce the ways things can grab input from each other (and adding ways to notify each other). I would prefer to avoid the complexities that arise from adding per-device/capability granularity to input capturing.

I wonder if it would be possible to remove or make the requested capabilities optional, defaulting to "give me all events I could care", or perhaps that's been the way to interpret those all along?

whot:
> I wonder if it would be possible to remove or make the requested capabilities optional, defaulting to "give me all events I could care", or perhaps that's been the way to interpret those all along?

Just for the record, we can differ between **capturing** (portal) and **transmitting** (libei) events. In the case of synergy, we cannot **transmit** touch events, they're not supported by synergy. But we could still **capture** touch events (and discard them) once the pointer barrier is triggered.

I'm not sure that's the best user-visible behaviour tough. I'm thinking of the use-case of e.g. forwarding a tablet device. This should probably work like USB forwarding, i.e. per-device only. The tablet device generally makes the pointer barriers more "interesting" - a tablet in relative mode can hit the pointer barrier without a pointer capability being necessary anywhere. Especially in mutter where tablets control separate cursors.

But since we don't have tablet support right now, removing the capabilities in `CreateSession` and defaulting to "capture everything" is doable, yes.

garnacho:
> Just for the record, we can differ between **capturing** (portal) and **transmitting** (libei) events. In the case of synergy, we cannot **transmit** touch events, they're not supported by synergy. But we could still **capture** touch events (and discard them) once the pointer barrier is triggered.

Aha, so the capabilities reported here about capturing and compositor/libei can still think otherwise about the events sent.

> 
> I'm not sure that's the best user-visible behaviour tough. I'm thinking of the use-case of e.g. forwarding a tablet device. This should probably work like USB forwarding, i.e. per-device only. The tablet device generally makes the pointer barriers more "interesting" - a tablet in relative mode can hit the pointer barrier without a pointer capability being necessary anywhere. Especially in mutter where tablets control separate cursors.

I didn't get there as there's the overall TABLET capability AFAICS, but indeed, multiple pointers make it all more fun :). If it is possible to have different pointing devices trigger multiple barriers capturing input towards different computers, questions like "what happens if all request KEYBOARD capability" start to pop up.

For the specific usecase of tablets, I tend to see more sense in a "switch host" pad button action like there is for cycling displays.

> 
> But since we don't have tablet support right now, removing the capabilities in `CreateSession` and defaulting to "capture everything" is doable, yes.

Cool, time would tell if it's simple or simplistic :), could always be extended at some point later on.

whot:
> Aha, so the capabilities reported here about capturing and compositor/libei can still think otherwise about the events sent.

correct. the libei events must be a subset of the capabilities in the portal, but otherwise they're independent.

libei also has the concept of seats (decided by the compositor) which cater for the multi-pointer scenario. This portal does **not** handle the case where you want to capture only one of the seats, pointer barriers apply to all (provided the compositor agrees of course). libei however splits devices by seat so you can separate the event stream per seat.

as for the tablet capability: I'm somewhat planning to add this to libei together with gestures and whatever else libinput does already (in a similar way anyway). since libei is supposed to be a generic input emulation layer, it should provide the various devices we know of already. it's not implemented anywhere but think of this as libinput-over-the-wire and that's fairly close.

marler8997:
Dropped by to describe a use case I have for this.

I work on a remote sharing application that emulates multiple cursors.  One user is the owner and controls the "real cursor" and the other cursors are "fake" that just are rendered in addition to the real cursor.  The problem occurs when a remote user owns the real cursor and the local user moves their mouse around.  The X server will continue to listen to the local mouse and move the cursor around.  This can be addressed by telling the X server to stop listening for events from the local mouse input devices, however, our application still needs those events to control a fake cursor and retake control when the local user clicks a mouse button.  So when this occurs, our application needs to listen for mouse input outside of the compositor since we had to disable it.  This is where we could make use of the proposed CaptureInput portal.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,549 @@
...
+              The current cursor position. Note that this position may be
+              outside the Zones available to this session - this indicates
+              movement larger than a single pixel.
+
+              For example, a fast movement against a barrier on the right edge
+              of a screen may logically put the cursor dozens of pixels into
+              the (non-existing) screen on the other side of the barrier.
+              It is the application's responsibility to adjust the cursor
+              position as necessary.
+            </para></listitem>
```
whot, line 505: @jadahl  - this is clearly copy/paste but I'm wondering whether there's use-case for including the cursor position on the `Deactivated` signal. Seems superfluous since the only thing the client can do is paint the cursor where it left, and it's unlikely to stay there anyway.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,549 @@
...
+        emitted.
+
+        Input events will not be captured until a subsequent
+        org.freedesktop.portal.InputCapture.Enable() call.
+    -->
+    <method name="Disable">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
```
whot, line 333: having implemented this in libportal - I'm questioning the need for `Disable` and `Release` to be a `Request`. I mostly did it for consistency here, and `Enable` has *some* use-case for delayed requests though this could be worked around by making that too a simple method and have the client rely on `Disabled` instead.

data/org.freedesktop.portal.InputCapture.xml
whot: Removed in the latest version

data/org.freedesktop.portal.InputCapture.xml
whot: Switched to normal method calls in the latest version .`Enable` too, see the commit message.

whot (commented):
I've rebased this (much has happened, including meson support) and squashed it together into one commit to make this a bit less unwieldly. This is a minor API break again due to renaming for `zones_id`.

It's now also based on #868.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,529 @@
...
+        <variablelist>
+          <varlistentry>
+            <term>zones a(uuii)</term>
+            <listitem><para>
+              An array of retions, each specifying that zone's width,
+              height, x/y offset.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>serial u</term>
```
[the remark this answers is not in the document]
whot: `s/zones_id/zone_set/` which is less terrible...

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -0,0 +1,431 @@
...
+        @options: Vardict with optional further information
+
+        The ZonesChanged signal is emitted when one or more of the zones
+        available **to this session** change. An application should immediately call
+        org.freedesktop.portal.InputCapture.GetZones() to update its state of the zones.
+    -->
+    <signal name="ZonesChanged">
+      <arg type="o" name="session_handle" direction="out"/>
+      <arg type="a{sv}" name="options" direction="out"/>
+    </signal>
```
whot, line 408: This signal should include the `zone_set` number as used in the `SetPointerBarrier` request, so applications can be sure of which zones are invalidated. Needs a fair bit of spec comments because (for easy of implementation) the zone_set should be associated with zones but if an application never calls `GetZones` it may never see a particular number invalidated later. Plus the theoretical race condition of calling `GetZones` while the signal is already in-flight, so the application gets the new (valid) `zone_set` returned together with the signal for the old one.

The difficult bit here is simply writing the spec :)

marler8997:
@whot we spoke a bit about my use case a while back on IRC. I've created a privileged daemon that uses EVIOCGRAB, which I believe is what you recommended (hope I'm remembering that correctly). Will this input portal allow something equivalent to EVIOCGRAB? Was wondering if we'll be able to use this in the future instead of going through a custom daemon.

whot:
> Will this input portal allow something equivalent to EVIOCGRAB?

Two comments here:
First, with this portal, an application can *capture* input, even when sandboxed. IIRC your use-case requires *replaying* input, which would be #762.

Second, EVIOCGRAB or not is orthogonal to either portal (or outside of scope, depending on viewpoint). Once you set up a EIS connection through either portal, the event flow goes via the [libei](https://gitlab.freedesktop.org/libinput/libei/) socket to your application or from your application in the case of #762. libei removes the question of *how* the events are generated, that is up to you or the compositor.

Your application could open an evdev device, EVIOCGRAB it and then use #762 to feed the events into the compositor without having to create a uinput device. You still need the permissions on the source devices though to grab it, so it only solves half the problem. That could possibly be solved with a potential evdev or device portal.

Your application could request key events to be captured via this portal and then replayed via #762, which would complete the circle. But this portal doesn't currently provide permanent keyboard capture and it definitely doesn't provide per-keyboard capture. So while theoretically possible, it's definitely not in the current patchset and I'm not sure that type of fine-grained handling is wanted.

marler8997:
> First, with this portal, an application can capture input, even when sandboxed. IIRC your use-case requires replaying input, which would be #762 

It's true that our application needs to simulate input from remote users, however, we are currently using XTestEvent for that on X11-based distros which works for the time being (we'll see what happens when we start supporting Wayland).  The problem we solve with our own privileged daemon was figuring out how to intercept input events before they get to the compositor.  The reason for doing this is we allow the remote user to control the real cursor and while this is happeneing, we don't want the local mouse to get to the compositor and have it move the cursor, instead, we render a "fake cursor" to the screen during this time and move it according to the intercepted input.

The daemon's full source code is here for reference: https://github.com/tupleapp/tuple-launch/blob/master/flatpak-launch.zig#L189

> Your application could open an evdev device, EVIOCGRAB it and then use https://github.com/flatpak/xdg-desktop-portal/pull/762 to feed the events into the compositor without having to create a uinput device.

Yes that's what our privileged daemon does.  The application can request it to open "mouse devices" which it will send back to the application over a unix socket, at which point the application will call EVIOCGRAB when a remote user takes control of the real cursor.

It sounds like this new input portal won't be able to support EVIOCGRAB since it uses "libei" and I'm guessing it's not using evdev and may not have access to this ioctl?

whot:
> It sounds like this new input portal won't be able to support EVIOCGRAB since it uses "libei" and I'm guessing it's not using evdev and may not have access to this ioctl?

libei plus the portals are a replacement for XTEST, not for uinput. Think of it that way, then it'll make more sense.

marler8997:
> libei plus the portals are a replacement for XTEST, not for uinput. Think of it that way, then it'll make more sense.

It still feels like there's a misunderstanding here.  My understanding is that XTEST and uinput are for providing virtual inputs to the system.  We already have that problem solved by using XTEST.  Our problem is how to capture and intercept inputs, not provide them.

marler8997:
We're on the same page with what XTEST/UInput/libei are.  I *think* I see where the misunderstanding is here.  Our use case is not one where we are trying to control a "remote machine".  We are making our screen available to another host.

Our use case is that while a remote user is controlling our mouse, we don't want the "local mouse device" to move real cursor.  Instead, we want to intercept the input from the "local mouse device" and have it control a "fake cursor" that we render to the local screen.  There's no "remote desktop" here, it's all on our own desktop.

whot:
Oh, right. So in that case - what would be needed here is a new trigger for the CaptureInput portal so you can effectively capture input on request by the application. Once input capturing starts, the compositor would forward events to you instead of controlling the normal cursor and you can do with them what you want. That's definitely within scope of this portal, it's just not there at this point.

At the same time you'd be using the RemoteDesktop portal for the remote machine to control your local cursor (instead of uinput). This makes it a bit tricky because you don't want the remote input to go straight through capturing back to you. So there needs to be *something* *somewhere* to exclude remote desktop input from being re-captured. I'm not sure that's a portal API though, this seems much like compositor policy? I don't know.

whot:
> Having some way of preventing the input from ever come back to us like you suggest would be even better.

Since this is all controlled by the compositor, it's fairly easy to cut out that feedback loop there - just don't forward RemoteDesktop input into CaptureInput. But that is a compositor decision/implementation, it cannot be controlled by the application directly.

> if the InputCapture portal were to add the features you're suggesting, I should be expecting support from Wayland-only distros correct?

These portals will never be supported by Xorg, simply because there's no-one working on that anymore and it's unlikely it'll ever see a new release anyway. So it's Wayland only, and there too it'll be compositor-specific (like anything Wayland).

Conan-Kudo:
What's keeping this pull request in draft status? It seems pretty complete at this point...

whot:
It all hinges on libei which hasn't had a stable release yet because the protocol is not stable. That's being worked on, but meanwhile we keep everything else in draft, just in case.

orowith2os:
Is there an EmulatedInput equivalent for this PR? Would that be the changes to the RemoteDesktop portal? 

Also, I'm not familiar with these portals, but could I request access to and from keyboards, cursors, or screens? I'm wondering if I could utilize this for some form of global keybinds, but for apps that want to view all input. 

I imagine the InputCapture portal would work for the keybinds (viewing) part, and RemoteDesktop would be for emulating input (say a macro), right? Could I request access to *only* the input and output types I need?

I brought this up on the Matrix, effectively duplicating this and the accessibility request on Android, so I'm quite interested in it. 😋

orowith2os:
> It all hinges on libei which hasn't had a stable release yet because the protocol is not stable.

If you follow semver, could you not make a breaking change on every minor release (0.x -> 0.y) until 1.0? Could be useful to make a usable release for this portal, so that it gets some real world testing.

albsch:
> > It all hinges on libei which hasn't had a stable release yet because the protocol is not stable.
> 
> If you follow semver, could you not make a breaking change on every minor release (0.x -> 0.y) until 1.0? Could be useful to make a usable release for this portal, so that it gets some real world testing.

That's what semver is for, no?

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

whot:
ftr, libei 1.0.0RC1 was released last week and the API is now stable.

however, this PR was updated to remove any dependencies on libei itself - just like #762 there is no libei-specific code here anymore, it's just a plain fd that gets passed around.

Moving out of Draft, I think this is ready

jadahl (commented):
Seems I have 3 comments that I didn't know of, lets see what they are (can't see in github UI).

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -0,0 +1,552 @@
...
+          <varlistentry>
+            <term>barrier_id u</term>
+            <listitem><para>
+              The non-zero id of this barrier. This id is used in the
+              #org.freedesktop.portal.InputCapture::Activated signal to inform
+              which barrier triggered input capture.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>position iiii</term>
```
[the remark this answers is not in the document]
whot: good point, same for the `cursor_position dd` -> `cursor_position (dd)` then?

Conan-Kudo:
Is there anything else left for merging this in? It seems like all the feedback has been addressed. I'm trying to understand where this is so that I can ask the KDE folks to look to implement this in `xdg-desktop-portal-kde`.

src/input-capture.c
```diff
@@ -494,20 +500,20 @@ static XdpOptionKey input_capture_set_pointer_barriers_options[] = {
...
+handle_set_pointer_barriers (XdpDbusInputCapture   *object,
                              GDBusMethodInvocation *invocation,
-                             const char *arg_session_handle,
-                             GVariant *arg_options,
-                             GVariant *arg_barriers,
-                             guint arg_zone_set)
+                             const char            *arg_session_handle,
+                             GVariant              *arg_options,
+                             GVariant              *arg_barriers,
+                             guint                  arg_zone_set)
```
jadahl, line 508: `uint32_t`.

src/input-capture.c
```diff
@@ -89,9 +89,9 @@ is_input_capture_session (Session *session)
 }
 
 static InputCaptureSession *
```
jadahl, line 91: The fixup! is on the wrong commit.

GeorgesStavracas:
Took the liberty to push cosmetic code style changes

PR #715, opened 2022-02-25
Some build system cleanups

whot:
~~Removal of `AM_MAINTAINER_MODE` and~~ collating the various XML files into a single location so adding to it is a single line instead of several (previously unsorted) Makefile lists.


configure.ac
```diff
@@ -21,7 +21,6 @@ AC_PROG_SED
 
 # Enable silent rules is available
 AM_SILENT_RULES([yes])
-AM_MAINTAINER_MODE([enable])
```
smcv, line 24: Some distributions like to use `--disable-maintainer-mode` to suppress rebuilds, and for those that don't, having the option available but enabled by default is very similar to not having the option. Saving 1 line doesn't seem like a huge benefit.

The upstream project doesn't have generated files in version control, but some downstreams import official source code releases (`make dist` tarballs) into a VCS, in which case all the Autotools cruft is in the VCS.

In the long term, as we gradually drop support for backporting the latest flatpak, xdg-desktop-portal, etc. to ancient distributions, I'd be very tempted to move to Meson (like flatpak/flatpak#2241) which would make this irrelevant.

doc/Makefile.am
```diff
@@ -54,56 +54,7 @@ portal_files = 								\
...
-xml_files = 								\
-	portal-org.freedesktop.portal.Account.xml 			\
-	portal-org.freedesktop.portal.Background.xml			\
-	portal-org.freedesktop.portal.Camera.xml			\
-	portal-org.freedesktop.portal.Device.xml			\
-	portal-org.freedesktop.portal.Documents.xml			\
-	portal-org.freedesktop.portal.Email.xml 			\
-	portal-org.freedesktop.portal.FileChooser.xml 			\
-	portal-org.freedesktop.portal.FileTransfer.xml			\
-	portal-org.freedesktop.portal.Flatpak.UpdateMonitor.xml 			\
```
smcv: > Note that this removes the
> FlatpakUpdateMonitor from the explicitly mentioned files but
> gdbus-codegen seems to do the right thing with this now.

Was this true as long ago as GLib 2.60?

smcv (commented):
In general this seems like an improvement.

configure.ac
smcv: See also #716

whot: >  I'd be very tempted to move to Meson

just time for a quick reply today: I have *most* of a meson port of xdg-desktop-portal ready, can push this on monday for discussion

doc/Makefile.am
whot: I tried this in a Fedora 30 podman which uses 2.60.7 and the answer is "no, apparently not". However: the `UpdateMonitor` file was only used in `doc/Makefile.am`, `git grep` shows nothing else referencing it. So the only drawback here appears to be that the documentation can't be built on old versions? And that's only because of the `<xi:include>`, without that it'd happily complete.

configure.ac
whot: I've dropped the patch to remove `AM_MAINTAINER_MODE` now

doc/Makefile.am
whot: re-visited this and: `UpdateMonitor` wasn't used for generation of files (the code always relied on this to be generated) but for listing the results of the gdbus-codegen files, to be used e.g in `CLEANFILES`. Updated the patchset accordingly.

GeorgesStavracas (approved):
Thanks, this looks great

PR #717, opened 2022-02-28
Add meson support

### LINKED ISSUE #716: Build using Meson?
Similar to flatpak/flatpak#2241, xdg-desktop-portal could build using Meson, to get a less annoying build system that does not rely on adding autogenerated goo to the tarball releases.

Unlike Flatpak, xdg-desktop-portal git HEAD already needs Meson indirectly: it requires GLib 2.60.0 or later, and GLib 2.60.0 required Meson 0.48.0 or newer. So nobody can backport the latest xdg-desktop-portal git HEAD to ancient distributions *anyway*, unless they would still be in a position to backport it if it was using Meson.

In .deb world, a Meson 0.48 dependency would mean the oldest target release for backporting would be Debian 10 (already superseded by Debian 11), or Ubuntu 20.04 LTS (will be superseded by 22.04 LTS within the next 2 months or so). I don't know what the equivalent RPM distributions look like.

(This assumes that people needing retrocomputing don't backport Meson - which is also relatively straightforward, and in my Steam Runtime day job I have it working in an Ubuntu 12.04-based environment, running on a backported copy of Python 3.5.)

whot:
This currently sits on top of #715, just ignore the first few patches. Pushing so the effort doesn't get duplicated, happy for someone to take this over though, just let me know.

It's a draft, there are a few bits missing and the test suite doesn't work yet, at least not all of it.

Keeping meson *and* autotools is a bit of a struggle, there's a bit of cleanup possible (or at least less jumping through hoops) if we can drop autotools at the same time.

~~atm this is build-tested only.~~

Fixes #716


document-portal/meson.build
```diff
@@ -0,0 +1,63 @@
+# Note: this can be flattened into the main meson.build file but the
```
smcv, line 1: Folding `document-portal/meson.build` into `./meson.build` is not really desirable anyway: in Meson it's conventional to have one `meson.build` per directory, and that doesn't have the same disadvantages that it would in Autotools (Meson/ninja always behaves like non-recursive `make`, even when it's "recursive" at the `meson.build` level).

document-portal/meson.build
```diff
@@ -0,0 +1,63 @@
...
+# code includes "document-portal/foo" and it's too hard to get meson and
+# autotools to agree on this
```
smcv, line 3: The Meson equivalent of `-I$(top_srcdir) -I$(top_builddir)` seems to be to have this in the top-level `meson.build`:

```
common_include_directories = include_directories('.')
```

and then use `include_directories : common_include_directories` in libraries and executables built by subdirectories.

meson.build
```diff
@@ -0,0 +1,470 @@
+project('xdg-desktop-portal', 'c',
+         version: '1.13.0',
+         license: 'LGPLv2+',
+         default_options: ['warning_level=2'],
+         meson_version: '>= 0.59.0')
```
smcv, line 5: I'd like to keep xdg-desktop-portal backportable to at least Debian 11 (Meson 0.56.2), and ideally Debian 10 (Meson 0.49.0) if possible. I'm sure developers of other long-term-supported distributions like RHEL and SLES would have similar requests.

meson.build
```diff
@@ -0,0 +1,470 @@
...
+if dir_dataroot == ''
+    dir_dataroot = dir_datadir
+endif
+
+###### various include directories we'll use later
+# These are set here so meson handles the relative paths correctly,
+# makes life easier for us
+
+inc_builddir = include_directories('.')  # config.h
+inc_src = include_directories('src')
```
smcv, line 42: This naming is misleading: `include_directories('.')` is the Meson equivalent of `-I$(top_builddir) -I$(top_srcdir)`, and `include_directories('src')` is the equivalent of `-I$(top_builddir)/src -I$(top_srcdir)/src`.

meson.build
```diff
@@ -0,0 +1,470 @@
...
+
+    custom_target('html',
+            build_by_default: true,
+            depends: [built_src_portal, built_src_impl],
+            input: portal_docs_xml,
+            output: 'portal-docs.html',
+            command: [prog_xmlto, get_option('xmlto-flags'), 'xhtml-nochunks', '-m', xslfile, '@INPUT@'])
+endif
+
+# FIXME: coverage testing  is missing
```
smcv, line 463: Meson automatically provides the `b_coverage` option, so we don't need to write any code to make it possible to track test coverage.

tests/backend/access.c
```diff
@@ -4,7 +4,7 @@
 
 #include <gio/gio.h>
 
-#include "src/xdp-impl-dbus.h"
+#include "xdp-impl-dbus.h"
```
smcv, line 7: This would be unnecessary with `include_directories : common_include_directories`.

tests/meson.build
```diff
@@ -0,0 +1,147 @@
...
+subdir('portals')
+subdir('services')
+
+test_db = executable('testdb',
+                    ['testdb.c'] + src_db,
+                    dependencies: [deps_common],
+                    include_directories: [inc_builddir])
+# This relies on dbs/meson.build to complete first which I think is always
+# true. Can't figure out how to add a configure_file() output as dependency
+# for the test.
```
smcv, line 18: Meson doesn't guarantee that directory A completes before directory B, so if `dbs/` needs to be *built* first, then there needs to be an explicit dependency.

If the file only needs to exist *before running the test* then I think that's automatic.

smcv (commented):
(Not a full review)

document-portal/meson.build
whot: The biggest issue I found with using `subdir()` in meson is that it only works as long as all components play along. For normal builds that's fine but gdbus-codegen produces multiple logically separate entities (in our case: code and docs) so using `subdir` means having to reach across directories and using variables defined in other `meson.build` files which is harder to read. Flattening things out can help with that, sometimes.

tests/backend/access.c
whot: true, but otoh it seems like a more correct approach to have the include directory-less and adjust the include path as necessary? Even for the autotools build.

tests/meson.build
whot: that's the case here, I'll amend the comment

meson.build
```diff
@@ -0,0 +1,470 @@
...
+endif
+
+have_pipewire = dep_pipewire.found()
+if have_pipewire
+    config_h.set('HAVE_PIPEWIRE', 1)
+endif
+
+build_docbook = false
+prog_xmlto = find_program('xmlto', required: get_option('docbook-docs'))
+if not get_option('docbook-docs').disabled() and prog_xmlto.found()
```
eli-schwartz, line 109: I don't understand the purpose of this conditional, the entire point of a feature option is so that if it is disabled, prog_xmlto will NOT be found -- it will log to the effect of "Program xmlto skipped: feature docbook docs is disabled".

meson.build
```diff
@@ -0,0 +1,470 @@
...
+
+doc_portal_service_file = files('document-portal/org.freedesktop.portal.Documents.service.in')
+permission_portal_service_file = files('document-portal/org.freedesktop.impl.portal.PermissionStore.service.in')
+desktop_portal_service_file = files('src/org.freedesktop.portal.Desktop.service.in')
+
+dbus_service_files = [
+    doc_portal_service_file,
+    permission_portal_service_file,
+    desktop_portal_service_file,
+]
```
eli-schwartz, line 152: Why do these each have their own variable which is never used other than to treat as opaque members of an array?

Also note that you can pass strings directly to configure_file, the point of a files() object is to remember which directory it can be found in *iff* you are going to use it from a meson.build in another directory.

tests/meson.build
eli-schwartz: > Can't figure out how to add a configure_file() output as dependency

configure_file is, by definition, created at configure time. No build rules are generated for it. It can never fail to exist on time.

whot: :facepalm:

yep, that'd explain it. removed the comment, thanks

meson.build
whot: the first two are used in `tests/services/meson.build`, hence the files object so I don't have to mangle the directories. The third one is indeed unused but added for consistency.

This has partially changed now since the two `document-portal` files are now in the `document-portal/meson.build` file, the `Desktop.service` file is now just a string since there's no group for consistency anymore.

meson.build
whot: I'm thinking this may be more correct as:
```
if get_option('docbook-docs').auto() and prog_xmlto.found()
```
because it's for the case of `auto` where you may have xmlto but we still need to the fs check to check for the xml file to be present. unless there's some other way?

meson.build
```diff
@@ -0,0 +1,470 @@
+project('xdg-desktop-portal', 'c',
+         version: '1.13.0',
+         license: 'LGPLv2+',
```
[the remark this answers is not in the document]
whot: oops, fixed locally

whot:
Thanks heaps for the reviews, I've changed this now to per-directory `meson.build` files. Still incomplete, there are a few things missing (`po` for example). The test suite runs and passes some tests, but `test-portals` fails with `Failed to launch test-backends`. And it doesn't actually abort correctly, it just sits at 100 cpu and eventually times out. 

I tried debugging this but it looks like it's related to gtest somehow which, ideally, one of you would know the immediate fix for :)
The executable path is correct, but something else must be missing.

meson.build
eli-schwartz: No? You check if the file is present and then in the enabled() case you error out and in the auto case you leave `build_docbook = false`.

It is correct to skip only in the .disabled() case, and my point was that `prog_xmlto.found()` implicitly checks the disabled case *already*.

eli-schwartz: No, my suggestion is correct. Delete the `not get_option('docbook-docs').disabled() and`

Just use:

```meson
if prog_xmlto.found()
```

whot: oh, indeed, thanks! changed locally, will be in the next push

whot:
Ok, this should be ready now. `diff -r` shows the installed tree is identical to the autotools one, bar "binary files differ" and the docs having different random identifiers.

meson.build
whot: 0.56.2 was an easy one, no warnings from meson. 0.49 is going to be more work, there are several warnings that need to be worked around (some of which now deprecated upstream so we'd have to do version checks). So this mostly depends on how badly we need 0.49 support.

mwleeds:
@whot looks like this will need rebasing to fix conflicts and account for recent changes on main, sorry

mwleeds:
Thanks for doing this by the way!

whot:
updated!

whot:
Still struggling with the test suite, best I can tell there is a race condition in the portal tests - sometimes they fail, sometimes they succeed but I've been unable so far to narrow down what exactly the problem is.

document-portal/meson.build
```diff
@@ -0,0 +1,102 @@
...
+    'permission-db.c',
+    'permission-db.h',
+    'gvdb/gvdb-reader.h',
+    'gvdb/gvdb-format.h',
+    'gvdb/gvdb-reader.c',
+    'gvdb/gvdb-builder.h',
+    'gvdb/gvdb-builder.c',
+)
+
+src_xdg_permission_store = [
```
tristan957, line 18:
```meson
src_xdg_permission_store = [
  files(...),
  ....,
]

File: document-portal/meson.build [hunk 12]
```diff
@@ -0,0 +1,102 @@
...
+configure_file(input: 'xdg-permission-store.service.in',
+    output: '@BASENAME@',
+    configuration: base_config,
+    install: true,
+    install_dir: dir_systemd_userunit,
+)
+
+built_src_document_portal = gnome.gdbus_codegen('document-portal-dbus',
+    sources: [
+        '../data/org.freedesktop.portal.Documents.xml',
```
tristan957, line 46: Probably want to use files() here.

meson.build
```diff
@@ -0,0 +1,179 @@
...
+config_h.set_quoted('DATADIR', dir_datadir)
+config_h.set_quoted('LIBEXECDIR', dir_libexecdir)
+config_h.set_quoted('LOCALEDIR', dir_localedir)
+config_h.set_quoted('GETTEXT_PACKAGE', 'xdg-desktop-portal')
+config_h.set_quoted('PACKAGE_STRING', 'xdg-desktop-portal @0@'.format(meson.project_version()))
+if cc.has_function('renameat2')
+    config_h.set('HAVE_RENAMEAT2', 1)
+endif
+
+dep_glib = dependency('glib-2.0')
```
tristan957, line 83: May want to think about adding fallback wraps

document-portal/meson.build
eli-schwartz: files() only has semantic meaning when assigned to a variable and reused in another subdir's meson.build.

document-portal/meson.build
whot: done, without the `files()` bit given @eli-schwartz comment below

whot:
Finally found the race condition, see commit [98107fb](https://github.com/flatpak/xdg-desktop-portal/pull/717/commits/98107fba12da2cc0bdce609ac2a8cca6643c729b). With this change meson's test suite now runs most tests successfully - the remaining failure in the openuri tests (`test_open_directory`) is somewhat unrelated: `xdg-desktop-portal` ends up  printing this error:
```
(/home/whot/code/xdg-desktop-portal/build/tests/../src/xdg-desktop-portal:1223670): GLib-GIO-CRITICAL **: 11:11:13.105: g_app_launch_context_launch_failed: assertion 'startup_notify_id != NULL' failed                                                                                                                     
```
and because criticals are a fatal error, our `teardown` fails because we have the wrong exit code for `xdg-desktop-portal`.
This is unrelated to the meson patches though, it fails in the autotools test suite with the same error (except there it's obscured due to an earlier failure).

I'd really appreciate a proper review now, thanks!

whot:
gentle ping

meson.build
```diff
@@ -0,0 +1,179 @@
...
+        build_docbook = true
+    elif get_option('docbook-docs').enabled()
+        error('Flatpak development files are required to build DocBook docs')
+    endif
+endif
+
+enable_installed_tests = get_option('installed-tests')
+
+summary({
+    'Enable docbook documentation': build_docbook,
```
tristan957, line 142: Indent.

meson.build
```diff
@@ -0,0 +1,179 @@
...
+if dir_dataroot == ''
+    dir_dataroot = dir_datadir
+endif
+
+dir_installed_tests = dir_prefix / dir_libexecdir / 'installed-tests' / meson.project_name()
+dir_installed_tests_data = dir_prefix / dir_datadir / 'installed-tests' / meson.project_name()
+dir_docs = dir_datadir / 'doc' / meson.project_name()
+
+summary({
+    'DBus service dir': dir_dbus_service,
```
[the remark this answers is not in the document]
whot: fixed, thanks.

meson.build
```diff
@@ -0,0 +1,179 @@
...
+
+have_libsystemd = dep_libsystemd.found()
+if have_libsystemd
+    config_h.set('HAVE_LIBSYSTEMD', 1)
+endif
+
+build_docbook = false
+prog_xmlto = find_program('xmlto', required: get_option('docbook-docs'))
+if prog_xmlto.found()
+    fs = import ('fs')
```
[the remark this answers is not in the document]
whot: fixed, thanks.

meson.build
whot: fixed, thanks.

meson_options.txt
```diff
@@ -0,0 +1,44 @@
...
+       value: 'auto',
+       description: 'Enable Geoclue support. Needed for location portal')
+option('pipewire',
+       type: 'feature',
+       value: 'auto',
+       description: 'Enable PipeWire support. Needed for screen cast portal')
+option('systemd',
+       type: 'feature',
+       value: 'auto',
+       description: 'Enable systemd support.')
```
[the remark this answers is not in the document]
whot: fixed, thanks.

whot:
gentle ping again, this is blocking #764 and that again is blocking #714 from integrating nicely.

Note also that this is merely **adding** meson, it's not removing autotools yet. so merging this is low-risk, we can fix any issue later.

tests/backend/access.c
GeorgesStavracas: I don't have strong opinions on this, and don't see any problem in the approach taken here

src/meson.build
```diff
@@ -0,0 +1,196 @@
...
+    'secret.h',
+    'flatpak-instance.c',
+    'flatpak-instance.h',
+    'portal-impl.h',
+    'portal-impl.c',
+    built_src_portal,
+    built_src_impl,
+    built_src_geoclue,
+    built_resources,
+]
```
GeorgesStavracas, line 100: Do you really need to include header files here, and in the variables below? I don't think they're necessary, only the .c files

src/meson.build
```diff
@@ -0,0 +1,196 @@
+###### xdg-desktop-portal
+
+built_src_portal = gnome.gdbus_codegen('xdp-dbus',
```
GeorgesStavracas, line 3: I appreciate this is essentially a matter of taste, but I would very much prefer if you could suffix variables instead of prefixing them. E.g. `portal_built_sources`, `impl_built_sources`, `xdg_desktop_portal_src`, etc

Makefile.am
```diff
@@ -12,6 +12,25 @@ DISTCLEANFILES= $(NULL)
...
+              tests/dbs/meson.build \
+              tests/meson.build \
+              tests/portals/meson.build \
+              tests/services/meson.build \
+              tests/share/applications/meson.build \
+              tests/share/meson.build \
+              tests/template.test.in \
+              $(NULL)
+
+
```
GeorgesStavracas, line 33: Remove one of these empty lines please

GeorgesStavracas (changes requested):
This looks mostly good to me, only requested superficial changes. I'll merge this after they're applied.

src/meson.build
eli-schwartz: https://mesonbuild.com/FAQ.html#do-i-need-to-add-my-headers-to-the-sources-list-like-in-autotools

You do not. If you do, it has no effect other than to assert at configure time that the files didn't mysteriously get deleted, and expose the headers identically to the `extra_files: ` kwarg, which is documented as:
> Not used for the build itself but are shown as source files in IDEs that group files by targets (such as Visual Studio)

GeorgesStavracas (approved):
Wonderful, I still have a few comments but I think it'll be more productive if I just open a follow-up pull request

GeorgesStavracas:
@whot I took the liberty to slightly change and unify the style of meson files, and some other smaller aesthetical changes. I hope you don't mind that.

meson.build
```diff
@@ -0,0 +1,184 @@
...
+                         required: get_option('geoclue'))
+libportal_dep = dependency('libportal',
+                           required: get_option('libportal'))
+pipewire_dep = dependency('libpipewire-0.3',
+                          version: '>= 0.2.90',
+                          required: get_option('pipewire'))
+libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
+
+bwrap = find_program('bwrap')
+config_h.set_quoted('BWRAP', bwrap.full_path())
```
jtojnar, line 102: Is this used anywhere? I do not see it in the code.

whot: no, apparently not. It's only used as compiler flag to `xdg-desktop-portal-validate-icon` but the `config.h` entry is not necessary. Please file a PR to remove it, thanks.

jtojnar: Done in https://github.com/flatpak/xdg-desktop-portal/pull/860

PR #718, opened 2022-03-01
Preparatory PR for setting unsandboxed process app IDs

mwleeds:
As discussed in #581, this is the work needed to make this changes possible.

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -95,7 +95,11 @@ send_response_in_thread_func (GTask        *task,
       g_autofree char *ruri = NULL;
       g_autoptr(GError) error = NULL;
 
-      ruri = register_document (image, xdp_app_info_get_id (request->app_info), FALSE, FALSE, FALSE, &error);
+      if (xdp_app_info_is_host (request->app_info))
+        ruri = g_strdup (image);
+      else
+        ruri = register_document (image, xdp_app_info_get_id (request->app_info), FALSE, FALSE, FALSE, &error);
+
       if (ruri == NULL)
         g_warning ("Failed to register %s: %s", image, error->message);
       else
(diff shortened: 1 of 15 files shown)


src/xdp-utils.c
```diff
@@ -929,6 +929,9 @@ xdp_get_alternate_document_path (const char *path, const char *app_id)
 {
   int len;
 
+  if (g_str_equal (app_id, ""))
```
GeorgesStavracas, line 932: `g_str_equal()` is **not** NULL-safe, is it possible that `app_id` is NULL here? If so, it'd be more appropriate to use `app_id == NULL || g_str_equal(app_id, "")`, or perhaps `g_strcmp0()`

src/account.c
```diff
@@ -95,7 +95,11 @@ send_response_in_thread_func (GTask        *task,
       g_autofree char *ruri = NULL;
       g_autoptr(GError) error = NULL;
 
-      ruri = register_document (image, xdp_app_info_get_id (request->app_info), FALSE, FALSE, FALSE, &error);
+      if (xdp_app_info_is_host (request->app_info))
+        ruri = g_strdup (image);
+      else
+        ruri = register_document (image, xdp_app_info_get_id (request->app_info), FALSE, FALSE, FALSE, &error);
+
```
GeorgesStavracas, line 102: I wonder if `register_document()` should receive `XdpAppInfo`, and implement this host/sandbox behaviour internally?

GeorgesStavracas (commented):
Couple of comments

src/xdp-utils.c
mwleeds: No, `xdp_app_info_get_id` is guaranteed to return a non-NULL value. I have some docs written to document that guarantee that I'm going to push with the follow-up PR.

src/account.c
mwleeds: The reason I didn't do that is there is no `XdpAppInfo` available in `open-uri.c`; we only have the app ID in that case.

PR #719, opened 2022-03-02
Get app ID via libsystemd for unsandboxed apps

mwleeds:
This is a rework of #581 and dependent on #718, but I'm marking it as a draft because sadly it doesn't work for the one case I actually need it: GNOME Software.

In the case of, say, non-Flatpak Epiphany, the app ID can be successfully gleaned from the systemd unit which looks like `app-gnome-org.gnome.Epiphany-182352.scope`. But in the case of gnome-software (which runs as `/usr/bin/gnome-software --gapplication-service`) the unit looks like `app-gnome-gnome\x2dsoftware\x2dservice-9502.scope` (where 9502 is the pid in my case) so parsing that in any way is not going to get us the correct app ID of `org.gnome.Software`. So I think we need to look for a better way to get app IDs...

### DIFF
--- a/src/sd-escape.c
+++ b/src/sd-escape.c
@@ -1,0 +1,356 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "sd-escape.h"
+#include <errno.h>
+
+/* The following is copied from these files in systemd with minor
+ * modifications:
+ * - src/basic/escape.c
+ * - src/basic/utf8.c
+ * - src/basic/hexdecoct.c
+ */
+
+static int unhexchar(char c) {
+
+        if (c >= '0' && c <= '9')
+                return c - '0';
+
+        if (c >= 'a' && c <= 'f')
+                return c - 'a' + 10;
+
+        if (c >= 'A' && c <= 'F')
+                return c - 'A' + 10;
+
+        return -EINVAL;
+}
+
+static gboolean
+unichar_is_valid(guint32 ch) {
+
+        if (ch >= 0x110000) /* End of unicode space */
+                return FALSE;
+        if ((ch & 0xFFFFF800) == 0xD800) /* Reserved area for UTF-16 */
+                return FALSE;
+        if ((ch >= 0xFDD0) && (ch <= 0xFDEF)) /* Reserved */
+                return FALSE;
+        if ((ch & 0xFFFE) == 0xFFFE) /* BOM (Byte Order Mark) */
+                return FALSE;
+
+        return TRUE;
+}
+
+static int
+unoctchar(char c) {
+
+        if (c >= '0' && c <= '7')
+                return c - '0';
+
+        return -EINVAL;
+}
+
+int cunescape_one(const char *p, gsize length, guint32 *ret, gboolean *eight_bit, gboolean accept_nul) {
+        int r = 1;
+
+        g_assert(p);
+        g_assert(ret);
+
+        /* Unescapes C style. Returns the unescaped character in ret.
+         * Sets *eight_bit to true if the escaped sequence either fits in
+         * one byte in UTF-8 or is a non-unicode literal byte and should
+         * instead be copied directly.
+         */
+
+        if (length != G_MAXSIZE && length < 1)
+                return -EINVAL;
+
+        switch (p[0]) {
+
+        case 'a':
+                *ret = '\a';
+                break;
+        case 'b':
+                *ret = '\b';
+                break;
+        case 'f':
+                *ret = '\f';
+                break;
+        case 'n':
+                *ret = '\n';
+                break;
+        case 'r':
+                *ret = '\r';
+                break;
+        case 't':
+                *ret = '\t';
+                break;
+        case 'v':
+                *ret = '\v';
+                break;
+        case '\\':
+                *ret = '\\';
+                break;
+        case '"':
+                *ret = '"';
+                break;
+        case '\'':
+                *ret = '\'';
+                break;
+
+        case 's':
+                /* This is an extension of the XDG syntax files */
+                *ret = ' ';
+                break;
+
+        case 'x': {
+                /* hexadecimal encoding */
+                int a, b;
+
+                if (length != G_MAXSIZE && length < 3)
+                        return -EINVAL;
+
+                a = unhexchar(p[1]);
+                if (a < 0)
+                        return -EINVAL;
+
+                b = unhexchar(p[2]);
+                if (b < 0)
+                        return -EINVAL;
+
+                /* Don't allow NUL bytes */
+                if (a == 0 && b == 0 && !accept_nul)
+                        return -EINVAL;
+
+                *ret = (a << 4U) | b;
+                *eight_bit = TRUE;
+                r = 3;
+                break;
+        }
+
+        case 'u': {
+                /* C++11 style 16bit unicode */
+
+                int a[4];
+                gsize i;
+                guint32 c;
+
+                if (length != G_MAXSIZE && length < 5)
+                        return -EINVAL;
+
+                for (i = 0; i < 4; i++) {
+                        a[i] = unhexchar(p[1 + i]);
+                        if (a[i] < 0)
+                                return a[i];
+                }
+
+                c = ((guint32) a[0] << 12U) | ((guint32) a[1] << 8U) | ((guint32) a[2] << 4U) | (guint32) a[3];
+
+                /* Don't allow 0 chars */
+                if (c == 0 && !accept_nul)
+                        return -EINVAL;
+
+                *ret = c;
+                r = 5;
+                break;
+        }
+
+        case 'U': {
+                /* C++11 style 32bit unicode */
+
+                int a[8];
+                gsize i;
+                guint32 c;
+
+                if (length != G_MAXSIZE && length < 9)
+                        return -EINVAL;
+
+                for (i = 0; i < 8; i++) {
+                        a[i] = unhexchar(p[1 + i]);
+                        if (a[i] < 0)
+                                return a[i];
+                }
+
+                c = ((guint32) a[0] << 28U) | ((guint32) a[1] << 24U) | ((guint32) a[2] << 20U) | ((guint32) a[3] << 16U) |
+                    ((guint32) a[4] << 12U) | ((guint32) a[5] <<  8U) | ((guint32) a[6] <<  4U) |  (guint32) a[7];
+
+                /* Don't allow 0 chars */
+                if (c == 0 && !accept_nul)
+                        return -EINVAL;
+
+                /* Don't allow invalid code points */
+                if (!unichar_is_valid(c))
+                        return -EINVAL;
+
+                *ret = c;
+                r = 9;
+                break;
+        }
+
+        case '0':
+        case '1':
+        case '2':
+        case '3':
+        case '4':
+        case '5':
+        case '6':
+        case '7': {
+                /* octal encoding */
+                int a, b, c;
+                guint32 m;
+
+                if (length != G_MAXSIZE && length < 3)
+                        return -EINVAL;
+
+                a = unoctchar(p[0]);
+                if (a < 0)
+                        return -EINVAL;
+
+                b = unoctchar(p[1]);
+                if (b < 0)
+                        return -EINVAL;
+
+                c = unoctchar(p[2]);
+                if (c < 0)
+                        return -EINVAL;
+
+                /* don't allow NUL bytes */
+                if (a == 0 && b == 0 && c == 0 && !accept_nul)
+                        return -EINVAL;
+
+                /* Don't allow bytes above 255 */
+                m = ((guint32) a << 6U) | ((guint32) b << 3U) | (guint32) c;
+                if (m > 255)
+                        return -EINVAL;
+
+                *ret = m;
+                *eight_bit = TRUE;
+                r = 3;
+                break;
+        }
+
+        default:
+                return -EINVAL;
+        }
+
+        return r;
+}
+
+/**
+ * utf8_encode_unichar() - Encode single UCS-4 character as UTF-8
+ * @out_utf8: output buffer of at least 4 bytes or NULL
+ * @g: UCS-4 character to encode
+ *
+ * This encodes a single UCS-4 character as UTF-8 and writes it into @out_utf8.
+ * The length of the character is returned. It is not zero-terminated! If the
+ * output buffer is NULL, only the length is returned.
+ *
+ * Returns: The length in bytes that the UTF-8 representation does or would
+ *          occupy.
+ */
+static gsize
+utf8_encode_unichar(char *out_utf8, guint32 g) {
+
+        if (g < (1 << 7)) {
+                if (out_utf8)
+                        out_utf8[0] = g & 0x7f;
+                return 1;
+        } else if (g < (1 << 11)) {
+                if (out_utf8) {
+                        out_utf8[0] = 0xc0 | ((g >> 6) & 0x1f);
+                        out_utf8[1] = 0x80 | (g & 0x3f);
+                }
+                return 2;
+        } else if (g < (1 << 16)) {
+                if (out_utf8) {
+                        out_utf8[0] = 0xe0 | ((g >> 12) & 0x0f);
+                        out_utf8[1] = 0x80 | ((g >> 6) & 0x3f);
+                        out_utf8[2] = 0x80 | (g & 0x3f);
+                }
+                return 3;
+        } else if (g < (1 << 21)) {
+                if (out_utf8) {
+                        out_utf8[0] = 0xf0 | ((g >> 18) & 0x07);
+                        out_utf8[1] = 0x80 | ((g >> 12) & 0x3f);
+                        out_utf8[2] = 0x80 | ((g >> 6) & 0x3f);
+                        out_utf8[3] = 0x80 | (g & 0x3f);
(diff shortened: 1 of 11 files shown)


mwleeds:
We probably need to do something like what gnome-shell does but the algorithm there is pretty complicated: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-window-tracker.c#L380-468

mwleeds:
Some notes from @refi64 on this topic:

> Okay so the gist of it for non-Flatpak apps was:
> 
> - check if an app ID was set via Wayland protocols (not sure if external clients can check this)
> - check for some GTK-specitic X properties
> - check cgroups I think? (there's something in the code about checking known applications but I couldn't figure it out)
> - check via the values passed to the fd.o startup notification protocol if it was used
> - there's another check I didn't understand involving what I think was called a MetaWindowGroup? Probably something internal to mutter?
> 
> Now I think I recall later finding some edge cases that stemmed from some Qt apps using different desktop file names but somehow still being associated by the compositor
> 
> Looking through some other comments at the time I had made, there were some other quirks too:
> 
> - some Wayland apps never set an app ID and thus couldn't be matched to anything, my solution at the time was using the executable basename but I'm not sure what Shell does
> - any app IDs need to also be matched against the StartupWMClass values even if the desktop file name doesn't match
> - And on Xorg you're supposed to check WM_CLASS before the GTK specific props

swick:
To me this looks like an issue with gnome-software. The file it installs in `/etc/xdg/autostart` is called `gnome-software-service.desktop` which is exactly what systemd then uses as the appid. Renaming it to `org.gnome.Software.desktop` should fix it.

The question I have however is why this is relevant. Shouldn't the portal be able to work with an appid of `gnome\x2dsoftware\x2dservice`?

carlocastoldi:
> To me this looks like an issue with gnome-software. The file it installs in /etc/xdg/autostart is called gnome-software-service.desktop which is exactly what systemd then uses as the appid. Renaming it to org.gnome.Software.desktop should fix it.

I feel like that is the situation as well. Gnome Software doesn't seem to follow the [XDG standardization for application](https://systemd.io/DESKTOP_ENVIRONMENTS/). Either because of the .desktop file name or, perhaps, it doesn't set the [`application_id`](https://docs.gtk.org/gtk3/ctor.Application.new.html) properly? (doubt)

> The question I have however is why this is relevant. Shouldn't the portal be able to work with an appid of gnome\x2dsoftware\x2dservice?

yes, indeed. (?) Since we are talking about an un-sandboxed app it doesn't seem to be that relevant...

mwleeds:
> > To me this looks like an issue with gnome-software. The file it installs in /etc/xdg/autostart is called gnome-software-service.desktop which is exactly what systemd then uses as the appid. Renaming it to org.gnome.Software.desktop should fix it.
> 
> I feel like that is the situation as well. Gnome Software doesn't seem to follow the [XDG standardization for application](https://systemd.io/DESKTOP_ENVIRONMENTS/). Either because of the .desktop file name or, perhaps, it doesn't set the [`application_id`](https://docs.gtk.org/gtk3/ctor.Application.new.html) properly? (doubt)
> 
> > The question I have however is why this is relevant. Shouldn't the portal be able to work with an appid of gnome\x2dsoftware\x2dservice?
> 
> yes, indeed. (?) Since we are talking about an un-sandboxed app it doesn't seem to be that relevant...
> 
> EDIT: wait, browsing Gnome Software repo i see they expect 3 `.desktop` files:
> 
> * `org.gnome.Software.desktop` (in `/usr/share/applications/`, for example)
> * `gnome-software-local-file.desktop` (in `/usr/share/applications/`, for example)
> * `gnome-software-service.desktop` (in `/etc/xdg/autostart/, for example)
> 
> could it be that the autostarted service is not an actual GUI application? Should it really be associated with the AppID `org.gnome.Software`?

Yes, `gnome-software-service.desktop` is named that way since it is distinct from `org.gnome.Software.desktop`. The former has `Exec=/usr/bin/gnome-software --gapplication-service` and the latter `Exec=gnome-software %U`. But the `gnome-software --gapplication-service` process that gets started by `gnome-session` is the graphical application, it just starts out running in the background and you only use it as a GUI if you launch it from `gnome-shell` or it is otherwise launched without `--gapplication-service`.

The reason we need the correct app ID, `org.gnome.Software` not `gnome-software-service`, is that we need to be able to check the app ID against an allowlist in the dynamic launcher portal [implementation](https://github.com/flatpak/xdg-desktop-portal-gtk/pull/371/files#diff-6373575aed2524053d89d8497a76ed48043ef4795c1357b3f696f8803083baa2R352). We also need it to be correct because if we use `gnome-software-service` as the app ID but then the user later disables the autostart launcher and launches Software from the shell, the app ID would be read as `org.gnome.Software` and the stored permissions wouldn't be properly granted any more. Similarly for any other reason that we might later get the right app ID, such as improving the app ID algorithm in xdg-desktop-portal.

carlocastoldi:
Mmh I see. Just one more question I can't answer since at the moment I'm not on Gnome: I don't fully understand what happens when `gnome-session` starts GS's service in the background and _then_ the user launches GS from `gnome-shell` without the parameter. Is the systemd slice changed or is it the same of the process launched in background?

In case that it changes, shouldn't the background service be identified as a [`background.slice`](https://www.freedesktop.org/software/systemd/man/systemd.special.html#Special%20User%20Slice%20Units)?
If it doesn't, do you think it's possible to change the name of the systemd unit only once the GUI app is actually launched?

mwleeds:
> Mmh I see. Just one more question I can't answer since at the moment I'm not on Gnome: I don't fully understand what happens when `gnome-session` starts GS's service in the background and _then_ the user launches GS from `gnome-shell` without the parameter. Is the systemd slice changed or is it the same of the process launched in background?
> 
> In case that it changes, shouldn't the background service be identified as a [`background.slice`](https://www.freedesktop.org/software/systemd/man/systemd.special.html#Special%20User%20Slice%20Units)?

When you launch `gnome-software`, from the shell or otherwise, the process you launched finds that there's a running instance and exits, and the running instance gets focused. Both before and after you launch Software the running process is in e.g. `/user.slice/user-1000.slice/user@1000.service/app.slice/app-gnome-gnome\x2dsoftware\x2dservice-13765.scope` so it's never in `background.slice`.

>  If it doesn't, do you think it's possible to change the name of the systemd unit only once the GUI app is actually launched?

I don't know of any way to make that happen, but even if we could, we want the correct app ID even when Software is running in the background. Portal operations are generally in response to user interaction but it's not inconceivable we would need to use x-d-p in the background, and certainly we would at least connect to the D-Bus service even if we didn't call any methods.

swick:
> The reason we need the correct app ID, org.gnome.Software not gnome-software-service, is that we need to be able to check the app ID against an allowlist in the dynamic launcher portal [implementation](https://github.com/flatpak/xdg-desktop-portal-gtk/pull/371/files#diff-6373575aed2524053d89d8497a76ed48043ef4795c1357b3f696f8803083baa2R352).

Urgh, yeah, so this is not a problem in principal just a weird whitelist case.

> We also need it to be correct because if we use gnome-software-service as the app ID but then the user later disables the autostart launcher and launches Software from the shell, the app ID would be read as org.gnome.Software and the stored permissions wouldn't be properly granted any more.

Not in this case, if you start gnome-software from shell it it will find the running instance, tell it to open a window and then exit.

> Yes, gnome-software-service.desktop is named that way since it is distinct from org.gnome.Software.desktop

Was this actually a conscious decision? They are after all the same application, the autostart desktop file just starts it in the background.

mwleeds:
I proposed a rename of the autostart desktop file: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1261

It looks like `gnome-software-service.desktop` used to start a binary called `gnome-software-service` so that is why it was named as such.

mwleeds:
> > We also need it to be correct because if we use gnome-software-service as the app ID but then the user later disables the autostart launcher and launches Software from the shell, the app ID would be read as org.gnome.Software and the stored permissions wouldn't be properly granted any more.
> 
> Not in this case, if you start gnome-software from shell it it will find the running instance, tell it to open a window and then exit.

The hypothetical scenario I mentioned was that the autostart launcher was disabled by the user, and in that case there would be no running instance to find.

mwleeds:
> I proposed a rename of the autostart desktop file: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1261
> 

Even with that change the cgroup does not contain the app ID when the app is launched from the command line: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1261#note_1398950

I don't think that's necessarily a blocker for merging this but it makes development hard.

carlocastoldi:
> Even with that change the cgroup does not contain the app ID when the app is launched from the command line

yes, the systemd specification is tailored for desktop environments. If you want to launch a program from terminal such that it complies with the xdg systemd & DE specification you have to do it as described in this snippet: https://invent.kde.org/-/snippets/1111
(spoiler: with `systemd-run` or `xdg-open`).

I'd say it's still worth it to push the MR on Gnome Software

src/xdp-utils.c
```diff
@@ -144,11 +149,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+    /* Escape the unit name which may have \x hex codes in it, e.g.
+     * "app-gnome-org.gnome.Evolution\x2dalarm\x2dnotify-2437.scope"
+     */
+    app_id = g_match_info_fetch (match, 1);
+    if (cunescape (app_id, UNESCAPE_RELAX, &app_info->id) < 0)
+      app_info->id = g_strdup ("");
+  } else {
+    app_info->id = g_strdup ("");
+    return;
+  }
```
GeorgesStavracas, line 196: Code style: curly braces should be in a newline, like the rest of the file

src/xdp-utils.c
```diff
@@ -144,11 +149,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  if (g_regex_match (regex, unit, 0, &match)) {
+    /* Escape the unit name which may have \x hex codes in it, e.g.
+     * "app-gnome-org.gnome.Evolution\x2dalarm\x2dnotify-2437.scope"
+     */
+    app_id = g_match_info_fetch (match, 1);
+    if (cunescape (app_id, UNESCAPE_RELAX, &app_info->id) < 0)
+      app_info->id = g_strdup ("");
+  } else {
+    app_info->id = g_strdup ("");
+    return;
```
GeorgesStavracas, line 195: Seems like this `return` is not necessary

GeorgesStavracas (approved):
Had a couple of nits, but it generally LGTM now. Feel free to merge after addressing them.

src/xdp-utils.c
```diff
@@ -144,11 +149,62 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+      return;
+    }
+
+  /*
+   * From https://systemd.io/DESKTOP_ENVIRONMENTS/ the format is one of:
+   * app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
+   * app[-<launcher>]-<ApplicationID>-autostart.service -> no longer true since systemd v248
+   * app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
+   */
+  regex = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?\\..+?\\..+?)"
```
hadess, line 181: All the string manipulation code should be split and unit tests added.

src/sd-escape.c
```diff
@@ -0,0 +1,356 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "sd-escape.h"
+#include <errno.h>
+
+/* The following is copied from these files in systemd with minor
```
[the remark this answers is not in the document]
mwleeds: If I add unit tests for the unit name parsing that uses these helper functions, I don't really see the point in also testing the helper functions, especially given they are directly copied from somewhere they are being tested?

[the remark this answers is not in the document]
hadess: > If I add unit tests for the unit name parsing that uses these helper functions,

I don't see new unit tests in the PR, I looked 3 times, including just now. Is github's review tool hiding them from me?

> I don't really see the point in also testing the helper functions, especially given they are directly copied from somewhere they are being tested?

Because they can, and will, eventually diverge. Another option would be to export those functions from systemd so you don't have to test them in xdg-desktop-portal at all.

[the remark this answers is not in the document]
mwleeds: > > If I add unit tests for the unit name parsing that uses these helper functions,
> 
> I don't see new unit tests in the PR, I looked 3 times, including just now. Is github's review tool hiding them from me?
> 

They don't exist yet. I was saying I can add them.

> > I don't really see the point in also testing the helper functions, especially given they are directly copied from somewhere they are being tested?
> 
> Because they can, and will, eventually diverge. Another option would be to export those functions from systemd so you don't have to test them in xdg-desktop-portal at all.

Do you mean add new library API to libsystemd upstream? That might be a good idea but I don't have time for it, sorry.

I'm not sure why they would diverge though? The copy here will become out of date when any changes are made to the copy in systemd, but I wouldn't expect us to need to make our own downstream changes here. And even if we did, the functionality we need would be tested by the higher level unit tests I am going to add.

tests/test-xdp-utils.c
```diff
@@ -127,6 +127,46 @@ test_alternate_doc_path (void)
...
 }
 
+static void
+test_app_id_via_systemd_unit (void)
+{
+  g_autofree char *app_id = NULL;
+
+  app_id = _xdp_parse_app_id_from_unit_name ("app-not-a-well-formed-unit-name");
+  g_assert_cmpstr (app_id, ==, "");
+  g_clear_pointer (&app_id);
```
GeorgesStavracas, line 137: `g_clear_pointer (&app_id, g_free);`

(Here and all instances below)

src/xdp-utils.c
mwleeds: You're right, added a unit test

mwleeds:
(force-pushed to rebase on main)

src/xdp-utils.c
```diff
@@ -144,11 +149,79 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  else
+    {
+      app_id = g_strdup ("");
+    }
+
+  return g_steal_pointer (&app_id);
+}
+
+void
+get_app_id_from_pid (pid_t pid, XdpAppInfo *app_info)
```
hadess, line 193: This call seems to unconditionally assign `app_info->id`, so it might be good to add a guard like:
```
g_return_if_fail (app_info->id == NULL);
```

Also, a bit late, but maybe `set_appid_from_pid()` is a better name, and pass `app_info` as the first argument?

src/xdp-utils.c
```diff
@@ -144,11 +149,79 @@ xdp_app_info_new (XdpAppInfoKind kind)
...
+  g_autoptr(GError) error = NULL;
+  g_autofree char *app_id = NULL;
+
+  g_assert (g_str_has_prefix (unit, "app-"));
+
+  /*
+   * From https://systemd.io/DESKTOP_ENVIRONMENTS/ the format is one of:
+   * app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
+   * app[-<launcher>]-<ApplicationID>-autostart.service -> no longer true since systemd v248
+   * app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
```
hadess, line 166: The 2 formats above aren't tested in the test suite.

src/xdp-utils.c
mwleeds: Sure, makes sense on both counts

src/xdp-utils.c
mwleeds: I'm not sure what you mean?
- `app[-<launcher>]-<ApplicationID>-<RANDOM>.scope` tested by `app-gnome-org.gnome.Epiphany-182352.scope`
- `app[-<launcher>]-<ApplicationID>-autostart.service` tested by `app-gnome-org.gnome.SettingsDaemon.DiskUtilityNotify-autostart.service`
- `app[-<launcher>]-<ApplicationID>[@<RANDOM>].service` tested by `app-KDE-org.kde.okular@12345.service`

GeorgesStavracas (approved):
This PR continues to LGTM, the tests are *chef's kiss*. @hadess should we land this? This is blocking further portals.

GeorgesStavracas:
For the sake of progress, given that @mwleeds added corresponding unit tests, I'll take the liberty to go ahead and land this.

@hadess if you have other concerns about this PR, please open an issue for us to discuss there.

src/sd-escape.c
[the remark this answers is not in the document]
hadess: > Do you mean add new library API to libsystemd upstream? That might be a good idea but I don't have time for it, sorry.

https://github.com/systemd/systemd/issues/22718

> I'm not sure why they would diverge though? The copy here will become out of date when any changes are made to the copy in systemd, but I wouldn't expect us to need to make our own downstream changes here. And even if we did, the functionality we need would be tested by the higher level unit tests I am going to add.

FWIW, it always does eventually diverge. Because someone with good intentions fixes a bug they find, or simplifies the code and they don't know that it's copy/paste, and the reviewer forgets.

src/xdp-utils.c
hadess: I'm not sure how I came to that conclusion when I was reading the diff, but I was obviously not concentrating hard enough.

PR #722, opened 2022-03-07
Bring back in-house icon validator

mwleeds:
This reverts commit 2fb498f92f8e1c41049dedf887690083fa12f6d1 and makes
related changes so that we use our own copy of Flatpak's icon validator.

Here are the reasons for bringing back this copied code:
(1) To avoid a circular dependency between x-d-p and Flatpak. Debian has
been patching this code back in for years for this purpose.
(2) We don't want icon validation to be optional since it's a security
measure. Having our own binary makes it easier to make it required.
(3) For the new dynamic launcher portal, we need to determine the size
and format of icons. The validator already gathers this information so
it makes sense to just have it print those things to stdout. This means
our copy of this code will diverge from Flatpak.
(4) It is too small to be worth maintaining as a separate repo.


smcv (commented):
I think it would be useful if the commit message named a Flatpak version that `src/validate-icon.c` was taken from (`git log` says the most recent change was before 1.12.0, so any recent version is equivalent to any other).

smcv (commented):
The changes other than the actual icon validator look good to me. I haven't reviewed the code itself, in the hope that it's identical to Flatpak 1.12.0 to 1.12.6 (and 1.13.1)

mwleeds:
> I think it would be useful if the commit message named a Flatpak version that `src/validate-icon.c` was taken from (`git log` says the most recent change was before 1.12.0, so any recent version is equivalent to any other).

Makes sense, I added a note to the commit.

mwleeds:
> The changes other than the actual icon validator look good to me. I haven't reviewed the code itself, in the hope that it's identical to Flatpak 1.12.0 to 1.12.6 (and 1.13.1)

Yes it's identical to Flatpak 1.13.1

src/validate-icon.c
```diff
@@ -0,0 +1,249 @@
...
+  validate_icon[symlink_size] = 0;
+
+  add_args (args,
+            flatpak_get_bwrap (),
+            "--unshare-ipc",
+            "--unshare-net",
+            "--unshare-pid",
+            "--ro-bind", "/usr", "/usr",
+            "--ro-bind-try", "/etc/ld.so.cache", "/etc/ld.so.cache",
+            "--ro-bind", validate_icon, validate_icon,
```
hadess, line 161: I'm not certain this would be enough to cover all cases, in all distros.

This is missing the fontconfig cache which is almost certainly needed for SVG:
https://gitlab.gnome.org/GNOME/gnome-desktop/-/blob/master/libgnome-desktop/gnome-desktop-thumbnail-script.c#L587-L589

src/validate-icon.c
```diff
@@ -0,0 +1,249 @@
...
+      return 1;
+    }
+
+  validate_icon[symlink_size] = 0;
+
+  add_args (args,
+            flatpak_get_bwrap (),
+            "--unshare-ipc",
+            "--unshare-net",
+            "--unshare-pid",
```
hadess, line 158: Would be good to list the namespaces it actually uses, explaining why `--unshare-all` isn't used.

src/validate-icon.c
```diff
@@ -0,0 +1,249 @@
...
+      if (max_height < 16 || max_height > 4096)
+        {
+          g_printerr ("Bad height limit: %s\n", arg_height);
+          return 1;
+        }
+    }
+  else
+    {
+      /* Sanity check for vector files */
+      max_height = max_width = 4096;
```
hadess, line 72: "Dimension checks"?

Would also be good to move those magic numbers to a constant/`#define`

src/validate-icon.c
```diff
@@ -0,0 +1,249 @@
...
+{
+  GdkPixbufFormat *format;
+  int max_width, max_height;
+  int width, height;
+  const char *name;
+  const char *allowed_formats[] = { "png", "jpeg", "svg", NULL };
+  g_autoptr(GdkPixbuf) pixbuf = NULL;
+  g_autoptr(GError) error = NULL;
+
+  format = gdk_pixbuf_get_file_info (filename, &width, &height);
```
hadess, line 39: This will open and read a fair chunk of the file...

src/validate-icon.c
```diff
@@ -0,0 +1,249 @@
...
+      max_height = max_width = 4096;
+    }
+
+  if (width > max_width || height > max_height)
+    {
+      g_printerr ("Image too large (%dx%d). Max. size %dx%d\n", width, height, max_width, max_height);
+      return 1;
+    }
+
+  pixbuf = gdk_pixbuf_new_from_file (filename, &error);
```
hadess, line 81: And this will read it a second time.

You might be better off reimplementing this loop using a loader and waiting to fetch the size yourself, depending on the possibly malicious image.

hadess (commented):
A couple of random comments.

Please also use this in the binary to protect against malicious files that are designed to eat all your RAM: https://gitlab.gnome.org/GNOME/totem/-/blob/master/src/totem-resources.c

mwleeds:
> A couple of random comments.
> 
> Please also use this in the binary to protect against malicious files that are designed to eat all your RAM: https://gitlab.gnome.org/GNOME/totem/-/blob/master/src/totem-resources.c

This code is all just copied verbatim from Flatpak, so perhaps you could open an issue there to discuss these improvements you're suggesting?

PR #730, opened 2022-03-16
screencast: Analyze the value we get straight away

aleixpol:
Otherwise I get the following error message:
> Received restore data in invalid variant format ('(suv)'; expected '(suv)')

### DIFF
--- a/src/screen-cast.c
+++ b/src/screen-cast.c
@@ -1002,11 +1002,9 @@ replace_restore_data_by_token (ScreenCastSession *screen_cast_session,
     {
       if (g_strcmp0 (key, "restore_data") == 0)
         {
-          GVariant *child = g_variant_get_child_value (value, 0);
-
-          if (g_variant_check_format_string (child, RESTORE_DATA_TYPE, FALSE))
+          if (g_variant_check_format_string (value, RESTORE_DATA_TYPE, FALSE))
             {
-              screen_cast_session->restore_data = g_variant_ref_sink (child);
+              screen_cast_session->restore_data = g_variant_ref_sink (value);
               found_restore_data = TRUE;
             }
           else


src/screen-cast.c
```diff
@@ -1002,11 +1002,9 @@ replace_restore_data_by_token (ScreenCastSession *screen_cast_session,
     {
       if (g_strcmp0 (key, "restore_data") == 0)
         {
-          GVariant *child = g_variant_get_child_value (value, 0);
```
GeorgesStavracas, line 1005: I don't think this is correct. It is expected that portal implementations provide a variant of signature `v` here, with a single child containing the `(suv)` tuple. That's because we're iterating over an `a{sv}` vardict.

GeorgesStavracas (changes requested):

src/screen-cast.c
aleixpol: `persist_mode` is also in the `a{sv}` and we are not calling g_variant_get_child_value there.

GeorgesStavracas: Right

*sighs*

I don't know why I still try to code, this is clearly not my craft.

GeorgesStavracas (approved):

GeorgesStavracas:
Looks like GitHub is stuck, could you please force-push something to get a new CI run?

aleixpol:
```
To github.com:aleixpol/xdg-desktop-portal.git
 ! [remote rejected] work/global-shortcut -> work/global-shortcut (Internal Server Error)
error: failed to push some refs to 'github.com:aleixpol/xdg-desktop-portal.git'
```

Seems like github is down in general...
https://www.githubstatus.com < marks it as degraded

PR #733, opened 2022-03-18
fix building without systemd

Dudemanguy:
sd-escape.c and sd-escape.h were skipped when building without systemd,
but they contain functions needed for xdg-desktop-portal to build.
Fortunately, neither file actually requires anything specific to systemd
so they can just always be included as part of the sources.


smcv (approved):

GeorgesStavracas:
Perhaps, as a follow up, we should add a CI job to build without systemd? So we could catch regressions like this more easily

mwleeds:
I don't think this is the right approach, we should do #735 instead.

PR #744, opened 2022-03-25
DynamicLauncher: fit short_description on one line

wjt:
Currently the word 'desktop' is considered a single-word paragraph, and
the short description is truncated in the TOC.


mwleeds (commented):
I think I'd rather just keep the description and remove the newline character, to be consistent with other portal descriptions

GeorgesStavracas (approved):
@mwleeds me too, LGTM after your force-push

PR #746, opened 2022-03-25
realtime: Fix crash with WebKit processes

### LINKED ISSUE #709: Realtime portal crash when mapping pids
Note: this issue seems to be unrelated to https://github.com/flatpak/xdg-desktop-portal/commit/bceb762cffaf740b64c82a96b274d9d59948982e.

I noticed an xdg-desktop-portal 1.12.1 crash today:

```
Core was generated by `/usr/libexec/xdg-desktop-portal'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055f0372a0271 in map_pid_if_needed (app_info=<optimized out>, pid=<optimized out>, error=0x7f8dc206f728) at src/realtime.c:65
Downloading 0.00 MB source file /usr/src/debug/xdg-desktop-portal-1.12.1-1.fc35.x86_64/src/realtime.c
65	      g_warning ("Realtime error: %s", (*error)->message);
[Current thread is 1 (Thread 0x7f8dc2070640 (LWP 96629))]
(gdb) bt
#0  0x000055f0372a0271 in map_pid_if_needed (app_info=<optimized out>, pid=<optimized out>, error=0x7f8dc206f728)
    at src/realtime.c:65
#1  map_pid_if_needed (app_info=0x7f8dac003040, pid=pid@entry=0x7f8dc206f734, error=error@entry=0x7f8dc206f728)
    at src/realtime.c:58
#2  0x000055f0372a5dab in handle_make_thread_realtime_with_pid
    (object=<optimized out>, invocation=0x7f8db4016910, process=<optimized out>, thread=6, priority=5)
    at src/realtime.c:123
#3  0x00007f8dd0794c04 in ffi_call_unix64 () at ../src/x86/unix64.S:76
#4  0x00007f8dd0794108 in ffi_call
    (cif=cif@entry=0x7f8dc206f970, fn=fn@entry=0x55f0372a5d00 <handle_make_thread_realtime_with_pid>, rvalue=<optimized out>, avalue=avalue@entry=0x7f8dc206f880) at ../src/x86/ffi64.c:525
#7  0x00007f8dd0bbbaf4 in <emit signal ??? on instance ???>
    (node=<optimized out>, detail=0, instance=0x55f038429d70, emission_return=0x7f8dc206fba0, instance_and_params=0x7f8db8004e00) at ../gobject/gsignal.c:3781
    #5  0x00007f8dd0ba5bfd in g_cclosure_marshal_generic
    (closure=<optimized out>, return_gvalue=<optimized out>, n_param_values=<optimized out>, param_values=<optimized out>, invocation_hint=<optimized out>, marshal_data=<optimized out>) at ../gobject/gclosure.c:1534
    #6  0x00007f8dd0b9fc7f in g_closure_invoke
    (closure=0x55f03844d180, return_value=0x7f8dc206fb10, n_param_values=5, param_values=0x7f8db8004e00, invocation_hint=0x7f8dc206faf0) at ../gobject/gclosure.c:830
#8  0x000055f0372cddc4 in _xdp_realtime_skeleton_handle_method_call
    (connection=<optimized out>, sender=sender@entry=0x7f8db4015e30 ":1.1282", object_path=object_path@entry=0x55f0384e33d0 "/org/freedesktop/portal/desktop", interface_name=interface_name@entry=0x7f8db40145e0 "org.freedesktop.portal.Realtime", method_name=method_name@entry=0x7f8db4011b30 "MakeThreadRealtimeWithPID", parameters=parameters@entry=0x55f038433930, invocation=0x7f8db4016910, user_data=0x55f038429d70) at src/xdp-dbus.c:29856
#9  0x00007f8dd0d0b4ae in dispatch_in_thread_func
    (task=<optimized out>, source_object=<optimized out>, task_data=0x55f03841c800, cancellable=<optimized out>)
    at ../gio/gdbusinterfaceskeleton.c:540
#10 0x00007f8dd0c9950f in g_task_thread_pool_thread (thread_data=0x55f0383fbbf0, pool_data=<optimized out>)
    at ../gio/gtask.c:1430
#11 0x00007f8dd0ad3114 in g_thread_pool_thread_proxy (data=<optimized out>) at ../glib/gthreadpool.c:354
#12 0x00007f8dd0ad0842 in g_thread_proxy (data=0x55f0384caa40) at ../glib/gthread.c:827
#13 0x00007f8dd08b7a87 in start_thread (arg=<optimized out>) at pthread_create.c:435
#14 0x00007f8dd093c640 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

I have this open in gdb and can see that `*error = NULL`, so it sure looks like xdg_app_info_map_pids() failed without setting its error. But looking at the implementation of xdg_app_info_map_pids() and also map_pids(), that seems clearly impossible: I don't see any way they can fail to set the error when required. I'm stumped.

I do see one problem here: map_pid_if_needed() assumes that `error` is not NULL, which is not a great assumption. However, that's not the problem here, as the code never passes NULL, and I can see in gdb that it `error` is not NULL. (`*error`, however, is.)

TingPing:
Closes #709

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -1994,7 +1994,11 @@ xdg_app_info_load_brwap_info (XdpAppInfo *app_info,
   instance = xdp_app_info_get_instance (app_info);
 
   if (instance == NULL)
-    return 0;
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Could not find instance-id in process's /.flatpak-info");
+      return 0;
+    }
 
   path = g_build_filename (g_get_user_runtime_dir (),
                            ".flatpak",


TingPing:
Will need to backport to 1.12 also.

GeorgesStavracas:
There seems to be a stray merge commit in this PR, mind dropping it?

src/xdp-utils.c
```diff
@@ -1994,7 +1994,11 @@ xdg_app_info_load_brwap_info (XdpAppInfo *app_info,
   instance = xdp_app_info_get_instance (app_info);
 
   if (instance == NULL)
-    return 0;
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Could not find instance-id in processes /.flatpak-info");
```
mcatanzaro, line 1999: process's

mcatanzaro, line 1999:
```suggestion
                           "Could not find instance-id in process's /.flatpak-info");
```

mwleeds (approved):
LGTM. If this is causing a crash in the wild it would be nice to understand why we're failing to find the `instance-id`. It looks like this code path is only hit when the `XdpAppInfo` is for a Flatpak, so the only other reason I can think of is if the version of Flatpak is too old. I do see this note above the docs for `instance-id` in `doc/flatpak-metadata.xml` but 0.6 is quite ancient:
```
            <!-- In versions prior to 0.6.10 some of this information was
            in [Application], but for simplicity that isn't documented
            here. -->
```

TingPing:
Its due to WebKit's sandbox. This shouldn't be hit by Flatpak applications.

A patch fixing that upstream is here: https://bugs.webkit.org/show_bug.cgi?id=238403

src/xdp-utils.c
GeorgesStavracas: I'm not a native speaker, but in this case, wouldn't the appropriate form be `process'` (without the extra `s`)?

mwleeds:
> There seems to be a stray merge commit in this PR, mind dropping it?

I would think the "Rebase and Merge" button would take care of that, but I agree the GitHub UI giving you a way to merge in commits from the base branch with a merge commit is questionable.

TingPing:
I've backported this. Since it is a crash triggered by just using most WebKitGTK applications it would be a good idea to release a minor update soon.

src/xdp-utils.c
mcatanzaro: Honestly nobody agrees, so both ways are OK. I like `s's` because it is unambiguously singular, whereas `s'` can be unclear. [This page](https://oneminuteenglish.org/en/possessive-nouns-ending-in-s/) that I just found argues that `process'` is better because the word has more than one syllable. I've never even heard of that rule before. Then [this page](https://www.thesaurus.com/e/grammar/whats-the-rule-for-doing-a-possessive-after-the-word-s/) says the monosyllable rule only applies to historical and biblical names. I've *really* never heard of that rule before! So I've learned multiple new things about possessive nouns today....

When I was younger, I mostly saw people writing `s'` and that is what my teachers used in school, but they also didn't use Oxford commas, which nowadays would be a quite unfashionable omission, so I don't think I'll take style advice from them. :P Nowadays, I think I see `s's` more commonly than before, but maybe I'm totally wrong about that.

PR #750, opened 2022-03-27
Fix compilation of rewrite-launchers on older GLib versions, such as in Debian 11

smcv:
* Move g_string_replace() backport to a separate translation unit
    
    This will make it simpler to use in rewrite-launchers without
    unnecessarily pulling in libsystemd.

* rewrite-launchers: Provide backported g_string_replace()
    
    This fixes compilation on Debian 11.


smcv:
> This will make it simpler to use in rewrite-launchers without unnecessarily pulling in libsystemd

Obviously libsystemd is beneficial for xdp-utils and for the main xdg-desktop-portal service, but it's unnecessary for rewrite-launchers, which just needs to manipulate some .desktop files.

xdp-utils also pulls in sd-escape.[ch], and I think also json-glib, but I didn't get as far as confirming this, because this seemed a better solution than linking rewrite-launchers to libraries that it doesn't actually use.

PR #752, opened 2022-03-28
l10n: update Indonesian translation

rmnscnce:
This PR updates the Indonesian translation by fixing a couple of grammatical errors and changing some words to better fit the context*.

*) While `nyalakan` is indeed a correct substitute for `turn on` in English, it implies ignition as a part of the process, whereas nothing is (hopefully) being ignited as components like the speakers are being turned on. Therefore, `hidupkan` (lit. `to put alive`) is most likely a better fit.


GeorgesStavracas:
@cho2 what do you think?

cho2:
go ahead with this PR

GeorgesStavracas:
Thanks

PR #756, opened 2022-03-29
Change generated D-Bus code namespace to XdpDbus(Impl)

jadahl:
Tests compile xdp-dbus.c and xdp-impl-dbus.c, and also link to
libportal. This, as is, isn't possible without symbol conflicts, since
xdp-dbus.c share the same symbol namespace with libportal, meaning it's
not possible to use any types that have the same name, e.g. XdpSession,
as it's likely that the type instance from xdp-dbus.c will be used
instead of the one from libportal.

Fix this by changing the namespace of code generated with gdbus-codegen
to XdpDbus and XdpDbusImpl to avoid these conflicts. The Impl part is
changed too to make things consistent.

------

This was causing issues with tests added by @whot for the input capture portal, where `g_object_new (XDP_TYPE_SESSION, NULL)` failing due to `XDP_TYPE_SESSION` ending up in `xdp_session_get_type()` in `xdp-dbus.c`, instead of the one in libportal.

### DIFF
--- a/src/Makefile.am.inc
+++ b/src/Makefile.am.inc
@@ -22,7 +22,7 @@ BUILT_SOURCES += $(xdp_dbus_built_sources) $(xdp_impl_dbus_built_sources) $(geoc
 $(xdp_dbus_built_sources) : $(PORTAL_IFACE_FILES)
 	$(AM_V_GEN) $(GDBUS_CODEGEN)                            \
 		--interface-prefix org.freedesktop.portal.      \
-		--c-namespace Xdp                               \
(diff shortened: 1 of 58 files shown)


whot (approved):
Well, that was fun to review :smile: thanks heaps, this LGTM

GeorgesStavracas (approved):
What a ride

jadahl:
> What a ride

Was a weird thing to debug. What gave it away was `break xdp_session_get_type` saying it found two places to break.

PR #762, opened 2022-04-01
remote-desktop: add the ability to communicate via an EIS socket

whot:
This is intended as replacement for the `NotifyFoo` methods. libei
provides a more flexible and powerful method of sending input events to
the compositor.

A new method `ConnectToEIS` requests a file descriptor from the compositor
which can then be plugged into libei.

Once established, the communication between compositor and application
is direct, without the need to go through the portal process(es).

To avoid ambiguities between NotifyFoo and input events sent via libei,
any application that uses an EIS connection cannot use the NotifyFoo
methods.

One notable comment: I decided to pass the EIS connection down through the `impl.portal` instead of pipewire's approach of having the portal connect directly. Unlike the single pipewire daemon, an EIS implementation can be provided by multiple processes, so I figured it's easier to pass this down as close as possible to the entity that knows who the EIS implementation 

Eventually (and once libreis is ready) this will have the hooks to auto-set the app-id on the EIS connection as well though strictly speaking it's not necessary since the app-id is passed down to whoever needs to set up the socket anyway.

Noteworthy: there is no keysym event in libei at this point, so there's no equivalent to NotifyKeyboardKeysym

cc @jadahl

### DIFF
--- a/data/org.freedesktop.impl.portal.RemoteDesktop.xml
+++ b/data/org.freedesktop.impl.portal.RemoteDesktop.xml
@@ -351,6 +351,26 @@
       <arg type="u" name="slot" direction="in"/>
     </method>
     <!--
+        ConnectToEIS:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @app_id: App id of the application
+        @options: Vardict with optional further information
+        @fd: A file descriptor to an EIS implementation that can be passed to a
+        sender libei context
+
+        Request a connection to an EIS implementation.
+
+        This method is available in version 2 of this interface.
+    -->
+    <method name="ConnectToEIS">
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="h" name="fd" direction="out"/>
+    </method>
+    <!--
         AvailableDeviceTypes:
 
         A bitmask of available source types. Currently defined types are:
--- a/data/org.freedesktop.portal.RemoteDesktop.xml
+++ b/data/org.freedesktop.portal.RemoteDesktop.xml
@@ -396,6 +396,37 @@
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="u" name="slot" direction="in"/>
     </method>
+
+    <!--
+        ConnectToEIS:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @options: Vardict with optional further information
+        @fd: A file descriptor to an EIS implementation that can be passed to a libei sender context
+
+        Request a connection to an EIS implementation. The returned handle can
+        be passed to ei_setup_backend_fd() for a libei sender context to
+        complete the connection. All information about available devices and the
+        event flow is subject to libei. Please see the libei documentation for details.
+
+        This method may only be called once per session, where the EIS
+        implementation disconnects the session should be closed.
+
+        This method must be called after #org.freedesktop.portal.RemoteDesktop.Start()
+
+        Once an EIS connection is established, input events must be sent exclusively via
+        the EIS connection. Any events submitted via NotifyPointerMotion,
+        NotifyKeyboardKeycode and other Notify* methods will return an error.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="ConnectToEIS">
+      <annotation name="org.gtk.GDBus.C.Name" value="connect_to_eis"/>
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="h" name="fd" direction="out"/>
+    </method>
+
     <!--
         AvailableDeviceTypes:
 
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -774,7 +774,7 @@ handle_notify_pointer_motion (XdpDbusRemoteDesktop *object,
       g_dbus_method_invocation_return_error (invocation,
                                              G_DBUS_ERROR,
                                              G_DBUS_ERROR_FAILED,
-                                             "Session doesn't have access to a device of type: pointer");
+                                             "Session is not allowed to call NotifyPointer methods");
       return G_DBUS_METHOD_INVOCATION_HANDLED;
     }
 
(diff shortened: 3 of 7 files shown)


whot:
ftr, we have a working (though not yet complete) use-case in this [input-leap PR](https://github.com/input-leap/input-leap/pull/1524)

src/remote-desktop.c
```diff
@@ -1350,6 +1355,116 @@ handle_notify_touch_up (XdpDbusRemoteDesktop *object,
...
+                                                             NULL,
+                                                             &error))
+    {
+      g_warning ("Failed to ConnectToEIS: %s", error->message);
+      out_fd_list = g_unix_fd_list_new ();
+    }
+
+  /* FIXME: libreis_set_name() and friends */
+
+  remote_desktop_session->uses_eis = TRUE;
```
salmanmlk, line 1462: Will it make sense to do this only if the backend call succeeded so that we can allow the existing/slower input injection implementation as the fallback?

whot: yes, good point, I'll add that to my list

whot: done in the lastest commit now.

whot:
Marking this as ready. There's no immediate libei dependency other than the socket being passed, so even if libei takes another few weeks for the official "it's stable" API this bit is ready.

Squashed the commits together, there is no purpose keeping the various bugfixes we had in the git history after the merge.

whot:
ftr, libei is now declared stable. Not that it matters for this PR since we don't actually do anything with libei directly.

tests/test_remotedesktop.py
```diff
@@ -78,3 +80,92 @@ def test_remote_desktop_create_session_signal_closed(self):
...
+            parent_window="",
+            options=options,
+        )
+        assert response.response == 0
+
+        with self.assertRaises(dbus.exceptions.DBusException) as cm:
+            fd = rd_intf.ConnectToEIS(
+                session.handle, dbus.Dictionary({}, signature="sv")
+            )
+        assert "Purposely failing ConnectToEIS" in cm.exception.get_dbus_message()
```
jadahl, line 171: A potential extra test is

```python
        with self.assertRaises(dbus.exceptions.DBusException) as cm:
            fd = rd_intf.NotifyKeyboardKeycode(
                session.handle, dbus.Dictionary({}, signature="sv", 30, 1)
            )
        assert "Purposely failing to use old event notification" in cm.exception.get_dbus_message()
```

data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -23,7 +23,7 @@
 
       The Remote desktop portal allows to create remote desktop sessions.
 
-      This documentation describes version 1 of this interface.
+      This documentation describes version 2 of this interface.
```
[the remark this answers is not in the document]
whot: git rebase to the rescue!

data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -385,6 +385,37 @@
...
+        event flow is subject to libei. Please see the libei documentation for details.
+
+        This method may only be called once per session, where the EIS
+        implementation disconnects the session should be closed.
+
+        This method must be called after #org.freedesktop.portal.RemoteDesktop.Start()
+
+        Once an EIS connection is established, input events must be sent exclusively via
+        the EIS connection. Any events submitted via NotifyPointerMotion,
+        NotifyKeyboardKeycode and other Notify* methods will be silently discarded.
```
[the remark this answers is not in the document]
whot: fixed, thanks

tests/test_remotedesktop.py
whot: Added a terst for all the `NotifyFoo` funcs to return an error. fwiw, I'm not super-happy with the error message but I'd rather that minor bit doesn't hold up this PR

tests/test_remotedesktop.py
```diff
@@ -78,3 +78,154 @@ def test_remote_desktop_create_session_signal_closed(self):
...
+                func = getattr(rd_intf, notifyfunc["name"])
+                assert func is not None
+                func(
+                    session.handle,
+                    dbus.Dictionary({}, signature="sv"),
+                    *notifyfunc["args"]
+                )
+            # Not the best error message but...
+            assert (
+                "Session doesn't have access to a device of type"
```
jadahl, line 229: Maybe the error should be tweaked to be less specific, e.g. just  "Not allowed"?

jadahl (approved):
All lgtm, I think this is ready to land. Can consider tweaking the error message, but nut super urgent.

tests/test_remotedesktop.py
whot: Follow-up commit changes this to `Session is not allowed to call NotifyPointer methods` and the same for `Keyboard` and `Touch`

data/org.freedesktop.impl.portal.RemoteDesktop.xml
```diff
@@ -350,6 +350,26 @@
...
+        ConnectToEIS:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @app_id: App id of the application
+        @options: Vardict with optional further information
+        @fd: A file descriptor to an EIS implementation that can be passed to a
+        sender libei context
+
+        Request a connection to an EIS implementation.
+
+        This method is available in version 2 of this interface.
```
GeorgesStavracas, line 363: Adding new methods require bumping the interface version, I think - so that apps can then decide whether or not to call this method solely based on the version number.

src/remote-desktop.c
```diff
@@ -1374,6 +1476,8 @@ remote_desktop_iface_init (XdpDbusRemoteDesktopIface *iface)
   iface->handle_notify_touch_down = handle_notify_touch_down;
   iface->handle_notify_touch_motion = handle_notify_touch_motion;
   iface->handle_notify_touch_up = handle_notify_touch_up;
+
```
GeorgesStavracas, line 1479: Please remove this newline

GeorgesStavracas (changes requested):

data/org.freedesktop.impl.portal.RemoteDesktop.xml
jadahl: The version was bumped by the clipboard remote desktop integration pull request.

GeorgesStavracas: Ah my bad then.

src/remote-desktop.c
GeorgesStavracas: Let's not block on this

GeorgesStavracas (approved):

src/remote-desktop.c
whot: ftr, newline is there to maintain existing grouping, `create/select/start` are in one group and the various `notify` functions in another. Happy to remove it if you want me to though.

GeorgesStavracas: Nah, it's all good.

PR #764, opened 2022-04-06
Add a dbusmock-based Python test suite

### LINKED ISSUE #761: RFC: python test suite?
I'm just leaving this here because otherwise it'll go under: in #714 I've implemented a python-dbusmock-based test suite for portal testing. It's not a 100% overlap to the existing test suite (it's a bit hard to decipher what that all does, I'd need more time for this). But writing actual tests against a portal is significantly simpler than the current glib test suite approach.

See [this commit](https://github.com/flatpak/xdg-desktop-portal/pull/714/commits/3d9ccdae5c51c9857fca7c85ea145c0a212cb855) for the addition of the test suite + the InputCapture portal tests. Executed via `pytest` in the git tree.

I'd love to hear some comments about viability, etc.

whot:
For complex portals like the InputCapture portal, the current test suite
is difficult to use and adapt. A much simpler approach is to use
dbusmock to provide the impl.portal emulation, use Python to poke at the
actual DBus API of the portal and have xdg-desktop-portal sit in
between to negotiate the two.

This patch adds a test suite that provides enough infrastructure to add
tests for other portals in a similar manner - luckily the impl side is
relatively trivial to implement in dbusmock.

Currently implemented are the equivalents to the tests/email.c and
tests/trash.c tests. The email tests is missing the test for parallel
ComposeEmail calls though.

Easiest invoked via pytest, one of XDP_UNINSTALLED or LIBEXECDIR must be
set to run from the git tree or using the installed xdg-desktop-portal,
respectively.

Fixes #761

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -199,6 +199,28 @@ test(
   protocol: 'tap',
 )
 
+pytest = find_program('pytest-3', 'pytest', required: false)
+pymod = import('python')
+python = pymod.find_installation(
+  'python3',
+  modules: ['dbus', 'dbusmock', 'gi'],
+  required: false,
+)
+
+if pytest.found() and python.found()
+  test_env = environment()
+  test_env.set('XDP_UNINSTALLED', '1')
+
+  test(
+    'pytest',
+    pytest,
+    args: ['--verbose', '--log-level=DEBUG'],
+    env: test_env,
+    workdir: meson.project_source_root(),
+    suite: ['pytest'],
+  )
+endif
+
 if enable_installed_tests
   # autotools used to symlink to the host files, here we just install our version
   install_data(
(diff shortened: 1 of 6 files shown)


3v1n0:
Are you adding also the bits required to get this running in CI?

whot:
tbh, not until #717 is merged, adding this to autotools is too painful, with meson it's a few lines and then it's just part of the `meson  test` invocation.

GeorgesStavracas (commented):
I like what I see so far, but I'd like to see CI integration of these tests. Let's land https://github.com/flatpak/xdg-desktop-portal/pull/717 first, then think about this one

tests/meson.build
```diff
@@ -199,6 +199,29 @@ test(
   protocol: 'tap',
 )
 
+if meson.version().version_compare('>= 0.56.0')
```
GeorgesStavracas, line 202: I don't think this check is required, since xdg-desktop-portal already depends on meson >= 0.56.2

GeorgesStavracas: Took the liberty to fix this myself

Issue #686, opened 2021-12-27
Add API to know whether a certain portal is really available


nielsdg:
There's a ton of different APIs that is part of xdg-desktop-portal. This means that some DEs might not have always implemented the complete set that is part of the latest XDG portals release.

As an application developer, I'm interested in knowing which APIs specifically are supported by the current session, so that I may deal with it in an appropriate way (e.g. making a "Screenshot"-button insensitive, or trying a X11-based fallback if we're running in an X11 session).

Some use cases I can think of when checking for support of a specific XDG portal feature:

1. I can check if the `org.freedesktop.portal.Desktop` name is claimed on the session bus (and ping it), to know that xdg-desktop-portal is installed on the system, and that it's running
2. I can introspect the `org.freedesktop.portal.Desktop` service, to make sure e.g. the screenshot portal exists in the version that is installed
3. I can't query if the impl of a specific portal exists, as the name on the bus is specific per implementation, and I ideally don't have to care about internal implementation of portals
4. I can't query if a certain feature of a portal exists (e.g. having a restore token available in the screencast portal)

At this point, I'm mostly (only) interested in solving point 3

mwleeds:
> I can't query if a certain feature of a portal exists (e.g. having a restore token available in the screencast portal)

I think you can by checking the `version` attribute: https://github.com/flatpak/xdg-desktop-portal/blob/9aad2ffc5b3da8af0f6fa842ca7121f69f1740d8/data/org.freedesktop.portal.ScreenCast.xml#L314

mwleeds:
I was asking about this functionality the other day and @TingPing said "I'm not against new API to query this" but it's not clear to me if that API should be in x-d-p or libportal. The use case I'm interested in is Epiphany being able to hide UI based on whether a dynamic launcher portal is available (#681).

TingPing:
libportal has no way of knowing on its own, xdp has to expose it somehow.

GeorgesStavracas:
Time for a meta-portal that allows querying which portals are implemented? :slightly_smiling_face:

TingPing:
The API I was thinking of was something like (don't focus on the name): 

`bool org.freedesktop.portal.PortalInfo.IsPortalSupported(string portal_name, uint portal_version)`

This might be annoying to implement, like we don't want to just hardcode a list that gets outdated, and have to be in each backend.

mwleeds:
> The API I was thinking of was something like (don't focus on the name):
> 
[...]

PR #766, opened 2022-04-10
screenshot: Add Impl:version property and query it

nielsdg:
Right now, we always assume screenshot portal implementations provide a
`PickColor()` implementation, which is not necessarily true.

This commit adds a `version` property to the impl that we can use to
propagate for our own D-Bus interface. That way, clients can properly
check whether the PickColor API is available.

Fixes https://github.com/flatpak/xdg-desktop-portal/issues/686 for me


nielsdg:
See also https://github.com/flatpak/xdg-desktop-portal-gtk/pull/378

PR #770, opened 2022-04-11
[WIP] ScreenCast: introduce app_id and window_title options

grulja:
Some applications, like Chromium, can optimize screen sharing performance when sharing its own window.

CC: @jadahl


swick:
What's the use case here? Why would Chromium use ScreenCast to share its own window?

grulja:
> What's the use case here? Why would Chromium use ScreenCast to share its own window?

The idea was that, similar to tab sharing, they can make it to perform better than PipeWire screen sharing. Anyway, I don't think this is something they are planning to implement in the near feature. It was just an idea that came to their mind during our regular meetings we used to have. I think this can be closed for now.

Issue #771, opened 2022-04-13
xdg-desktop-portal crashes when using ScreenCast / Remote desktop portal


JonasE-TV:
When using the [experimental Wayland support of TeamViewer](https://community.teamviewer.com/English/discussion/122410/teamviewer-support-on-wayland-experimental-state-%F0%9F%A7%AA), xdg-desktop-portal segfaults pretty regularly (tested with Gnome).

This seems to happen for more recent versions of xdg-desktop-portal:

- crashes on Ubuntu 22.04 Beta - xdg-desktop-portal 1.14.2-1ubuntu1 ([bug report on launchpad](https://bugs.launchpad.net/ubuntu/+source/xdg-desktop-portal/+bug/1966945))
- crashes on Fedora 35 - xdg-desktop-portal 1.12.3-1.fc35
- seems not to crash on fresh Fedora 35 installation - xdg-desktop-portal 1.10.1-1.fc35
- seems not to crash on: Debian 11 - xdg-dekstop-portal 1.8.1-1

Not sure whether this is really an xdg-desktop-portal issue or glib/gslice, but I am happy to provide further information/help if possible.

Backtrace:
```
[luser@fedora ~]$ coredumpctl gdb 2686
           PID: 2686 (xdg-desktop-por)
           UID: 1000 (luser)
           GID: 1000 (luser)
        Signal: 11 (SEGV)
     Timestamp: Mon 2022-04-11 15:25:30 CEST (38min ago)
  Command Line: /usr/libexec/xdg-desktop-portal
    Executable: /usr/libexec/xdg-desktop-portal
 Control Group: /user.slice/user-1000.slice/user@1000.service/session.slice/xdg-desktop-portal.service
          Unit: user@1000.service
     User Unit: xdg-desktop-portal.service
         Slice: user-1000.slice
     Owner UID: 1000 (luser)
       Boot ID: 8b79b2ca196d48159bcd90f1ceb0dd2d
    Machine ID: d4ac315bdd7041a5b92ee66dd4a3072d
      Hostname: fedora
       Storage: /var/lib/systemd/coredump/core.xdg-desktop-por.1000.8b79b2ca196d48159bcd90f1ceb0dd2d.2686.1649683530000000.zst (present)
     Disk Size: 462.8K
       Message: Process 2686 (xdg-desktop-por) of user 1000 dumped core.
                
[...]

PR #772, opened 2022-04-14
Disable GSlice

GeorgesStavracas:
(Warning: this was only compile-tested)

GSlice is practically unmaintained, and these days various malloc
implementations have caught up to GSlice. There have been reports of
GSlice crashing on multithreaded setups for many years now too [1].

Having the allocator crash an app is already really bad, but crashing
a critical user session service like xdg-desktop-portal is even worse,
since it provides important funcionality for a desktop environment to
work.

Force the GSlice allocator to always use malloc internally.

Closes https://github.com/flatpak/xdg-desktop-portal/issues/771

[1] https://gitlab.gnome.org/GNOME/glib/-/issues/1079


GeorgesStavracas:
Switched to a draft. Please don't merge yet.

PR #774, opened 2022-04-14
remote-desktop: Make relationship with screen-cast more clear

mwleeds:


data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -34,13 +34,15 @@
...
 
-        To also get a screen content, call the
-        org.freedesktop.portal.ScreenCast.SelectSources() with the
-        #org.freedesktop.portal.Session object created with this method.
+        A remote desktop session may only be started and stopped with this interface,
+        but you can use the #org.freedesktop.portal.Session object created with this
+        method together with certain methods on the #org.freedesktop.portal.ScreenCast
+        interface. Specifically, you can call
+        org.freedesktop.portal.ScreenCast.SelectSources() to also get screen content,
+        or org.freedesktop.portal.ScreenCast.OpenPipewireRemote() to acquire a file
```
GeorgesStavracas, line 44: `or` → `and`

I think it's worth adding something like `See #org.freedesktop.portal.ScreenCast for more information on how to operate it`, because one cannot call OpenPipeWireRemote() without calling SelectSources() first.

GeorgesStavracas (changes requested):

GeorgesStavracas (approved):

PR #778, opened 2022-04-19
background-portal: Stop sending app-id quoted

sophie-h:
Since #696, the arguments sent to background portal implementations
contained the app-id in quoted form. Implementations, in turn, created
invalid autostart files.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -372,10 +372,14 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
 
           g_ptr_array_add (args, g_strdup_printf ("--command=%s", quoted_command));
 
-          /* Always quote the app ID to make rewriting the file simpler in case
-           * the app is renamed.
+          /* Always quote the app ID if quote_escape is enabled to make
+           * rewriting the file simpler in case the app is renamed.
            */
-          g_ptr_array_add (args, g_shell_quote (app_info->id));
+          if (quote_escape)
+            g_ptr_array_add (args, g_shell_quote (app_info->id));
+          else
+            g_ptr_array_add (args, g_strdup (app_info->id));
+
           for (i = 1; commandline[i]; i++)
             g_ptr_array_add (args, maybe_quote (commandline[i], quote_escape));
         }


sophie-h:
While 28ec93ccc2cc56b664edbdf37d96bc83fe6bd048 mentions

> First parse the argv so it's unquoted then quote any arguments that need
it. Don't do so in the background portal since the backend handles it.

I think the commit misses excluding the app-id from quoting for the background portal.

src/xdp-utils.c
```diff
@@ -372,10 +372,15 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
...
-           * the app is renamed.
+          /* Always quote the app ID if quote_escape is enabled to make
+           * rewriting the file simpler in case the app is renamed.
            */
-          g_ptr_array_add (args, g_shell_quote (app_info->id));
+          if (quote_escape) {
+            g_ptr_array_add (args, g_shell_quote (app_info->id));
+          } else {
+            g_ptr_array_add (args, g_strdup (app_info->id));
+          }
```
smcv, line 382: The actual code change looks fine to me.

Flatpak uses GNU-style indentation:

```suggestion
          if (quote_escape)
            g_ptr_array_add (args, g_shell_quote (app_info->id));
          else
            g_ptr_array_add (args, g_strdup (app_info->id));
```

or if one of the blocks needed more than one line, it would look like this:

```suggestion
          if (quote_escape)
            {
              g_ptr_array_add (args, g_shell_quote (app_info->id));
            }
          else
            {
              g_ptr_array_add (args, g_strdup (app_info->id));
            }
```

sophie-h:
Thanks for fixing the coding style!

sophie-h:
Thanks for fixing the coding style!

sophie-h:
@mleeds @GeorgesStavracas You authored/merged the original changes. Maybe you can have a look at this?

As I see it the implications of this bug are that, for example, background processes for scheduled backups are not started. IIRC the autostart file is never rewritten. That means if anyone sets up the app with 1.14.0 to 1.14.3 the autostart file might never get automatically fixed. Therefore it feels somewhat urgent for me to release this fix.

sophie-h:
> Please squash the second commit with the first one. LGTM otherwise.

Done :)

PR #786, opened 2022-05-04
Add Persistence notification flags

3v1n0:
FDO Notifications had persistence flags, we don't support them in the portal, but applications may want to be able to still use them.

While transient notifications can be shown as they are, resident ones could instead be bothering an user, so add a further permission check in such case.

### DIFF
--- a/data/org.freedesktop.portal.Notification.xml
+++ b/data/org.freedesktop.portal.Notification.xml
@@ -98,7 +98,7 @@
             <term>persistence s</term>
             <listitem><para>
               The persistence of the notification, if server supports it.
-              Supported values: normal, transient.
+              Supported values: normal, transient, resident.
             </para></listitem>
           </varlistentry>
           <varlistentry>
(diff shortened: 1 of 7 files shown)


3v1n0:
I wanted to add tests, but sadly this is a bit complex at the moment because of the `xdp_app_info_is_host` checks.

So, I thought about defining a `XDP_APP_INFO_KIND_TEST` app that somewhat works, but requires further tests changes.

So, any suggestion here?

3v1n0:
@GeorgesStavracas hey, could you have a look at this?

3v1n0:
> So, I thought about defining a `XDP_APP_INFO_KIND_TEST` app that somewhat works, but requires further tests changes.

So, the paradigm to make things more testable is in #904, so once that lands this will be easily testable.

swick:
@3v1n0 is this still relevant with the Notifications v2 portal soon landing?

jsparber:
I think this can be closed since notification v2 has a new property `display-hint` which does the same thing as proposed here. See: https://github.com/flatpak/xdg-desktop-portal/pull/1298

GeorgesStavracas:
Thanks @jsparber for checking this. @3v1n0 as always feel free to reopen if you think this still applies!

PR #787, opened 2022-05-04
CI: Use Ubuntu 22.04 for tests

3v1n0:
It's stable now, so let's use it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -58,11 +58,29 @@ jobs:
         $RUN_CMD apt-get update
         $RUN_CMD apt-get upgrade -y
         $RUN_CMD apt-get install -y --no-install-recommends \
-          make automake autoconf libtool gettext autopoint ${{ matrix.compiler }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -121,10 +139,21 @@ jobs:
...
+          autopoint \
+          gcc \
+          gtk-doc-tools \
+          shared-mime-info \
+          desktop-file-utils \
+          libglib2.0-dev \
+          libjson-glib-dev \
+          libfontconfig1-dev \
+          libgdk-pixbuf2.0-dev \
+          libsystemd-dev
```
whot, line 156: at the risk of being unnecessarily annoying, but this would be nice as sorted list - the one above is but this one isn't

GeorgesStavracas:
Obsoleted by https://github.com/flatpak/xdg-desktop-portal/commit/05356b4e9c7a5c37945d204e30a06408a9eb115e

PR #790, opened 2022-05-05
CI: Run all FUSE tests

3v1n0:
While we were already running most of the tests requiring FUSE, we were skipping the tests driven by `tests/test-document-fuse.sh`, so add missing deps and increase timeouts to ensure we're running them all.

/cc @TingPing


.github/workflows/check.yml
```diff
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   DEBIAN_FRONTEND: noninteractive
-  TESTS_TIMEOUT: 5 # in minutes
+  TESTS_TIMEOUT: 15 # in minutes
```
3v1n0, line 7: To reduce runtime we could probably reduce the number of parallel tests in `test-document-fuse.sh`, when `TEST_IN_CI` is set.

3v1n0: I've done it, as it may just take too much time (up to 20 minutes in CI) to run it in normal mode, we can maybe add scheduled jobs that run this with normal timeouts?

GeorgesStavracas (approved):

GeorgesStavracas (approved):

PR #791, opened 2022-05-05
notification: Add platform-data (with activation_token) to ActionInvoked

3v1n0:
Since some time FDO Notifications introduced the ActivationToken signal
to ensure that clients will receive an X11-startup notification ID or a
wayland xdg-activation token in order to support stealing-focus
prevention protocols.

However this isn't supported by xdg-desktop-portal's.
So fill the gap by adding a further activation token.

While it could be added by as an optional parameter, it's just better to
make it explicit.

It needs libportal from https://github.com/flatpak/libportal/pull/88

Implemented in:
 - https://github.com/flatpak/xdg-desktop-portal-gtk/pull/380

### DIFF
--- a/data/org.freedesktop.impl.portal.Notification.xml
+++ b/data/org.freedesktop.impl.portal.Notification.xml
@@ -77,5 +77,43 @@
       <arg type="s" name="action"/>
       <arg type="av" name="parameter"/>
     </signal>
+
+  <!--
+        ActionInvoked2:
+        @app_id: App id of the application
+        @id: the application-provided ID for the notification
+        @action: the name of the action
+        @platform_data: array which will contain the platform data
+          such as desktop-startup-id.
+        @parameter: array which will contain the target parameter
+          for the action, if one was specified
+
+        Send to the application when a non-exported action is
+        activated.
+
+        Supported keys in the @platform_data vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem>
+              <para>
+              A token that can be used to activate windows or applications
+              when action is invoked.
+              </para>
+              <para>
+              This can be either an X11-style startup ID or a Wayland
+              xdg-activation token.
+              </para>
+            </listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <signal name="ActionInvoked2">
+      <arg type="s" name="app_id"/>
+      <arg type="s" name="id"/>
+      <arg type="s" name="action"/>
+      <arg type="a{sv}" name="platform_data"/>
+      <arg type="av" name="parameter"/>
+    </signal>
   </interface>
 </node>
--- a/data/org.freedesktop.portal.Notification.xml
+++ b/data/org.freedesktop.portal.Notification.xml
@@ -167,6 +167,42 @@
       <arg type="s" name="action"/>
       <arg type="av" name="parameter"/>
     </signal>
+
+    <!--
+        ActionInvoked2:
+        @id: the application-provided ID for the notification
+        @action: the name of the action
+        @platform_data: array which will contain the platform data
+          such as desktop-startup-id.
+        @parameter: array which will contain the target parameter
+          for the action, if one was specified
+
+        Send to the application when a non-exported action is
+        activated.
+
+        Supported keys in the @platform_data vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>activation_token s</term>
+            <listitem>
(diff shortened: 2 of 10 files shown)


3v1n0:
@matthiasclasen wdyt about this?

At the current time I decided to do an api-break on libportal too, otherwise it could be exposed as an another signal, but it looked overkill to me.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -165,6 +167,7 @@
     <signal name="ActionInvoked">
       <arg type="s" name="id"/>
       <arg type="s" name="action"/>
+      <arg type="s" name="activation_token"/>
```
jadahl, line 170: Doesn't this break all existing users of this signal?

To avoid that, we can create a `ActionInvoked2`, and while at it, make it pass a `a{sv}` to make things a bit more future proof.

smcv: > Doesn't this break all existing users of this signal?

Yes it does. (</dbus-maintainer>)

3v1n0: Yeah, indeed, reason why I bumped the protocol version.

I was quickly discussing this with @matthiasclasen, but indeed another versioned signal could be added instead.

But my point was that clients should check on version before.

@jadahl I agree could be just an `a{sv}` platform data though.

jadahl: Apps that didn't update will not check the version though, as from their perspective, version 2 doesn't exist. And as there is no "binding" that lets the portal send the right signature to old applications, they would just fail / crash.

3v1n0: This is fixed now, added the platform-data dict.

3v1n0:
I wanted to add a test for checking on the old interface too, but to use libportal there too, we'd need to make a way to override the proto version, and sadly that can't be done once we've the proxy up.

So can only test it manually.

A6GibKm:
Hello, whats the state of this PR? Does the gnome shell know what to do with the activation token, so I can try testing this?

data/org.freedesktop.impl.portal.Notification.xml
```diff
@@ -77,5 +77,43 @@
...
+              </para>
+            </listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <signal name="ActionInvoked2">
+      <arg type="s" name="app_id"/>
+      <arg type="s" name="id"/>
+      <arg type="s" name="action"/>
+      <arg type="a{sv}" name="platform_data"/>
```
jadahl, line 115: Perhaps name this `options`? It's what the equivalent thing is called in other portals.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -167,6 +167,42 @@
...
+              xdg-activation token.
+              </para>
+            </listitem>
+          </varlistentry>
+        </variablelist>
+    -->
+    <signal name="ActionInvoked2">
+      <arg type="s" name="id"/>
+      <arg type="s" name="action"/>
+      <arg type="a{sv}" name="platform_data"/>
```
jadahl, line 203: Same about calling it `options` for consistency.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -167,6 +167,42 @@
...
+      <arg type="av" name="parameter"/>
+    </signal>
     <property name="version" type="u" access="read"/>
```
ebassi, line 206: Can we not bump the version, here? If version == 2, then ActionInvoked has more parameters, otherwise it doesn't.

Unless we assume that all portal implementations aren't checking the version.

I really don't want to have methods called `Foo2`; that's a great way to cause legibility issues, as well as the potential for typos. Might as well call this signal with another name entirely.

jadahl: That's not how it works; the apps don't "bind" a version it's compatible with, so we cannot change the type signature.

How do you imagine "app1" that is using the current version of the API to continue to work with a new xdg-desktop-portal if the type signature changes?

I think we just have to live with the fact that there will be a "Foo2" here and there. It's also why it's good to more or less always have a vararg so one can add more properties like this without changing the type signature.

jsparber:
The user can now obtain an activation token since notification v2 is now merged. https://github.com/flatpak/xdg-desktop-portal/pull/1298 So i think this can be closed.

PR #793, opened 2022-05-07
Screenshot: Allow deleting the file

bilelmoussaoui:
In some context where the screenshot could contain sensitive information
like a QR code, it would make sense to allow the application that scans
the resulting screenshot to remove it just after it is done with it.

Use case: https://gitlab.gnome.org/World/Authenticator/-/issues/280

### DIFF
--- a/src/documents.c
+++ b/src/documents.c
@@ -32,6 +32,7 @@
 
 #include "xdp-dbus.h"
 #include "xdp-utils.h"
+#include "documents.h"
 #include "document-enums.h"
 
 static XdpDbusDocuments *documents = NULL;
@@ -53,10 +54,7 @@ init_document_proxy (GDBusConnection *connection)
 char *
 register_document (const char *uri,
                    const char *app_id,
-                   gboolean for_save,
-                   gboolean writable,
-                   gboolean directory,
-                   gboolean deletable,
+                   DocumentFlags flags,
                    GError **error)
 {
   g_autofree char *doc_id = NULL;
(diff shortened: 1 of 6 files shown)


src/documents.c
```diff
@@ -56,6 +56,7 @@ register_document (const char *uri,
                    gboolean for_save,
                    gboolean writable,
                    gboolean directory,
+                   gboolean deletable,
```
GeorgesStavracas, line 59: I think this reached a point where it's better to switch to flags, e.g.:

```c
typedef enum {
  DOCUMENT_FLAG_FOR_SAVE,
  DOCUMENT_FLAG_WRITABLE,
  DOCUMENT_FLAG_DIRECTORY,
  DOCUMENT_FLAG_DELETABLE,
} DocumentFlags;
```

bilelmoussaoui: I agree here, I will clean up the PR later today

matthiasclasen:
This makes sense to me, and I agree with switching to flags

GeorgesStavracas (changes requested):
I apparently forgot to submit this review

src/documents.h
```diff
@@ -22,14 +22,19 @@
 
 #include <gio/gio.h>
 
+typedef enum {
+  DOCUMENT_FLAG_NONE      = (1 << 0),
```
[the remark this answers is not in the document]
bilelmoussaoui: Thanks, fixed!

src/documents.h
```diff
@@ -22,13 +22,19 @@
 
 #include <gio/gio.h>
 
+typedef enum {
+  DOCUMENT_FLAG_NONE      = 0,
+  DOCUMENT_FLAG_FOR_SAVE  = (1 << 1),
+  DOCUMENT_FLAG_WRITABLE  = (1 << 2),
+  DOCUMENT_FLAG_DIRECTORY = (1 << 3),
+  DOCUMENT_FLAG_DELETABLE = (1 << 4),
```
GeorgesStavracas, line 30: Sorry for the persistence here, but now that NONE = 0, the first entry should start with `1 << 0`, and the rest should be updated

bilelmoussaoui: No worries, I am not very familiar with C "best practices" yet.

GeorgesStavracas (approved):

Issue #769, opened 2022-04-11
parse_app_info_from_snap() sets the app id to "snap.$AppName", but gnome-shell's access dialog compares it to $DesktopFile


oSoMoN:
(originally incorrectly reported [against gnome-shell](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2264))

When testing the [proposed WebExtensions portal](https://github.com/flatpak/xdg-desktop-portal/pull/705) with [Firefox packaged as a snap](https://phabricator.services.mozilla.com/D140803), the access dialog isn't displayed because the caller's appId is `"snap.firefox"`, whereas the desktop id that [gnome-shell compares it to](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/accessDialog.js#L143) is `"firefox_firefox"`.

Other portals also use the app id when calling into `xdp_impl_access_call_access_dialog_sync()`, so it looks like these calls will consistently be denied for applications running as snaps.

The implementation of `parse_app_info_from_snap()` parses the output of `snap routine portal-info $PID`, which for firefox looks like this:

    [Snap Info]
    InstanceName=firefox
    AppName=firefox
    DesktopFile=firefox_firefox.desktop
    HasNetworkStatus=false

So instead of setting the app id to `"snap." + $AppName`, it looks like the value of `$DesktopFile` (trimmed of the ".desktop" suffix) should be used. Not sure what assumptions this would break elsewhere though.

oSoMoN:
One of these assumptions is in `app_has_file_access()` in `document-portal/document-portal.c`:

```C
if (g_str_has_prefix (target_app_id, "snap."))
  {
    res = get_output (&error, "snap", "routine", "file-access",
                      target_app_id + strlen ("snap."), path, NULL);
  }
```

jhenstridge:
There is some background on the selection of IDs in this PR:

jhenstridge:
There is some background on the selection of the ID in PR https://github.com/flatpak/xdg-desktop-portal/pull/155. The main goals were to pick something that (a) would not conflict with IDs used by Flatpak apps, and (b) would represent the security domain of the client.

[...]

PR #798, opened 2022-05-12
xdp-utils: Include the application name in the SNAP AppInfo ID

3v1n0:
Snaps may contain mutiple applications in the whole snap, but we're
currently granting permissions and emitting IDs for the main snap in all
the cases, so that an application inside the snap could benefit of the
same permissions of all the others.

Other than providing wrong information about the application-id to our
customers, that may not be able to retrieve the desktop-id properly.

As per this, always include the application name in the snapped
applications ids.

It may help in future with #769

Also add support for migrating old permissions.

Ideally we may just include it to snaps that use sub-applications names, but I think it's safer to just use it in all the cases. Also because it would provide a better way to compute the original desktop ID to customers.

/cc @jhenstridge

### DIFF
--- a/src/permissions.c
+++ b/src/permissions.c
@@ -26,6 +26,55 @@
 
 static XdpDbusImplPermissionStore *permission_store = NULL;
 
+static gboolean
+maybe_check_base_snap_permissions (const char   *app_id,
+                                   const char   *table,
+                                   const char   *id,
+                                   char       ***out_permissions)
+{
+  const char *app_name_part;
+
+  /* We used to check permissions for the base snap, so if those were granted
+   * already we should just return those, migrating to the new name.
+   * But we do it only for the main application.
+   */
+  if (g_str_has_prefix (app_id, "snap.") &&
+      (app_name_part = strrchr (app_id, '_')))
+    {
+      g_autofree char *base_app_id = NULL;
+      const char *app_name = app_name_part + 1;
+      size_t app_name_part_len = strlen (app_name_part);
+      size_t app_len;
+
+      if (app_name_part_len < 2)
+        return FALSE;
+
+      app_len = strlen (app_id) - app_name_part_len;
+      if (!app_len)
+        return FALSE;
+
+      g_assert (strchr (app_id, '_') == app_name_part);
+      base_app_id = g_strndup (app_id, app_len);
+
+      if (!g_str_equal (app_name, base_app_id + strlen ("snap.")))
+        return FALSE;
+
+      *out_permissions = get_permissions_sync (base_app_id, table, id);
+
+      if (*out_permissions)
+        {
+          static const char **empty = { NULL };
+
+          set_permissions_sync (app_id, table, id, *out_permissions);
+          set_permissions_sync (base_app_id, table, id, empty);
+        }
+
+      return TRUE;
+    }
+
+  return FALSE;
+}
+
 char **
 get_permissions_sync (const char *app_id,
                       const char *table,
@@ -46,12 +95,19 @@ get_permissions_sync (const char *app_id,
     {
       g_dbus_error_strip_remote_error (error);
       g_debug ("No '%s' permissions found: %s", table, error->message);
+
+      if (maybe_check_base_snap_permissions (app_id, table, id, &permissions))
+          return g_steal_pointer (&permissions);
+
       return NULL;
     }
 
   if (!g_variant_lookup (out_perms, app_id, "^a&s", &permissions))
     {
       g_debug ("No permissions stored for: %s %s, app %s", table, id, app_id);
+
+      if (maybe_check_base_snap_permissions (app_id, table, id, &permissions))
+          return g_steal_pointer (&permissions);
 
       return NULL;
     }
@@ -142,12 +198,13 @@ get_permission_sync (const char *app_id,
                      const char *id)
 {
   g_auto(GStrv) perms = NULL;
+  Permission ret = PERMISSION_UNSET;
 
   perms = get_permissions_sync (app_id, table, id);
   if (perms)
-    return permissions_to_tristate (perms);
+    ret = permissions_to_tristate (perms);
 
-  return PERMISSION_UNSET;
+  return ret;
 }
 
 void set_permission_sync (const char *app_id,
--- a/src/permissions.c
+++ b/src/permissions.c
@@ -198,13 +198,12 @@ get_permission_sync (const char *app_id,
                      const char *id)
 {
   g_auto(GStrv) perms = NULL;
-  Permission ret = PERMISSION_UNSET;
 
   perms = get_permissions_sync (app_id, table, id);
   if (perms)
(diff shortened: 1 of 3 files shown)


3v1n0:
@jhenstridge can you maybe check this?

src/permissions.c
```diff
@@ -26,6 +26,55 @@
...
+                                   const char   *id,
+                                   char       ***out_permissions)
+{
+  const char *app_name_part;
+
+  /* We used to check permissions for the base snap, so if those were granted
+   * already we should just return those, migrating to the new name.
+   * But we do it only for the main application.
+   */
+  if (g_str_has_prefix (app_id, "snap.") &&
```
robert-ancell, line 41: Some functions here would make it much clearer what all the string processing is doing.

src/permissions.c
```diff
@@ -26,6 +26,55 @@
 
 static XdpDbusImplPermissionStore *permission_store = NULL;
 
+static gboolean
```
robert-ancell, line 29: Why not return GStrv and get rid of `out_permissions`?

src/permissions.c
```diff
@@ -142,12 +198,13 @@ get_permission_sync (const char *app_id,
                      const char *id)
 {
   g_auto(GStrv) perms = NULL;
+  Permission ret = PERMISSION_UNSET;
```
robert-ancell, line 201: The changes in this function don't seem to do anything - remove from the PR?

src/permissions.c
3v1n0: Well I liked more to check the return value and based on that react on caller function, but both are fine options.

src/xdp-utils.c
```diff
@@ -801,8 +840,15 @@ parse_app_info_from_snap (pid_t pid, GError **error)
...
+  app_name = g_key_file_get_string (metadata, SNAP_METADATA_GROUP_INFO,
+                                    SNAP_METADATA_KEY_APP_NAME, error);
+  if (app_name == NULL)
+    {
+      return NULL;
+    }
+
   app_info = xdp_app_info_new (XDP_APP_INFO_KIND_SNAP);
-  app_info->id = g_strconcat ("snap.", snap_name, NULL);
+  app_info->id = g_strconcat ("snap.", snap_name, "_", app_name, NULL);
```
jhenstridge, line 851: I think changing the ID like this is a bad idea. The reason I only used the snap name in `app_info->id` was that there is effectively no isolation between different applications within the same snap package. If the user grants access to some information to one application within the snap, there is nothing to prevent that app from passing that information on to the second app. So it makes most sense to record the permissions at the level of the package, rather than application.

I know there are rough edges in the portal UI for snaps at present, but I think a lot of this is better handled by passing desktop file IDs to the impl services (since the impl service isn't generally making its own access decisions, and is usually expecting a desktop ID anyway).

src/xdp-utils.c
```diff
@@ -392,6 +392,44 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
...
+
+          g_ptr_array_add (args, g_strdup ("-c"));
+
+          joined_commands = g_strjoinv (" ", (GStrv) commandline);
+          g_ptr_array_add (args, g_shell_quote (joined_commands));
+        }
+
+      g_ptr_array_add (args, NULL);
+
       return (char **)g_ptr_array_free (g_steal_pointer (&args), FALSE);
```
jhenstridge, line 433: This code seems to be unrelated to the rest of PR's purpose. We'll need to sort out a snap implementation for this utility function, but it'd be better to handle that in a separate PR.

jhenstridge (changes requested):
I have concerns about this proposed change. Including the application name in the identifier used to make access decisions would only make sense if two applications in the same snap were meaningfully isolated from each other.

3v1n0:
> Do you still need this, Marco? The negative feedback above leads me to suspect this is probably not actually needed?

So, I was actually thinking at this and then I discovered you pinged me here... As in fact we are still in troubles, because we are still not able to identify what application did a request and many portals end up being broken in the snap world.

At the same time we have the issue that as per #155 we wanted to use snap IDs for permissions and desktop ID for impl requests, and this is getting complicated.

Thus, I'm still considering whether there are better options than this.

src/xdp-utils.c
3v1n0: > I think changing the ID like this is a bad idea. The reason I only used the snap name in app_info->id was that there is effectively no isolation between different applications within the same snap package. If the user grants access to some information to one application within the snap, there is nothing to prevent that app from passing that information on to the second app. So it makes most sense to record the permissions at the level of the package, rather than application.

Mhmh, I was re-reading this, but this is only partially true... For example, it's true that if a snap application opens a file through a portal and stores it in a snap-space the other app can read, then the security is compromised, but for what it concerns other per-process transient access (notifications, camera access), those requests are still confined to the requesting process, then if the process stores the data in a place that another application in the snap can read, then it's not anymore accessing to external resources, but to the snap-area resources that both applications must have the permission to access.

Now, I see the point of just avoiding to ask the user a request for something that the application may compromise within the snap, but I still feel it would provide more protection for the non malicious cases.

> I know there are rough edges in the portal UI for snaps at present, but I think a lot of this is better handled by passing desktop file IDs to the impl services (since the impl service isn't generally making its own access decisions, and is usually expecting a desktop ID anyway).

But isn't this still creating the same false security-feeling?

3v1n0:
I think we can close this as per #1764

src/xdp-utils.c
3v1n0: @jhenstridge I was discussing this again recently (even more), but...

> If the user grants access to some information to one application within the snap, there is nothing to prevent that app from passing that information on to the second app

This is potentially true, but this is the same issue also for the whole snap permissions design, no? A snap app that has `home` interface can share files in there to another app in the snap that has not such privilege (through common path for example), so why this should be different in portals?

AdrianVovk: > But isn't this still creating the same false security-feeling?

I'd also like to reinforce this point.

By passing the desktop-app-id into the backend you're just punting the problem onto the user. If it asks "Allow $APP to access $RESOURCE" then the user doesn't expect all the other apps in the same snap package to gain access to the same resource. At least if we don't grant the permission explicitly, the apps will have to work around the portals rather than just being directly granted permission by the portals.

I don't really see a way you'd be able to effectively communicate to the user that granting permissions to app A will also grant permissions to app B. And it's the same either doing it explicitly by making xdg-desktop-portal maintain the separate concept of a "permissions ID" (as in #1764), or implicitly through the weak isolation between apps.

So if the end result is the same (user is prompted to grant permission for some app A, but there's no way to prevent app B from using the same permission), then I'd rather xdg-desktop-portal be clueless about this. So we don't end up having to juggle additional concepts like "permission IDs" and instead live in a world where only app IDs exist. Especially because shipping multiple apps in a single snap package is actually something that's relatively rare, right?

PR #802, opened 2022-05-16
Stop xdg-desktop-portal when the session ends

### LINKED ISSUE #801: xdg-desktop-portal is not stopped when logging out of the desktop session
I'm having a problem caused by `xdg-desktop-portal` staying alive after the graphical session is closed.

If a new graphical session is started afterwards and has a different value for DISPLAY or XAUTHORITY (as [is the case with SDDM](https://github.com/sddm/sddm/blob/develop/src/common/XAuth.cpp#L74)) then some calls to the portal API can fail (I'm having problems with `OpenURI` in particular). The problem disappears if the user kills or restarts the `xdg-desktop-portal` process.

The GNOME session works around this problem by [restarting dbus on shutdown](https://gitlab.gnome.org/GNOME/gnome-session/-/blob/main/data/gnome-session-shutdown.target#L26), thus ensuring that the new desktop session will have a new portal instance. The comment on that same file suggests adding `PartOf=graphical-session.target` to the relevant services to fix this problem.

bertogg:
Fixes #801

### DIFF
--- a/src/xdg-desktop-portal.service.in
+++ b/src/xdg-desktop-portal.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=Portal service
+PartOf=graphical-session.target
 
 [Service]
 Type=dbus


GeorgesStavracas (approved):

GeorgesStavracas:
Thanks!

PR #811, opened 2022-06-09
utils: Fix typo

debarshiray:
On success lookup_ns_from_pid_fd() is meant to return 0, which is what
FALSE is also defined as.  Therefore, this typo could lead to
surprises in case of a programmer error.


GeorgesStavracas (approved):

GeorgesStavracas:
Thanks!

PR #816, opened 2022-06-20
Add a CI run for the meson build

whot:
Duplicate of the existing autotools run, just using meson instead for build/test/install. However - meson has its own TAP parser which is unhappy with the `G_MESSAGES_DEBUG='all'` output that floods stdout (see also d661854d4c6298976f61438cc6d9bbc1ac9a967f). Adding `g_log_writer_default_set_use_stderr(TRUE)` fixes that and that's probably fine for anything in `tests/`, but I also had to add this to the document portal and the permission store (afaict they're dbus-activated) where it's not fine. 

Any hints on how to convince glib to spew debug to stderr instead would be appreciated, there must be something obvious I'm missing here.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -75,8 +75,8 @@ jobs:
           libtool \
           llvm \
           make \
-          shared-mime-info \
-          python3-gi
+          python3-gi \
+          shared-mime-info
 
     - name: Check out xdg-desktop-portal
       uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


smcv:
> Adding g_log_writer_default_set_use_stderr(TRUE) fixes that and that's probably fine for anything in tests/, but I also had to add this to the document portal and the permission store (afaict they're dbus-activated) where it's not fine.

Why is this not fine for D-Bus-activated services? dbus-daemon runs D-Bus-activated services with both stdout and stderr connected to the systemd Journal, and I suspect dbus-broker does the same.

> Any hints on how to convince glib to spew debug to stderr instead would be appreciated

That's exactly why I added `g_log_writer_default_set_use_stderr()`.

whot:
> Why is this not fine for D-Bus-activated services?

Primarily because this would change the current behaviour of the document portal and the permission store for the sole purpose of the test suite. If that's fine, then nothing really stands in the way.

3v1n0:
Please, also take in mind the changes that are part of this: https://github.com/flatpak/xdg-desktop-portal/pull/790/files

whot:
Depends on #817 because right now the various thread sanitize jobs are failing. Otherwise this should be fine now with the meson builds running in the CI.

whot:
> Please, also take in mind the changes that are part of this: https://github.com/flatpak/xdg-desktop-portal/pull/790/files

No worries, switched this one back to Draft, will apply the changes once #790 is merged and #817 is fixed.

(Sorry, didn't see this until after I had typed my comment above and went to check my emails.)

whot:
Rebased on top of #790 and #817, this should be ready to go now

PR #817, opened 2022-06-22
CI: actually run the thread sanitizer jobs

whot:
The github syntax for matrix.include requires it to be a single value,
not an array. The result of this was that the value wasn't "thread" but
rather "Array" and never triggered the condition we had for it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,9 +14,6 @@ jobs:
       matrix:
         compiler: ['gcc', 'clang']
         sanitizer: ['address']
-        include:
-          - compiler: gcc
-            sanitizer: 'thread'
 
     env:
       UBUNTU_VERSION: impish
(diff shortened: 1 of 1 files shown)


whot:
Well, guess there are some thread sanitizier issues after all but I'm unsure how many of those are false positives or just don't matter, cc @3v1n0

3v1n0:
Oh, good catch, thanks!

GeorgesStavracas:
Gave my best to try and make this test pass, but I can't spent more time on this. @3v1n0 please fix the test, or drop it since it's not helping so far

whot:
I pushed a commit to remove them from the CI run to this branch - as separate patch. If @3v1n0 can fix the issues, then it's easy enough to just revert that commit and get them back.

GeorgesStavracas (approved):

GeorgesStavracas:
GitHub must be on a vendetta against Meson, I just can't merge these branches

whot:
I wonder if it's to do with the CI run? It's hanging at the now-removed targets. I remember github has some safety checks so as non-maintainer (developer?) you cannot add CI targets, maybe the same is true for removing them? Do you as maintainer see some way to cancel or approve that CI run?

GeorgesStavracas:
I don't see that, and I apparently don't have enough permissions on this repository to change its settings

PR #819, opened 2022-06-23
Adds the option to open a folder as writable

A6GibKm:
Many apps need to select a folder to store their related data, e.g. set
a Downloads folder for Fragments. This requires write permissions.

See https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/38.


A6GibKm:
Note that there is a validation for the options dictionary, any option which is not in the list will be ignored, in this case even if `writable=True` is passed it will be ignored without the proposed changes.

A6GibKm:
This MR was wrong. We were getting "writable" from the impl regardless of this.

Issue #820, opened 2022-06-23
xdg-desktop-portal fails when selecting a remote URI from a Flatpak/Snap


sergio-costas:
If the user has a remote drive mounted using Gvfs (using SFTP or SMB, for example), and a containerized application calls any of the methods in `org.freedesktop.portal.FileChooser`, the file chooser do show the remote units:

![Captura desde 2022-06-23 15-30-59](https://user-images.githubusercontent.com/105285003/175357855-e332e7dd-9348-4dc4-8431-36c5f55bf696.png)

If the user chooses a local file, the xdg-desktop-portal works as expected and "tunnels" the file:

![Captura desde 2022-06-23 15-36-21](https://user-images.githubusercontent.com/105285003/175358217-db76c9c3-2105-4fb9-b519-c3573d58be95.png)

But if the user chooses a file in the remote drive, the portal returns no URI:

![Captura desde 2022-06-23 15-33-53](https://user-images.githubusercontent.com/105285003/175358345-c1255a04-4011-4c9d-bd5f-13cf739ee350.png)

Also, `xdg-desktop-portal` shows an error in the terminal where it is running (of course, this requires to kill it previously and launch it from a terminal):

![Captura desde 2022-06-23 15-32-36](https://user-images.githubusercontent.com/105285003/175358562-4671d229-1ea4-44df-a7c6-2286650097fd.png)

The test code used (a python program that calls directly to the DBus API) and the JSON file to build a Flatpak is here:

[test_portal_python.zip](https://github.com/flatpak/xdg-desktop-portal/files/8969425/test_portal_python.zip)

But it also happens with Firefox in Flatpak: trying to save a page or a PDF in a remote drive mounted using Gvfs fails, and shows the same error in `xdg-desktop-portal`.

sergio-costas:
A proposed solution was to manually map any remote file to the corresponding Gvfs FUSE mount. This code is available at https://github.com/flatpak/xdg-desktop-portal/pull/814

With this code, it is possible to save PDF files from Firefox (both in Flatpak and Snap containers) in remote drives, without having to enable VFS in `xdg-desktop-portal`.

GeorgesStavracas:
The work moved to xdg-desktop-portal-gnome so let's track it there

Issue #213, opened 2018-08-27
ensure the file chooser only sends file: uris back


matthiasclasen:
So far, we've used GIO_USE_VFS=local to guarantee this, but we do want to let the user access gvfs. That means that we need to translate non-file uris to their local (fuse) equivalents, since we can't rely on  things like trash: or recent: to work inside the sandbox.

matthiasclasen:
A quick test shows that the file chooser already returns file: uris, at least for recent files.

hadess:
This little song and dance gets a file:/// URI for most gvfs backends:
https://gitlab.gnome.org/GNOME/totem/blob/master/src/backend/bacon-video-widget.c#L4045
`get_target_uri()` is the interesting one.

Be-ing:
I am stumbling over this as well. I am working on [rfd](https://github.com/PolyMeilex/rfd/), a cross platform Rust library abstracting over file dialogs on different OSes. On macOS and Windows, the file dialogs only return file paths. But the XDG Desktop Portal allows the file dialog to return arbitrary URIs. I can easily strip off the `file://` prefix, but if the user picks something with a different prefix, the best this cross platform abstraction can do is act as if no file was picked or ask the user to pick again.

lissyx:
Until this issue is fixed, UX should not show anything to the user: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/48

sergio-costas:
proposed patch in xdg-desktop-portal-gnome that fixes this: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/67

GeorgesStavracas:
This has been pushed into a documentation change. Backends are responsible for returning only `file://` URIs.

sergio-costas:
Thanks!

PR #822, opened 2022-06-27
documents: Add support for remote files

sergio-costas:
By default, xdg-desktop-portal runs with "GIO_USE_VFS=local" to guarantee that it only sends local URIs, but that also means that the user can't access files that are mounted locally using GVfs.

This patch filters all the URIs sent and replaces, wherever possible, any remote URI (like sftp://, smb://...) with the corresponding one in the local filesystem created by GVfs.

Fix https://github.com/flatpak/xdg-desktop-portal/issues/213
Fix https://github.com/flatpak/xdg-desktop-portal/issues/820

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -2349,3 +2349,67 @@ xdp_validate_serialized_icon (GVariant  *v,
 
   return TRUE;
 }
+
+#if G_ENCODE_VERSION (GLIB_MAJOR_VERSION, GLIB_MINOR_VERSION) >= G_ENCODE_VERSION (2, 66)
+gchar *
+xdp_transform_remote_uri_into_local (const gchar *uri)
+{
+  gchar *pathtext = NULL;
+  gchar *uristring = NULL;
+  gchar *fuse_mountpoint = NULL;
+  GUri *file_uri = NULL;
+  GString *gvfs_folder = NULL;
+
+  file_uri = g_uri_parse (uri, G_URI_FLAGS_NONE, NULL);
+  if (file_uri == NULL)
+    {
+      return g_strdup(uri);
+    }
+  if ((0 != g_strcmp0 (g_uri_get_scheme (file_uri), "file")) &&
+      (g_uri_get_host (file_uri) != NULL))
+    {
+      if (g_get_user_runtime_dir() == g_get_user_cache_dir ())
+        fuse_mountpoint = g_build_filename (g_get_home_dir(), ".gvfs", NULL);
+      else
+        fuse_mountpoint = g_build_filename (g_get_user_runtime_dir(), "gvfs", NULL);
+      gvfs_folder = g_string_new ("");
+      g_string_printf (gvfs_folder,
+                       "%s:host=%s",
+                       g_uri_get_scheme (file_uri),
+                       g_uri_get_host (file_uri));
+      if (g_uri_get_port (file_uri) != -1)
+        {
+          g_string_append_printf (gvfs_folder,
+                                  ",port=%d",
+                                  g_uri_get_port (file_uri));
+        }
+      if (g_uri_get_user (file_uri) != NULL)
+        {
+          g_string_append_printf (gvfs_folder,
+                                  ",user=%s",
+                                  g_uri_get_user (file_uri));
+        }
+      pathtext = g_build_filename (fuse_mountpoint,
+                                   gvfs_folder->str,
+                                   g_uri_get_path (file_uri),
+                                   NULL);
+      g_string_free (gvfs_folder, TRUE);
+      uristring = g_uri_join (G_URI_FLAGS_NONE,
+                              "file",
+                              NULL,
+                              NULL,
+                              -1,
+                                pathtext,
+                              NULL,
+                              NULL);
+      g_free (pathtext);
+    }
+  else
+    {
+      uristring = g_strdup (uri);
+    }
+  g_uri_unref (file_uri);
+
+  return uristring;
+}
+#endif
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -2354,61 +2354,50 @@ xdp_validate_serialized_icon (GVariant  *v,
 gchar *
 xdp_transform_remote_uri_into_local (const gchar *uri)
 {
-  gchar *pathtext = NULL;
+  g_autofree gchar *pathtext = NULL;
+  g_autofree gchar *fuse_mountpoint = NULL;
   gchar *uristring = NULL;
-  gchar *fuse_mountpoint = NULL;
-  GUri *file_uri = NULL;
-  GString *gvfs_folder = NULL;
+  g_autoptr(GUri) file_uri = NULL;
+  g_autoptr(GString) gvfs_folder = NULL;
 
   file_uri = g_uri_parse (uri, G_URI_FLAGS_NONE, NULL);
-  if (file_uri == NULL)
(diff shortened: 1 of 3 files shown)


sergio-costas:
I think that this patch is much better than the original.

sergio-costas:
@hadess What do you think about this proposal?

advancingu:
So what's the next step here? What needs to be done to get a merge decision?

lissyx:
Gentle ping?

jhenstridge:
I think this is something we want to support, but the implementation is the wrong approach. The better fix is probably to remove this code:

https://github.com/flatpak/xdg-desktop-portal/blob/750b9e5d6d7cdf451bb4f9f4fc97682fafa1a364/src/xdg-desktop-portal.c#L380-L381

This code in `documents.c`:

https://github.com/flatpak/xdg-desktop-portal/blob/750b9e5d6d7cdf451bb4f9f4fc97682fafa1a364/src/documents.c#L79-L80

... will already rewrite GVFS URIs to the corresponding FUSE file paths if we're not forcing `GIO_USE_VFS=local`:

```python
>>> file = Gio.File.new_for_uri('smb://elzar/media/foo.txt')
>>> file.get_path()
'/run/user/1000/gvfs/smb-share:server=elzar,share=media/foo.txt'
```

Outside of the `register_document` code, the only other place that uses GIO to with non-file URIs is the OpenURI portal. I think the code is fairly safe, since it codes its own lookup of non-local URIs to only search for scheme handlers. There is however a `g_file_new_for_uri` call within the `g_app_info_launch_uris` implementation. I think this is probably also safe, since it just seems to be formatting the URI to pass it on the command line to the application.

I think the root cause for the bug reports was this change in xdg-desktop-portal-gtk (which was then inherited by x-d-p-gnome):

https://github.com/flatpak/xdg-desktop-portal-gtk/commit/ce520f7bb4a60e669669402c9ffae9aeb4799943

The impl service started passing us GVFS URIs without us ever testing whether it worked. Given that change is now 4 years old, it's probably not worth trying to update the contract between x-d-p and the impl services to indicate whether VFS URIs are supported or not.

src/xdp-utils.c
```diff
@@ -2349,3 +2349,67 @@ xdp_validate_serialized_icon (GVariant  *v,
 
   return TRUE;
 }
+
+#if G_ENCODE_VERSION (GLIB_MAJOR_VERSION, GLIB_MINOR_VERSION) >= G_ENCODE_VERSION (2, 66)
+gchar *
+xdp_transform_remote_uri_into_local (const gchar *uri)
+{
+  gchar *pathtext = NULL;
```
robert-ancell, line 2357: Use `g_autofree gchar *` to remove the free calls below.

src/xdp-utils.c
```diff
@@ -2349,3 +2349,67 @@ xdp_validate_serialized_icon (GVariant  *v,
...
+  GUri *file_uri = NULL;
+  GString *gvfs_folder = NULL;
+
+  file_uri = g_uri_parse (uri, G_URI_FLAGS_NONE, NULL);
+  if (file_uri == NULL)
+    {
+      return g_strdup(uri);
+    }
+  if ((0 != g_strcmp0 (g_uri_get_scheme (file_uri), "file")) &&
+      (g_uri_get_host (file_uri) != NULL))
```
robert-ancell, line 2369: This would be easier to read if you returned on all the cases where this function doesn't work, and the contents of the function don't need to be indented then. i.e.

```
if (file_uri == NULL || g_strcmp0 (g_uri_get_scheme (file_uri), "file") == 0 || g_uri_get_host (file_uri) == NULL)
  {
    return g_strdup (uri);
  }

// Do all the remote URI code processing here
```

src/xdp-utils.c
```diff
@@ -2349,3 +2349,67 @@ xdp_validate_serialized_icon (GVariant  *v,
...
+    {
+      if (g_get_user_runtime_dir() == g_get_user_cache_dir ())
```
robert-ancell, line 2371: Use g_strcmp0 here? Comparing two pointers doesn't seem to be what you want here.

sergio-costas:
@jhenstridge I fully agree in that the right solution was to remove that. In fact, I proposed that too, but they were reluctant to remove it. That's why I prepared this other patch, at least as a temporary solution. But of course, if you think that it is safe to remove that line and fully enable GVFS, I will be very happy to burn this MR and bury the remains very, very deep :-)

src/xdp-utils.c
sergio-costas: OK, not so rookie... I just copy-pasted the code from Gvfs...

imagen

Should I send a patch?

src/xdp-utils.c
sergio-costas: Done with pathtext and fuse_mountpoint, and used g_autoptr for fileuri and gvfs_folder.

src/xdp-utils.c
sergio-costas: Done.

jhenstridge:
@sergio-costas: I think it is going to be hard to reproduce the GVFS logic, and even if we do it'd risk getting out of sync.

It also seems the path translation doesn't always match the logic in this PR. While it seems to be correct for e.g. accessing my camera via a `gphoto2://` URI, the smb URI I mentioned above is a bit different:

```python
>>> file = Gio.File.new_for_uri('smb://elzar/media/foo.txt')
>>> file.get_path()
'/run/user/1000/gvfs/smb-share:server=elzar,share=media/foo.txt'
```

Your code looks like it'd instead generate the file path `/run/user/1000/gvfs/smb:host=elzar/media/foo.txt`. This second path doesn't actually work to access the files, so it's not just a case of having multiple paths representing the same resource.

Even if we add special case logic for SMB, it's quite possible that there are yet more GVFS backends that don't follow the common pattern. At some point it must be easier to just turn gvfs back on, and ensure we are careful with URIs supplied by sandboxed applications (which it already looks like we do).

sergio-costas:
That's a good point, indeed.

src/xdp-utils.c
robert-ancell: Yes, you should probably open a MR for that in gvfs. This code may work because the pointers happen to be the same, but it seems like even if it works it could break at any point if glib changes the way it stores these values.

sergio-costas: Sent https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/151

advancingu:
Friendly ping to ask what needs to be done to get this moving again? The PR has been open for three months.

sergio-costas:
So... I see two possible ways of fixing this:

- we can finally enable using GVFS by removing `g_setenv ("GIO_USE_VFS", "local", TRUE);`

- or we can create a very simple daemon that allows to transform remote URIs into local ones using GVFS using DBus, running outside the flatpak/snap, and let the container security manage whether the contained program can or cannot access that local file. That way, the GVFS code that has access to remote places is isolated and has a very specific interface. A call to this DBus interface would replace the code in this MR.

Obviously, the most elegant option is the first one, but the second one can be used temporarily while there are still doubts about the implications of enabling the whole GVFS inside the portal.

But, as usual, is just an idea...

sergio-costas:
I discovered a simpler way of fixing this: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/67

The point is that xdg-desktop-portal-gtk does the translation from SMB/SFTP/other remote file systems into a local URI using FUSE before sending the list to xdg-desktop-portal, but xdg-desktop-portal-gnome doesn't.

PR #823, opened 2022-06-27
Don't show odd errors when using remote URIs

sergio-costas:
When a remote URI is selected in the portal, and the application
is inside a flatpak or a snap, it is not possible to get a local
path.

This MR detects this and shows an specific error, instead of the
odd errors that previously were shown.


sergio-costas:
This patch is the bare minimum for https://github.com/flatpak/xdg-desktop-portal/issues/820 and https://github.com/flatpak/xdg-desktop-portal/issues/213 . Ideally, apply https://github.com/flatpak/xdg-desktop-portal/pull/822 instead.

robert-ancell (approved):
LGTM!

src/documents.c
```diff
@@ -77,6 +77,12 @@ register_document (const char *uri,
 
   file = g_file_new_for_uri (uri);
   path = g_file_get_path (file);
+  if (path == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
+                   "URI %s not supported", uri);
```
mcatanzaro: I agree this looks good. We could make the error message more specific, though:

```suggestion
                   "URI %s not supported by the document portal", uri);
```

(It's also quite unfortunate to see xdg-desktop-portal is not translating GError messages, since these are likely to wind up displayed in UIs. But that seems to be a project convention, and consistency is most important, so I won't ask you to change that.)

mcatanzaro:
Looks like there are conflicts that need to be resolved, which is not surprising after 3 years.

src/documents.c
sergio-costas: Done. Thanks!

sergio-costas:
@swick Done.

sergio-costas:
@swick the mergebot removed this from merging queue due to a failed test, but now it passed.

PR #832, opened 2022-07-12
ScreenCast.xml: Embed monitor connector as stream property

salmanmlk:
If a screencast session is screencasting from multiple monitors,
it is not possible to distinguish which stream (or PipeWire stream
node id) belongs to which connected monitor (such information may be useful
to the caller when if they want to embed the stream in a window that
and name it after the monitor connector, for example). This change adds
optional metadata for monitor streams to also return the monitor connector
name with the stream.

Ideally, this metadata should equip the caller to get more information about
display from the following Wayland interfaces:

https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output
https://wayland.app/protocols/xdg-output-unstable-v1


data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -260,6 +260,13 @@
               This property was added in version 3 of this interface.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>output_names</term>
```
jadahl, line 264: `output_name` (not plural).

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -260,6 +260,13 @@
...
+            <listitem><para>
+            If a monitor is being screencasted, then this property contains the
+            wayland output name. Only available for monitor streams.
```
jadahl, line 267: The indentation seems a bit off; also, in some way, we should reference what the name corresponds to. By no means are anything directly tied to Wayland here, the screen cast API can just as well be used on an X11 session. Thus, we should probably list at least those two. For X11, I guess the xrandr name is fine. For Wayland, it'd be better to be a bit more specific, since "wayland output name" is a bit ambiguous.

jadahl (changes requested):

data/org.freedesktop.portal.ScreenCast.xml
salmanmlk: Done.

data/org.freedesktop.portal.ScreenCast.xml
salmanmlk: Thanks for the comment.

- Adjusted the indentation.
- Generalized the description to mention both X11 and Wayland.
- Added more specific details about what this string would contain in context of Wayland.

GeorgesStavracas:
I wonder if we could use PipeWire's stream properties for this instead?

salmanmlk:
> I wonder if we could use PipeWire's stream properties for this instead?

Do you mean leverage these properties to add a custom, say, `output`: https://pipewire.github.io/pipewire/structpw__properties.html ? I see set / get method there that could be used by compositor to set and clients to get the property but I am not sure if custom properties are allowed by pipewire already. Let me know if I understood correctly.

swick:
Has anything changed here regarding pipewire or is this still desirable?

grulja:
> Has anything changed here regarding pipewire or is this still desirable?

Just FYI: from what I remember @salmanmlk has moved to a different project within Google and he no longer works on Google Remote Desktop so I don't think he wants to continue with this change.

mcatanzaro:
Well this merge request is not useful if no portal backends or applications will use it. So if Salman isn't working on this anymore, we should probably simply close this?

PR #834, opened 2022-07-19
realtime: fix priority type in MakeThreadHighPriority

wtay:
The RTKit argument for the priority in MakeThreadHighPriority is
supposed to be an int, not a uint.

This avoids errors like: Jul 19 11:16:34 wtay rtkit-daemon[1474]:
Failed to parse MakeThreadHighPriority() method call: Argument 2 is
specified to be of type "int32", but is actually of type "uint32"


GeorgesStavracas (approved):
LGTM

PR #836, opened 2022-07-23
realtime: Fix property types

Kontrabant:
The types for the properties MaxRealtimePriority and RTTimeUSecMax are incorrect in the D-Bus xml interface file: MaxRealtimePriority should be type 'i' (int32) and RTTimeUSecMax should be type 'x' (int64).  This changes them to match the types used in the realtime.c source file as well as matching the types used by rtkit for these properties.

### DIFF
--- a/data/org.freedesktop.portal.Realtime.xml
+++ b/data/org.freedesktop.portal.Realtime.xml
@@ -56,9 +56,9 @@
         <arg type="i" name="priority" direction="in"/>
       </method>
 
-      <property name="MaxRealtimePriority" type="x" access="read"/>
+      <property name="MaxRealtimePriority" type="i" access="read"/>
       <property name="MinNiceLevel" type="i" access="read"/>
-      <property name="RTTimeUSecMax" type="i" access="read"/>
+      <property name="RTTimeUSecMax" type="x" access="read"/>
 
       <property name="version" type="u" access="read"/>
     </interface>


Kontrabant:
The build failures seem to be something failing internally in the build system and aren't related to this pull.

robert-ancell (changes requested):
Since these are already in use, you'll need to make new properties (e.g. MaxRealtimePriority and MaxRealtimePriority2) that have the different types otherwise this may break existing clients.

Kontrabant:
The problem with doing that is that this interface is then no longer a 1:1 drop-in replacement for RealtimeKit1, which seems to be the original intention of it.

Beyond syntactical correctness with RealtimeKit1, the RTTimeUSecMax property is currently being truncated from 64->32 bits with no warning, so it shouldn't be used as-is.

From a backwards compatibility standpoint, I question if anyone is using this interface prior to Pipewire starting to implement support for it given the other breaking bugs that were only very recently fixed or are being fixed (#834 and #841), and looking at the Pipewire code, this interface patch wouldn't impact them at all.

That said, if it's that big of an issue, it seems that it would be better to just bump the interface version with the corrected types and deprecate the broken version vs introducing two sets of properties, one of which doesn't directly map to the RealtimeKit1 types and has a truncation issue, and the other of which has property name mismatches with RealtimeKit1.

TingPing:
WebKit was using it but the API wasn't actually working. I think this breakage may be fine.

PR #841, opened 2022-07-26
Issue #837 - Remap thread ids

### LINKED ISSUE #837: Thread ID not remapped in RTKit portal
The RTKit portal remap the process id, but not the thread ID.

As a consequence I get "Error not found" when I try to use in inside Flatpak.

This is when I'm trying to implement portal support in Pipewire.
See https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/1332

hfiguiere:
Closes #837 

This is tested from inside flatpak, using freedesktop-sdk 22.08beta, with a Pipewire patch: https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/1332#note_1481517

Also a rebuilt version of the Ardour flatpak.

__Update__: this reworked version works fine. Ignore what follows.

__Update2__: Refactored the code a bit more.

~~Two things about this PR:~~

~~- I now get permission error instead of not found error, and Bustle no long shows  the call to `org.freedesktop.RealtimeKit1.MakeThreadRealtimeWithPID` on the system bus.~~

Before (system-bus):
image
note the thread id isn't consistent.
The error:
image

After (session-bus):
image
image
~~The pid and thread id are consistent with inside the sandbox. The PID in the error is the one of `xdg-dbus-proxy`~~

~~- I am not sure is the way to remap the thread ID is correct. It works for me (F36) and it sound like the difference between a LWP and the PID is the same on both side. Now that I think of it, it's incorrect of the PID happen to wrap around... If there is a better way I'd love to.~~

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -2195,8 +2195,7 @@ xdg_app_info_map_tids (XdpAppInfo  *app_info,
                        guint        n_tids,
                        GError     **error)
 {
-  char proc_dir[31] = {0, };
-  snprintf (proc_dir, sizeof(proc_dir), "/proc/%u/task", (guint) owner_pid);
+  g_autofree char *proc_dir = g_strdup_printf ("/proc/%u/task", (guint) owner_pid);
   return app_info_map_pids (app_info, proc_dir, tids, n_tids, error);
 }
 
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -2136,6 +2136,58 @@ xdp_app_info_ensure_pidns (XdpAppInfo  *app_info,
 }
 
 
+gboolean
+xdg_app_info_map_tids (XdpAppInfo  *app_info,
+                       pid_t        owner_pid,
+                       pid_t       *tids,
+                       guint        n_tids,
+                       GError     **error)
+{
+  char proc_dir[31] = {0, };
+  gboolean ok;
+  DIR *proc;
+  uid_t uid;
+  ino_t ns;
+
+  g_return_val_if_fail (app_info != NULL, FALSE);
+  g_return_val_if_fail (tids != NULL, FALSE);
+
+  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
+                           "Mapping pids is not supported.");
+      return FALSE;
+    }
+
+  snprintf (proc_dir, sizeof(proc_dir), "/proc/%u/task", (guint) owner_pid);
+  proc = opendir (proc_dir);
+  if (proc == NULL)
+    {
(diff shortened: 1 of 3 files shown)


hfiguiere:
Actually the error was caused by starting the modified `xdg-desktop-portal` from toolbx instead of host. It seems I get the right tid but still get a not found error.

hfiguiere:
note sure how to get CI to run properly. It looks like a setup issue.

hfiguiere:
@TingPing @GeorgesStavracas what's the next step? I can't seem to be able to have CI run.

src/xdp-utils.c
```diff
@@ -2187,6 +2187,28 @@ xdp_app_info_map_pids (XdpAppInfo  *app_info,
...
+
+gboolean
+xdg_app_info_map_tids (XdpAppInfo  *app_info,
+                       pid_t        owner_pid,
+                       pid_t       *tids,
+                       guint        n_tids,
+                       GError     **error)
+{
+  char proc_dir[31] = {0, };
+  snprintf (proc_dir, sizeof(proc_dir), "/proc/%u/task", (guint) owner_pid);
```
TingPing, line 2199:
```suggestion
  g_autofree char *proc_dir = g_strdup_printf ("/proc/%d/task", owner_pid);
```

Was there a reason you used %u, at a glance its an int?

TingPing:
I'm not familiar with how the remapping should work tbh.

Ideally we could add a test for this.

src/xdp-utils.c
hfiguiere: Because I did a lot of copy/paste assuming the code was right, like this https://github.com/flatpak/xdg-desktop-portal/blob/main/src/xdp-utils.c#L1776
(it's not the only place)

hfiguiere:
> I'm not familiar with how the remapping should work tbh.

Then who would be? Git's blame shows @gicmo 

The thing is #600 just doesn't work, because it doesn't handle the thread id that need the same remapping.

Note that I went from "not found error" with Bustle showing the unmapped tid was sent to the host to "no error" with the proper tid sent. (my first patch didn't work either).

hfiguiere:
> Ideally we could add a test for this.

I didn't find tests for the previous iterations of the code.

src/xdp-utils.c
TingPing: It's probably not a problem. I don't think negative pids exist.

TingPing:
> I didn't find tests for the previous iterations of the code.

`xdg-desktop-portal` is very under tested sadly.

> The thing is https://github.com/flatpak/xdg-desktop-portal/pull/600 just doesn't work, because it doesn't handle the thread id that need the same remapping.

> Note that I went from "not found error" with Bustle showing the unmapped tid was sent to the host to "no error" with the proper tid sent. (my first patch didn't work either).

Yeah I'm not trying to block this, just saying ideally it would get a test.

src/xdp-utils.c
hfiguiere: changed to use `g_strdup_printf()`

Issue #649, opened 2021-10-16
Screenshot portal without prompt


lionirdeadman:
This may seem dangerous but for screenshot applications, this sounds necessary from a design POV. Having a screenshot application which asks twice to screenshot would be quite awkward, I think.

Related : https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1970 , https://github.com/flameshot-org/flameshot/issues/1910

lionirdeadman:
I think it should be strictly easy to control so it shouldn't be metadata (and require something like Flatseal to change).

I think it should be a one-time prompt like there is for the background or camera permission. I suppose it wouldn't be a bad idea to make the permanency of it optional as to make the use case for it more flexible though.

All3xJ:
Is there any news about this annoying issue?

Midnex:
I hope you only have to give permission one time, and then from there it can do it everytime. This would be more secure than the old method, but less annoying then the current.

kurobeats:
Can I express how powerless this issue makes me feel in relation to Gnome development. The tone from the Gnome folks here: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1970 indicates to me this issue will not be resolved/is not considered an issue. Creating a Gnome related bug report is akin to shouting into the void, isn't it?

DamirPorobic:
They don't see the issue there but it's probably the same folks that doesn't take much screenshots so they don't feel the pain. As long as they're not made aware about the user frustration coming from this, they won't fix it am afraid.

mcatanzaro:
> Can I express how powerless this issue makes me feel in relation to Gnome development. The tone from the Gnome folks here: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1970 indicates to me this issue will not be resolved/is not considered an issue. Creating a Gnome related bug report is akin to shouting into the void, isn't it?

Honestly, your expectations are misplaced. Bypassing the screenshot portal is unacceptable, as it defeats the point of having the portal in the first place. Applications should not be able to screenshot your desktop without your permission. Removing the backdoor should not be controversial.

If you read up in this issue, we already have agreement on the path forward to enhance the screenshot portal. It's just waiting for a motivated developer to tackle it.

VitalyAnkh:
@kurobeats @DamirPorobic
[...]

PR #851, opened 2022-08-05
Screenshot permission

GeorgesStavracas:
This PR implements a new permission table and value for screenshots, which allows portal implementations to skip the sharing dialog.

Closes https://github.com/flatpak/xdg-desktop-portal/issues/649


GeorgesStavracas:
GNOME Settings can handle this permission now: https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/1412

Captura de tela de 2022-08-05 19-37-30

(Still WIP but it works)

GeorgesStavracas:
GNOME portal implementation ready: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/50
KDE doesn't seem to need this, they always skip the dialog for non-interactive requests anyway.

This somehow seems to be working, and that's highly suspicious:

[2022-08-05 20-50-25.webm](https://user-images.githubusercontent.com/3518204/183225222-a3e6bc0c-9172-4f7b-b7ac-466a73041134.webm)

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -49,7 +49,9 @@
             <term>interactive b</term>
             <listitem><para>
               Hint whether the dialog should offer customization before taking a screenshot.
-              Defaults to no.
+              Defaults to no. Since version 3 of #org.freedesktop.portal.Screenshot,
+              non-interactive screenshots are mediated by the permission store, so
+              implementations should not to show any additional interaction.
```
bilelmoussaoui, line 54:
```suggestion
              Defaults to no. Since version 3 of #org.freedesktop.portal.Screenshot,
              non-interactive screenshots are mediated by the permission store, so
              implementations should not show any additional interaction.
```

GeorgesStavracas: Good catch

GeorgesStavracas:
Let's go with this for now.

data/org.freedesktop.portal.Screenshot.xml
```diff
@@ -29,7 +29,7 @@
       via the document portal, and the returned URI will point
       into the document portal fuse filesystem in /run/user/$UID/doc/.
 
-      This documentation describes version 2 of this interface.
+      This documentation describes version 3 of this interface.
```
jadahl, line 32: What's the purpose of this version bump?

jadahl:
A bit unfortunate that a pull request with API changes affecting all potential portal backends only get 3 days, 2 which are on the weekend for review, before being landed without any "lgtm".

data/org.freedesktop.portal.Screenshot.xml
GeorgesStavracas: It's documented in the commit message: 

> Portal implementations will need to be adapted to this change,
> even though no actual D-Bus API change is introduced. Basically
> they'll have to skip showing any kind of dialog if the screenshot
> portal version is equal or greater than 3.

In other words, even if the D-Bus API is intact, v3 of the portal asks for the `screenshot` permissions, v2 doesn't, and implementations need to know this to act

GeorgesStavracas:
> A bit unfortunate that a pull request with API changes affecting all potential portal backends only get 3 days, 2 which are on the weekend for review, before being landed without any "lgtm".

It is indeed unfortunate, but as you can see, when I'm given the silence treatment after asking for reviews, landing pull requests immediately make people act.

jadahl:
I didn't see any asking for reviews, and 3 days is extremely short for API changes like this.

GeorgesStavracas:
This is not in any release yet, so post-merge review is also an option. If you think the commits here are not satisfactory, they can be reverted.

GeorgesStavracas:
As you wish, sir: https://github.com/flatpak/xdg-desktop-portal/pull/853

All3xJ:
Love this! Thanks!!

PR #852, opened 2022-08-08
Add Clipboard support for RemoteDesktop via org.freedesktop.portal.Clipboard

### LINKED ISSUE #843: Proposal:  Add Clipboard Control to RemoteDesktop
I am interested in adding clipboard control to the RemoteDesktop api.  There is a relevant discussion in #743 , where it suggests the RemoteDesktop might be a good place for this stuff to live. If there isn’t anyone working on something similar already, I’d be happy to submit a draft of it.  As an initial thought, I think we could elevate what already exists inside mutter’s RemoteDesktop api.  E.g. the following endpoints would be added:
    EnableClipboard
    DisableClipboard
    SetSelection
    SelectionWrite
    SelectionWriteDone
    SelectionRead
    SelectionOwnerChanged (signal)
    SelectionTransfer (signal)

Does anyone have any thoughts?

maxhbooth:
This PR is to add clipboard controls for RemoteDesktop as described in #843.

[xdg-desktop-portal-gnome](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/53)

Fix #843


gasinvein:
It it possible to use this without RemoteDesktop? I believe a Clipboard portal would be useful on its own, as a more convenient way to interact with the clipboard on Wayland.

jadahl:
> It it possible to use this without RemoteDesktop? I believe a Clipboard portal would be useful on its own, as a more convenient way to interact with the clipboard on Wayland.

Why would you want that? Clipboard in the form that remote desktop applications need should not be exposed to applications for convenience, since it avoids limiting clipboard content to the focused application and user interaction on that application. Doing that has very large privacy implications.

gasinvein:
> > It it possible to use this without RemoteDesktop? I believe a Clipboard portal would be useful on its own, as a more convenient way to interact with the clipboard on Wayland.
> 
> Why would you want that? 

One example I have in mind is a translation app that has a feature to translate selected text and put translation back to clipboard; this can be triggered without interacting with the app main window, via hotkey or tray icon menu.

> Clipboard in the form that remote desktop applications need should not be exposed to applications for convenience, since it avoids limiting clipboard content to the focused application and user interaction on that application. Doing that has very large privacy implications.

I suppose user has to grant the app permission to interact with clipboard first?

gasinvein:
> > One example I have in mind is a translation app that has a feature to translate selected text
> 
> In #843 it was mentioned that this interface could be useful for the `InputCapture` portal in the future, which might solve this use case?

Sorry, I don't know anything about the proposed `InputCapture` portal.

data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -38,8 +38,8 @@
 
         A remote desktop session may only be started and stopped with this interface,
         but you can use the #org.freedesktop.portal.Session object created with this
-        method together with certain methods on the #org.freedesktop.portal.ScreenCast
-        interface. Specifically, you can call
+        method together with certain methods on the #org.freedesktop.portal.ScreenCast and
+        #org.freedesktop.portal.Clipboard interfaces. Specifically, you can call
         org.freedesktop.portal.ScreenCast.SelectSources() to also get screen content,
```
jadahl, line 43: Also explicitly mention that you can call `#org.freedesktop.portal.Clipboard.RequestClipboard()` to request clipboard integration.

src/remote-desktop.c
```diff
@@ -56,13 +56,14 @@ G_DEFINE_TYPE_WITH_CODE (RemoteDesktop, remote_desktop,
                          G_IMPLEMENT_INTERFACE (XDP_DBUS_TYPE_REMOTE_DESKTOP,
                                                 remote_desktop_iface_init))
 
-typedef enum _RemoteDesktopSessionState
-{
+typedef enum _RemoteDesktopSessionState {
```
jadahl, line 59: Unrelated change - remove.

src/clipboard.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+
+#pragma once
+
+#include <gio/gio.h>
+
+#include "session.h"
+
+GDBusInterfaceSkeleton *clipboard_create(GDBusConnection *connection,
```
jadahl, line 25: Missing space before `(`.

src/remote-desktop.c
```diff
@@ -124,6 +128,26 @@ remote_desktop_session_can_select_sources (RemoteDesktopSession *session)
   g_assert_not_reached ();
 }
 
+gboolean remote_desktop_session_can_enable_clipboard(
+    RemoteDesktopSession *session) {
```
jadahl, line 132: Seems most of the changes to this file needs to be changed to follow the existing coding style.

src/clipboard.c
```diff
@@ -0,0 +1,719 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <gio/gunixfdlist.h>
```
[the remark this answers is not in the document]
jadahl: I don't think there is a formatter that does it perfectly, but if available, try to find the "GNU" coding style; it'll make it closer to what is expected. See https://developer.gnome.org/documentation/guidelines/programming/coding-style.html and the "GNU" variant.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,150 @@
...
+    @session_handle: Object path for the #org.freedesktop.portal.Session object
+    @options: Options for the clipboard
+    @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
+    Enables clipboard control for the given remote desktop session.
+
+    Supported keys in the @options vardict include:
+    <variablelist>
+      <varlistentry>
+        <term>handle_token s</term>
```
jadahl, line 41: I think we can skip using `o.f.p.Request` and friends; it was a mistake it was used for `SelectDevices` and `SelectSources` in the other interfaces - the Request object / response signal is primarily used for requests that can take an indefinite amount of time due to a user query, but that is not the case for this method, which is just for setting up a request that at a later point in time will result in a user being potentially queried.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,150 @@
...
+            object path element. See the #org.freedesktop.portal.Request documentation for
+            more information about the @handle.
+          </para></listitem>
+        </varlistentry>
+      </variablelist>
+     -->
+    <method name="SetSelection">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
```
jadahl, line 76: Same here, we don't need to bother with Request/handle/... - just make it a plain D-Bus method. The same applies to all the methods below as well.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,150 @@
...
+    </method>
+    <!-- 
+      SetSelection:
+      @session_handle: Object path for the #org.freedesktop.portal.Session object
+      @options: Vardict with optional further information
+      @handle: Object path for the #org.freedesktop.portal.Request object representing this call
+
```
jadahl, line 60: Missing documentation for what the method actually does. This applies to the other methods that follows this one as well.

jadahl (changes requested):
Haven't looked close at the actual C code yet, as if we're dropping the Request method dance, it'll likely look different enough to invalidate any review I do today.

tests/templates/clipboard.py
```diff
@@ -0,0 +1,171 @@
...
+    self, handle, session_handle, app_id, serial, success, options, cb_success, cb_error
+):
+    try:
+        logger.debug(
+            f"SelectionWriteDone({handle}, {session_handle}, {app_id}, {serial}, {success}, {options})"
+        )
+
+        response = Response(self.response, {})
+
+        request = ImplRequest(self, BUS_NAME, handle)
```
jadahl, line 121: Appears that tests needs to be updated.

jadahl (changes requested):

garnacho:
I've read through the changes and the proposed API makes sense to me. I feel there should be further documentation on the expected order of events after the `SelectionTransfer` signal, e.g. how to react to transfer requests of unrelated mime types or if no FD can be obtained.

In that regard, I mildly wonder if passing an FD should be implicitly considered success, writing on an end of a pipe accounts on the other side reading or the pipe will grow full and writes will stall, and having to wait to the end of the transfer for a `SelectionWriteDone` verifying the success means the other end needs to cache the full result before doing anything useful with it.

pnowack:
Since the portal clipboard API here is based on the private clipboard API between mutter and g-r-d, there is something to note here:

There are thoughts about changing the `SelectionWrite` and `SelectionWriteDone` methods in a way by using lists of what they actually currently have for arguments.
The reason is to group calls to kill latency when handling multiple clipboard content requests.
Due to time constraints, this was so far not realized for mutter and g-r-d (other tasks got priority and I only was made aware of this PR today) and so these changes did not went in yet.

For the portal API we probably wanna change the handling in that direction, so that mutter would not have two paths for handling selection writes and their results.
For `SelectionWriteDone` this would mean, a list would be submitted, where each element contains the serial and its sucess value.
I don't remember my thoughts about `SelectionWrite` to what it should be changed to be able handle multiple requests at the same time, but using a list of elements of (serial, fd) might be sufficient for this.

matthiasclasen:
Since this is only about remote desktop sessions, I don't think it should be called org.freedesktop.portal.Clipboard. That name is confusing and implies that this also covers local clipboards, which are handled via display server protocols already.

jadahl:
> Since this is only about remote desktop sessions, I don't think it should be called org.freedesktop.portal.Clipboard. That name is confusing and implies that this also covers local clipboards, which are handled via display server protocols already.

This is not a separate clipboard portal, but a way to hook up clipboard to the remote desktop portal. Since we might want clipboard to other portals (e.g. input capture), it makes sense to keep it separate, and not tied to remote desktop, otherwise we need to copy paste the whole API in every place its used, including follow up fixes / changes.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,180 @@
...
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+    <!-- 
+        SetSelection:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @options: Vardict with optional further information
+
+        Sets the owner of the clipboard formats in 'mime_types' in @options to
+        the session, i.e. this session has data for the advertised clipboard formats.
```
jadahl, line 54: Could be good to link to `org.freedesktop.portal.FileTransfer` to point out when to use the `application/vnd.portal.filetransfer` mimetype, e.g. when transferring files via otherwise done via file lists.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,180 @@
...
+        </variablelist>
+     -->
+    <method name="SetSelection">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+    <!--
+        SelectionWrite:
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @serials:  The list of serials of requests that this is answer is directed towards
```
jadahl, line 75: Stray space after `:`

src/clipboard.c
```diff
@@ -0,0 +1,538 @@
...
+handle_set_selection (XdpDbusClipboard *object,
+                      GDBusMethodInvocation *invocation,
+                      const char *arg_session_handle,
+                      GVariant *arg_options)
+{
+  Call *call = call_from_invocation (invocation);
+  Session *session;
+  GVariantBuilder options_builder;
+  GVariant *options;
+  g_autoptr (GError) error = NULL;
```
jadahl, line 117: The coding style for `g_autoptr` in xdg-desktop-portal is to not have a space before `(`.

src/clipboard.c
```diff
@@ -0,0 +1,538 @@
...
+
+static void
+selection_owner_changed_cb (XdpDbusImplClipboard *impl,
+                            const char *arg_session_handle,
+                            GVariant *arg_options,
+                            gpointer data)
+{
+  GDBusConnection *connection =
+      g_dbus_proxy_get_connection (G_DBUS_PROXY (impl));
+  g_autoptr (Session) session = lookup_session (arg_session_handle);
```
jadahl, line 485: Might want to move the assignment out of the declaration, and then NULL-check it.

The signal we're handling is from the `impl` but the session is local and might have been destroyed already.

src/clipboard.c
```diff
@@ -0,0 +1,538 @@
...
+}
+
+static void
+selection_owner_changed_cb (XdpDbusImplClipboard *impl,
+                            const char *arg_session_handle,
+                            GVariant *arg_options,
+                            gpointer data)
+{
+  GDBusConnection *connection =
+      g_dbus_proxy_get_connection (G_DBUS_PROXY (impl));
```
jadahl, line 484: Wrong indentation.

src/clipboard.c
```diff
@@ -0,0 +1,538 @@
...
+  g_autoptr (Session) session = lookup_session (arg_session_handle);
+
+  RemoteDesktopSession *remote_desktop_session =
+      (RemoteDesktopSession *)session;
```
jadahl, line 488: Wrong indentation.

src/clipboard.c
```diff
@@ -0,0 +1,538 @@
...
+
+static void
+selection_transfer_cb (XdpDbusImplClipboard *impl,
+                       const char *arg_session_handle,
+                       const char *arg_mime_type,
+                       guint arg_serial,
+                       gpointer data)
+{
+  GDBusConnection *connection =
+      g_dbus_proxy_get_connection (G_DBUS_PROXY (impl));
```
jadahl, line 455: Should shift only two spaces, not four.

jadahl (changes requested):
Starting to look pretty good to me, but wouldn't mind some second opinion from fellow xdg-desktop-portal developers.

Also, regarding the commit history, xdg-desktop-portal uses linear git history, meaning we don't squash merge requests upon merging. This means commits be split into logical chunks. E.g. the state machine cleanup in remote-desktop/screen-cast should be its own commit, descsribing the change, while the introduction of the clipboard parts should be its own.

data/org.freedesktop.impl.portal.Clipboard.xml
```diff
@@ -0,0 +1,181 @@
...
+        Answer to 'SelectionTransfer' signal(s). Transfers the clipboard content for one or more 
+        serials via file descriptors.
+        It is the Callee that creates the file descriptors.
+
+        May only be called if clipboard access was given after starting the session.
+    -->
+    <method name="SelectionWrite">
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a(u)" name="serials" direction="in"/>
```
smcv, line 90: This is an array of structs (tuples) where each one has only one element, like this:

```
[(1234,), (4567,), ...]
```

Are you sure that's what you want? Structs with only one element are usually not very useful, for the same reason that a Python `Tuple[int]` is not very useful: you could just use the int directly.

If you just want an array of integers:

```
[1234, 4567, ...]
```

then the type for that is `au`.

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -0,0 +1,183 @@
...
+        Answer to 'SelectionTransfer' signal(s). Transfers the clipboard content for one or more 
+        serials via file descriptors.
+        It is the Callee that creates the file descriptors.
+
+        May only be called if clipboard access was given after starting the session.
+    -->
+    <method name="SelectionWrite">
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="a(u)" name="serials" direction="in"/>
```
smcv, line 92: Same here.

tests/test_clipboard.py
```diff
@@ -0,0 +1,100 @@
...
+        clipboard_interface.SetSelection(session.handle, {})
+
+    def test_clipboard_selection_write(self):
+        params = {"force-clipoboard-enabled": True}
+        session, _ = self.start_session(params)
+        clipboard_interface = self.get_dbus_interface()
+
+        serials = []
+        # causes "'dbus.UInt32' is not iterable" exception
+        # serials.append(dbus.UInt32(1234))
```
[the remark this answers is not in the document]
smcv: You've told dbus-python that the type of `serials` is D-Bus type `a(u)`, which is a `dbus.Array` of `dbus.Struct` each containing only a `dbus.UInt32`.

dbus-python can (and does) try to remap the objects you gave it to the type that's expected, but it needs something that is "the right shape", which in this case would be an `Iterable[Iterable[int]]` where each of the "inner" iterables yields only one item.

If you changed the type to `au` then an `Iterable[int]` would be appropriate.

src/clipboard.c
```diff
@@ -0,0 +1,554 @@
...
+  out_fd_list = g_unix_fd_list_new ();
+  g_variant_builder_init (&out_request_fds_builder,
+                          (const GVariantType *)"a{uh}");
+
+  if (request_fds && fd_list) {
+    GVariantIter iter;
+    g_variant_iter_init (&iter, request_fds);
+    while (g_variant_iter_loop (&iter, "{uh}", &serial, &fd_id))
+      {
+        // fd_id = g_variant_get_handle (fd_handle);
```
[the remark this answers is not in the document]
smcv: A handle *is* a numeric ID. `g_variant_get_handle` is similar to `g_variant_get_uint32`: it looks at a `GVariant *` of type `h` and returns the handle (numeric ID). If you're using convenience APIs like `g_variant_iter_loop` that do the unpacking for you, then you usually don't need functions like `g_variant_get_handle`.

After that, you check that the handle (numeric ID) is in-range for the `GUnixFDList`, and then use `g_unix_fd_list_get` or `g_unix_fd_list_peek_fds` to get the actual file descriptor.

jadahl:
This needs a rebase to drop the autotools parts.

data/org.freedesktop.impl.portal.Clipboard.xml
```diff
@@ -0,0 +1,181 @@
...
+        Answer to 'SelectionTransfer' signal(s). Transfers the clipboard content for one or more 
+        serials via file descriptors.
+        It is the Callee that creates the file descriptors.
+
+        May only be called if clipboard access was given after starting the session.
+    -->
+    <method name="SelectionWrite">
+      <annotation name="org.gtk.GDBus.C.UnixFD" value="true"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="au" name="serials" direction="in"/>
```
jadahl, line 90: Discussed this some more with @pnowack and the conclusion is that the API here doesn't need this to handle the potential latency improvements. The main reason this could be useful was somewhat badly implemented applications that re-request the clipboard without there being any changes to it that would make it needed, but that is not really a good enough reason to complicate things.

For the record, the way this is intended to be handled in remote controlled screen share servers by making these method calls asynchronously. The calls themselves are fast, but handling things synchronously can cause problems if there are many requests coming in.

In other words, go back to the original API with a single serial, and a single file descriptor. The same goes for `SelectionWriteDone`.

jadahl: The same applies to both the `.portal.` and `.impl.portal.` interfaces.

src/remote-desktop.c
```diff
@@ -59,11 +59,6 @@ G_DEFINE_TYPE_WITH_CODE (RemoteDesktop, remote_desktop,
 typedef enum _RemoteDesktopSessionState
```
jadahl, line 59: Commit message should be e.g. `remote-desktop: Simplify state machine`

with e.g.

```
Remove some state that didn't help much keeping track of separately. We'll
want to be able to prepare more in an arbitrary order before starting so
just have a single state for not having started (`INIT`).
```

data/meson.build
```diff
@@ -8,6 +8,7 @@ portal_sources = files(
   'org.freedesktop.portal.Account.xml',
```
jadahl, line 8: Commit message header should be e.g. `Add Clipboard interface` (describing the change). The commit message body already gives more details so IMO good enough.

tests/__init__.py
```diff
@@ -367,6 +367,17 @@ def start_impl_portal(self, params=None, portal=None):
```
jadahl, line 367: Commit message nit: should be in the form of a description of the change, e.g. `remote-desktop: Add clipboard tests`

jadahl (changes requested):
I think this is more or less ready to land. I'd like to go back to a simpler previous version (see comment about SelectionWrite(Done) about what/why), but other than that it seems ready.

data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -148,6 +151,13 @@
               A bitmask of the devices selected by the user.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>clipboard_enabled b</term>
+            <listitem><para>
+              A boolean for whether the clipboard was enabled ('true') or not ('false'). 
+              See the #org.freedesktop.portal.Clipboard documentation for more information.
+            </para></listitem>
+          </varlistentry>
```
jadahl, line 160: This needs to be specified in `org.freedesktop.impl.portal.RemoteDesktop` as well. Both here and there should this describe what version of `org.freedesktop.impl.portal.RemoteDesktop` it was added.

The `This documentation describes version 1 of this interface.` higher up in this file (and the '.impl' variant) needs the version to be bumped to `2`.

src/remote-desktop.c
```diff
@@ -138,6 +142,24 @@ remote_desktop_session_can_select_devices (RemoteDesktopSession *session)
...
+
+  switch (session->state)
+    {
+    case REMOTE_DESKTOP_SESSION_STATE_INIT:
+      return TRUE;
+    case REMOTE_DESKTOP_SESSION_STATE_STARTED:
+    case REMOTE_DESKTOP_SESSION_STATE_CLOSED:
+      return FALSE;
+    }
+
```
jadahl: We should also check that the version advertised by the `impl` to be `>= 2`.

jadahl (changes requested):

tests/portals/test/test.portal
```diff
@@ -0,0 +1,3 @@
+[portal]
```
ebassi, line 1: We replaced `test.portal` with the `test.portal.in` template, which generates the portal file at build configuration time. You can drop this file.

jadahl (approved):
Everything lgtm. Planning to land this within some days, e.g. monday next week, if no-one objects before that.

PR #853, opened 2022-08-08
Screenshot permission, attempt 2

GeorgesStavracas:
This PR implements a new permission table and value for screenshots, which allows portal implementations to skip the sharing dialog.

Closes https://github.com/flatpak/xdg-desktop-portal/issues/649

### DIFF
--- a/data/org.freedesktop.impl.portal.Screenshot.xml
+++ b/data/org.freedesktop.impl.portal.Screenshot.xml
@@ -24,6 +24,8 @@
       @short_description: Screenshot portal backend interface
 
       This screenshot portal lets sandboxed applications request a screenshot.
+
+      This documentation describes version 2 of this interface.
   -->
   <interface name="org.freedesktop.impl.portal.Screenshot">
     <!--
@@ -50,6 +52,15 @@
             <listitem><para>
               Hint whether the dialog should offer customization before taking a screenshot.
               Defaults to no.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>permission_store_checked b</term>
+            <listitem><para>
+              Hint whether the screenshot portal has checked the 'screenshot' permission for
+              the requesting app. Defaults to no.
+
+              This option was added in version 2 of this interface.
             </para></listitem>
           </varlistentry>
         </variablelist>
@@ -103,5 +114,7 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <property name="version" type="u" access="read"/>
   </interface>
 </node>
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -51,6 +51,7 @@ struct _ScreenshotClass
 };
 
 static XdpDbusImplScreenshot *impl;
+static XdpDbusImplAccess *access_impl;
 static Screenshot *screenshot;
 
 GType screenshot_get_type (void) G_GNUC_CONST;
@@ -353,13 +354,14 @@ screenshot_class_init (ScreenshotClass *klass)
 
 GDBusInterfaceSkeleton *
 screenshot_create (GDBusConnection *connection,
-                   const char *dbus_name)
+                   const char *dbus_name_access,
+                   const char *dbus_name_screenshot)
 {
   g_autoptr(GError) error = NULL;
 
   impl = xdp_dbus_impl_screenshot_proxy_new_sync (connection,
                                                   G_DBUS_PROXY_FLAGS_NONE,
-                                                  dbus_name,
+                                                  dbus_name_screenshot,
                                                   DESKTOP_PORTAL_OBJECT_PATH,
                                                   NULL,
                                                   &error);
@@ -372,6 +374,13 @@ screenshot_create (GDBusConnection *connection,
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (impl), G_MAXINT);
 
   screenshot = g_object_new (screenshot_get_type (), NULL);
+
+  access_impl = xdp_dbus_impl_access_proxy_new_sync (connection,
+                                                     G_DBUS_PROXY_FLAGS_NONE,
+                                                     dbus_name_access,
+                                                     DESKTOP_PORTAL_OBJECT_PATH,
+                                                     NULL,
+                                                     &error);
 
   return G_DBUS_INTERFACE_SKELETON (screenshot);
 }
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -318,6 +318,7 @@ query_impl:
                              g_variant_new_boolean (TRUE));
     }
 
+  g_debug ("Calling Screenshot with interactive=%d", interactive);
   xdp_dbus_impl_screenshot_call_screenshot (impl,
                                             request->id,
                                             app_id,
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -175,18 +175,23 @@ static XdpOptionKey screenshot_options[] = {
   { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL }
 };
 
-static gboolean
-handle_screenshot (XdpDbusScreenshot *object,
-                   GDBusMethodInvocation *invocation,
-                   const gchar *arg_parent_window,
-                   GVariant *arg_options)
+static void
+handle_screenshot_in_thread_func (GTask *task,
+                                  gpointer source_object,
+                                  gpointer task_data,
+                                  GCancellable *cancellable)
 {
-  Request *request = request_from_invocation (invocation);
+  Request *request = (Request *)task_data;
   g_autoptr(GError) error = NULL;
   g_autoptr(XdpDbusImplRequest) impl_request = NULL;
   GVariantBuilder opt_builder;
+  GVariant *options;
+  const char *parent_window;
 
   REQUEST_AUTOLOCK (request);
+
+  parent_window = ((const char *)g_object_get_data (G_OBJECT (request), "parent-window"));
+  options = ((GVariant *)g_object_get_data (G_OBJECT (request), "options"));
 
   impl_request =
     xdp_dbus_impl_request_proxy_new_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (impl)),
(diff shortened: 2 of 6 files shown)


robert-ancell:
I reviewed the code and it looks good to me, thanks Georges!

jadahl (commented):
lgtm, just a minor leak spotted

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -49,7 +49,9 @@
             <term>interactive b</term>
             <listitem><para>
               Hint whether the dialog should offer customization before taking a screenshot.
-              Defaults to no.
+              Defaults to no. Since version 3 of #org.freedesktop.portal.Screenshot,
+              non-interactive screenshots are mediated by the permission store, so
+              implementations should not show any additional interaction.
```
jadahl, line 54: Thinking some more about this, it'd be more correct if `org.freedesktop.impl.portal.Screenshot` had a version set by the backend. That'd avoid breaking backends when the frontend is upgraded.

src/screenshot.c
```diff
@@ -186,14 +192,100 @@ handle_screenshot_in_thread_func (GTask *task,
...
+                             "deny_label", g_variant_new_string (_("Deny")));
+      g_variant_builder_add (&access_opt_builder, "{sv}",
+                             "grant_label", g_variant_new_string (_("Allow")));
+      g_variant_builder_add (&access_opt_builder, "{sv}",
+                             "icon", g_variant_new_string ("applets-screenshooter-symbolic"));
+
+      if (g_strcmp0 (app_id, "") != 0)
+        {
+          g_autoptr(GDesktopAppInfo) info = NULL;
+          const gchar *id;
```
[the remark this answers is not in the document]
GeorgesStavracas: Good catch. This code was mostly copied from the Wallpaper portal, which was also leaking, so I pushed an extra commit addressing that too.

data/org.freedesktop.impl.portal.Screenshot.xml
jadahl: A quick test shows that

```diff
diff --git a/data/org.freedesktop.impl.portal.Screenshot.xml b/data/org.freedesktop.impl.portal.Screenshot.xml
index cd769b5..1ae1608 100644
--- a/data/org.freedesktop.impl.portal.Screenshot.xml
+++ b/data/org.freedesktop.impl.portal.Screenshot.xml
@@ -105,5 +105,6 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+    <property name="version" type="u" access="read"/>
   </interface>
 </node>
diff --git a/src/screenshot.c b/src/screenshot.c
index a28339c..ab13a6a 100644
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -206,6 +206,9 @@ handle_screenshot_in_thread_func (GTask *task,
   parent_window = ((const char *)g_object_get_data (G_OBJECT (request), "parent-window"));
   options = ((GVariant *)g_object_get_data (G_OBJECT (request), "options"));
 
+  if (xdp_dbus_impl_screenshot_get_version (impl) == 0)
+    goto query_impl;
+
   permission = get_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID);
 
   if (!g_variant_lookup (options, "interactive", "b", &interactive))
@@ -286,6 +289,8 @@ handle_screenshot_in_thread_func (GTask *task,
         }
     }
 
+query_impl:
+
   impl_request =
     xdp_dbus_impl_request_proxy_new_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (impl)),
                                           G_DBUS_PROXY_FLAGS_NONE,
```

here, and

```diff
diff --git a/src/screenshot.c b/src/screenshot.c
index 1dca27d..27fab30 100644
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -304,6 +304,7 @@ screenshot_init (GDBusConnection *bus,
   GDBusInterfaceSkeleton *helper;
 
   helper = G_DBUS_INTERFACE_SKELETON (xdp_impl_screenshot_skeleton_new ());
+  xdp_impl_screenshot_set_version (XDP_IMPL_SCREENSHOT (helper), 1);
 
   g_signal_connect (helper, "handle-screenshot", G_CALLBACK (handle_screenshot), NULL);
   g_signal_connect (helper, "handle-pick-color", G_CALLBACK (handle_pick_color), NULL);
```

in xdg-desktop-portal-gnome avoids any regressions.

jadahl: Also, adding a `skip_dialog` boolean instead of the reverse version check might be a bit cleaner too?

GeorgesStavracas: > Also, adding a skip_dialog boolean instead of the reverse version check might be a bit cleaner too?

I've just quickly discussed this with Allan and my understanding is that this should have been a permission from the get-go; the dialog on the non-interactive case was a mistake. The expectation is that interactive=false will only be set by apps that are themselves screenshot editors, like Flameshot. Consequently, there shouldn't be any case where an app would request non-interactive screenshots when they aren't editors themselves. So I don't think this should be supported from the API.

jadahl: `skip_dialog` would only be part of the impl portal, as a backward compatible way to the new behavior, it wouldn't affect anything app facing.

jadahl: To put my word into diffs, it'd be

```diff
diff --git a/src/screenshot.c b/src/screenshot.c
index ab13a6a..5dc7bc2 100644
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -197,6 +197,7 @@ handle_screenshot_in_thread_func (GTask *task,
   gboolean interactive;
   const char *parent_window;
   const char *app_id;
+  gboolean skip_dialog = FALSE;
 
   REQUEST_AUTOLOCK (request);
 
@@ -289,6 +290,8 @@ handle_screenshot_in_thread_func (GTask *task,
         }
     }
 
+  skip_dialog = TRUE;
+
 query_impl:
 
   impl_request =
@@ -310,6 +313,12 @@ query_impl:
                       screenshot_options, G_N_ELEMENTS (screenshot_options),
                       NULL);
 
+  if (skip_dialog)
+    {
+      g_variant_builder_add (&opt_builder, "{sv}", "skip_dialog",
+                             g_variant_new_boolean (TRUE));
+    }
+
   g_debug ("Calling Screenshot with interactive=%d", interactive);
   xdp_dbus_impl_screenshot_call_screenshot (impl,
                                             request->id,
```

and

```diff
diff --git a/src/screenshot.c b/src/screenshot.c
index 27fab30..caed19a 100644
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -39,8 +39,6 @@ typedef struct {
   const char *retval;
 } ScreenshotDialogHandle;
 
-static uint32_t screenshot_portal_version = 0;
-
 static void
 screenshot_dialog_handle_free (gpointer data)
 {
@@ -158,6 +156,7 @@ handle_screenshot (XdpImplScreenshot *object,
   ScreenshotDialogHandle *handle;
   gboolean modal;
   gboolean interactive;
+  gboolean skip_dialog;
   GtkWindow *dialog;
   GdkSurface *surface;
   GdkDisplay *display;
@@ -173,6 +172,8 @@ handle_screenshot (XdpImplScreenshot *object,
 
   if (!g_variant_lookup (arg_options, "interactive", "b", &interactive))
     interactive = FALSE;
+  if (!g_variant_lookup (arg_options, "skip_dialog", "b", &skip_dialog))
+    skip_dialog = FALSE;
 
   if (arg_parent_window)
     {
@@ -193,7 +194,7 @@ handle_screenshot (XdpImplScreenshot *object,
   g_object_ref_sink (fake_parent);
 
   dialog = GTK_WINDOW (screenshot_dialog_new (arg_app_id,
-                                              screenshot_portal_version,
+                                              skip_dialog,
                                               interactive,
                                               shell));
   gtk_window_set_transient_for (dialog, GTK_WINDOW (fake_parent));
@@ -299,7 +300,6 @@ gboolean
 screenshot_init (GDBusConnection *bus,
                  GError **error)
 {
-  g_autoptr(GVariant) version_variant = NULL;
   g_autoptr(GVariant) result = NULL;
   GDBusInterfaceSkeleton *helper;
 
@@ -336,16 +336,6 @@ screenshot_init (GDBusConnection *bus,
                                         NULL,
                                         error);
 
-  version_variant = g_variant_get_child_value (result, 0);
-
-  if (!version_variant)
-    return FALSE;
-
-  g_variant_get_child (version_variant, 0, "u", &screenshot_portal_version);
-
-  if (screenshot_portal_version == 0)
-    return FALSE;
-
   g_debug ("providing %s", g_dbus_interface_skeleton_get_info (helper)->name);
 
   return TRUE;
diff --git a/src/screenshotdialog.c b/src/screenshotdialog.c
index b42251c..f2c3f5c 100644
--- a/src/screenshotdialog.c
+++ b/src/screenshotdialog.c
@@ -416,7 +416,7 @@ screenshot_dialog_class_init (ScreenshotDialogClass *class)
 
 ScreenshotDialog *
 screenshot_dialog_new (const char *app_id,
-                       uint32_t screenshot_portal_version,
+                       gboolean skip_dialog,
                        gboolean interactive,
                        OrgGnomeShellScreenshot *shell)
 {
@@ -441,7 +441,7 @@ screenshot_dialog_new (const char *app_id,
   gtk_label_set_label (GTK_LABEL (dialog->heading), heading);
 
   dialog->shell = g_object_ref (shell);
-  dialog->skip_dialog = screenshot_portal_version >= 3 && !interactive;
+  dialog->skip_dialog = skip_dialog;
 
   if (interactive)
     show_options (dialog);
diff --git a/src/screenshotdialog.h b/src/screenshotdialog.h
index 98b336c..eeea773 100644
--- a/src/screenshotdialog.h
+++ b/src/screenshotdialog.h
@@ -8,6 +8,6 @@ typedef struct _ScreenshotDialog ScreenshotDialog;
 typedef struct _ScreenshotDialogClass ScreenshotDialogClass;
 
 ScreenshotDialog * screenshot_dialog_new (const char *app_id,
-                                          uint32_t screenshot_portal_version,
+                                          gboolean skip_dialog,
                                           gboolean interactive,
                                           OrgGnomeShellScreenshot *shell);
```

on top of the above diffs. The point here is to not tie a xdg-desktop-portal version to a xdg-desktop-portal-gnome/xdg-desktop-portal-kde/... version.

jadahl: Maybe `skip_dialog` should be `permission_store_checked` or something instead? I assume that e.g. KDE did their own permission checking in the non-interactive path?

GeorgesStavracas: That opens a couple of questions:

 * What should happen if the user denies the app the permission to take non-interactive screenshots? Should we continue to abort the screenshot call, or should it pass "skip_dialog=false"? (I think the former is the better option)
 * Should the version property should still be added?

GeorgesStavracas: > I assume that e.g. KDE did their own permission checking in the non-interactive path?

AFAICS xdg-desktop-portal-kde is more liberal about that and doesn't show any dialogs in the non-interactive case anyway

jadahl: > * What should happen if the user denies the app the permission to take non-interactive screenshots? Should we continue to abort the screenshot call, or should it pass "skip_dialog=false"? (I think the former is the better option)

I'd say: if the backend has the new enough impl version, always check the permission store and never ask for a dialog, aka exactly what the new intended behavior is in this PR; if the backend is too old, assume it will take care of all permission checking, as is the case without this PR.

> 
> * Should the version property should still be added?

Only the backend/impl would need a version and version bump in the way I have in mind now.

The frontend/app facing API didn't change -> no API bump.

The backend/impl API did change -> API bump. Since there is no version, we need to add it to bump the API version.

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -52,6 +54,13 @@
               Defaults to no.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>permission_store_checked b</term>
+            <listitem><para>
+              Hint whether the screenshot portal has the 'screenshot' permission. Defaults
+              to no.
```
jadahl, line 61: Should we say "added in version 1"?

GeorgesStavracas: Sure

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -52,6 +54,15 @@
...
```
jadahl, line 61: Sorry, missed this but the portal doesn't have the permission, but it has checked the permission. I suggest
```
              Hint whether the screenshot portal has checked the 'screenshot' permission for
              the requesting app. Defaults to no.
```

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -24,6 +24,8 @@
       @short_description: Screenshot portal backend interface
 
       This screenshot portal lets sandboxed applications request a screenshot.
+
+      This documentation describes version 1 of this interface.
```
jadahl, line 28: The last comment; is it risky to "start" at 1, when this is the de facto second version?

GeorgesStavracas: Good point, it's probably more prudent to jump to 2 already

jadahl:
lgtm

GeorgesStavracas:
Thanks @robert-ancell and @jadahl

DamirPorobic:
Is anything changing for the caller or does this affects only the internal implementation of the portal?

All3xJ:
Thanks a lot! :D

PR #857, opened 2022-08-11
doc: Update camera documentation.

robert-ancell:
Add a description of what AccessCamera() is doing.
Note that there are currently no supported options in OpenPipeWireRemote().

### DIFF
--- a/data/org.freedesktop.portal.Camera.xml
+++ b/data/org.freedesktop.portal.Camera.xml
@@ -30,6 +30,8 @@
         @options: Vardict with optional further information
         @handle: Object path for the #org.freedesktop.portal.Request object representing this call
 
+        Request to gain access to the camera.
+
         Supported keys in the @options vardict include:
         <variablelist>
           <varlistentry>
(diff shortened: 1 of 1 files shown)


data/org.freedesktop.portal.Camera.xml
```diff
@@ -30,6 +30,9 @@
         @options: Vardict with optional further information
         @handle: Object path for the #org.freedesktop.portal.Request object representing this call
 
+        Request to gain access to the camera. The portal will present the user with a dialog to
+        allow or deny this request.
+
```
GeorgesStavracas, line 35: I think only the "Request to gain access to the camera." is enough. The fact that a dialog may be presented is an implementation detail of the GNOME access portal.

GeorgesStavracas (changes requested):

data/org.freedesktop.portal.Camera.xml
robert-ancell: I added the dialog note as other portals note this. However those cases seem to be more substantial dialogs (file chooser, print etc). In this case it is a yes/no dialog and the other cases don't mention this, so agreed - will remove.

GeorgesStavracas (approved):

PR #863, opened 2022-08-19
A few fixes to the tests

whot:
Together with a include sort that is unrelated, but won't hurt :)


GeorgesStavracas (approved):
All looks good, thanks

PR #864, opened 2022-08-19
Draft: Add a SessionFactory portal

whot:
Motivated by the discussion in https://github.com/flatpak/libportal/pull/102

This disentangles session creation from any specific portal. Previously,
a `Session` was created by a portal's call to `CreateSession` but sessions
are more complicated than that. The `RemoteDesktop`/`ScreenCast` sessions
are of largely overlapping functionality but the `Location` session is
slightly different (or at least not as commonly combined, I guess).

This patch introduces a new portal: the `SessionFactory` with a single
call `CreateSession`. Once a session has been created, it can be used with
the various portals that require session, effectively adding that portal
to the existing session so all interactions can be terminated with a
single `Close()`/`Closed` signal.

*Note that implementation is missing, it's a fair bit of work so I'd rather work on this once we agree it's the way to go.*

cc @jadahl

### DIFF
--- a/Makefile.am.inc
+++ b/Makefile.am.inc
@@ -25,6 +25,7 @@ PORTAL_IFACE_FILES =\
 	$(top_srcdir)/data/org.freedesktop.portal.Screenshot.xml \
 	$(top_srcdir)/data/org.freedesktop.portal.Secret.xml \
 	$(top_srcdir)/data/org.freedesktop.portal.Session.xml \
+	$(top_srcdir)/data/org.freedesktop.portal.SessionFactory.xml \
 	$(top_srcdir)/data/org.freedesktop.portal.Settings.xml \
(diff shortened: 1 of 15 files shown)


jadahl:
There is one pending PR that has non-trivial logic attached to the `CreateSession` call: https://github.com/flatpak/xdg-desktop-portal/pull/711

Cc: @aleixpol

whot:
Same is true for the `Location` portal too btw, but in all cases I think we can work around this by simply accepting an existing `session_handle` as argument to `CreateSession`. The actual creation of the session is secondary anyway.

whot:
I'm closing this one, it gets across the point but it's not immediately needed. The `SessionFactory`  in particular can be avoided by supporting the `session_handle` in every portal's `CreateSession` request. However, there is no immediate need for having a single session covering multiple portals (RD/SC excepted but that's a different can of worms).

Let's postpone this until it's actually needed.

Issue #508, opened 2020-06-30
xdg-portal does not forward the session closed event


ofourdan:
## Summary

The API states that a `org.freedesktop.portal.Session::Closed` signal is sent when the session is closed from the service side.

Yet, if I terminate an existing session (in gnome-shell, select "Screen is being shared" → "Turn off"), I see no such event, not even with dbus-monitor.

## Steps to reproduce

1. Save https://gitlab.gnome.org/snippets/1122 as xdp-pointer-events.py
2. Run python xdp-pointer-events.py
3. Select pointer in the list
4. Click [Share] to start the RemoteDesktop session
5. Run `dbus-monitor`
6. From gnome-shell, select "Screen is being shared" → "Turn off" to terminate terminate the session
7. Search for the `Closed` signal(s) in the `dbus-monitor` logs

## Actual results

The signal `org.gnome.Mutter.RemoteDesktop.Session::Closed` is found but no `org.freedesktop.portal.Session::Closed`

## Expected results

The `org.freedesktop.portal.Session::Closed` is forwarded to the client as per the [specification](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#idm13)

## Additional information

It's entirely possible that I am missing something here, but it seems to me that the signal is emitted by `mutter` but not forwarded to the client by `xdg-portal`.

jindal-arpit:
Any update on this issue?
We are also facing the same problem. Without this signal, our session remains open even after the screen sharing is closed.

jindal-arpit:
Does not seems like it is fixed. I could still reproduce it with the above mentioned steps.

whot:
Remaining issue is a bug in xdg-desktop-portal-gnome, no longer in xdg-desktop-portal (see #868). Should be fixed with
https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/52

GeorgesStavracas:
@whot tests landed, and the xdg-desktop-portal-gnome bug fix was merged too, so this is probably fixed. I'll probably backport the GNOME changes and release to stable this week.

PR #868, opened 2022-08-23
Add tests for Session.Close() and Session.Closed

whot:
This sits on top of #863 

Add support for session in the python portal tests and a minimal RemoteDesktop test case that creates a session and verifies that it was closed in the impl as well. And then the same the other way with the `Closed` signal bubbling up.

Motivated by #508, verifies that bug isn't actually x-d-p's fault anymore.


GeorgesStavracas:
Took the liberty to manually rebase this. Reviewing...

GeorgesStavracas (approved):

PR #872, opened 2022-08-25
CI: build the docs with meson

whot:
It's the fuuuuture! Also, meson has better autodetection of features, so we don't need to install *all* dependencies just to build the docs.


GeorgesStavracas (approved):
No objections, looks fantastic

Issue #847, opened 2022-07-31
FileChooser: Portal can only open, not just choose files


sophie-h:
This issue is basically an upstreaming of https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/26

Currently, OpenFile() is designed to make files/folders available as readable in the sandbox. Following problems exist with the current design/definition

1. The definition does not mention how error handling works. IIRC KDE shows an error when a file is not readable. GNOME currently behaves the same way as if the user has nothing selected. The latter means, there is no way for the app to handle the error.
2. There is no way to select files/folders that are not readable. My use-case is a backup application where you want to exclude folders that are not readable anyways. However, that's not possible with the OpenFile() portal.

I see two things that probably should happen:

1. Either clarify that portal implementations should handle errors for unreadable folders/files and implement them where missing. Or, add error feedback to the portal that can be handled by the app. This is also a question for designers.
2. I think in some sense the OpenFile() portal behaves correctly since it is supposed to make the file/folder available in a readable form to the app. However, I think there are enough use-cases where the app just wants the path. This will especially be true if we really start using the file portal on the host system. Maybe we need a flag for OpenFile() or a ChooseFile() function that only passes the path through? (Unclear if it should pass a path through, that's not available in the Flatpak.)

matthiasclasen:
My take on this: the portal can only open files, since thats what it says on the label (literally, in the filechooser button): "Open".

The fact that the backends currently let you select files or folders that the frontend then refuses to return to the caller feels like a quality-of-implementation issue in the affected backends. They should only let the user select objects that are acceptable.

ZanderBrown:
Files vs folders, a classic UNIX tale…

Given how disperate the implementations are it would seem we still have a documentation issue, or alternatively this is a feature request.

Also [Matthias Clasen's downstream comment](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/26#note_1431374) seems to imply this should work with folders:

> In any case, those permissions are enforced by the portal frontend when adding the files/directories to the document store.

And if I understand correctly, the main issue seems to be that implementations don't know (or don't tell apps?) when that adding-to-document-store failed? Though it's possible I'm not following.

sophie-h:
And another day on which, someone got bitten by the problem because we dropped the most basic features from our ecosystem. https://gitlab.gnome.org/World/pika-backup/-/issues/138#note_1600670

[...]

Issue #788, opened 2022-05-05
Get real path from FileChooser portal


JakobDev:
The file chooser Portal gives you currently access to a file/directory the user selects. The file can be accessed from inside the sandbox in the directory /run/user/(uid)/doc. The file access itself works great, but the path is the Problem. Many Apps display the Path e.g. in a recent files menu or showing the file the user is currently editing.

Showing the /run path is a very bad user experience.  The thing remains from the real path is the file name. So the User gets in a recent files list a list of /run/random_stuff/filename. If you have multiple files with the same name in the list, it is very confusing. e.g. the file which contains the list for needed Python packages is named requirements.txt. If you work with multiple requirements.txt, you can't tell from looking at the list with 7run paths, which is which.

Another example would be a text editor, that shows the Path of the file that is open. If the User has the File some time open in the Background and get's back the work, it is hard to say which file he/she is currently editing.

vchernin:
https://github.com/flatpak/xdg-desktop-portal/issues/475

GeorgesStavracas:
Duplicate of #475

Issue #475, opened 2020-04-17
Real directory name not accessible from file opened via document portal


chergert:
Currently, when accessing documents via the document portal we're provided with a file in the format of `/run/user/1000/doc/abcdef/filename.txt`. This is fine and a nice way to implement things using a FUSE mount to ensure access.

However, it's very problematic for applications like text editors. We want to provide information to the user about what file was accessed (including the path). We also want to allow the user to open a file browser or terminal where that file exists. Both of these things are impossible currently unless we grant access to `$HOME` or get lucky with access like `xdg-documents` or something like that.

I know it's a bit extra work, even for applications, but perhaps we could add a new `G_FILE_ATTRIBUTE_STANDARD_DIRNAME` that could grab an extended attribute from the FUSE driver.

My main argument is that I *don't think* there is value to obscuring a directory path from the application even if that is not how they access it. As a user, browsing to a directory and selecting a file certainly doesn't convey to me that the application wouldn't know where that file was (even if it accesses it via another path). It feels very much like an implementation detail due to how FUSE works, and if so, it would be nice to find a workaround for applications to exist with stronger sandboxing even for this use case.

hadess:
> We also want to allow the user to open a file browser or terminal where that file exists.

The `OpenDirectory` portal does just that:
> Asks to open the directory containing a local file in the file browser.

I'm not sure why the application needs to know what the name of the directory the file is in is.

hadess:
Also, adding a `G_FILE_ATTRIBUTE_STANDARD_DIRNAME` attribute won't fix the terminal use case either.

chergert:
> I'm not sure why the application needs to know what the name of the directory the file is in is.

Because in a text editor it's helpful to give context to the user as you might have multiple files with the same name in different directories.

For example, `Makefile`.

hadess:
So, this use case: https://github.com/flatpak/xdg-desktop-portal/issues/463 ?

chergert:
Additionally, it's nice to open a new `GtkFileChooserNative` in the same location as the focused file. If you can't actually get that location, you always end up at `$HOME` or `Recent` or something annoying like that.

hadess:
> Additionally, it's nice to open a new `GtkFileChooserNative` in the same location as the focused file. If you can't actually get that location, you always end up at `$HOME` or `Recent` or something annoying like that.

I've filed this as https://github.com/flatpak/xdg-desktop-portal/issues/477

[...]

PR #874, opened 2022-08-29
filechooser: Support current_folder with OpenFile

### LINKED ISSUE #796: FileChooser: Add current_folder option to OpenFile
One typical use-case would be a text editor where the open dialog pre-selects the folder of the currently open file.

sophie-h:
- [x] [xdg-desktop-portal-gnome](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/54)
- [x] xdg-desktop-portal-gtk: https://github.com/flatpak/xdg-desktop-portal-gtk/pull/396
- [x] GTK 4: Already sends `current_folder` option for OpenFile
- [x] Ashpd: https://github.com/bilelmoussaoui/ashpd/pull/90

Closes #796

### DIFF
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -273,6 +273,15 @@
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>current_folder ay</term>
+          <listitem>
+            <para>
+              Suggested folder to select the files from. The byte array is
+              expected to be null-terminated.
+            </para>
+          </listitem>
+        </varlistentry>
       </variablelist>
     -->
     <method name="SaveFile">
--- a/data/org.freedesktop.impl.portal.FileChooser.xml
+++ b/data/org.freedesktop.impl.portal.FileChooser.xml
@@ -118,6 +118,12 @@
               Whether the file is opened with write access. Default is no.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>current_folder ay</term>
+            <listitem><para>
+              A suggested folder to open the files from.
+            </para></listitem>
+          </varlistentry>
         </variablelist>
     -->
     <method name="OpenFile">
--- a/src/file-chooser.c
+++ b/src/file-chooser.c
@@ -455,7 +455,8 @@ static XdpOptionKey open_file_options[] = {
   { "directory", G_VARIANT_TYPE_BOOLEAN, NULL },
   { "filters", (const GVariantType *)"a(sa(us))", validate_filters },
   { "current_filter", (const GVariantType *)"(sa(us))", validate_current_filter },
-  { "choices", (const GVariantType *)"a(ssa(ss)s)", validate_choices }
+  { "choices", (const GVariantType *)"a(ssa(ss)s)", validate_choices },
+  { "current_folder", G_VARIANT_TYPE_BYTESTRING, NULL }
 };
 
 static gboolean


matthiasclasen:
I think current_folder is problematic in portal api - What filesystem view does this refer to, the one inside the sandbox, or the one outside? Sandboxed apps may not have any idea about the outside view.

matthiasclasen:
Depending on the use case, it may be possible to pass a reference to a file in the document store, and say: 'open in the folder that this file comes from'

sophie-h:
> What filesystem view does this refer to

I mean this is not any different from the `current_folder` option for `SaveFile`. It's the host location.

> Sandboxed apps may not have any idea about the outside view.

That's not really true. For example, `$HOME` is identical to the host location. Therefore, things like `get_user_special_dir` should work, I guess? That would allow a music app to open the file chooser in the Music folder for example. There are also some [`HOST_`](https://github.com/flatpak/flatpak/issues/2424#issuecomment-447320614) variables, but those are probably less useful in this case.

Further, there are suggestions to use the portal for GtkFileChooserNative on the host. We will lose this feature on the host if it is not supported by the portal.

Also, there have been requests about exposing the file path for opened files #475 #788. This would make this option and the existing options more useful.

I do agree that we need some more sophisticated solutions for file portals, see #847 for example. But I think those things are more like long-shot and should involve some careful design work.

matthiasclasen:
An alternative suggestion that does not need any new api, and will still address a number of related use cases: Make the portal backend remember the last folder, per-app.

sophie-h:
> An alternative suggestion that does not need any new api, and will still address a number of related use cases: Make the portal backend remember the last folder, per-app.

Probably better than no pre-selection at all. However, it would not cover the case of having a new app and using the open folder for example. Also, there are cases where more fine-grained control could be needed. For example, a text editor with several tabs/windows, and the developer wants to open the folder that contains the current file.

Realistically, apps currently do things like `--filesystem=host` (Text Editor of GNOME does that) for being able to show the file path. I currently don't see how we can get around providing the host file path for opened files as well. Or otherwise, all this discussion about tight sandboxing will not apply to many apps anyways.

felinira:
There are also applications like Bottles that allow you to choose a wine executable from within the applications data directory (to run it). Expecting users to ever find the sandbox folder is not really an option though as it is hidden well out of sight.

Preselecting folders is expected behavior for many users and I have already seen countless examples where users were confused about what to open and where it is.

btzy:
Given that `SaveFile` already supports ` current_folder`, I think it's only reasonable that `OpenFile` does the same thing.  This is probably also the way that gives the application developer the most flexibility in deciding which folder to show, whether it is some specific system or data directory, the directory containing some previously opened file, or something else.

ebassi:
> An alternative suggestion that does not need any new api, and will still address a number of related use cases: Make the portal backend remember the last folder, per-app.

This was implemented in the GNOME portal, incidentally. I wonder if we should hoist the implementation in xdg-desktop-portal, but that has some implications about storing settings.

redtide:
> I don't really understand what opening the file chooser with the last selected folder has to do with this. Those features are orthogonal to each other and have very different use cases and reasons to exist.

Yes, please forget the last used, just suggest a path like in SaveFile.

Also in other platforms are used to set the directory from where to open a file, and this should be done from the app, which it will take care of whatever path it is and tell the filechooser to open it, last used or else.

Slapbox:
This would be a really big help to have available for OpenFile. Is there any chance that might happen?

@GeorgesStavracas @alexlarsson can you tell who would need to review this? Is there anything the community can do to help get the ball rolling on this?

ebassi:
See also: #1007, which tries to convert paths known to the document storage.

ebassi:
Considering that SaveFile is already set up to do this, I think we should merge this pull request; though adding the same mechanism that PR #1007 uses to map paths passed to the portal to paths in the document storage would probably be a good idea as well. Maybe we want a separate PR for that, though.

Slapbox:
Thanks for looking into this @ebassi! As a Kubuntu user this is my #1 wishlist item.

data/org.freedesktop.portal.FileChooser.xml
```diff
@@ -273,6 +273,15 @@
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term>current_folder ay</term>
+          <listitem>
+            <para>
+              Suggested folder to select the files from. The byte array is
+              expected to be null-terminated.
```
matthiasclasen, line 281: The documentation would be a lot more useful if it stated what filesystem view will be used to resolve this path - is it a host path ? Or something visible inside the sandbox?

It would also be good to clarify 'suggested' a bit more. I am reading it to mean:
- portal implementations may ignore this
- this is not meant to limit the users freedom to provide whatever file he chooses (e.g. it is fine for him/her to navigate to another folder before making a selection)

matthiasclasen: Also, aren't byte arrays always null-terminated?

btzy: Though, I don't know why we want to require null termination of the byte arrays.  The null at the end of a string is not considered to be part of the string itself, so similar logic dictates that byte arrays should not semantically include the null byte.  (Whether or not the marshalling format will place a null at the end of the byte array is a different matter, but I don't think D-Bus users should not rely on that.)

smcv: > The documentation would be a lot more useful if it stated what filesystem view will be used to resolve this path - is it a host path ? Or something visible inside the sandbox?

I agree that to make this interface useful, the documentation needs to answer this question.

> It would also be good to clarify 'suggested' a bit more

I agree.


> Also, aren't byte arrays always null-terminated?

D-Bus doesn't require byte arrays to be zero-terminated. A byte array can be any arbitrary blob of bytes, similar to any other integer array but with 1-byte elements; it's like a `struct { unsigned char *location, size_t length }` in C.

As a higher-level, application-layer thing, an interface can define specific semantics for a byte array, like "it contains the bytes of a valid JPEG". In this particular case, the interface is using the convention for representing "bytestrings" (conceptually a string in an arbitrary ASCII superset, similar to `(type filename)` in GObject-Introspection), which is to encode them in the D-Bus message as a byte array with length `strlen(str) + 1`, containing the `strlen(str)` non-zero bytes of the string, followed by exactly one byte with value zero.

In the D-Bus Specification, the LinuxSecurityLabel is currently the only standardized bytestring, and is described as "the non-zero bytes of [whatever] in an unspecified ASCII-compatible encoding, followed by a single zero byte". I think that's a good way to phrase it.

"Null-terminated" is a somewhat misleading term for this, because this is byte value 0 (ASCII mnemonic `NUL`, not a `NULL` pointer.

It's correct to use the bytestring convention for Unix filenames, environment variable names/values, and any similar string-like thing that (unfortunately) cannot be guaranteed to be UTF-8 (and might instead be a legacy national character set like Latin-1).

> Whether or not the marshalling format will place a null at the end of the byte array is a different matter

The underlying marshalling format does not guarantee that there will be a zero after the end of a byte array, and also does not guarantee that there will not be a zero in the middle of the byte array. The higher-level (application-level) bytestring convention is that there is exactly one zero in the length-counted content of the array, as the last item.

Sodivad:
Support in Qt and KDE portal

https://codereview.qt-project.org/c/qt/qtbase/+/478997
https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/199

ebassi:
I opened https://github.com/flatpak/xdg-desktop-portal/pull/1045 which contains:

- documentation updates
- mapping from document storage to host paths

ebassi:
PR #1045 was merged, so we can close this one.

Thanks to @sophie-h for the work, and thanks everyone for the reviews.

Issue #789, opened 2022-05-05
`org.freedesktop.portal.Settings.Read` ("color-scheme") returns `Variant[Variant[uint]]`, while `*.impl.*.Read` returns just `Variant[uint]`


SerVB:
This can be seen in the log in the following picture:

![grafik](https://user-images.githubusercontent.com/20105593/166912322-4b200fd1-d5a4-4f51-9940-646d0d511164.png)

First call was made to:
```
Bus name: org.freedesktop.portal.Desktop
Object path: /org/freedesktop/portal/desktop
Interface: org.freedesktop.portal.Settings
Method: Read
Args: "org.freedesktop.appearance", "color-scheme"
```

Reply: `Variant: [Variant(uint): 2]`.

Second call was made to:
```
Bus name: org.freedesktop.impl.desktop.kde   <-- different
Object path: /org/freedesktop/portal/desktop
Interface: org.freedesktop.impl.portal.Settings   <-- different
Method: Read
Args: "org.freedesktop.appearance", "color-scheme"
```

Reply: `Variant(uint): 2`.

I don't understand why results are different since XMLs look equal to me ([org.freedesktop.portal.Settings.xml](https://github.com/flatpak/xdg-desktop-portal/blob/d7a304a00697d7d608821253cd013f3b97ac0fb6/data/org.freedesktop.portal.Settings.xml) and [org.freedesktop.impl.portal.Settings.xml](https://github.com/flatpak/xdg-desktop-portal/blob/d7a304a00697d7d608821253cd013f3b97ac0fb6/data/org.freedesktop.impl.portal.Settings.xml)). The second reply looks OK, the first one doesn't (I would say it doesn't match the docs). Probably it's a bug in the first case?

Just in case, OS info:
```
Operating System: Manjaro Linux
KDE Plasma Version: 5.24.4
KDE Frameworks Version: 5.92.0
Qt Version: 5.15.3
Kernel Version: 5.17.1-3-MANJARO (64-bit)
Graphics Platform: X11
```

robert-ancell:
The bug is in `settings_handle_read` - it reads the value as a variant from the backend, and instead of passing it directly back the caller it puts it inside another variant first.

grulja:
I can reproduce and I think it explains why sometimes when using QGnomePlatform it gets the light theme even though the dark variant is set.

PR #877, opened 2022-09-01
settings: Fix values from read being variants inside variants.

robert-ancell:
The backend returns a variant, and this was mistakenly being put inside another
variant when returned to the original caller.

Fixes https://github.com/flatpak/xdg-desktop-portal/issues/789

### DIFF
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -49,7 +49,7 @@
     implementation details that are undocumented. If you are a
     toolkit and want to use this please open an issue.
 
-    This documentation describes version 1 of this interface.
+    This documentation describes version 2 of this interface.
   -->
   <interface name="org.freedesktop.portal.Settings">
 
@@ -73,8 +73,29 @@
       @value: The value @key is set to.
 
       Reads a single value. Returns an error on any unknown namespace or key.
+
+      Deprecated, use ReadOne instead. The value argument was intended to have
+      the value inside one layer of variant as it is in ReadOne, for example
+      `&lt;string "hello"&gt;` in GVariant text notation; but it is actually
+      returned inside two layers of variant, for example
+      `&lt;&lt;string "hello"&gt;&gt;`.
     -->
     <method name='Read'>
+      <annotation name="org.freedesktop.DBus.Deprecated" value="true"/>
+      <arg name='namespace' type='s'/>
+      <arg name='key' type='s'/>
+      <arg name='value' direction='out' type='v'/>
+    </method>
+
+    <!--
+      ReadOne:
+      @namespace: Namespace to look up @key in.
+      @key: The key to get.
+      @value: The value @key is set to.
+
+      Reads a single value which may be any valid DBus type. Returns an error on any unknown namespace or key.
+    -->
+    <method name='ReadOne'>
       <arg name='namespace' type='s'/>
       <arg name='key' type='s'/>
       <arg name='value' direction='out' type='v'/>
--- a/src/settings.c
+++ b/src/settings.c
@@ -129,6 +129,42 @@ settings_handle_read (XdpDbusSettings       *object,
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
+static gboolean
+settings_handle_read_one (XdpDbusSettings       *object,
+                          GDBusMethodInvocation *invocation,
+                          const char            *arg_namespace,
+                          const char            *arg_key)
+{
+  int i;
+
+  g_debug ("ReadOne %s %s", arg_namespace, arg_key);
+
+  for (i = 0; i < n_impls; i++)
+    {
+      g_autoptr(GError) error = NULL;
+      g_autoptr(GVariant) impl_value = NULL;
+
+      if (!xdp_dbus_impl_settings_call_read_sync (impls[i], arg_namespace,
+                                                  arg_key, &impl_value, NULL, &error))
+        {
+          /* A key not being found is expected, continue to our implementation */
+          g_debug ("Failed to Read() from Settings implementation: %s", error->message);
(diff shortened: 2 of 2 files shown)


robert-ancell:
This is a behaviour change, so I'm not sure if it's acceptable to make at this point.

robert-ancell:
Updated with the new method.

A6GibKm:
I think it would be good to document exactly how Read was wrong.

src/settings.c
```diff
@@ -129,6 +129,42 @@ settings_handle_read (XdpDbusSettings       *object,
...
+settings_handle_read_one (XdpDbusSettings       *object,
+                          GDBusMethodInvocation *invocation,
+                          const char            *arg_namespace,
+                          const char            *arg_key)
+{
+  int i;
+
+  g_debug ("ReadOne %s %s", arg_namespace, arg_key);
+
+ for (i = 0; i < n_impls; i++)
```
jadahl, line 142: Wrong indentation.

data/org.freedesktop.portal.Settings.xml
```diff
@@ -73,8 +73,25 @@
...
+      <arg name='value' direction='out' type='v'/>
+    </method>
+
+    <!--
+      ReadOne:
+      @namespace: Namespace to look up @key in.
+      @key: The key to get.
+      @value: The value @key is set to.
+
+      Reads a single value. Returns an error on any unknown namespace or key.
```
jadahl, line 92: I suppose this needs a interface version bump.

data/org.freedesktop.portal.Settings.xml
```diff
@@ -73,8 +73,25 @@
       @value: The value @key is set to.
 
       Reads a single value. Returns an error on any unknown namespace or key.
```
jadahl, line 75: Maybe document what it actually returns too?

jadahl (commented):
@TingPing added this backend, so CC:ing.

data/org.freedesktop.portal.Settings.xml
robert-ancell: I'm not sure what you mean here - document that it's a variant inside another variant? I find it quite hard to work clearly, which is why the deprecation line below suggests it's broken and you should use ReadOne instead.

A6GibKm: image

Correcting the already existing docs would be enough imo, ` OUT v `-> ` OUT (v)`

smcv: No, `(v)` is a variant inside a (pointless) tuple. The existing `Read` method is already correctly returning a variant: the problem is that if the setting is (for example) a string, instead of the intended result (a variant containing a string), it's returning a variant containing a variant containing a string. In GLib's GVariant notation, instead of `<'hello'>`, it's returning `<<'hello'>>`.

Return type `(v)` would be `(<something>,)`, which is incompatible, and still doesn't give you any more information about what the *something* is.

A6GibKm: well, it has to be documented anyways that it returns variant(variant(what you actually want))

data/org.freedesktop.portal.Settings.xml
smcv, line 92: Perhaps "Reads a single value, similar to the standard `org.freedesktop.DBus.Properties.Get` method" or something, to emphasize that they're in the same encoding?

data/org.freedesktop.portal.Settings.xml
```diff
@@ -73,8 +73,25 @@
...
+
+      Deprecated for ReadOne. The value argument incorrectly returns the value inside two variants instead of one.
```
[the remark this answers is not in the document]
smcv: Or maybe the mention of GVariant text notation makes this less clear to D-Bus novices, I don't know. The variant type is just generally quite difficult to write clearly and concisely about.

smcv:
With hindsight, I sort of wish D-Bus variants didn't allow directly nesting another variant inside, because there's essentially no reasonable use-case for doing that. dbus-python also has some awkward API corners as a result of nested variants being possible. It's decades too late to change that, though.

data/org.freedesktop.portal.Settings.xml
robert-ancell: I copied the "may be any valid DBus type" from the org.freedesktop.DBus.Properties documentation which hopefully indicates what this variant is expected to contain.

data/org.freedesktop.portal.Settings.xml
```diff
@@ -73,8 +73,25 @@
       @value: The value @key is set to.
 
       Reads a single value. Returns an error on any unknown namespace or key.
+
+      Deprecated, use ReadOne instead. The value argument was intended to have the value inside one layer of variant as it is in ReadOne, for example `&lt;string "hello"&gt;` in GVariant text notation; but it is actually returned inside two layers of variant, for example `&lt;&lt;string "hello"&gt;&gt;`.
```
jadahl, line 77: Should use some line breaks somewhere.

jadahl (approved):
lgtm, just a nit

WhyNotHugo:
> With hindsight, I sort of wish D-Bus variants didn't allow directly nesting another variant inside, because there's essentially no reasonable use-case for doing that.

Maybe d-bus libraries can show a warning when nesting variants, and eventually drop support for it in a year or two? It'd definitely a situation which bit me a few times and I also don't think there are any deliberate usages for this "feature".

smcv:
> Maybe d-bus libraries can show a warning when nesting variants, and eventually drop support for it in a year or two?

That would be an incompatible change to documented behaviour, and the maintainers of widely-used libraries (should) take backwards-compatibility seriously, so: no, not unless you can provide a time machine.

PR #882, opened 2022-09-13
Added support for 'mapping' Snaps

ashuntu:
Allows Snaps to have their pids "mapped" if each pid app info matches, namely for use cases such as [this](https://github.com/FeralInteractive/gamemode/pull/385).

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -2151,7 +2151,22 @@ app_info_map_pids (XdpAppInfo  *app_info,
   g_return_val_if_fail (app_info != NULL, FALSE);
   g_return_val_if_fail (pids != NULL, FALSE);
 
-  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+  if (app_info->kind == XDP_APP_INFO_KIND_SNAP)
+    {
+      for (int i = 0; i < n_pids; i++)
+        {
+          XdpAppInfo *info = xdp_get_app_info_from_pid(pids[i], error);
+          if (strcmp(info->id, app_info->id) != 0 || info->kind != app_info->kind)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
+                                   "App info of pids do not match.");
+              return FALSE;
+            }
+        }
+
+      return TRUE;
+    }
+  else if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
     {
       g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
                            "Mapping pids is not supported.");


sergio-costas:
Looks good to me.

sergio-costas (approved):
LGTM

src/xdp-utils.c
```diff
@@ -2151,7 +2151,22 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
+        {
+          XdpAppInfo *info = xdp_get_app_info_from_pid (pids[i], error);
+          if (strcmp (info->id, app_info->id) != 0 || info->kind != app_info->kind)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
+                                   "App info of pids do not match.");
+              return FALSE;
+            }
+        }
+
```
kenvandine, line 2166: nit: remove extra blank line here.

src/xdp-utils.c
```diff
@@ -2151,7 +2151,22 @@ app_info_map_pids (XdpAppInfo  *app_info,
   g_return_val_if_fail (app_info != NULL, FALSE);
   g_return_val_if_fail (pids != NULL, FALSE);
 
-  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+  if (app_info->kind == XDP_APP_INFO_KIND_SNAP)
+    {
+      for (int i = 0; i < n_pids; i++)
+        {
+          XdpAppInfo *info = xdp_get_app_info_from_pid (pids[i], error);
```
3v1n0, line 2158: You're potentially overwriting `error` here, you need to check if info is valid too.

src/xdp-utils.c
```diff
@@ -2151,7 +2151,22 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
+          if (strcmp (info->id, app_info->id) != 0 || info->kind != app_info->kind)
```
3v1n0, line 2159:
```
if (info && info->kind != app_info->kind && g_strcmp0 (info->id, app_info->id) != 0)
```

3v1n0 (changes requested):

src/xdp-utils.c
3v1n0: This still stands.

If `xdp_get_app_info_from_pid` fails, you'll have the `error` set, thus the next `g_set_error_literal` will fail as you're overwriting it.

So I think it would be better to propagate it instead.

src/xdp-utils.c
```diff
@@ -2151,7 +2151,21 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
-  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+  if (app_info->kind == XDP_APP_INFO_KIND_SNAP)
+    {
+      for (int i = 0; i < n_pids; i++)
+        {
+          XdpAppInfo *info = xdp_get_app_info_from_pid (pids[i], error);
+          if (!info || info->kind != app_info->kind || g_strcmp0 (info->id, app_info->id) != 0)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
+                                   "App info of pids do not match.");
```
3v1n0, line 2162:
```suggestion
          if (!info)
            return FALSE;
   
          if (info->kind != app_info->kind || g_strcmp0 (info->id, app_info->id) != 0)
            {
              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
                                   "App info of pids do not match.");
```

In case in the `(!info)` case (when also `error` is set). you can use `g_propagate_prefixed_error`.

3v1n0 (changes requested):

src/xdp-utils.c
ashuntu:
```suggestion
          g_autoptr(GError) local_error = NULL;
          XdpAppInfo *info = xdp_get_app_info_from_pid (pids[i], &local_error);
          if (!info)
            {
              g_propagate_prefixed_error(error, g_steal_pointer(&local_error), "Invalid app");
              return FALSE;
            }

          if (info->kind != app_info->kind || g_strcmp0 (info->id, app_info->id) != 0)
            {
              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
                                   "App info of pids do not match.");
              return FALSE;
            }
```

Something like this?

3v1n0: Yes, a part from using correct spacing `g_propagate_prefixed_error (error, g_steal_pointer (&local_error), "Can't find app for pid %" G_PID_FORMAT ": ", pid[i]);`

3v1n0: In general though, you can also decide to skip the error: do you want all the pids to be mapped, right?

ashuntu: Yeah, each pid just needs to be confirmed to belong to the Snap

src/xdp-utils.c
```diff
@@ -2151,7 +2151,29 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
-  if (app_info->kind != XDP_APP_INFO_KIND_FLATPAK)
+  if (app_info->kind == XDP_APP_INFO_KIND_SNAP)
+    {
+      for (int i = 0; i < n_pids; i++)
+        {
+          g_autoptr(GError) local_error = NULL;
+          XdpAppInfo *info = xdp_get_app_info_from_pid (pids[i], &local_error);
+          if (!info)
+            {
+              g_propagate_prefixed_error (error, g_steal_pointer(&local_error),
```
3v1n0, line 2162: nit:

```suggestion

          if (!info)
            {
              g_propagate_prefixed_error (error, g_steal_pointer (&local_error),
```

src/xdp-utils.c
```diff
@@ -2151,7 +2151,29 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
+                                          "Can't find app for pid %i: ", pids[i]);
+              return FALSE;
+            }
+
+          if (info->kind != app_info->kind || g_strcmp0 (info->id, app_info->id) != 0)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
+                                   "App info of pids do not match.");
```
3v1n0, line 2170:
```suggestion
                                   "App info or pids do not match.");
```

3v1n0 (commented):
Looks good, a part few minor details.

3v1n0:
Once you've applied the suggestions, please, squash all the commits and write a commit explaining the rationale for this clearer.

src/xdp-utils.c
ashuntu: Fixed in https://github.com/flatpak/xdg-desktop-portal/commit/3055e448bd61dd551be6fce48d15ce70751182f6

src/xdp-utils.c
ashuntu: Fixed in https://github.com/flatpak/xdg-desktop-portal/commit/3055e448bd61dd551be6fce48d15ce70751182f6

src/xdp-utils.c
ashuntu: I don't believe "or" would be the right word here, I used "of" since `app_info` and `info` are derived from the pid. I changed this to "App info from pid does not match." to maybe be more clear in the most recent commit, but please let me know if I'm wrong.

src/xdp-utils.c
```diff
@@ -2151,7 +2151,29 @@ app_info_map_pids (XdpAppInfo  *app_info,
...
+          if (!info)
+            {
+              g_propagate_prefixed_error (error, g_steal_pointer (&local_error),
+                                          "Can't find app for pid %i: ", pids[i]);
+              return FALSE;
+            }
+
+          if (info->kind != app_info->kind || g_strcmp0 (info->id, app_info->id) != 0)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED,
```
3v1n0, line 2169: Using `G_IO_ERROR_NOT_SUPPORTED` or `G_IO_ERROR_NOT_FOUND` maybe?

3v1n0 (commented):
Code looks good now, see if we prefer having a better error code.

As per the rest, please squash the two commits together and provide a clear commit message (and PR descriptions) so that also who is less familiar with the change could understand it.

See: https://wiki.gnome.org/Git/CommitMessages

src/xdp-utils.c
ashuntu: I changed it to `G_IO_ERROR_NOT_SUPPORTED` to match what it does for Flatpak, and squashed the commits into 1.

Issue #889, opened 2022-09-22
Global Shortcuts follow-ups


GeorgesStavracas:
* [ ] Fix documentation generation
 * [ ] Improve D-Bus API documentation
 * [ ] Validate input options
 * [ ] Review API once again
   * [ ] Move the `shortcuts` option from `CreateSession()` to `BindShortcuts()`?

PR #890, opened 2022-09-22
Global Shortcuts cleanups, part 1

GeorgesStavracas:
See commits. This PR only contains the easy parts of https://github.com/flatpak/xdg-desktop-portal/issues/889.


GeorgesStavracas:
@aleixpol @jadahl please review, this should be the easy one

GeorgesStavracas:
Draft because I just realized there's more to cleanup

aleixpol:
Thanks for going through it. The changes so far look good to me.

data/org.freedesktop.impl.portal.GlobalShortcuts.xml
```diff
@@ -41,19 +43,12 @@
...
             <listitem><para>
-              The different shortcut ids that the app needs. For those that had already been bound on previous runs, they are mapped to user-readable descriptions of the trigger. This can be used by the app to describe the shortcut to the user.
-
-              By default these shortcuts are not bound. To bind them, BindShortcuts needs to be called for the portal
-              to assign a trigger to a shortcut.
-
-              They are mapped to a list of pairs including:
-                - description: User-readable text describing what the shortcut does
-                - shortcut: A string that contains a suggestion of a default shortcut, defined as described in the "shortcuts" xdg-spec.
+              The session id. A string representing the created screen cast session.
```
DaPigGuy, line 48:
```suggestion
              The session id. A string representing the created global shortcuts session.
```

GeorgesStavracas: Oops

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -88,6 +82,30 @@
             </para></listitem>
           </varlistentry>
         </variablelist>
+
+        Shortcuts contain the following properties:
+        <variablelist>
+          <varlistentry>
+            <term>trigger s</term>
```
jadahl, line 104: Should this ever be something passed from the app? I assume it's rather something that is always set by the portal implementation, and handed to the app so it can be presented in some way.

jadahl:
> Thanks for going through it. The changes so far look good to me.

@aleixpol  This PR makes the CreateSession() not have any shortcuts as port of the method call; was there any reason that it was there originally?

IGN-Styly:
Please Refactor a bit this code. it may be blotish but necessary. THX for this.

data/org.freedesktop.portal.GlobalShortcuts.xml
GeorgesStavracas: Now that I look into it... I wonder what's that even supposed to do. I suppose this was meant to be some sort of "title" or "name" of the shortcut, like "Create Source" or "Hide Something"?

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -23,27 +23,21 @@
...
-          - description: User-readable text describing what the shortcut does.
-          - shortcut: A string that contains a suggestion to the portal of a default shortcut, defined as described in the "shortcuts" xdg-spec.
-          - trigger: User-readable text describing how to trigger the shortcut for the client to render.
-      </para>
+      This portal lets applications create global shortcuts sessions, and
+      register shortcuts to them. These shortcuts are activated regardless of
+      the focused state of the application window.
+
+      All global shortcuts are bound to a session, and all sessions are bound to
+      the application that created them.
```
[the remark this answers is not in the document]
GeorgesStavracas: Brought that back.

data/org.freedesktop.portal.GlobalShortcuts.xml
GeorgesStavracas: @aleixpol could you please clarify?

jadahl: Isn't it meant to let the application present in e.g. some configuration view what the global keybinding was bound to?

aleixpol: The app provides a shortcut as specified by:
https://github.com/freedesktop/xdg-specs/blob/master/shortcuts/shortcuts.txt

Then the portal replies with a trigger description to be rendered on the app to help the user.

(not sure if this is what you were asking)

GeorgesStavracas: I think the question is, why is there:

```
            <term>trigger s</term>
            <listitem><para>
              User-readable text describing how to trigger the shortcut for the
              client to render.
            </para></listitem>
```

when there already is

```
            <term>shortcut s</term>
            <listitem><para>
              The suggested shortcut trigger, defined as described by the "shortcuts"
              XDG specification. Optional.
            </para></listitem>
```

aleixpol: What @jadahl said. ^

GeorgesStavracas: I'm not sure that separating `trigger` from `shortcut` is a good idea. Portal implementations could derive a user-visible string from a shortcuts, couldn't they?

GeorgesStavracas: I'm concerned with the fact that we can't validate that `trigger` is equal to `shortcut`

aleixpol: It's perfectly valid that it isn't. Big part of this whole process is for the portal to show a dialog that will allow the user to change the shortcut eventually. Maybe we can rename `shortcut` to `default_shortcut`?

aleixpol:
> > Thanks for going through it. The changes so far look good to me.
> 
> @aleixpol This PR makes the CreateSession() not have any shortcuts as port of the method call; was there any reason that it was there originally?

It was there so not to have an uninitialised session for a period.

I understand that you prefer to need to call ListShortcuts first? Is it to make the CreateSession signatures generic?

data/org.freedesktop.portal.GlobalShortcuts.xml
jadahl: Maybe `shortcut_hint`?

jadahl:
> I understand that you prefer to need to call ListShortcuts first? Is it to make the CreateSession signatures generic?

I don't have a preference, just want to make sure some design goal you had in mind is not lost with that change. The motivation @GeorgesStavracas gave was that it's common practice (so far) to have relatively small "CreateSession" methods, and set up things that require more data in separate methods.

data/org.freedesktop.portal.GlobalShortcuts.xml
jadahl: My suggestion here is to split up the vararg specification to two different ones: one "passed to portal", that contains `shortcut`/`shortcut_hint`/`default_shortcut`/whatever, and another that the portal passes to the app, that contains `trigger`.

That makes it clear that the app passes the suggested shortcut according to the machine readable format, and that the portal passes the trigger in a human readable format back with the chosen trigger.

GeorgesStavracas:
> I understand that you prefer to need to call ListShortcuts first?

I don't see how that could possibly be the case. Consumers of the API create a session, bind a list shortcuts, and get a result.

aleixpol:
> I don't see how that could possibly be the case. Consumers of the API create a session, bind a list shortcuts, and get a result.

Correct, I meant `BindShortcuts`.

I'm okay with keeping a more minimal CreateSession method, even though it means it's empty for a period in time until the bindings are set up.

data/org.freedesktop.portal.GlobalShortcuts.xml
aleixpol: > My suggestion here is to split up the vararg specification to two different ones:...

Makes sense to me. This is something I also considered doing, did not do it because I feared it might overcomplicate the spec, but it clearly still is confusing...

GeorgesStavracas: Sounds good

GeorgesStavracas:
Sorry, I guess I'm an idiot, but the difference between `trigger` and `shortcut` is still not entirely clear.

I understand that `shortcut` is supposed to be a hint that applications pass, to let the portal know which keycombo it suggests for a shortcut.

What I don't really understand is why `trigger` exists, like, at all. So, let's take this mockup as an example:

image

Specifically, let's look at the first row:

image

Here we have:

 * `Take screenshot`, which IIUC maps to the `description` property
 * `Print`, which IIUC maps to the `shortcut` property
 * ???, which maps to `trigger` property

Please ELI5 because so far I just don't see how this all connects together. In the mockup above, should the application pass `{ description: 'Take screenshot', trigger: 'Print', shortcut: 'Print' }`?

swick:
`shortcut` is a string according to the xdg Shortcuts Specification, e.g `CTRL+SHIFT+a` or `Print`. `trigger` is for the UI/the user e.g. `Control and Shift and a` or `Left joystick up`.

GeorgesStavracas:
Okay, then my earlier concern still applies; there is nothing actually forcing `shortcut` and `trigger` to actually match. Applications can, for example, pass:

```
{
  description: 'Take screenshot',
  trigger: 'Print',
  shortcut: 'CTRL+ALT+m',
}
```

and there is nothing we could do prevent that, I think. Which to me means this approach / API is broken already.

aleixpol:
> Applications can, for example, pass

Applications should never pass a `shortcut`, it's for the portal to tell the application what the shortcut is for a said action, so it can be rendered on the UI.

swick:
> "what it's actually bound to" implies following XDG shortcuts, "user-visible description of a shortcut" means exactly not that.

`shortcut` is just a hint for what the shortcut could be bound to. The user then decides to bind it to any trigger. Could be the hint, could be the X button on a controller, anything that the backend supports. `trigger` then contains a "User-readable text describing how to trigger the shortcut for the client to render".  This string in `trigger` represents "what it's actually bound to" as well as  

 `Control and Shift and a` for example is what it's bound to, "what it's actually bound to" tells you *what* it is, "user-visible", just not necessarily "a user visible description of the shortcut hint".

For example, `shortcut` could be `CTRL+ALT+m` and the user binds it to another keyboard shortcut (`CTRL+x`) then the `trigger` could be `Control and x` or `Ctrl X` or whatever the backend thinks is a good representation of `CTRL+x`.

> Applications should never pass a shortcut, it's for the portal to tell the application what the shortcut is for a said action, so it can be rendered on the UI.

I think you're confusing `shortcut` and `trigger`.

jadahl:
Perhaps calling the application provided hint `preferred_shortcut_hint`, and the description from the portal `selected_shortcut_description` would help clarify things.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -199,6 +221,24 @@
...
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>trigger_description s</term>
+            <listitem><para>
+              User-readable text describing how to trigger the shortcut for the
+              client to render.
+            </para></listitem>
+          </varlistentry>
+        </variablelist>
```
jadahl, line 241: Can't we just refer to the list in BindShortcuts?

GeorgesStavracas: Sure. Done.

PR #893, opened 2022-09-28
Fix Meson build of 1.15.x on Debian 11 (GLib 2.66)

smcv:
For the benefit of Flatpak apps that need recent portal features, it's useful for newer xdg-desktop-portal releases to be backportable to LTS distributions like Debian stable, Ubuntu LTS and RHEL (in Debian, I maintain an official backport of the versions we aim to have in Debian 12, backported for Debian 11 users).

* tests: Don't apply Meson's strict TAP parsing for older GLib
    
    In older GLib, we don't have a straightforward way to avoid debug
    messages going to stdout, which would be tolerated by Autotools' TAP
    parser but not by Meson's stricter TAP parser.

* tests: Include src/glib-backports.h where needed
    
    The tests call g_log_writer_default_set_use_stderr(), but that function
    isn't available in older GLib.
    
    We have to include this as src/glib-backports.h, because
    tests/glib-backports.h contains different backports.

* glib-backports: Declare stub function as inline
    
    This avoids `-Wunused-function` warnings in translation units that use
    glib-backports.h but do not call this particular function.

### DIFF
--- a/src/glib-backports.h
+++ b/src/glib-backports.h
@@ -53,7 +53,7 @@ guint g_string_replace (GString     *string,
 #endif
 
 #if !GLIB_CHECK_VERSION (2, 68, 0)
-static void
+static inline void
 g_log_writer_default_set_use_stderr (gboolean use_stderr)
 {
   /* Does nothing because outside of the tests we don't really care that it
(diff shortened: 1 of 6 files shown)


GeorgesStavracas (changes requested):
Only one small nit

GeorgesStavracas (approved):

Issue #626, opened 2021-09-05
Add support for querying whether gamemode is enabled


mikix:
The current gamemode portal API is designed for an app to turn gamemode on or off for itself. All the query methods take a specific PID.

But it would be useful for an app to query whether gamemode is enabled on the system at all.

(My use case is a backup app that would like to politely avoid scheduling any automatic backups while gamemode is active.)

The non-portal gamemode API exposes a property `ClientCount` that can be used for this. Either proxying that or exposing a boolean property would both do the trick.

If folks agree, I'm happy to do the work.

PR #900, opened 2022-10-11
gamemode: add Active property

mikix:
This allows sandboxed apps to tell if any process on the system has enabled Game Mode. Which is useful to modify behavior, like not starting expensive operations.

Fixes https://github.com/flatpak/xdg-desktop-portal/issues/626

How do we feel about the property name? Should it be simply `IsEnabled`? Or even simpler `Enabled`?


src/gamemode.c
```diff
@@ -535,6 +535,21 @@ handle_unregister_game_by_pidfd (XdpDbusGameMode *object,
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
+static void
+client_properties_changed (GDBusProxy* proxy,
+                           GVariant* changed_properties,
+                           char** invalidated_properties)
```
GeorgesStavracas, line 541: The code style of the project places the asterisks together with the variable name, not the type. E.g. `char **invalidated_properties`.

src/gamemode.c
```diff
@@ -535,6 +535,21 @@ handle_unregister_game_by_pidfd (XdpDbusGameMode *object,
...
+{
+  g_autoptr(GVariant) count = NULL;
+  gboolean enabled = FALSE;
+
+  count = g_dbus_proxy_get_cached_property (proxy, "ClientCount");
```
GeorgesStavracas, line 546: You should probably check if any one of the invalidated properties is `ClientCount` before updating the D-Bus property. Something like `if (!g_strv_contains (invalidated_properties, "ClientCount")) \n return;`

GeorgesStavracas (changes requested):
This doesn't seem controversial to me, but I want to hear from the people who originally introduced this portal first (@gicmo ?)

src/gamemode.c
mikix: Ah thank you. Fixed.

src/gamemode.c
mikix: Ah, that's why I added [G_DBUS_PROXY_FLAGS_GET_INVALIDATED_PROPERTIES](https://docs.gtk.org/gio/flags.DBusProxyFlags.html#get_invalidated_properties) -- because the invalidated properties dance is annoying if you really do need the value.

But I could change that and query for the value when it's invalidated if you don't like using the flag.

mikix: I could add a bit of optimization here and check if ClientCount is actually in changed properties. But I was going for simplicity over 100% correctness. But that might not be the style move here.

gicmo:
> This doesn't seem controversial to me, but I want to hear from the people who originally introduced this portal first (@gicmo ?)

Sounds like a useful addition. I think I would have called it `Active` instead of `IsGameModeEnabled`, but it is not a big deal.

src/gamemode.c
GeorgesStavracas: It's not a massive deal, but in my experience we have to be careful not to update properties when they didn't actually change, as this could lead to all sorts of odd behaviors on apps.

GeorgesStavracas:
@mikix as per @gicmo's comment, could you please rename it to `Active`?

mikix:
OK, made three changes:
- Renamed property to `Active`
- Only update it when `ClientCount` changes, not any property
- Check `ClientCount` on portal startup, just in case we got restarted or who knows

src/gamemode.c
```diff
@@ -562,5 +604,11 @@ game_mode_create (GDBusConnection *connection)
   gamemode = g_object_new (game_mode_get_type (), NULL);
   gamemode->client = client;
 
+  g_signal_connect (client, "g-properties-changed",
+                    G_CALLBACK (client_properties_changed), NULL);
+
+  // Should not normally happen but just in case, check for active games on startup
```
GeorgesStavracas, line 610: I don't think this comment is particularly necessary, please remove it

src/gamemode.c
```diff
@@ -535,6 +535,47 @@ handle_unregister_game_by_pidfd (XdpDbusGameMode *object,
...
+                           GVariant *changed_properties,
+                           char **invalidated_properties)
+{
+  const gchar *key;
+  GVariant *value = NULL;
+  g_autoptr(GVariantIter) iter;
+
+  g_variant_get (changed_properties, "a{sv}", &iter);
+  while (g_variant_iter_next (iter, "{&sv}", &key, &value))
+  {
```
GeorgesStavracas, line 569: This whole hash table iteration is unnecessary :slightly_smiling_face: 

The `g-properties-changed` signal conveniently passes the list of invalidated properties for us, that's the `invalidated_properties` array. You can simply check if the `ClientCount` string is in that array. Something like this:

```c
  int n_clients;

  if (!g_strv_contains (invalidated_properties, "ClientCount"))
    return;

  g_variant_lookup (changed_properties, "ClientCount", "i", &n_clients);
  xdp_dbus_game_mode_set_active (XDP_DBUS_GAME_MODE (gamemode), n_clients > 0);
```

src/gamemode.c
```diff
@@ -535,6 +535,47 @@ handle_unregister_game_by_pidfd (XdpDbusGameMode *object,
...
+
+static void
+update_active_state_from_cache (GDBusProxy *proxy)
+{
+  g_autoptr(GVariant) client_count = NULL;
+
+  client_count = g_dbus_proxy_get_cached_property (proxy, "ClientCount");
+
+  if (client_count != NULL)
+    update_active_state (client_count);
```
GeorgesStavracas, line 555: I guess you can update it in place here, instead of using a separate function.

src/gamemode.c
```diff
@@ -535,6 +535,47 @@ handle_unregister_game_by_pidfd (XdpDbusGameMode *object,
...
 }
 
+
+/* properties */
+static void
+update_active_state (GVariant *client_count)
+{
+  gboolean enabled = g_variant_get_int32 (client_count) > 0;
+  xdp_dbus_game_mode_set_active (XDP_DBUS_GAME_MODE (gamemode), enabled);
+}
```
GeorgesStavracas, line 545: If you adhered to the previous comments, this function is now unused, and can be removed

GeorgesStavracas (changes requested):

GeorgesStavracas:
Thanks. I'm happy with the overall approach here, the comments are mostly about the code itself, not the functionality.

src/gamemode.c
mikix: Done :+1:

src/gamemode.c
mikix: ### invalidated properties
Unless I misunderstand you, I don't think `invalidated_properties` does what you say.

`invalidated_properties` and `changed_properties` are two different lists, [meaning different things](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-properties).

The invalidated list is basically the other side on the bus being lazy and merely signaling that a property changed but not sending the value. Then the client has to go ask for the value if it actually cares.

Since that's a little annoying when you do care about the value, I added the flag `G_DBUS_PROXY_FLAGS_GET_INVALIDATED_PROPERTIES` so that the invalidated properties list can be ignored (I assume it's always empty) and I can just check the changed properties list instead.

### dictionary lookups

But yeah, I can use the dictionary variant calls instead of visibly looping through the list. Latest update calls `g_variant_lookup_value` instead of looping itself. And then passes the variant to `update_active_state`.

src/gamemode.c
mikix: I'm going to push back a little on this, but if you still disagree, I'm happy to inline this method.

I like having this method be separate because it has the "business logic" of greater-than-zero. Which is admittedly trivial logic. But still, it felt appropriate to have only one place that knew how to map ClientCount to Active state.

But it's not important, and I'm happy to inline.

GeorgesStavracas (approved):

Issue #899, opened 2022-10-07
Improve background apps


GeorgesStavracas:
After a long series of discussions on how to improve the handling of background apps, some consensus was reached, and parts of it involve xdg-desktop-portal.

The primary responsibility of xdg-desktop-portal is to determine which applications are considered to be running in background. Currently, this is done through the following steps:

 * x-d-p lists which Flatpak instances are running (let's call it I)
 * x-d-p asks backends for a list of running apps (let's call it R)
 * x-d-p determines that background apps B = I - (I ∩ R)

With that, x-d-p goes ahead and notified the backend about new background apps. The backend potentially shows an UI, and if the user denies that app the permission to run in background, x-d-p kills the application.

While this works, there are a few problems in this approach:

 1. At any point, there is no public, canonical list of background apps. It makes it impossible for backends to list these background apps anywhere in the UI.
 2. The GNOME backend, in particular, shows a notification that is both annoying and not very helpful. Without (1), it's all it can do.

In order to improve this situation, we agreed on the following changes:

 * [x] Stop interacting with the Access portal when apps call `RequestBackground()`
 * [x] Expose the list of background applications
   * [x] It must be in a separate D-Bus connection, we must not leak it to Flatpak sandboxes
 * [x] Add a new `SetStatus()` method to the Background portal
 * [x] Re-evaluate the frequency and heuristics used to poll background apps
 * [x] Notify backends about these changes

sophie-h:
Short question about daemons/monitors: Does "list of background applications" mean that a second process started in the Flatpak is treated as part of the original app?

hadess:
I'll need to review this more next week, but can we allowlist flatpak-builder, and can we find a way to also allowlist debug sessions started with, say, `--command=bash`?

GeorgesStavracas:
The only limit is how many PRs people can open for these QoL improvements. There's plenty on my plate already, but PRs are opened for that, I'll be happy to land.

sophie-h:
I tried to summarize the current approach and a bunch of questions I have. Hope that's okay.

- Every app can just run in the background by default.
- Every app can just activate autostart by default.
- Autostart and 'run in background' can both be disabled and re-enabled in Settings.
- For the re-enabling autostart to work, the portal stores the last autostart instructions on the drive.
- We have a background app list that shells can use.

1. Should we maybe add the simple state description, which is shown in the background app list mockups, as a new function to the background portal?
[...]

PR #901, opened 2022-10-11
Background apps monitoring

GeorgesStavracas:
libportal counterpart: https://github.com/flatpak/libportal/pull/115

See commits.

Most of the heavy lifting is pre-existing - the background portal already does the monitoring. This PR merely exposes this in a separate D-Bus name / path  / interface, in a way that's consumable by host system services. This is just the bare minimum to get this whole things up and running.

No backend or client changes are needed, all this work is restricted to xdg-desktop-portal itself.

This should be enough to implement richer interfaces around background running apps, such as [this proposal](https://gitlab.gnome.org/Teams/Design/os-mockups/-/issues/191). With this new background monitor interface, implementing such proposal is much easier.

TODO:

 * [x] Add a way for apps to set a status
 * [x] Review Access portal usage
 * [x] Fine tune sleep times and grace period

Related: https://github.com/flatpak/xdg-desktop-portal/issues/899


data/org.freedesktop.background.Monitor.xml
```diff
@@ -0,0 +1,52 @@
...
+      <variablelist>
+        <varlistentry>
+          <term>app_id s</term>
+          <listitem><para>
+            App id of the application.
+          </para></listitem>
+        </varlistentry>
+      </variablelist>
+    -->
+    <property name="BackgroundApps" type="a{sv}" access="read"/>
```
jadahl, line 48: Probably want this to be a `aa{sv}` to make it an array of dictionaries, one dictionary per app.

GeorgesStavracas: Ooof, that was the intention :facepalm:

GeorgesStavracas: Done. Please please please double check if my usage of GVariantBuilder is correct - as far as I can see, the D-Bus property has the expected result, but I'm never confident when it comes to variants

src/background-monitor.c
```diff
@@ -0,0 +1,164 @@
...
+                                   "exit-on-close", TRUE,
+                                   NULL);
+
+  if (!g_initable_init (G_INITABLE (self->connection), cancellable, error))
+    return FALSE;
+
+  g_dbus_interface_skeleton_set_flags (G_DBUS_INTERFACE_SKELETON (initable),
+                                       G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
+
+  /* TODO: dos it need to listen to 'g-authorize-method'? */
```
jadahl, line 107: I guess not? But I guess this needs double checking that the interface isn't reachable or discoverable from sandboxes. I wonder if we can add tests that run flatpak sandboxes that try to escape to this interface.

src/background-monitor.c
```diff
@@ -0,0 +1,164 @@
...
+                                   "flags", G_DBUS_CONNECTION_FLAGS_AUTHENTICATION_CLIENT |
+#if GLIB_CHECK_VERSION(2,74,0)
+                                            G_DBUS_CONNECTION_FLAGS_CROSS_NAMESPACE |
+#endif
+                                            G_DBUS_CONNECTION_FLAGS_MESSAGE_BUS_CONNECTION,
+                                   "exit-on-close", TRUE,
+                                   NULL);
+
+  if (!g_initable_init (G_INITABLE (self->connection), cancellable, error))
+    return FALSE;
```
jadahl, line 102: These two calls can be shrunk into a single `g_initable_new()`.

src/background-monitor.c
```diff
@@ -0,0 +1,164 @@
...
+{
+  iface->init = background_monitor_initable_init;
+}
+
+static void
+background_monitor_finalize (GObject *object)
+{
+  BackgroundMonitor *self = (BackgroundMonitor *)object;
+
+  g_dbus_connection_flush_sync (self->connection, NULL, NULL);
```
jadahl, line 134: This will crash if `g_dbus_address_get_for_bus_sync()` failed on init.

src/background-monitor.c
```diff
@@ -0,0 +1,164 @@
...
+BackgroundMonitor *
+background_monitor_new (GCancellable  *cancellable,
+                        GError       **error)
+{
+  return g_initable_new (BACKGROUND_TYPE_MONITOR,
+                         cancellable,
+                         error,
+                         NULL);
+}
+
```
jadahl, line 164: Trailing newline

src/background.c
```diff
@@ -600,6 +600,44 @@ remove_outdated_instances (int stamp)
...
 
+static void
+update_background_monitor_properties (GHashTable *app_states)
+{
+  GVariantBuilder builder;
+  GHashTableIter iter;
+  InstanceData *data;
+  char *id;
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE_ARRAY);
```
jadahl, line 611: Perhaps `G_VARIANT_TYPE ("aa{sv}");` would be "safer"; e.g. I'm not sure what happens if the array is empty? Or will this some how be magically turned into the right type by the D-Bus marshaling code?

jadahl (commented):
Looks pretty straight forward to me.

src/background.c
GeorgesStavracas: Done

src/background-monitor.c
GeorgesStavracas: Done

data/org.freedesktop.background.Monitor.xml
```diff
@@ -0,0 +1,58 @@
...
+
+      <variablelist>
+        <varlistentry>
+          <term>app_id s</term>
+          <listitem><para>
+            App id of the application.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>message s</term>
```
GeorgesStavracas, line 47: Opted for `message` here, but naming suggestions are welcomed

GeorgesStavracas:
Okay, from an API / services point of view, I think this is ready again. I'm still not entirely happy with the sleep times, but this iteration feels a lot more responsive compared to 30s + 30s.

data/org.freedesktop.background.Monitor.xml
```diff
@@ -0,0 +1,64 @@
...
+          <term>app_id s</term>
+          <listitem><para>
+            App id of the application.
+          </para></listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>instance s</term>
+          <listitem><para>
+            The Flatpak instance of the application.
+          </para></listitem>
```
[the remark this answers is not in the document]
jadahl: Perhaps leave it out until there is a user?

src/background.c
```diff
@@ -647,10 +647,11 @@ background_monitor (gpointer data)
     {
       g_main_context_iteration (monitor_context, TRUE);
       /* We check twice, to avoid killing unlucky apps hit at a bad time */
-      sleep (30);
+      sleep (5);
       check_background_apps ();
-      sleep (30);
+      sleep (10);
```
jadahl, line 652: So we check, sleep 10s, check again. That means you went from 30 to 10 seconds grace period until killing, if you're unlucky?

GeorgesStavracas: That *would* have been the case, if we didn't change the policy together. This PR makes the background portal more lenient with these apps, and paired with https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/56, we only kill apps explicitly forbidden from running in background.

data/org.freedesktop.background.Monitor.xml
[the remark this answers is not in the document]
GeorgesStavracas: I'm about 80% convinced we'll need it to implement the "Stop this app" part of the proposed mockups...

src/background.c
jadahl: That's a good point; it allows us to be more agressive when checking.

data/org.freedesktop.background.Monitor.xml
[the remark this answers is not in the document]
GeorgesStavracas: Yup, it's required for implementing killing the app.

GeorgesStavracas:
I've written a little demo python app to showcase this new background monitoring service: https://gitlab.gnome.org/-/snippets/4305

To run, download the file, then run it in a terminal.

Here's an example of it working:

[Gravação de tela de 2022-10-20 13-20-37.webm](https://user-images.githubusercontent.com/3518204/197013567-c47fa97f-b8f2-46bc-87be-e4695ef6e6ea.webm)

We probably need to be more aggressive with the detection timings...

GeorgesStavracas:
Last push should make it a lot more eager in detecting background apps!

src/background-monitor.c
matthiasclasen: The idea was that we need a separate unique name to avoid confusing the proxy filtering?

src/background.c
```diff
@@ -600,6 +600,46 @@ remove_outdated_instances (int stamp)
...
+      if (!flatpak_instance_is_running (data->instance))
+        continue;
+
+      id = flatpak_instance_get_id (data->instance);
+      app_id = flatpak_instance_get_app (data->instance);
+      g_assert (app_id != NULL);
+
+      g_variant_builder_init (&app_builder, G_VARIANT_TYPE_VARDICT);
+      g_variant_builder_add (&app_builder, "{sv}", "app_id", g_variant_new_string (app_id));
+      g_variant_builder_add (&app_builder, "{sv}", "instance", g_variant_new_string (id));
```
matthiasclasen, line 384: The "instance" key was not mentioned in the interface description, I think.

data/org.freedesktop.background.Monitor.xml
matthiasclasen: It would be useful to provide some guidance:

This should be a single line that can be presented to the user in a list. Not a full sentence or paragraph

data/org.freedesktop.portal.Background.xml
```diff
@@ -103,6 +103,29 @@
...
+
+      Sets the status of the application running in background.
+
+      Supported keys in the @options vardict include:
+      <variablelist>
+        <varlistentry>
+          <term>message s</term>
+          <listitem><para>
+            A string that will be used as the status message of the
+            application. Must be shorter than 96 characters.
```
matthiasclasen, line 118: Ah, you have some of that here

src/background.c
```diff
@@ -379,6 +382,8 @@ update_background_monitor_properties (void)
       g_variant_builder_init (&app_builder, G_VARIANT_TYPE_VARDICT);
       g_variant_builder_add (&app_builder, "{sv}", "app_id", g_variant_new_string (app_id));
       g_variant_builder_add (&app_builder, "{sv}", "instance", g_variant_new_string (id));
+      if (data->status_message)
+        g_variant_builder_add (&app_builder, "{sv}", "message", g_variant_new_string (data->status_message));
```
matthiasclasen, line 386: i guess the "message" key should also be mentioned in the xml interface description

matthiasclasen:
Looks fine, apart from the minor comments I left.

src/background-monitor.c
GeorgesStavracas: Yes, we need a unique name, and to achieve that with GDBus the easiest approach was creating a separate connection.

src/background.c
```diff
@@ -204,419 +220,151 @@ set_permission (const char *app_id,
...
-                                          gpointer task_data,
-                                          GCancellable *cancellable)
+/* The background monitor is running in a dedicated thread.
+ *
+ * We rely on the RunningApplicationsChanged signal from the backend to get
+ * notified about applications that start or stop having open windows, and on
+ * file monitoring to learn about flatpak instances appearing and disappearing.
+ *
+ * When either of these changes happens, we wake up the background monitor
+ * thread, and it will do a check the state of applications a few times, with
```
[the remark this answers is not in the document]
GeorgesStavracas: This is a pre-existing comment that I just shuffled around :) But I'll fix it

src/background.c
GeorgesStavracas: It's described in `org.freedesktop.background.Monitor` already

data/org.freedesktop.portal.Background.xml
GeorgesStavracas: Yes, but I've applied your previous suggestion, and made the validation a bit more strict on the portal too

src/background.c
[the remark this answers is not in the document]
GeorgesStavracas: Added an extra commit adjusting this comment

src/background.c
GeorgesStavracas: It's described in `org.freedesktop.background.Monitor` already, in the introductory commit. Not sure if you're thinking of `org.freedesktop.portal.Background`.

data/org.freedesktop.background.Monitor.xml
GeorgesStavracas: Yup

GeorgesStavracas:
ACKed on the Matrix channel, let's land

PR #909, opened 2022-11-09
xdp-utils: Add AppInfo Kind TEST for testing purposes

3v1n0:
Make possible to test applications by allow mocking various applications properties and capabilities.

This is simpler for now, but will help various cases in future.

We can then avoid doing "empty" checks assuming we're in the test cases, replacing them by explicit kind tests, while be able to do permission checks with more granularity.

It should help testing PR's such as https://github.com/flatpak/xdg-desktop-portal/pull/786 and many other cases were we need custom behavior for testing.

Splitting this from #904 for easier reviewing.

### DIFF
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -680,7 +680,7 @@ handle_request_install_token (XdpDbusDynamicLauncher *object,
    * app was launched from the CLI:
    * https://github.com/flatpak/xdg-desktop-portal/pull/719#issuecomment-1057412221
    */
-  if (xdp_app_info_is_host (call->app_info) && g_str_equal (app_id, ""))
+  if (xdp_app_info_get_kind (call->app_info) == XDP_APP_INFO_KIND_TEST)
     {
       response = 0;
     }
--- a/src/location.c
+++ b/src/location.c
@@ -507,6 +507,7 @@ handle_start_in_thread_func (GTask *task,
   if (!get_location_permissions (request->app_info, &accuracy, &last_used))
     {
       guint access_response = 2;
+      g_autoptr(GAppInfo) info = NULL;
       g_autoptr(GVariant) access_results = NULL;
       g_autoptr(XdpDbusImplRequest) impl_request = NULL;
       GVariantBuilder access_opt_builder;
@@ -530,24 +531,21 @@ handle_start_in_thread_func (GTask *task,
       g_variant_builder_add (&access_opt_builder, "{sv}",
                              "icon", g_variant_new_string ("find-location-symbolic"));
 
-      if (g_str_equal (app_id, ""))
+      info = xdp_app_info_load_app_info (request->app_info);
+      if (info == NULL)
         {
           title = g_strdup (_("Grant Access to Your Location?"));
           subtitle = g_strdup (_("An application wants to use your location."));
         }
       else
         {
-          g_autofree char *id = NULL;
-          g_autoptr(GDesktopAppInfo) info = NULL;
           const char *name;
 
-          id = g_strconcat (app_id, ".desktop", NULL);
-          info = g_desktop_app_info_new (id);
-          name = g_app_info_get_display_name (G_APP_INFO (info));
-
+          name = g_app_info_get_display_name (info);
           title = g_strdup_printf (_("Give %s Access to Your Location?"), name);
-          if (g_desktop_app_info_has_key (info, "X-Geoclue-Reason"))
-            subtitle = g_desktop_app_info_get_string (info, "X-Geoclue-Reason");
+          if (G_DESKTOP_APP_INFO (info) &&
+              g_desktop_app_info_has_key (G_DESKTOP_APP_INFO (info), "X-Geoclue-Reason"))
+            subtitle = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (info), "X-Geoclue-Reason");
           else
             subtitle = g_strdup_printf (_("%s wants to use your location."), name);
         }
--- a/src/xdp-test-utils.h
+++ b/src/xdp-test-utils.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright © 2022, Canonical Ltd.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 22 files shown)


src/xdp-utils.c
```diff
@@ -808,6 +856,48 @@ parse_app_info_from_snap (pid_t pid, GError **error)
...
+  g_autofree char *test_file_app_info = NULL;
+  g_autofree char *key_data = NULL;
+  size_t key_data_len;
+
+  g_assert (g_get_user_data_dir () != NULL);
+
+  test_file_app_info =
+    g_build_filename (g_get_user_data_dir (), TEST_METADATA_FILE_NAME, NULL);
+
+  if (!g_file_get_contents (test_file_app_info, &key_data, &key_data_len, NULL))
```
jadahl, line 874: NULL -> error

src/location.c
```diff
@@ -530,24 +531,21 @@ handle_start_in_thread_func (GTask *task,
...
 
-          id = g_strconcat (app_id, ".desktop", NULL);
-          info = g_desktop_app_info_new (id);
-          name = g_app_info_get_display_name (G_APP_INFO (info));
-
+          name = g_app_info_get_display_name (info);
           title = g_strdup_printf (_("Give %s Access to Your Location?"), name);
-          if (g_desktop_app_info_has_key (info, "X-Geoclue-Reason"))
-            subtitle = g_desktop_app_info_get_string (info, "X-Geoclue-Reason");
+          if (G_DESKTOP_APP_INFO (info) &&
```
jadahl, line 546: This cast and NULL check seems unnecessary.

src/xdp-utils.c
```diff
@@ -1204,11 +1204,11 @@ xdp_set_documents_mountpoint (const char *path)
 
 /* alternate_document_path converts a file path  */
 char *
-xdp_get_alternate_document_path (const char *path, const char *app_id)
+xdp_get_alternate_document_path (const char *path, const char *app_id, XdpAppInfoKind kind)
```
jadahl, line 1207: One argument per line

src/xdp-utils.h
```diff
@@ -114,7 +114,7 @@ char       *xdp_app_info_get_tryexec_path (XdpAppInfo  *app_info);
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(XdpAppInfo, xdp_app_info_unref)
 
 void  xdp_set_documents_mountpoint    (const char *path);
-char *xdp_get_alternate_document_path (const char *path, const char *app_id);
+char *xdp_get_alternate_document_path (const char *path, const char *app_id, XdpAppInfoKind kind);
```
jadahl, line 117: One argument per line

src/dynamic-launcher.c
```diff
@@ -680,7 +680,7 @@ handle_request_install_token (XdpDbusDynamicLauncher *object,
    * app was launched from the CLI:
    * https://github.com/flatpak/xdg-desktop-portal/pull/719#issuecomment-1057412221
    */
-  if (xdp_app_info_is_host (call->app_info) && g_str_equal (app_id, ""))
+  if (xdp_app_info_get_kind (call->app_info) == XDP_APP_INFO_KIND_TEST)
```
jadahl, line 683: This looks like it'll break the check for unsandboxed apps.

src/xdp-utils.c
```diff
@@ -1204,11 +1204,11 @@ xdp_set_documents_mountpoint (const char *path)
...
 {
   int len;
 
-  if (g_str_equal (app_id, ""))
+  if (g_str_equal (app_id, "") || kind == XDP_APP_INFO_KIND_TEST)
```
jadahl, line 1211: Can't this be handled by the tests calling `xdp_set_documents_mountpoint()`?

tests/test-portals.c
```diff
@@ -151,9 +240,18 @@ global_setup (void)
   g_mkdtemp (outdir);
   g_debug ("outdir: %s\n", outdir);
 
+  /* FIXME: Instead of this we should really use G_TEST_OPTION_ISOLATE_DIRS
+   * but it would require more rework to ensure that some desktop files are
+   * still found.
+   */
+  g_assert_no_errno (g_chdir (outdir));
```
jadahl, line 247: Urg. Could you expand?

src/dynamic-launcher.c
3v1n0: unsandboxed apps may have an app-id aren't they? In any case I can check for both HOST and TEST ones to be sure.

src/location.c
3v1n0: it was meant to be a an `IS` check, but however it's always true in the current code but `GAppInfo` may also be implemented differently.

src/xdp-test-utils.h
```diff
@@ -0,0 +1,28 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Marco Trevisan <marco.trevisan@canonical.com>
+ */
+
+#define TEST_METADATA_FILE_NAME ".xdg_app_info"
```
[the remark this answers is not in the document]
3v1n0: Well, it's a file that goes in a build or temporary path anyways... No specific reason, nor a reson to keep it visible. Whatever works for us.

src/xdp-utils.c
3v1n0: So, the point here was that the function cal also return `NULL` when there is not an error... Like we accept that the file does not exists, so we don't have a a test-app, but it's not an error.

And the code calling this take care of it (I think other types do the same).

tests/test-portals.c
3v1n0: So.... Ideally I was still hacking to find a solution to avoid this and take the `G_TEST_OPTION_ISOLATE_DIRS` route, but in practice.

By using `G_TEST_OPTION_ISOLATE_DIRS` each single test has a separate working temporary directory, and I ideally would like to create our test-app-info file there, so that there's no risk that a test interfere with another one that is running afterwards.

However, right now some tests can't work with such flag, and so I kept using the `outdir` that has been used so far as the "test working directory". This is far from being ideal though, as we the have the right machinery to keep tests isolated without relying on a single place.

swick:
With https://github.com/flatpak/xdg-desktop-portal/pull/1369 merged, we now have a XdpAppInfo specifically for tests. This MR can be closed.

mcatanzaro:
> With https://github.com/flatpak/xdg-desktop-portal/pull/1369 merged, we now have a XdpAppInfo specifically for tests. This MR can be closed.

I don't have permission to close pull requests here. Marco or a maintainer will need to do so.

PR #914, opened 2022-11-22
global-shortcuts: Responses should succeed if nothing wrong happened

aleixpol:
We were always returning 2, which is not correct.

### DIFF
--- a/data/org.freedesktop.portal.GlobalShortcuts.xml
+++ b/data/org.freedesktop.portal.GlobalShortcuts.xml
@@ -167,7 +167,7 @@
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="request_handle" direction="out"/>
 
-      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QMap&lt;QString,QVariantMap&gt;"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QList&lt;QPair&lt;QString,QVariantMap&gt;&gt;"/>
       <annotation name="org.qtproject.QtDBus.QtTypeName.In3" value="QVariantMap"/>
     </method>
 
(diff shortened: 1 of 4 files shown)


src/global-shortcuts.c
```diff
@@ -394,6 +395,7 @@ shortcuts_listed_cb (GObject *source_object,
           results = g_variant_ref_sink (g_variant_builder_end (&results_builder));
         }
 
+      response = 0;
```
jadahl, line 398: These tends to be in the impl method call, e.g. https://github.com/flatpak/xdg-desktop-portal/blob/8255fabc37cd6a943c07a0b469e5a1a944aeaecf/data/org.freedesktop.impl.portal.ScreenCast.xml#L250 and https://github.com/flatpak/xdg-desktop-portal/blob/8255fabc37cd6a943c07a0b469e5a1a944aeaecf/data/org.freedesktop.impl.portal.Inhibit.xml#L85.

There hasn't been a release yet, so we can still add the `response` argument where it's missing.

aleixpol: Addressed. Added a couple of extra changes of things that cropped up when testing it.

data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -160,10 +160,23 @@
...
         @session_handle: Object path for the #org.freedesktop.impl.portal.Session object representing the session
-        @options: Unused so far
+        @options: See below
 
         Lists all shortcuts.
 
+        Supported keys in the @options vardict include:
+        <variablelist>
+          <varlistentry>
+            <term>handle_token s</term>
```
jadahl, line 170: Shouldn't `handle_token` be added to `BindShortcuts` too?

jadahl (approved):
lgtm

aleixpol:
Hi @jadahl, thanks for your review. Please note I do not have powers to merge so somebody will have to press the button for the changes to go in.

jadahl:
> Hi @jadahl, thanks for your review. Please note I do not have powers to merge so somebody will have to press the button for the changes to go in.

I don't have any merge rights either.

src/global-shortcuts.c
```diff
@@ -445,8 +446,7 @@ handle_list_shortcuts (XdpDbusGlobalShortcuts *object,
                                                       request->id,
                                                       arg_session_handle,
                                                       NULL,
-                                                      shortcuts_listed_cb,
-                                                      g_object_ref (request));
+                                                      shortcuts_listed_cb, g_object_ref (request));
```
GeorgesStavracas, line 450: Please don't change this

GeorgesStavracas (changes requested):
One comments, LGTM otherwise

src/global-shortcuts.c
aleixpol: Addressed.

GeorgesStavracas (approved):

PR #921, opened 2022-11-28
Use a dbus service name as its app_id

aleixpol:
Also decode the unit name for dbus services and offer to the implementation for some extra context of what is going on.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -189,6 +189,40 @@ _xdp_parse_app_id_from_unit_name (const char *unit)
 
   return g_steal_pointer (&app_id);
 }
+
+char *
+_xdp_parse_app_id_from_dbus_name (const char *unit)
+{
+  g_autoptr(GRegex) regex = NULL;
+  g_autoptr(GMatchInfo) match = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *app_id = NULL;
+
+  g_assert (g_str_has_prefix (unit, "dbus-"));
+
+  // Parses a unit name such as dbus-:1.2-org.kde.kdeconnect@7.service
+  regex = g_regex_new ("^dbus-:[[:digit:]]+\\.[[:digit:]]+-"
+                       "(.+?\\..+?\\..+?)(?:[@\\-][[:alnum:]]*)\\.service$",
+                       0, 0, &error);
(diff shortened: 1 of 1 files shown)


swick:
I wonder if dbus-broker should use the app prefix if there is a corresponding desktop file. The dbus prefixed name seems like an implementation detail we should probably not rely on.

aleixpol:
I'm not sure, do you know who should be asked?

I keep going back-and-forth on this though. I think that ideally it should be the full .service filename, as in `org.gnome.Identity.service` or `org.kde.kdeconnect.service`.

As is, it's not giving us much information anyway.

swick:
https://github.com/bus1/dbus-broker/issues/302

src/xdp-utils.c
```diff
@@ -189,6 +189,40 @@ _xdp_parse_app_id_from_unit_name (const char *unit)
...
+_xdp_parse_app_id_from_dbus_name (const char *unit)
+{
+  g_autoptr(GRegex) regex = NULL;
+  g_autoptr(GMatchInfo) match = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *app_id = NULL;
+
+  g_assert (g_str_has_prefix (unit, "dbus-"));
+
+  // Parses a unit name such as dbus-:1.2-org.kde.kdeconnect@7.service
```
smcv, line 203: This naming convention for systemd units is specific to dbus-broker.

On systems that have systemd activation available, the reference implementation (dbus-daemon) does not guarantee that activated services are in any sort of systemd service or scope: D-Bus services with `SystemdService=foo.service` are run as `foo.service` (probably also true for dbus-broker), and D-Bus services without `SystemdService` are run via "traditional activation" as children of `dbus.service` (unlike dbus-broker).

smcv: Having said that, if a unit name matches this pattern then you can probably be quite confident that it's named according to how dbus-broker does it.

ebassi:
@aleixpol Could you please rebase this pull request?

swick:
I still believe that we should fix this in dbus-broker instead of adding more non-standard cgroup naming schemes.

swick:
I'm poking dbus-broker upstream again to see if we can move forward with changing the naming scheme for transient units.

PR #926, opened 2022-12-07
CI: make the autotools vs meson decision a matrix field

whot:
This sits on top of #925 

Instead of two near-identical workflow jobs, push the buildsystem into the matrix and make the various `run:` commands dependent on that matrix value. This removes *some* duplication, albeit not that much. But it's probably better than having this separately.

In particular, I'm interested in the `CI: only pass the sanitize CFLAGS through for autotools` commit because that is what makes building libei from git harder in #714.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -153,7 +153,6 @@ jobs:
 
     - name: Upload test logs
       uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):
LGTM but needs manual rebase

Issue #691, opened 2022-01-14
Add `viewport` attribute in docs


TheEvilSkeleton:
Is it possible to add `<meta name="viewport" content="width=device-width, initial-scale=1.0">` in the HTML? I have no idea how to add tags.

I wish to improve the docs by making it mobile-friendly.

GeorgesStavracas:
I have no idea either. Somebody with web knowledge, please help.

bilelmoussaoui:
i don't think you can easily do that as the html is generated by xmlto from the xml file
so you would have to add some post processing stage that would modify the html output and append the meta tag

PR #931, opened 2022-12-12
doc: Add viewport meta tag

YarnSphere:
Fixes: #691

### DIFF
--- a/doc/xmlto-config.xsl
+++ b/doc/xmlto-config.xsl
@@ -1,6 +1,13 @@
 <?xml version='1.0'?>
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                 xmlns:fo="http://www.w3.org/1999/XSL/Format"
+                exclude-result-prefixes="fo"
                 version="1.0">
   <xsl:param name="html.stylesheet" select="'docbook.css'"/>
+
+  <xsl:template name="system.head.content">
+    <meta xmlns="http://www.w3.org/1999/xhtml"
+          name="viewport"
+          content="width=device-width, initial-scale=1.0"/>
+  </xsl:template>
 </xsl:stylesheet>


doc/xmlto-config.xsl
```diff
@@ -1,6 +1,13 @@
 <?xml version='1.0'?>
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                 xmlns:fo="http://www.w3.org/1999/XSL/Format"
+                exclude-result-prefixes="fo"
                 version="1.0">
   <xsl:param name="html.stylesheet" select="'docbook.css'"/>
+
+  <xsl:template name="system.head.content">
+    <meta xmlns="http://www.w3.org/1999/xhtml"
```
YarnSphere, line 9: Without the `xmlns` attribute here, the generated HTML was:
```
<meta xmlns="" name="viewport" content="width=device-width, initial-scale=1.0"/>
```

doc/xmlto-config.xsl
```diff
@@ -1,6 +1,13 @@
 <?xml version='1.0'?>
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                 xmlns:fo="http://www.w3.org/1999/XSL/Format"
+                exclude-result-prefixes="fo"
```
YarnSphere, line 4: Without this line, the generated HTML was:
```html
<meta xmlns:fo="http://www.w3.org/1999/XSL/Format" name="viewport" content="width=device-width, initial-scale=1.0"/>
```

Another option (which I _believe_ is fine, and probably cleaner) is to remove the `xmlns:fo` line together with this newly added one. I don't think that namespace is being used anywhere and should be useless.

TheEvilSkeleton (approved):
Works locally.

PR #932, opened 2022-12-14
doc: Make docs fully responsive

TheEvilSkeleton:
This makes the docs page fully responsive.


GeorgesStavracas (approved):
Since you're touching docs, how do you feel about prettifying them a bit more? Not in this PR of course

TheEvilSkeleton:
> Since you're touching docs, how do you feel about prettifying them a bit more? Not in this PR of course

Definitely. I'm contributing to the docs for that reason :)

PR #941, opened 2023-01-03
file-chooser: Set writable flag correctly

A6GibKm:
Previously this was evaluating as if (gvariant != NULL) which is always true if the option is present, this caused all files to be passed down as read-write regardless of what the impl, app and user requested.

I tested this change locally using Bustle and one can see that the permission store gets 

```
([handle 0], uint32 7, 'com.belmoussaoui.ashpd.demo', ['read', 'grant-permissions'])
```

rather than 

```
([handle 0], uint32 7, 'com.belmoussaoui.ashpd.demo', ['read', 'write', 'grant-permissions'])
```

when the user ticks the "read-only" choice in the file picker.

### DIFF
--- a/src/file-chooser.c
+++ b/src/file-chooser.c
@@ -79,6 +79,7 @@ send_response_in_thread_func (GTask        *task,
   const char **uris;
   GVariant *choices;
   GVariant *current_filter;
+  GVariant *writable;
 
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):

GeorgesStavracas:
Looks correct to me - let's see if CI is happy

PR #945, opened 2023-01-11
Create be.po

Bugomol:
Belarusian translation for xdg-desktop-portal


A6GibKm:
You forgot to update `po/LINGUAS`.

GeorgesStavracas (approved):

Issue #906, opened 2022-11-03
Replace UseIn with more descriptive declarative keys


ebassi:
The backend to use is chosen on a per-interface basis, so it is OK to split functionality between backends. For example, it's fine to choose `foo.portal` for `o.fd.impl.portal.One` and `bar.portal` for `o.fd.impl.portal.Two`.

GNOME is a concrete example of this: GNOME-appropriate behaviour for various interfaces is distributed between x-d-p-gnome (`gnome.portal`), x-d-p-gtk (`gtk.portal`), gnome-shell (`gnome-shell.portal`) and gnome-keyring (`gnome-keyring.portal`).

The current priority/fallback behaviour is:

* First, x-d-p enumerates all portals whose `.portal` file declares that they are suitable for the `$XDG_CURRENT_DESKTOP` via the `UseIn` field.
* If this results in only finding one implementation of an interface, it is used, and we stop here: for example, in a GNOME environment, the only implementation of `o.fd.impl.portal.Secret` is `gnome-keyring.portal`.
* If this results in finding more than one implementation of an interface, the first one in lexicographic order is used, and we stop here.
    * For GNOME users, this happens to do the right thing for `x-d-p-gtk` vs. `x-d-p-gnome`, because `gnome` happens to sort lexicographically lower than `gtk`, but that's more like a lucky accident than by design.
    * For users of non-GNOME GTK-based desktop environments like XFCE, this does not necessarily do the right thing: if `x-d-p-gtk` and `x-d-p-gnome` are both installed, maybe XFCE users would prefer to use `-gtk`? (Although arguably that should be solved by having `UseIn=GNOME;XFCE;etc.;` in `-gtk`.)
* If this results in finding *no* implementations of an interface, then the fallback is to ignore the `UseIn` field, and enumerate all portals indiscriminately. Again, the first one in lexicographic order is used.
* If there are still no portals found, then x-d-p does not implement the interface.

This is not ideal. Ideally, we would be able to distinguish between:

* implementations that *can be* used from any desktop environment, but are less ideal in some desktop environments
    * for example, a GTK `FileChooser` is primarily for use with GTK-based desktops like GNOME, MATE, Cinnamon, LXDE and XFCE; it will look out-of-place in a Qt-based desktop, so ideally in a Qt-based desktop like LXQT, we would use something provided by the desktop instead; but it will still *work* fine in any desktop environment, so it's always better than nothing, and we can safely use it as a fallback
[...]

PR #955, opened 2023-01-19
Add portals configuration file for desktop environments

ebassi:
Replace the `UseIn` key in each portal implementation with a per-desktop configuration file.

Desktop environments (either upstream or downstream) are supposed to ship the configuration file to describe which portal implementation to use for each portal interface.

Fixes: #906


ebassi:
Open questions:

- deprecation policy for `UseIn`; add a warning, or just silently discard it?

The CI failure on 16.04 is unrelated, and seems to be an infrastructure issue.

src/portal-impl.c
```diff
@@ -270,6 +304,119 @@ load_installed_portals (gboolean opt_verbose)
...
+                 path,
+                 desktops[i]);
+
+      ifaces = g_key_file_get_keys (key_file, "preferred", NULL, NULL);
+
+      g_autoptr(GPtrArray) interfaces = g_ptr_array_new_full (g_strv_length (ifaces) + 1, NULL);
+
+      if (ifaces != NULL)
+        {
+          for (size_t i = 0; ifaces[i] != NULL; i++)
```
GeorgesStavracas: Instead of nesting an if and a for loop, I'd rather have `for (size_t i = 0; ifaces != NULL && ifaces[i] != NULL; i++)`

GeorgesStavracas (changes requested):

src/portal-impl.c
```diff
@@ -270,6 +304,119 @@ load_installed_portals (gboolean opt_verbose)
...
+          conf->ifaces = (PortalInterface **) g_ptr_array_steal (interfaces, NULL);
+
+          config = g_steal_pointer (&conf);
+          return;
+        }
+    }
+}
+
+static gboolean
+portal_impl_matches_config (const PortalImplementation *impl, const char *interface)
```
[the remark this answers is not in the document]
ebassi: The rest of the file is kind of inconsistent, so I've tried following the existing style. Should I fix the rest of the definitions as well?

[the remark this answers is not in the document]
GeorgesStavracas: That would be appreciated, if you're up for doing that, please do as a separate commit

src/portal-impl.c
```diff
@@ -316,65 +369,48 @@ load_portal_configuration (gboolean opt_verbose)
   if (portal_dir == NULL)
     portal_dir = DATADIR "/xdg-desktop-portal";
 
-  conf = g_new0 (PortalConfig, 1);
+  user_portal_dir = g_build_filename (g_get_user_config_dir (),
```
Sodivad, line 377: For system config the data_dir is used, for symmetry I think get_user_data_dir would make more sense.

src/portal-impl.c
```diff
@@ -316,65 +369,48 @@ load_portal_configuration (gboolean opt_verbose)
...
+                     portals_conf,
+                     desktops[i]);
 
-      g_autoptr(GPtrArray) interfaces = g_ptr_array_new_full (g_strv_length (ifaces) + 1, NULL);
+          config = g_steal_pointer (&conf);
+          return;
+        }
 
-      if (ifaces != NULL)
+      conf = load_portal_configuration_for_dir (opt_verbose, user_portal_dir, portals_conf);
```
Sodivad, line 398: Isn't this the wrong way around? Normally user configuration overrides system configuration while here system configuration takes precedence.

src/portal-impl.c
```diff
@@ -198,18 +309,163 @@ load_installed_portals (gboolean opt_verbose)
...
+
+      for (size_t i = 0; ifaces[i] != NULL; i++)
+        {
+          g_autoptr(PortalInterface) interface = g_new0 (PortalInterface, 1);
+
+          /* Special case "default" as the fallback */
+          if (strcmp (ifaces[i], "default") == 0)
+            interface->dbus_name = g_strdup ("default");
+          else
+            interface->dbus_name = g_strdup (ifaces[i]);
```
Sodivad, line 337: Aren't both branches the same?

src/portal-impl.c
```diff
@@ -220,14 +476,22 @@ find_portal_implementation (const char *interface)
...
+            {
+              g_debug ("Using %s.portal for %s in %s (config)", impl->source, interface, desktops[i]);
+              return impl;
+            }
+
+          /* Fallback */
+          if (impl->use_in != NULL && g_strv_case_contains ((const char **)impl->use_in, desktops[i]))
             {
-              g_debug ("Using %s for %s in %s", impl->source, interface, desktops[i]);
+              g_debug ("Using %s.portal for %s in %s (fallback)", impl->source, interface, desktops[i]);
```
Sodivad, line 488: I think this needs to be in a separate loop otherwise you could use a fallback even though we have  a portal that matches the config.

Sodivad (commented):
This doesn't seem to work as I expect
```
[preferred]
default=kde
org.freedesktop.impl.portal.FileChooser=gtk
```

src/portal-impl.c
```diff
@@ -41,6 +41,84 @@ portal_implementation_free (PortalImplementation *impl)
...
+  g_ptr_array_add (valid_desktops, NULL);
+  g_free (tmp);
+
+  tmp = (char **) g_ptr_array_steal (valid_desktops, NULL);
+  g_ptr_array_unref (valid_desktops);
+
+  return tmp;
+}
+
+static const char **
```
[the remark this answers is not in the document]
ebassi: I can switch, but since I lifted this code from GLib, I'm trying to keep the changes minimal, in case we have to rebase it.

src/portal-impl.c
```diff
@@ -29,6 +29,35 @@
...
+  g_free (iface);
+}
+
+static void
+portal_config_free (PortalConfig *config)
+{
+  g_free (config->source);
+  for (size_t i = 0; i < config->n_ifaces; i++)
+    portal_interface_free (config->ifaces[i]);
+  g_free (config);
```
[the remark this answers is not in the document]
ebassi: `g_free()` is NULL-safe, guaranteed by the C standard, and clearing the pointer right before destruction is just going to be extra work.

Anyway, I can change the rest of the destructors to do that.

src/portal-impl.c
ebassi: There's also the option of using sysconfdir for the system configuration. I am fine with either.

src/portal-impl.c
ebassi: Indeed, I left this in because of the test suite and never rolled it back.

There's also the option of merging multiple layers of configuration, with the last one winning, but I'll leave that for another pull request.

src/portal-impl.c
ebassi: Yes, though it makes it clear which one it is; I'm also open to changing the default key.

src/portal-impl.c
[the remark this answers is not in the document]
GeorgesStavracas: Sure, I'm familiar with these guarantees, I've asked this not because of NULL-safety but rather to keep things a bit more consistent from now on :slightly_smiling_face:

matthiasclasen:
> Open questions:
> 
>     * deprecation policy for `UseIn`; add a warning, or just silently discard it?
> 

I think a warning would be appropriate.

matthiasclasen:
I sadly don't find `UseIn` mentioned in any docs we ship, so portal configuration is probably entirely undocumented.

It might be nice to start changing that by including a manpage for the portals.conf file format.

src/portal-impl.c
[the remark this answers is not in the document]
ebassi: Switching generates a bunch of compiler warnings that I'm not entirely sure are correct.

src/portal-impl.c
```diff
@@ -41,6 +41,84 @@ portal_implementation_free (PortalImplementation *impl)
...
+static char **
+get_valid_current_desktops (const char *value)
+{
+  if (value == NULL)
+    value = g_getenv ("XDG_CURRENT_DESKTOP");
+  if (value == NULL)
+    value = "";
+
+  char **tmp = g_strsplit (value, G_SEARCHPATH_SEPARATOR_S, 0);
+  GPtrArray *valid_desktops = g_ptr_array_new_full (g_strv_length (tmp) + 1, g_free);
```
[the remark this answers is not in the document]
ebassi: Done

src/portal-impl.c
```diff
@@ -270,6 +304,119 @@ load_installed_portals (gboolean opt_verbose)
...
+        continue;
+
+      if (opt_verbose)
+        g_debug ("Loading portal configuration file '%s' for desktop '%s'",
+                 path,
+                 desktops[i]);
+
+      ifaces = g_key_file_get_keys (key_file, "preferred", NULL, NULL);
+
+      g_autoptr(GPtrArray) interfaces = g_ptr_array_new_full (g_strv_length (ifaces) + 1, NULL);
```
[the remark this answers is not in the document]
ebassi: Done.

src/portal-impl.c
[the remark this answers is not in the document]
ebassi: Done, in a separate commit.

src/portal-impl.c
ebassi: Once you hoist the variable declarations, the condition becomes useful to avoid allocating stuff pointlessly.

src/portal-impl.c
ebassi: I switch to using `sysconfdir`, as it does make much more sense.

There's also the option of using:

- `sysconfdir` for site configuration
- `prefix` / `sysconfdir` for packaging configuration

src/portal-impl.c
ebassi: Fixed.

src/portal-impl.c
ebassi: Indeed. Done.

ebassi:
> I think a warning would be appropriate.

Switched to a warning.

> It might be nice to start changing that by including a manpage for the portals.conf file format.

Done. I've added an optional dependency on `rst2man` to generate the man page. I can also commit the generated man page.

GeorgesStavracas (approved):

rpigott:
Can I ask why XDG_DATA_HOME / XDG_DATA_DIRS was chosen for this config instead of XDG_CONFIG_HOME / XDG_COFNIG_DIRS? Isn't this a user configuration?

WhyNotHugo:
With this merged in, if I'm using a `portals.conf` file to define which portals to use, I no longer need to set the `XDG_CURRENT_DESKTOP` variable for `xdg-desktop-portal`, right?

Sodivad:
XDG_CURRENT_DESKTOP affects more than just portal backend selection

ebassi:
You should not be setting `XDG_CURRENT_DESKTOP` yourself: it's up to whatever governs your session to do so.

The `portals.conf` file is a replacement for the `UseIn` key inside the portal implementation files; it is also a replacement for people disabling portal support.

If you don't have `XDG_CURRENT_DESKTOP` defined in your session, and if you don't have any session shipping a `${DESKTOP}-portals.conf` then xdg-desktop-portal will read `${XDG_CONFIG_HOME}/xdg-desktop-portal/portals.conf`.

For more information, xdg-desktop-portal ships a `portals.conf(5)` man page: https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals-conf.rst

WhyNotHugo:
What else is it used for?

WhyNotHugo:
> You should not be setting XDG_CURRENT_DESKTOP yourself: it's up to whatever governs your session to do so.

I'm not using a session governor, so I have to set this manually.

With the new `portals.conf` in my `$XDG_CONFIG_HOME`, I'm wondering if I need to set `$XDG_CURRENT_DESKTOP` for anything else or if it would no longer be necessary.

smcv:
@WhyNotHugo:
> I'm not using a session governor, so I have to set this manually.

If you are building a desktop environment out of individual pieces, then yes, it's up to you to be your own system integrator, because there is nobody else in that scenario who can take responsibility for doing this for you.

> With the new `portals.conf` in my `$XDG_CONFIG_HOME`, I'm wondering if I need to set `$XDG_CURRENT_DESKTOP` for anything else or if it would no longer be necessary.

`XDG_CURRENT_DESKTOP` is not a new invention, and is used by several other freedesktop.org specifications. The ones I know about are [`mimeapps.list`](https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html) (desktop-specific defaults for MIME and URI handlers, so that by default you'll get GNOME apps in a GNOME session and KDE apps in a Plasma session if both are installed) and the `OnlyShowIn`/`NotShowIn` keys in the [desktop entry specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/) (so apps that are inappropriate for a particular desktop environment won't show there). I'm sure there are others.

Again, if you're building a desktop environment out of individual pieces, then it's up to you to carry out the system-integrator role.

Sodivad:
xdg-utils are also using `XDG_CURRENT_DESKTOP` to choose a strategy to fulfill what you requested a tool to do. Note that for instance there are also applications that don't implement the mimeapps.list spec themselves but just do `xdg-open $file`

smcv:
As a hint for system integrators who are assembling their own desktop environments, the usual way to set `XDG_CURRENT_DESKTOP=Foo:Bar` for your desktop session is to put `DesktopNames=Foo;Bar;` into `/usr/share/xsessions/foo.desktop` or `/usr/share/wayland-sessions/foo.desktop`, which will make many display managers (lightdm, gdm, sddm, probably others) set `XDG_CURRENT_DESKTOP` for the session.

Budgie (`XDG_CURRENT_DESKTOP=Budgie:GNOME`) is a good example of the non-trivial case where the desktop environment declares that it is a variant of something better-known, so that various specifications and components will use Budgie-specific behaviour if they know one, falling back to GNOME-specific behaviour or non-desktop-specific defaults otherwise.

rpigott:
@WhyNotHugo See [1] for example, which filters desktop entries as intended by the spec.

[1] https://codeberg.org/dnkl/fuzzel/pulls/235#issue-316968

rpigott:
@rpigott 
> It's both. The original version in this PR only searched the configuration directories, and my followup #1082 added the data directories as a lower priority.

Ah, I didn't notice that the config dirs were checked. Thanks.

WhyNotHugo:
Thanks for all the detailed replies! :)

> It's both. The original version in this PR only searched the configuration directories, and my follow-up https://github.com/flatpak/xdg-desktop-portal/pull/1082 added the data directories as a lower priority.
> 
> XDG_CONFIG_HOME / XDG_CONFIG_DIRS are read as the highest priority, so that user configuration overrides OS vendor or upstream defaults; but if there is no user or sysadmin configuration, the OS vendor or upstream project for the desktop environment needs to be able to provide reasonable defaults, and those should live in /usr.

This makes perfect sense. It can be configured by the distribution, overridden by the system administrator and overridden by the end-user in a way that's pretty standard.

Should upstream compositors ship the configuration for their portals? Portals like `xdg-desktop-portal-wlr` are used by multiple compositors. Is the expectation that each compositor ships the corresponding portal file? 

From what I can tell, for general purpose portals (e.g.: darkman, a dark mode / light mode daemon that exposes its value via the settings portal, or xdp-access-portal, a standalone `org.freedesktop.impl.portal.Access` implementation) there's no correct place to ship any configuration, and it is entirely the responsibility of the user to configure it.

However, given that only the first matching configuration file is used, the user would have to copy their existing one (e.g.: from /etc or /usr) and append any customisations into that (also making sure to keep their copy in sync in future).

Wouldn't it be best to read all configuration files, and each one overrides the previous one? So that users can _append_ configurations easily?

PR #964, opened 2023-02-08
utils: Match all valid App IDs from systemd unit names

swick:
There is no restriction to reverse DNS App IDs in the systemd spec and we don't rely on it either.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -153,7 +153,8 @@ xdp_app_info_new (XdpAppInfoKind kind)
 char *
 _xdp_parse_app_id_from_unit_name (const char *unit)
 {
-  g_autoptr(GRegex) regex = NULL;
+  g_autoptr(GRegex) regex1 = NULL;
+  g_autoptr(GRegex) regex2 = NULL;
   g_autoptr(GMatchInfo) match = NULL;
   g_autoptr(GError) error = NULL;
   g_autofree char *app_id = NULL;
@@ -163,16 +164,24 @@ _xdp_parse_app_id_from_unit_name (const char *unit)
   /*
    * From https://systemd.io/DESKTOP_ENVIRONMENTS/ the format is one of:
    * app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
-   * app[-<launcher>]-<ApplicationID>-autostart.service -> no longer true since systemd v248
-   * app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
    * app[-<launcher>]-<ApplicationID>-<RANDOM>.slice
    */
-  regex = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?\\..+?\\..+?)"
-                       "(?:[@\\-][[:alnum:]]*)?(?:-autostart)?"
-                       "(?:\\.scope|\\.service|\\.slice)$",
-                       0, 0, &error);
+  regex1 = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?)(?:\\-[[:alnum:]]*)(?:\\.scope|\\.slice)$", 0, 0, &error);
   g_assert (error == NULL);
-  if (g_regex_match (regex, unit, 0, &match))
+  /*
+   * app[-<launcher>]-<ApplicationID>-autostart.service -> no longer true since systemd v248
+   * app[-<launcher>]-<ApplicationID>[@<RANDOM>].service
+   */
+  regex2 = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?)(?:@[[:alnum:]]*|\\-autostart)?\\.service$", 0, 0, &error);
+  g_assert (error == NULL);
+
+  if (!g_regex_match (regex1, unit, 0, &match))
+    g_clear_pointer (&match, g_match_info_unref);
+
+  if (match == NULL && !g_regex_match (regex2, unit, 0, &match))
+    g_clear_pointer (&match, g_match_info_unref);
+
+  if (match != NULL)
     {
       const char *escaped_app_id;
       /* Unescape the unit name which may have \x hex codes in it, e.g.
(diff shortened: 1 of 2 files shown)


smcv:
Does this introduce an ambiguity for `app-foo-bar.scope`? That could either mean launcher *foo* started app *bar* (more realistically `app-gnome-firefox.scope`), or an unspecified launcher started instance *bar* of app *foo* (more realistically `app-firefox-2PnQw.scope`.

swick:
Really good point and I was about to file an issue about it but reading the spec again, I think the `.service` and `.scope` distinction here is important:

> Application units should follow the scheme app[-<launcher>]-<ApplicationID>[@<RANDOM>].service or app[-<launcher>]-<ApplicationID>-<RANDOM>.scope

I'll adjust the code accordingly.

swick:
Should be all good now.

GeorgesStavracas:
Simon, if you're okay with this branch, feel free to land. Code LGTM from a quick glance.

src/xdp-utils.c
```diff
@@ -163,16 +164,19 @@ _xdp_parse_app_id_from_unit_name (const char *unit)
...
-  regex = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?\\..+?\\..+?)"
-                       "(?:[@\\-][[:alnum:]]*)?(?:-autostart)?"
-                       "(?:\\.scope|\\.service|\\.slice)$",
-                       0, 0, &error);
+  regex2 = g_regex_new ("^app-(?:[[:alnum:]]+\\-)?(.+?)(?:@[[:alnum:]]*|\\-autostart)?\\.service$", 0, 0, &error);
   g_assert (error == NULL);
-  if (g_regex_match (regex, unit, 0, &match))
+
+  if (g_regex_match (regex1, unit, 0, &match) ||
+      g_regex_match (regex2, unit, 0, &match))
```
smcv, line 179: This leaks:

> Note that if match_info is not NULL then it is created even if the function returns FALSE, i.e. you must free it regardless if regular expression actually matched

The cleanest way I can think of is something like this:

```suggestion
  if (!g_regex_match (regex1, unit, 0, &match))
    g_clear_pointer (&match, g_match_info_unref);

  if (match == NULL && !g_regex_match (regex2, unit, 0, &match))
    g_clear_pointer (&match, g_match_info_unref);

  if (match != NULL)
```

tests/test-xdp-utils.c
```diff
@@ -146,11 +146,6 @@ test_app_id_via_systemd_unit (void)
   g_assert_cmpstr (app_id, ==, "org.gnome.Epiphany");
   g_clear_pointer (&app_id, g_free);
 
-  app_id = _xdp_parse_app_id_from_unit_name ("app-glib-spice\\x2dvdagent-1839.scope");
-  /* App IDs must have two periods */
-  g_assert_cmpstr (app_id, ==, "");
-  g_clear_pointer (&app_id, g_free);
-
```
smcv, line 153: I'd keep this test-case, and change the expected result to

```
  g_assert_cmpstr (app_id, ==, "spice-vdagent");
```

which would make it very obvious that this commit is a behaviour change.

tests/test-xdp-utils.c
```diff
@@ -170,6 +165,14 @@ test_app_id_via_systemd_unit (void)
   app_id = _xdp_parse_app_id_from_unit_name ("app-com.obsproject.Studio-d70acc38b5154a3a8b4a60accc4b15f4.scope");
   g_assert_cmpstr (app_id, ==, "com.obsproject.Studio");
   g_clear_pointer (&app_id, g_free);
+
+  app_id = _xdp_parse_app_id_from_unit_name ("app-foo-bar.scope");
+  g_assert_cmpstr (app_id, ==, "foo");
```
smcv, line 170: Might be more obviously correct if you used `app-firefox-jcfppqx.scope`, which is functionally equivalent.

tests/test-xdp-utils.c
```diff
@@ -170,6 +165,14 @@ test_app_id_via_systemd_unit (void)
...
+  g_clear_pointer (&app_id, g_free);
+
+  app_id = _xdp_parse_app_id_from_unit_name ("app-foo-bar.service");
+  g_assert_cmpstr (app_id, ==, "bar");
```
smcv, line 174: Similarly, might be more obviously correct if you used `app-gnome-firefox.service`.

smcv (changes requested):
Having a test for `app-firefox.service` (the simplest possible unit name now allowed) would also be nice: now that we don't require dots, it's functionally equivalent to `app-org.kde.amarok.service`, but that's not immediately obvious.

src/xdp-utils.c
swick: Good catch!

swick:
Thanks Simon! I think everything is addressed.

smcv (approved):

smcv:
> reading the spec again, I think the .service and .scope distinction here is important

Yes, this is pretty subtle. It might be better if the spec specifically mentioned that `app[-<launcher>]-<ApplicationID>.scope` isn't allowed, because if it was, it would introduce this parsing ambiguity.

src/xdp-utils.c
```diff
@@ -163,16 +164,24 @@ _xdp_parse_app_id_from_unit_name (const char *unit)
   /*
    * From https://systemd.io/DESKTOP_ENVIRONMENTS/ the format is one of:
    * app[-<launcher>]-<ApplicationID>-<RANDOM>.scope
+   * app[-<launcher>]-<ApplicationID>-<RANDOM>.slice
```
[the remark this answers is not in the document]
swick: No idea tbh. The document says

> TODO: Define the name of slices that should be used. This could be app-\<launcher\>-
<ApplicationID\>-\<RANDOM\>.slice.

So I very much hope nothing uses it but I also don't think matching on it makes anything worse.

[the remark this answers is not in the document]
smcv: Yeah, certainly not a blocker. If an app process does somehow exist in `app[-<launcher>]-<ApplicationID>-<RANDOM>.slice` then its app ID is clearly meant to be `<ApplicationID>`.

PR #976, opened 2023-03-01
docs: Update supported versions table

ebassi:
Now that 1.16 is out, we can drop 1.12 from the list of supported versions.


GeorgesStavracas:
Should we wait for branching before landing this too?

ebassi:
Since 1.16 is already out, I think it's fair to update the file now; but the main development branch isn't out yet, so maybe we should wait until branching. It's up to you, really.

GeorgesStavracas:
:+1:

PR #977, opened 2023-03-01
build: Drop Autotools

ebassi:
None of the supported versions and platforms requires Autotools, so we can drop it from the build and from the CI pipeline.


ebassi:
This should be merged after branching the main development branch, so we can keep Autotools for 1.16.

GeorgesStavracas:
I removed the requirement for Ubuntu 16.04 so we can eventually land this. After landing, we can bring back the required CI flows.

GeorgesStavracas:
(To be clearer, I have not reviewed this yet)

smcv:
> I removed the requirement for Ubuntu 16.04 so we can eventually land this

Ubuntu 16.04 is EOL for everything except subscription-based updates, so I think it's reasonable to say that Ubuntu 16.04 users should have no expectation of being able to run the latest x-d-p.

GeorgesStavracas:
I'm pretty sure we can drop a bunch of GLib compatibility code that we copied in-tree too

ebassi:
Yeah, I wanted to drop the compatibility code in a separate MR.

GeorgesStavracas:
Sounds good!

GeorgesStavracas (approved):

PR #978, opened 2023-03-02
email: validate addresses following the HTML specs

### LINKED ISSUE #944: email portal: email address validation is too strict
The current email address validation regex [`^\w+([-+.]\w+)*@\w+([-.]\w+)*\.\w+([-.]\w+)*$`](https://github.com/flatpak/xdg-desktop-portal/blob/d7cfc16a6dc11cb925116fe36c5ee23e7075849c/src/email.c#L121) is too strict.
It refuses all slightly esoteric (but valid) email addresses such as `some--address@example.com`.
It also refuses the mailbox format `Some Name <some.name@example.com>` which is required by [RFC2368](https://www.rfc-editor.org/rfc/rfc2368), the format referenced in the [API reference](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-method-org-freedesktop-portal-Email.ComposeEmail).

In my opinion validating addresses/mailboxes is best left to the email composer itself.
Would you accept a patch that removes the validation altogether?

Or if you insist on keeping the validation:
 * Should we follow [RFC2368](https://www.rfc-editor.org/rfc/rfc2368) and accept mailboxes, comments etc?
 * Should we instead follow [RFC6068](https://www.rfc-editor.org/rfc/rfc6068) which only allows plain email addresses?
 * Should we try and accept both?

nicolas-guichard:
If we validate email addresses, at least do it by following some existing specification and document it.

It remains on app developers to:
 * make sure addresses are ASCII-only, by converting international labels to Punycode
 * remove anything but the addresses, `Some Name <name@example.com>` is not accepted

Fixes #944 (closed in a hurry by @matthiasclasen but @smcv agreed there was a bug to solve)

### DIFF
--- a/data/org.freedesktop.portal.Email.xml
+++ b/data/org.freedesktop.portal.Email.xml
@@ -64,13 +64,15 @@
           <varlistentry>
             <term>address s</term>
             <listitem><para>
-              The email address to send to.
+              The email address to send to. Must conform to the HTML5 definition of
+              <ulink url="https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address">valid email address</ulink>.
             </para></listitem>
           </varlistentry>
           <varlistentry>
             <term>addresses as</term>
             <listitem><para>
-              Email addresses to send to. This will be used in addition to address.
+              Email addresses to send to. This will be used in addition to address and must
+              pass the same validation.
             </para><para>
               This option was added in version 3.
             </para></listitem>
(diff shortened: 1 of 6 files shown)


smcv:
The regular expression is terrifying and I haven't reviewed it in detail, but I think this is a good approach.

ebassi (approved):
Looks good to me, too.

PR #984, opened 2023-03-13
A few fixes for the pytest test suite

whot:
After successfully paging everything out about this test suite, I struggled to run it - because the defaults only work on autotools but not on meson. This patchset fixes (and documents) this a bit better. Once #977 is merged we can probably drop a few of these.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -14,9 +14,9 @@
 # the dbusmock documentation for details on those templates.
 #
 # Environment variables:
-#   XDP_UNINSTALLED: run from $PWD (with an emulation of the glib test behavior)
-#   LIBEXECDIR: run xdg-desktop-portal from that dir, required if
-#               XDP_UNINSTALLED is unset
+#   G_TEST_BUILDDIR: override the path to the tests/ build
+#                    directory (default: $PWD)
+#   LIBEXECDIR: run xdg-desktop-portal from that dir
 #   XDP_DBUS_MONITOR: if set, starts dbus_monitor on the custom bus, useful
 #                     for debugging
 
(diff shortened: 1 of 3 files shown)


tests/__init__.py
```diff
@@ -375,11 +375,13 @@ def start_xdp(self):
...
-            xdp_path = builddir / "xdg-desktop-portal"
-            if not xdp_path.exists():
-                xdp_path = builddir / "src" / "xdg-desktop-portal"
+            xdp_path = os.getenv("XDG_DESKTOP_PORTAL_PATH", None)
+            if not xdp_path:
+                builddir = Path(os.getenv("G_TEST_BUILDDIR") or "tests") / ".."
+                # meson has x-d-p in a different build directory than the autotools build...
+                xdp_path = builddir / "xdg-desktop-portal"
+                if not xdp_path.exists():
+                    xdp_path = builddir / "src" / "xdg-desktop-portal"
```
jadahl, line 384: Any point in keeping the fall-back?

whot: the primary reason is that I don't want to clean this up until autotools is properly removed. Then we can look at what else relies on G_TEST_BUILDDIR and somesuch and remove all vestiges. Otherwise it's twice the work for little benefit.

ebassi: The Autotools build has been dropped.

whot:
alright, significant rework now that autotools is gone, this is now optimised to be able to just run `pytest` in the meson builddir and have it do the right thing. Which makes debugging significantly easier. Has to be run in the builddir because otherwise it's too hard to find the correct paths.

tests/templates/meson.build
```diff
@@ -0,0 +1,13 @@
+template_files = [
+    '__init__.py',
+    'email.py',
+    'remotedesktop.py',
+]
+foreach tf : template_files
+  configure_file(
+      input: tf,
+      output: tf,
```
GeorgesStavracas, line 9: Nitpick: `tf` → `template_file`

tests/meson.build
```diff
@@ -219,15 +219,26 @@ python = pymod.find_installation(
...
+  pytest_files = [
+      '__init__.py',
+      'test_email.py',
+      'test_remotedesktop.py',
+      'test_trash.py',
+  ]
+  foreach pf : pytest_files
+      configure_file(
+          input: pf,
+          output: pf,
```
GeorgesStavracas, line 232: Nitpick: `pf` → `pytest_file`

tests/meson.build
```diff
@@ -219,15 +219,26 @@ python = pymod.find_installation(
...
 
 if pytest.found() and python.found()
-  test_env = environment()
-  test_env.set('XDP_UNINSTALLED', '1')
+  subdir('templates')
+  pytest_files = [
+      '__init__.py',
+      'test_email.py',
+      'test_remotedesktop.py',
+      'test_trash.py',
```
GeorgesStavracas, line 227: Please use 2 spaces of indentation

tests/meson.build
```diff
@@ -219,15 +219,26 @@ python = pymod.find_installation(
...
+  ]
+  foreach pf : pytest_files
+      configure_file(
+          input: pf,
+          output: pf,
+          copy: true,
+          install: false
+      )
+  endforeach
```
GeorgesStavracas, line 237: Please use 2 spaces of indentation

tests/templates/meson.build
```diff
@@ -0,0 +1,13 @@
...
+      copy: true,
+      install: false
+  )
+endforeach
```
GeorgesStavracas, line 13: Please use 2 spaces of indentation

GeorgesStavracas (changes requested):

tests/templates/meson.build
whot: All fixed

whot:
Indentations fixed (in this patch). I had a wrong local `.editorconfig` file so there are 4-space meson.build indentations in other files, I'll fix those separately.

GeorgesStavracas (approved):
Thank you!

PR #985, opened 2023-03-13
portal-impl: Only return found implementation if it launched

jadahl:
If no portal backend for a given interface is found, a fallback is always tried anyway, despite that fallback not being listed as compatible with the current desktop environment.

Sometimes it's good that a fallback is returned; e.g. the xdg-desktop-portal-gtk file chooser backend is technically usable anywhere, however, some backends might be specifically designed to only work in a specific desktop environment, e.g. xdg-desktop-portal-gnome.

In order to avoid creating portals with non-functional backends, make sure it's possible to create a proxy object for the interface and D-Bus name, and that it launched successfully (i.e. has no name owner after creating the proxy).

----

Opening this against xdg-desktop-portals-1.16 because hopefully https://github.com/flatpak/xdg-desktop-portal/pull/955 will make this obsolete.

Not sure how this will work on non-systemd systems; it relies on using `g_dbus_proxy_get_name_owner()` being set after D-Bus activation did its thing, which is true for the portals I have managed to test (xdg-desktop-portal-{gnome,gtk,kde})

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -29,9 +29,12 @@
 #include <glib.h>
 #include <gio/gio.h>
 
+#include "xdp-utils.h"
+
 static void
 portal_implementation_free (PortalImplementation *impl)
 {
+  g_clear_pointer (&impl->dummy_proxies, g_hash_table_unref);
   g_free (impl->source);
   g_free (impl->dbus_name);
   g_strfreev (impl->interfaces);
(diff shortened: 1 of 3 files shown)


jadahl:
Gather some logs on a system using dbus-broker (F38 beta):

```
method call time=1679309435.630877 sender=:1.2 -> destination=org.freedesktop.systemd1 serial=121 path=/org/freedesktop/system
d1; interface=org.freedesktop.systemd1.Manager; member=StartUnit
   string "xdg-desktop-portal-gnome.service"
   string "replace"

method return time=1679309435.633568 sender=:1.1 -> destination=:1.2 serial=1691 reply_serial=121
   object path "/org/freedesktop/systemd1/job/554"

signal time=1679309435.831621 sender=org.freedesktop.DBus -> destination=(null destination) serial=4294967295 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameOwnerChanged
   string "org.freedesktop.impl.portal.desktop.gnome"
   string ""
   string ":1.89"
```

In other words, it looks like `NameOwnerChanged` happens after the method is returned. For GDBus that's not a problem; it calls `GetNameOwner` after `StartServiceByName` returned and gets the name that way.

jadahl:
@smcv any concerns left?

GeorgesStavracas (approved):
Let's go with this for 1.16. For 1.18, the plan is to land Emmanuele's rework anyway.

PR #990, opened 2023-03-23
Remove transient permissions after 24h

GeorgesStavracas:
... and as long as the app is running.

Draft because Chromium seems still be able to restore this token, despite it bring dropped (!) - you can verify that by changing the timeout to, say, 5 seconds.

@grulja do you know why is that?


GeorgesStavracas:
Soft depends on https://github.com/flatpak/xdg-desktop-portal/pull/977 (I didn't even bother changing Autotools)

Issue #956, opened 2023-01-19
Settings portal: WindowDecorations key


orowith2os:
Currently the state of window decorations on Linux is a bit weird, with X11 having SSDs, Wayland not requiring SSDs and applications ignoring the lack of xdg-decoration support, or applications that draw client side decorations (like libadwaita apps, or Chrome/ium, Firefox) not following some system themes.

Having a WindowDecorations key can be useful to specify if the user wants the close/max/min buttons to show, and whether or not they should show on the left or right sides of the application headerbar. 
I don't believe this would be good as a Wayland or X11 protocol, but rather something independent of the display server.

On top of the main three buttons, maybe a Titlebar option too? So that compositors that don't implement xdg-decoration (but don't want CSDs) can disable it there.

Additionally, maybe this could include a filesystem link to some SVG files to be used for the icons (/run/user/UID/theme/close.svg)? This part might be out of scope, and applications might not agree on this part.

This could be done alongside org.freedesktop.appearance.color-scheme, as org.freedesktop.appearance.windowdecorations. 

color-scheme: https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.portal.Settings.xml#L32

grulja:
This is already sort of provided by xdg-desktop-portal-[gnome/gtk], where one can read `button-layout` property from `org.gnome.desktop.wm.preferences`. I use it with my CSD implementation in QGnomePlatform. In KDE you don't need this because it supports SSD. And the only remaining portal backend implementation is `xdg-desktop-portal-wlr` doesn't seem to implement the settings portal.

The "titlebar" option you mention, I think this is already something specified and communicated by the Wayland decoration protocol.

Even though I think this might be a good thing to have, I'm unsure if it would be used somewhere else besides GNOME where this is already available, but I might be wrong.

orowith2os:
The idea of having it in the core portal is so that its independent from the desktop. And even when SSDs are supported, apps can still draw their own decorations; GTK apps and Steam do this. 
Its especially useful for if/when Linux moves closer to the mobile space, as mobile shells might not want window controls or a title bar if their shell doesn't implement xdg-decoration.

orowith2os:
@grulja made #996, check that if you'd like to see the current take on it. It's heavily based off of the `button-layout` property from GNOME.

TingPing:
[...]

PR #996, opened 2023-03-29
Button placement key 

orowith2os:
Closes https://github.com/flatpak/xdg-desktop-portal/issues/956

This only handles the layout of window decorations, the window decorations visuals themselves would probably best be handled separately, if at all.

Client implementations:

- [ ] GTK
- [ ] QT
- [ ] Electron + ArmCord
- [ ] Firefox

Portal implementations: 

- [ ] GNOME: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/95
- [ ] KDE: https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/issues/14
- [ ] Cosmic, maybe?
- [ ] Pantheon: https://github.com/elementary/portals/issues/87

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -43,6 +43,29 @@
           Unknown values should be treated as 0 (no preference).
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>org.freedesktop.appearance button-placement (asas)</term>
+        <listitem><para>
+          Indicates the system's preferred arrangement of buttons on the titlebar.
+          In LTR locales, the first tuple is on the left (leading), and the second tuple is on the right (trailing).
+          In RTL locales, the first tuple is on the right (leading), and the second tuple is on the left (trailing).
+          Supported values are:
+          <simplelist>
+            <member>close: the close button</member>
+            <member>minimize: the minimize button</member>
+            <member>maximize: the maximize button</member>
+          </simplelist>
+          Unknown values must be ignored if the client does
+          not understand them. Duplicate buttons across the tuples and
+          string arrays are not allowed.
+          Examples:
+          <simplelist>
+            <member>`(["close","minimize","maximize"][])` would result in the close, minimize, and maximize buttons being drawn on the leading side of the window, in that order; there will be no buttons on the trailing side of the window.</member>
+            <member>`([]["close"])` would result in the close button being drawn on the trailing side of the window; there will be no buttons on the leading side of the window.</member>
+            <member>`(["close"]["maximize"])` would result in the close button being drawn on the leading side of the window, and the maximize button being drawn on the trailing side of the window.</member>
+          </simplelist>
+        </para></listitem>
+      </varlistentry>
     </variablelist>
 
     Implementations can provide other keys; they are entirely
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -43,6 +43,29 @@
           Unknown values should be treated as 0 (no preference).
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>org.freedesktop.appearance button-placement (asas)</term>
+        <listitem><para>
+          Indicates the system's preferred arrangement of buttons on the titlebar.
+          In LTR locales, the first tuple is on the left (leading), and the second tuple is on the right (trailing).
+          In RTL locales, the first tuple is on the right (leading), and the second tuple is on the left (trailing).
+          Supported values are:
+          <simplelist>
+            <member>close: the close button</member>
+            <member>minimize: the minimize button</member>
+            <member>maximize: the maximize button</member>
+          </simplelist>
+          Unknown values must be ignored if the client does
+          not understand them. Duplicate buttons across the tuples and
+          string arrays are not allowed.
+          Examples:
+          <simplelist>
+            <member>`(["close","minimize","maximize"][])` would result in the close, minimize, and maximize buttons being drawn on the leading side of the window, in that order; there will be no buttons on the trailing side of the window.</member>
+            <member>`([]["close"])` would result in the close button being drawn on the trailing side of the window; there will be no buttons on the leading side of the window.</member>
+            <member>`(["close"]["maximize"])` would result in the close button being drawn on the leading side of the window, and the maximize button being drawn on the trailing side of the window.</member>
+          </simplelist>
+        </para></listitem>
+      </varlistentry>
     </variablelist>
 
     Implementations can provide other keys; they are entirely


orowith2os:
I'm not entirely sure why workflows are failing, help would be appreciated there.

data/org.freedesktop.portal.Settings.xml
```diff
@@ -43,6 +43,24 @@
...
+            <member>minimize: the minimize button</member>
+            <member>maximize: the maximize button</member>
+            <member>close: the close button</member>
+          </simplelist>
+          Unknown values should be ignored. If a value is duplicated, it should be drawn again in the indicated area.
+          Read from left-to-right; a colon separates the left corner from the right corner, and button names are comma-separated.
+          If menu is not supplied, assume all buttons are to be drawn on the right.
+        </para></listitem>
+      </varlistentry>
+    </variablelist>
```
lleyton, line 63:
```suggestion
```

The closing tag is already present after

orowith2os: Whoops. Thanks.

data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -43,6 +43,23 @@
...
+        <listitem><para>
+          Arrangement of buttons on the titlebar.
+          Supported values are:
+          <simplelist>
+            <member>menu: application content, or titlebar</member>
+            <member>minimize: the minimize button</member>
+            <member>maximize: the maximize button</member>
+            <member>close: the close button</member>
+          </simplelist>
+          Unknown values should be ignored. If a value is duplicated, it should be drawn again in the indicated area.
```
lleyton, line 58: I think it would make sense to have this before the list of supported buttons. Right now, it looks kind of like it's describing an enum, rather than a string containing the button layout.

orowith2os:
I made https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/200 to add the appropriate mode to xdg-decoration.

data/org.freedesktop.impl.portal.Settings.xml
orowith2os: Wouldn't it also make sense to change the placement of the description on color-scheme too, then? Trying to understand how you'd ideally want it to be laid out.

lleyton: Well, color-scheme is describing an enum value so it's a bit different imo :p

data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -43,6 +43,23 @@
...
+          Read from left-to-right; a colon separates the left corner from the right corner, and button names are comma-separated.
+          If menu is not supplied, assume all buttons are to be drawn on the right.
```
orowith2os, line 60:
```suggestion
          Read from left-to-right; a colon separates the left corner from the right corner, and button names are comma-separated.
          Supported values are:
          <simplelist>
            <member>menu: application content, or titlebar</member>
            <member>minimize: the minimize button</member>
            <member>maximize: the maximize button</member>
            <member>close: the close button</member>
          </simplelist>
          Unknown values should be ignored. If a value other than menu is duplicated, it should be drawn again in the indicated area.
          Examples:
          <simplelist>
            <member>`close,maximize,minimize:menu` would result in the close, maximize, and minimize buttons being drawn in the left corner.</member>
            <member>`menu:minimize,maximize,close` would result in the close, maximize, and minimize buttons being drawn in the right corner.</member>
            <member>`close,maximize,minimize:menu:minimize,maximize,close` would result in the close, maximize, and minimize buttons being drawn in both the left and right corners.</member>
          </simplelist>
```

@lleyton how's this?

lleyton: lots better :)

orowith2os:
Update: still waiting on more input on how I should define this. The Wayland merge request is also useless here, was just a misunderstanding; server-side means apps shouldn't draw decorations at all, client-side means apps should draw them themselves. A none mode wasn't needed.

orowith2os:
Going to update and rework this just a bit, since I've learned quite a bit more about how the GTK preference works and cross-toolkit-ness.

jadahl:
This seems fine to add to me, it could be useful for hide e.g. minimize where it doesn't matter etc. What I'd like to point out is that it won't solve potential "annoyances" that people have with CSD apps hiding things so they can get another "frame" placed around them. Arguably the value for this settings entry in those DE:s should still be so that CSD-only applications can, if they want, add the appropriate window controls, so that if the DE uses the "minimize" concept a lot, the app knows it should add a minimize button. Maybe this was something you already had in mind already.

swick:
The description of the setting is probably too strong. It tells clients exactly which buttons should be drawn in which order.  It would probably be make more sense for it to be a hint: these actions make sense in the current environment. And then let the client figure out where to draw what buttons.

That also sidesteps issues with RTL layouts.

orowith2os:
> It would probably be make more sense for it to be a hint: these actions make sense in the current environment. And then let the client figure out where to draw what buttons.

I was thinking about this when reimplementing the button-layout logic from Tweaks, and yeah, a list of ideal actions would be better. I do think telling them *where they should be drawn* should be done though, so that on Pantheon you can get a close button on the left and on GNOME you get a close button on the right - while having no decorations on Sway and all decorations on the right on KDE.

orowith2os:
Updated, and accounted for some potential configurations such that they're invalid, when they would be valid before.

Now it should work for setups such as: Sway, KDE, Pantheon, and GNOME.

orowith2os:
Docs are finally building! Glad to see it making some more progress forward from where it was last time.

data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -43,6 +43,29 @@
           Unknown values should be treated as 0 (no preference).
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>org.freedesktop.appearance button-placement (asas)</term>
+        <listitem><para>
+          Indicates the system's preferred arrangement of buttons on the titlebar.
+          The first string array in the tuple is the left side of the window,
+          and the second string array is the right side of the window.
```
ebassi, line 51: "Left" and "right" are going to be "right" and "left" in right-to-left language locales.

We use "leading"/"start" and "trailing"/"end" for these cases; otherwise you'll have to expand that to something like this:

```suggestion
          The first string array in the tuple is the left side of the window,
          and the second string array is the right side of the window. In RTL
          locales, the first string array defines buttons in the right side of
          the window, and the second array defines buttons in the left side
          of the window.
```

ebassi (changes requested):

data/org.freedesktop.impl.portal.Settings.xml
orowith2os: I don't think I like this, then you'd get inconsistent behavior across languages. Is there no better solution? How does this even end up getting reversed in RTL locales? Can it not be described without a visual representation? 

The first string array should always be on the left, regardless of locale, and vice versa. It shouldn't depend on the language.

ebassi: That's not how window decorations work: RTL locales flip the order of the buttons on **every** platform.

lleyton (approved):
Generally looks okay, at least from a quick look over. Personally, I'd use an int enum to represent the button types, but that's just bikeshedding from my part :p

orowith2os:
@ebassi would you mind giving this an updated review, please? I'd like to see if there's anything else to nab before I start implementing it.

orowith2os:
Bugged @danirabbit on the Elementary Slack about it, so I think Elementary can be considered mostly neutral or in favor of this - they'd just hardcode `(["close"],["maximize"])` on their side of things. Here's a portal issue for them: https://github.com/elementary/portals/issues/87

Someone needs to handle it for KDE too: https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/issues/14

GNOME should be more or less fine, but someone needs to first expose a `([],["close"])` on the GNOME portal, then can possibly look at converting the gsettings key to that.

ilya-fedin:
btw, only GTK from client implementations mentioned in the PR seems realistic as both Firefox and Electron would rather use [GtkSettings](https://docs.gtk.org/gtk3/property.Settings.gtk-decoration-layout.html) I believe and Qt has no concept of titlebar settings. Qt has only a dummy decoration for the cases when xdg-decoration is not present but then it's better to spend time on adding libdecor support I believe what ends with GTK client implementation again.

ebassi (approved):

jadahl:
@orowith2os don't merge main into this branch, this repository uses a linear history.

orowith2os:
Fair enough, I can force push such that there's only one commit if/when this is ready to be merged :)

jadahl:
> Fair enough, I can force push such that there's only one commit if/when this is ready to be merged :)

Just do so now, no reason to wait to clean up the branch.

jadahl:
Next step before this can land would be to create proof of concepts and "acks" in "enough" toolkits.

orowith2os:
I'm sure @lleyton and @ItsJamie9494 should be able to handle the tauOS and libhelium side of things, and the work can easily be strapped onto the GNOME portal + GTK. I plan on making a Rust representation and reference for the GNOME portal soon, and tauOS can reuse it in their stuff.

ItsJamie9494:
I can work with GTK and GNOME, but I have nothing to do with libhelium and/or tauOS anymore

jadahl:
What about Qt (and/or https://github.com/leidav/adwaita-qt-decoration)? If it's GTK land only, then it's not gaining much compared to just forwarding the GTK gsetting, does it not?

jadahl:
Sorry, sent the wrong link to QtAdwaitaDecorations.

orowith2os:
I don't have anybody at hand to manage QT, but AFAIK that's all done with the platform plugins. You'd need to go through and update each one individually. 

If I/we can sort something out for a source code representation of this setting, I can release it as a library and reuse that, possibly. Or use it as a base for other implementations.

grulja:
In QAdwaitaDecorations I already use what's provided by `org.gnome.desktop.wm.preferences` (in particular `button-layout` key). There is not much I can gain changing that for something else since this project targets GNOME and this is expected to be provided all the time. For Qt in general, I can imagine this being part of the [decoration plugin](https://github.com/qt/qtwayland/blob/dev/src/client/qwaylandabstractdecoration_p.h).

orowith2os:
QAdwaitaDecorations should probably still look to use this portal at some point, in the (possibly unlikely) event it's used someplace that's not GNOME, and the GNOME preferences aren't available to be read.

data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -43,6 +43,29 @@
...
+          In RTL locales, the first tuple is on the right (leading), and the second tuple is on the left (trailing).
+          Supported values are:
+          <simplelist>
+            <member>close: the close button</member>
+            <member>minimize: the minimize button</member>
+            <member>maximize: the maximize button</member>
+          </simplelist>
+          Unknown values must be ignored if the client does
+          not understand them. Duplicate buttons across the tuples and
+          string arrays are not allowed.
```
orowith2os, line 60: @ebassi @jadahl do you both have any thoughts on other paths this could take? anything nicer to expose over dbus? anything that's easier to put into code? it is Not Fun™ as-is.

swick:
@grulja `org.gnome.desktop.wm.preferences` isn't something that's available inside the flatpak sandbox so this should help with providing the same data in a cross-platform way inside the sandbox.

grulja:
> @grulja `org.gnome.desktop.wm.preferences` isn't something that's available inside the flatpak sandbox so this should help with providing the same data in a cross-platform way inside the sandbox.

Sure it is, you can get it through the settings portal already and it doesn't make a difference whether you make the call inside or outside the sandbox. I'm using [it here ](https://github.com/qt/qtwayland/blob/dev/src/plugins/decorations/adwaita/qwaylandadwaitadecoration.cpp#L317)for example.

ilya-fedin:
> And then goes on to list `org.freedesktop.appearance color-scheme accent-color contrast`.

None of those settings provided via xsettings btw. org.gnome.desktop.wm.preferences.button-layout is provided (Gtk/DecorationLayout).

swick:
As @tytan652 pointed out to me, the docs also say

> Implementations can provide other keys; they are entirely implementation details that are undocumented. If you are a toolkit and want to use this please open an issue.

but I completely missed it.

orowith2os:
Reactivating this MR - are we good to add this, or does it still need more discussion? I would like to see environments such as tiling WMs able to remove window decorations from applications that draw CSDs to try and keep everything consistent.

ilya-fedin:
It seems they don't want to add this due to GTK being the only toolkit that has csd (i.e. the only client-side user). Personally I, too, miss this as a standardized way to provide the setting even if only gtk and multiple apps having csd on app level (rather than toolkit level) would read it, so that everyone is not required to read gtk code to know the key they need to provide via portal and to work without having xdg-desktop-portal-gtk (which is still doesn't seem to be present in most compositors' portals.conf, the user has to modify it), i.e. better discoverability for compositor developers and users wiling to configure their system in general.

jadahl:
> I would like to see environments such as tiling WMs able to remove window decorations from applications that draw CSDs to try and keep everything consistent.

This is largely covered by https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/258d8f88f2c8c25a830c6316f87d23ce1a0f12d9/stable/xdg-shell/xdg-shell.xml?page=2#L1176-1207.

> It seems they don't want to add this due to GTK being the only toolkit that has csd (i.e. the only client-side user).

Both of the largest toolkits, Qt and GTK, have client side decoration support. Where was it stated that this idea was "not wanted"?

ilya-fedin:
> This is largely covered by https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/258d8f88f2c8c25a830c6316f87d23ce1a0f12d9/stable/xdg-shell/xdg-shell.xml?page=2#L1176-1207

btw is there any toolkit supporting those? I haven't seen it being supported by neither gtk nor qt.

> Both of the largest toolkits, Qt and GTK, have client side decoration support.

Qt just draws a dummy one with "minimize, maximize, cross" on the right hardcoded. And, recently, since 6.8 it also has an adwaita-themed one which just reads GNOME's key. In other words, there seem to be no place in Qt to implement this spec.

> Where was it stated that this idea was "not wanted"?

From #956:

> Adding a setting here isn't for an "if/when" use-case. We should have real world examples of non-GNOME toolkits/platforms wanting/able to use this. Otherwise the, kinda private to GNOME, setting already works for GNOME.

And, well, there's still no real world need in it outside of GNOME as far as I'm aware, per checkboxes here at least:

> Client implementations:
> 
> GTK

A real one
> QT

libdecor-like plugin-based abstractions, needs it only for GNOME and reads its setting just fine, no app-accessible CSD APIs -> could be merged with the GTK item

> Electron + ArmCord

Just uses libdecor, no app-accessible CSD APIs -> could be replaced with libdecor or merged with the GTK item

> Firefox

Just uses GtkSettings -> could be merged with the GTK item

jadahl:
I'm somewhat reading your arguments as "it's tied to GNOME/GTK in the implementations", while the motivation is try to untie it from a global preference level.

ilya-fedin:
I rather mean that decoration drawing is delegated to the libraries in electron/firefox case and you would either have to rewrite decoration drawing in a major way (imagine replacing libdecor with custom solution in electron case) or just change the libraries (which just comes to libdecor/gtk).

In Qt case, it's delegated to plugins which are meant to be provided by desktop environments. I.e. it's meant to have plugins for exactly each DE which reads exactly its settings.

orowith2os:
> This is largely covered by https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/258d8f88f2c8c25a830c6316f87d23ce1a0f12d9/stable/xdg-shell/xdg-shell.xml?page=2#L1176-1207.

Here's the problem (well, a few) with that:
- compositors might support those capabilities, but won't want them exposed to the end user normally (GNOME)
- what if the user, or desktop, wants the buttons placed somewhere else, such as on the left? This is the case in Elementary.
- what should X11 clients do if they want to draw those items on the client side?

orowith2os:
I would like to unblock this a bit. Should I modify this a bit to state that this is a *hint*, not a *requirement*, of where to draw the buttons? And that if a string is in here at all, we can take that as a hint to enable or disable that button, as appropriate? 

Maybe specifying how custom implementations should add their keys, so that they don't conflict across toolkits, if it were to happen? Something like `namespace.blah.setting-name` as a string in here, for example.

tytan652:
To unblock it, I think a rewrite to rst is needed.

`hint` or `requirement`, application can decide to not follow it even if it's "required".  Application following it would be a good QOL.


> Maybe specifying how custom implementations should add their keys, so that they don't conflict across toolkits, if it were to happen? Something like `namespace.blah.setting-name` as a string in here, for example.

This PR is meant to set a common spec to follow, we can ask more toolkit folks on Matrix.

If toolkits ends up adding a custom key afterward the introduction of this settings rather than discussing to expand/improve the portal, I think we'll see when it happen.

-------------

Honestly about Qt, just ignore it with what they done until now. They clearly do not want to give freedom to the toolkit consumer unless you use QML which is a big issue.
They can rely on this settings to make bradient (default CSD) work better, rather than forcing the app to "look  GNOME".

swick:
If anyone is interested in this feature, please go ahead and open a new PR and:

* Update the dbus interface description
* Provide an implementation in one backend
* Provide an implementation to one toolkit (GTK is on board with this portal)

PR #1001, opened 2023-04-13
Drop pre-2.66 GLib compatibility layer

ebassi:
GLib 2.66 was released in September 2020.

Note: this pull request depends on #977.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -97,7 +97,7 @@ foreach h : check_headers
   config_h.set(h.get(1), cc.has_header(h.get(0)))
 endforeach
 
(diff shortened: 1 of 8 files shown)


GeorgesStavracas (approved):

smcv:
FYI this means there can no longer be straightforward backports of 1.17/1.18 to Ubuntu 20.04, Debian 10, RHEL/CentOS 8 and other similarly old LTS distributions. I think that's OK, we have to draw a line somewhere (and the current stable releases Ubuntu 22.04, Debian 11, and RHEL/CentOS 9 are new enough).

Issue #850, opened 2022-08-05
Extend persist/restore_token concept to RemoteDesktop


bjaraujo:
File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.portal.Error.InvalidArgument: Remote desktop sessions cannot persist

It would be important to remotely access unattended machines.

marler8997:
This would be a welcome addition to a "pair programming" application I work on.  It aims to make pair programming remotely as seamless as possible.  Having to give permission to our app each time the "host user" (the one sharing their screen) changes makes the transition less seamless.  Using a restore token means users could switch who is sharing their screen multiple times without having to ask the user for the same permissions each time.

jadahl:
I'm currently looking into how it can be done. The idea I have now is to use the same restore token like API that ScreenCast has, but make it part of SelectDevices. When used with SelectDevices, it'd also restore the screen cast source, in addition to devices, and and store the permission/source data in a `remotedesktop` permission store table, in contrast to the `screencast` table.

jadahl:
See https://github.com/flatpak/xdg-desktop-portal/pull/1004.

PR #1004, opened 2023-04-14
Persistent remote desktop sessions

jadahl:
This adds persistent remote desktop sessions similar to persistent
screen cast sessions. What it means is that applications can ask for
sessions to be "saved" either temporarily in the portal, or in the
permission store, and later be restored at will, without any interactive
dialog showing up.
    
It works the same way, by the Start() method returning a restore token,
and similarly to the screen cast variant, passing the same restore token
to the SelectDevices() method the next time.
    
Closes: https://github.com/flatpak/xdg-desktop-portal/issues/850

-----

Also contains a bunch of cleanups to the screen cast and remote desktop portal specification.

### DIFF
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -491,10 +491,10 @@ replace_remote_desktop_restore_token_with_data (Session *session,
     persist_mode = PERSIST_MODE_NONE;
 
   remote_desktop_session->persist_mode = persist_mode;
-  replace_restore_token_with_data (session,
-                                   REMOTE_DESKTOP_TABLE,
-                                   in_out_options,
-                                   &remote_desktop_session->restore_token);
+  xdp_session_persistence_replace_restore_token_with_data (session,
+                                                           REMOTE_DESKTOP_TABLE,
+                                                           in_out_options,
+                                                           &remote_desktop_session->restore_token);
 
   return TRUE;
 }
@@ -598,12 +598,12 @@ static void
 replace_restore_remote_desktop_data_with_token (RemoteDesktopSession *remote_desktop_session,
                                                 GVariant **in_out_results)
 {
-  replace_restore_data_with_token ((Session *) remote_desktop_session,
-                                   REMOTE_DESKTOP_TABLE,
-                                   in_out_results,
-                                   &remote_desktop_session->persist_mode,
-                                   &remote_desktop_session->restore_token,
-                                   &remote_desktop_session->restore_data);
+  xdp_session_persistence_replace_restore_data_with_token ((Session *) remote_desktop_session,
+                                                           REMOTE_DESKTOP_TABLE,
+                                                           in_out_results,
+                                                           &remote_desktop_session->persist_mode,
+                                                           &remote_desktop_session->restore_token,
+                                                           &remote_desktop_session->restore_data);
 }
 
 static gboolean
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -21,6 +21,7 @@
 #include "remote-desktop.h"
 #include "screen-cast.h"
 #include "request.h"
+#include "restore-token.h"
 #include "pipewire.h"
 #include "call.h"
 #include "session.h"
@@ -30,6 +31,8 @@
 
 #include <gio/gunixfdlist.h>
 #include <stdint.h>
+
+#define REMOTE_DESKTOP_TABLE "remote-desktop"
 
 typedef struct _RemoteDesktop RemoteDesktop;
 typedef struct _RemoteDesktopClass RemoteDesktopClass;
@@ -91,6 +94,10 @@ typedef struct _RemoteDesktopSession
   gboolean clipboard_enabled;
 
   gboolean uses_eis;
+
+  char *restore_token;
+  PersistMode persist_mode;
+  GVariant *restore_data;
 } RemoteDesktopSession;
 
 typedef struct _RemoteDesktopSessionClass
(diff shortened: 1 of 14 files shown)


jadahl:
For backend support in GNOME, see https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/88.

src/remote-desktop.c
```diff
@@ -30,6 +31,8 @@
 
 #include <stdint.h>
 
+#define REMOTE_DESKTOP_TABLE "remotedesktop"
```
GeorgesStavracas, line 34: `remote-desktop`?

src/remote-desktop.c
```diff
@@ -394,6 +465,7 @@ handle_select_devices (XdpDbusRemoteDesktop *object,
   g_autoptr(GError) error = NULL;
   g_autoptr(XdpDbusImplRequest) impl_request = NULL;
   GVariantBuilder options_builder;
+  GVariant *options;
```
GeorgesStavracas, line 468: Perhaps `g_autoptr(GVariant)` and use `g_steal_pointer()` below?

src/restore-token.h
```diff
@@ -0,0 +1,71 @@
...
+typedef enum _PersistMode
+{
+  PERSIST_MODE_NONE = 0,
+  PERSIST_MODE_TRANSIENT = 1,
+  PERSIST_MODE_PERSISTENT = 2,
+} PersistMode;
+
+void set_transient_permissions (const char *sender,
+                                const char *restore_token,
+                                GVariant *restore_data);
```
GeorgesStavracas, line 32: I've been staring at this for a while, and am starting to this it'd be better to rename these functions and he enum to something like:

```
typedef enum _XdpSessionPersistMode
{
   ...
} XdpSessionPersistMode;

void xdp_session_persistence_store_transient (...);
void xdp_session_persistence_store (...);
GVariant * xdp_session_persistence_find (...);
```

What do you think?

GeorgesStavracas (changes requested):
Left a few comments, but the general approach seems great and the code flow looks good to me. Thanks!

src/restore-token.h
jadahl: I guess you mean taking the `sender` from the session? Seems reasonable to me.

GeorgesStavracas: I was thinking only in renaming the functions, arguments would stay the same. But now that you mention, taking sender from the session might be a good idea too!

src/remote-desktop.c
jadahl: It made things a bit awkward, since we expect it to be consumed by `replace_remote_desktop_restore_token_with_data()` even on error, meaning can't `g_steal_pointer()` it in the argument, but would have to after.

Perhaps it'd be nicer to pass one `in` and one `out` `options` to that function instead.

src/restore-token.h
jadahl: Anything called `xdp_session_*()` I'd expect to take a "session as the first argument.

jadahl:
> Left a few comments, but the general approach seems great and the code flow looks good to me. Thanks!

And thanks for giving me an already established concept to just tweak a bit!

src/remote-desktop.c
GeorgesStavracas: Yeah, I think that's probably better. Not sure why I went with inout, I don't like it :stuck_out_tongue: 

This can be in separate commits btw, or even follow-up PRs, if it complicates things too much.

src/remote-desktop.c
```diff
@@ -539,7 +639,7 @@ start_done (GObject *source_object,
     {
       if (response == 0)
         {
-          if (!process_results (remote_desktop_session, results, &error))
+          if (!process_results (remote_desktop_session, &results, &error))
             {
               g_warning ("Could not start remote desktop session: %s",
                          error->message);
```
jadahl, line 645: Github doesn't let me comment on the correct line, so I'll add a reminder here instead.

`xdp_dbus_remote_desktop_set_version (XDP_DBUS_REMOTE_DESKTOP (remote_desktop), 1);`
needs to be changed to
`xdp_dbus_remote_desktop_set_version (XDP_DBUS_REMOTE_DESKTOP (remote_desktop), 2);`

whot: unrelated, but I've been using a `VERSION_V(1)` macro in [libei](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/util-version.h) and it really makes these lines a lot easier to find and replace...

whot (commented):
#762 has some extensions to the pytest test suite for testing RD, maybe it's worth stealing those so you can add the tests for restore tokens here?

bjaraujo:
Any updates on this?

jadahl:
> Any updates on this?

Was somewhat waiting for #852 to land so this could be rebased to include a clipboard part too. My plan is to get it into the coming release of xdg-desktop-portal.

src/remote-desktop.c
jadahl: That's better.

src/remote-desktop.c
jadahl: Sounds useful - could add that for all instances of version checking/setting (as a follow up, I humbly suggest)

jadahl:
> #762 has some extensions to the pytest test suite for testing RD, maybe it's worth stealing those so you can add the tests for restore tokens here?

I tried, and came so far that I was lacking the permission store mock, so stopped there for now. Will pick it up later, but would be nice if it didn't block this.

src/restore-token.h
jadahl: From Matrix:

> all of them
> the ones you listed already receive a Session* object; the others (`set_transient_permissions`, `delete_transient_permissions`, `get_transient_permissions`, etc...) can be adjusted to receive a Session* object too
> we always pass `session->sender` to the ones that receive `const char *sender`, for example

src/xdp-utils.c
```diff
@@ -972,6 +972,48 @@ xdp_connection_track_name_owners (GDBusConnection *connection,
                                       peer_died_cb, NULL);
 }
 
+gboolean
+xdp_inverse_filter_options (GVariant *options,
+                            GVariantBuilder *filtered,
+                            XdpOptionKey *unsupported_options,
+                            int n_unsupported_options,
+                            GError **error)
```
GeorgesStavracas, line 980: I think this ended up in the wrong commit?

src/remote-desktop.c
GeorgesStavracas: Let's do it as a separate PR please

src/xdp-utils.c
GeorgesStavracas: Actually no commit seems to use this

GeorgesStavracas (changes requested):

src/remote-desktop.c
GeorgesStavracas: Tracking this on https://github.com/flatpak/xdg-desktop-portal/issues/1049 now.

GeorgesStavracas (approved):
I cannot overstate how happy the last pair of commits make me feel. Clearly this is how this code should have been implemented initially. Thank you so much for doing it. Today is a good day.

src/screen-cast.c
```diff
@@ -569,117 +422,54 @@ validate_persist_mode (const char *key,
...
-                   "Remote desktop sessions cannot persist");
-      return FALSE;
+      if (persist_mode != PERSIST_MODE_NONE ||
+          xdp_variant_contains_key (options, "restore_token"))
+        {
+          g_set_error (error,
+                       XDG_DESKTOP_PORTAL_ERROR,
+                       XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                       "Remote desktop sessions cannot persist");
+          return FALSE;
```
axelkar, line 454: Seems that this PR didn't add persistence for combined remote desktop+screen cast sessions. Why? Would it be possible to do that?

axelkar: Ah, nevermind. I thought I had to add persistence to the `SelectSources` invocation as well. The error message is wrong though.

PR #1005, opened 2023-04-18
Add HACKING and CONTRIBUTING guidelines

GeorgesStavracas:
Only time will tell whether or not people will actually read this, but it exists.


smcv:
I'm not sure how much sense it makes to distinguish between guidelines for hacking and guidelines for contributing. In other projects I've tended to put them into a single file, which would be `CONTRIBUTING.md` with Github naming conventions.

smcv:
(The content of these looks good to me, I'm just not sure about the value of having two separate files, or whether a new contributor would know to look at a document named HACKING.)

GeorgesStavracas:
@smcv how about this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,56 @@
...
+- You have included updates to all appropriate documentation
+
+We use GitHub pull requests to review contributions. Please be kind and patient
+as code reviews can be long and minutious.
+
+## Development
+
+xdg-desktop-portal usually runs as a user session service, initialized before
+with the graphical session. xdg-desktop-portal initializes specific backends
+through D-Bus activation.
```
jadahl, line 19: Not really true is it? `xdg-desktop-portal` is D-Bus activated too thus starts as soon as anything requests it. In F38 it is effectively started immediately because gnome-software is autostarted and asks for `org.freedesktop.portal.Desktop`, but it doesn't have to be autostarted.

GeorgesStavracas: You're right! Thank you

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Building
+
+To build xdg-desktop-portal, first make sure you have the build dependencies
+installed through your distribution's package manager. With them installed,
+run:
+
+```
+$ meson setup . _build
+$ ninja -C _build
```
whot, line 31: fwiw, this should probably be
```
$ meson compile -C _build
```
It's available since 0.55 and we require 0.56.

PR #1015, opened 2023-05-04
Switch from ninja to meson compile in the docs

whot:
As of meson 0.55 this is the backend-agnostic compiler command so let's use it to standardize across projects. We require 0.56 anyway.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ run:
 
 ```
 $ meson setup . _build
-$ ninja -C _build
+$ meson compile -C _build
 ```
 
 ### Running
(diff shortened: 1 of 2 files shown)


GeorgesStavracas (approved):

eli-schwartz:
There's no actual advantage of `meson compile` unless you a) need to support the xcode and visual studio backends, or b) need to initialize an MSVC developer command prompt to get `cl.exe` to work...

whot:
fwiw, for me the main advantage is primarily for newcomers: having three meson commands to setup, build and install is easier than having a different compile command. Those who know what ninja is can invoke it directly but I like the consistency.

Plus given meson's development speed it'll probably have nice extra features in the future :)

PR #1016, opened 2023-05-09
Fix the pytest breakage and run the pytest test suite in the CI

whot:


.github/workflows/check.yml
```diff
@@ -90,6 +95,11 @@ jobs:
       env:
         MESON_OPTIONS: -Dinstalled-tests=true -Db_sanitize=${{ matrix.sanitizer }}
 
+    - name: Run dbus-daemon
+      run: |
+        $RUN_CMD $AS_USER sh -c "dbus-daemon --fork --session --print-address > dbus-env"
+        echo "DBUS_SESSION_BUS_ADDRESS=$($RUN_CMD $AS_USER cat dbus-env)" >> "$GITHUB_ENV"
```
smcv, line 101: The pytest test suite should really run a dbus-daemon itself, either directly from Python code or [by telling Meson to wrap it in `dbus-run-session`](https://sources.debian.org/src/dbus-python/1.3.2-4/test/meson.build/?hl=72#L72). If it did that, then the Github workflow wouldn't need to, and that would also be nicer for running the pytest test suite on a developer's system or during rpm/dpkg/etc. build.

whot: good point, thanks. though staring at this again I realised that we're using `dbusmock` anyway which starts its own session bus for the test cases anyway. Not sure yet why this doesn't work in the container, need to debug this

whot: yup, works without, that hunk is dropped now. thanks again!

jadahl:
> and run the pytest test suite in the CI

I might be blind, but which commit makes this happen?

smcv:
> I might be blind, but which commit makes this happen?

"CI: install the requirements for the pytest test suite" results in it being run as a build-time test, which is done or not done automatically, based on whether the requirements are present. Turning the pytest test into a `feature` (defaulting to `auto` presumably) and using `-Dpytest=enabled` in the Github workflows might make it more obvious that this is being done, but I don't think that should be a blocker here.

smcv (approved):

jadahl:
> "CI: install the requirements for the pytest test suite" results in it being run as a build-time test,

Thanks for clarifying.

Issue #1017, opened 2023-05-10
Settings portals should not be provided unconditionally


kennylevinsen:
With the new portal configuration, we can finally set per-desktop environment behavior, avoiding situations where e.g. xdg-desktop-portal-gnome is attempted called if it is installed, despite only working under GNOME.

The settings portal unfortunately disregards this, always providing all available backends at once (013c117a545cbc280ad0e0bc3f3184bebff80d45). This means that the GNOME portal may still be attempted activated outside of GNOME, which usually just leads to hangs until the service activation times out.

I believe the preferable solution would be to also respect the portal configuration, adjusting find_all_portal_implementations to return all implementations specified, rather than all implementations available.

matthiasclasen:
That is very much intentional, the idea is that you want settings to be merged, since there are desktop-specific settings in these portals. Why should a kde application miss its kde-specific settings, just because it runs under gnome?

jadahl:
> which usually just leads to hangs until the service activation times out

This should be fixed now.

> That is very much intentional, the idea is that you want settings to be merged, since there are desktop-specific settings in these portals.

If you want settings from `xdg-desktop-portal-gnome` but running outside of GNOME, it needs to run in a special "headless" mode where it doesn't try to be an actual fully fledged portal backend. We could I guess in theory detect this ourselves over in xdp-gnome and be settings only automatically.

kennylevinsen:
Hmm, discussion is going to be duplicated slightly between the PR and the issue.

> That is very much intentional, the idea is that you want settings to be merged, since there are desktop-specific settings in these portals. Why should a kde application miss its kde-specific settings, just because it runs under gnome?

Support for that should be maintained, but a user or DE might also want to instead pick which backends to use and what ordering they should have, should they so desire.

As an example, I might want to use Darkman (providing automatic dark-mode color-scheme changes) under sway, but let GNOME manage things on its own there without accidentally conflicting with the GNOME Settings interactions. Likewise, I may *not* want KDE or GNOME settings made available under another environment, having an alternate provider with more DE-appropriate settings instead.

I imagine such cases will only grow more common, and under the new system we can have our cake and eat it too.

WhyNotHugo:
> Why should a kde application miss its kde-specific settings, just because it runs under gnome?

[...]

PR #1018, opened 2023-05-10
Respect portal configuration for find_all_portal

kennylevinsen:
The settings portal is discovered using find_all_portal_implementations, which looked for any portal that provided the implementation, regardless of configuration or UseIn=.

The new portal configuration allows us to specify which portals should be considered for each interface. Update find_all_portal_implementations to respect the configuration.

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/1017

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -503,24 +503,21 @@ find_portal_implementation (const char *interface)
   GList *l;
   int i;
 
-  desktops = get_current_lowercase_desktops ();
-
-  for (i = 0; desktops[i] != NULL; i++)
+  for (l = implementations; l != NULL; l = l->next)
     {
-     for (l = implementations; l != NULL; l = l->next)
+      PortalImplementation *impl = l->data;
+
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
+        continue;
+
+      if (portal_impl_matches_config (impl, interface))
         {
-          PortalImplementation *impl = l->data;
-
-          if (!g_strv_contains ((const char **)impl->interfaces, interface))
-            continue;
-
-          if (portal_impl_matches_config (impl, interface))
-            {
-              g_debug ("Using %s.portal for %s in %s (config)", impl->source, interface, desktops[i]);
-              return impl;
-            }
+          g_debug ("Using %s.portal for %s (config)", impl->source, interface);
+          return impl;
         }
     }
+
+  desktops = get_current_lowercase_desktops ();
 
   /* Fallback to the old UseIn key */
   for (i = 0; desktops[i] != NULL; i++)
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -569,9 +569,12 @@ find_all_portal_implementations (const char *interface)
     {
       PortalImplementation *impl = l->data;
 
-      if (g_strv_contains ((const char **)impl->interfaces, interface))
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
+        continue;
+
+      if (portal_impl_matches_config (impl, interface))
         {
-          g_debug ("Using %s.portal for %s", impl->source, interface);
+          g_debug ("Using %s.portal for %s (config)", impl->source, interface);
           g_ptr_array_add (impls, impl);
         }
     }


matthiasclasen:
It is intentional that this this code uses all available settings backends.

See 013c117a545cbc280ad0e0bc3f3184bebff80d45

columbarius:
Doesn't the merging require an order? And do the settings implementations carry undocumented privat options for their toolkit?

src/portal-impl.c
```diff
@@ -551,19 +551,29 @@ find_portal_implementation (const char *interface)
...
   impls = g_ptr_array_new ();
 
-  for (l = implementations; l != NULL; l = l->next)
-    {
-      PortalImplementation *impl = l->data;
+  desktops = get_current_lowercase_desktops ();
 
-      if (g_strv_contains ((const char **)impl->interfaces, interface))
+  for (i = 0; desktops[i] != NULL; i++)
+    {
```
matthiasclasen, line 564: I don't see desktops[i] being used anywhere in the body of the loop. What am I missing?

kennylevinsen: Ah, I accidentally copied dead code from `find_portal_implementation` that was left behind when the UseIn= fallback was refactored to its own loop.

Fixed, and added a commit that removes the dead code at the source while at it.

kennylevinsen:
> That is not true at all. The settings portal was only using the currently configured backend before it was changed to the current implementation, to address complaints about missing settings.

I meant that it predated our new and shiny portal configuration file, not older means of configuration. UseIn= was too inflexible, so the current behavior of ignoring configuration and using all backends was indeed the best option - not pointing any fingers at that. But with the portal configuration file we have more options available to us.

kennylevinsen:
Note, this PR does not yet add ordering to find_all_portal_implementations, just simple filtering with wildcard.

jadahl:
FWIW, with https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/95, https://github.com/linuxmint/xdg-desktop-portal-xapp/pull/7 and similar ones elsewhere when needed, this shouldn't be needed.

ebassi:
I'm in favour of making all portals obey the configuration file: getting portal implementations to work in "adversarial" conditions just piles complexity into them, instead of centralising the configuration into the x-d-p proxy.

ebassi (approved):
I think this is good to go.

src/portal-impl.c
```diff
@@ -560,10 +560,13 @@ find_all_portal_implementations (const char *interface)
...
-      if (g_strv_contains ((const char **)impl->interfaces, interface))
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
+        continue;
+
+      if (portal_impl_matches_config (impl, interface))
         {
-          g_debug ("Using %s.portal for %s", impl->source, interface);
-          g_ptr_array_add (impls, impl);
+          g_debug ("Using %s.portal for %s (config)", impl->source, interface);
+          g_ptr_array_add(impls, impl);
```
GeorgesStavracas, line 565: Nitpick: space before `(`

src/portal-impl.c
```diff
@@ -494,25 +494,21 @@ find_portal_implementation (const char *interface)
...
-              return impl;
-            }
+      if (portal_impl_matches_config (impl, interface))
+        {
+          g_debug ("Using %s.portal for %s (config)", impl->source, interface);
+          return impl;
         }
     }
 
+  desktops = get_current_lowercase_desktops ();
```
GeorgesStavracas, line 511: Nitpick: please add an empty line after this line

GeorgesStavracas (changes requested):
Found a couple of tiny nitpicks, but if Emmanuele is happy with it, we're mostly good to go

kennylevinsen:
Style comments addressed and branch rebased.

GeorgesStavracas (approved):

grulja:
This change breaks Settings portal in GNOME. It makes it not to find any backend implementation (while there is xdp-gnome and xdp-gtk available) and when there is no backend implementation found, even though they are available and I can see them exposed on DBus, the settings portal interface will not be exposed and clients cannot basically get to any settings at all. This for example breaks changing light/dark theme in gnome-control-center.

Tested on Fedora 39 and git bisected to this commit.

ebassi:
@grulja Make sure the gnome-session package in F39 is shipping the `gnome-portals.conf` file added by [this merge request](https://gitlab.gnome.org/GNOME/gnome-session/-/merge_requests/95).

grulja:
Looks `gnome-sessions` hasn't been updated in Fedora 39 to GNOME 45, while the rest of packages are updated. That's most likely the issue. I will test to build it locally.

grulja:
I can see `gnome-session` is not updated to version 45 even in upstream, not only in Fedora. This is quite confusing and honesly when relying on such change you should make sure it's going to be picked up to avoid such breakage. I will open a pull request to `gnome-session` package in Fedora backporting this change.

ebassi:
> This is quite confusing and honesly when relying on such change you should make sure it's going to be picked up to avoid such breakage

Sure, but this is not the place to complain about GNOME not releasing gnome-session.

Issue #1011, opened 2023-04-28
OpenFile uses the /run/user/1000/doc/.../file.ext path instead of host path


xhorak:
The `OpenFile` method of `org.freedesktop.portal.OpenURI` portal uses the `/run/user/1000/doc/.../file.ext` path. 

For example when downloading file in Firefox into custom location the document portal entry is created and used when opening the file. This leads that the application which opens the file in the non-host location. This might be confusing for the user and could have some side effects.

For example the libreoffice opened from the `/run/user/...` location are read only:
![image](https://user-images.githubusercontent.com/2932697/235080921-86592548-9999-4ac7-8831-e1dedd03f937.png)

```
 ~/m/mozilla-central  ls -l ~/file-sample_100kB.odt
-rw-r--r--. 1 jhorak jhorak 609710 Apr 28 09:22 /home/jhorak/file-sample_100kB.odt
 ~/m/mozilla-central  ls -l /run/user/1000/doc/c2716c1b/file-sample_100kB.odt
-rw-r--r--. 1 jhorak jhorak 609710 Apr 28 09:22 /run/user/1000/doc/c2716c1b/file-sample_100kB.odt
```
even the permission allows the writing to the file - and the file was not opened before by the LO. Maybe it's unable to create lockfile in the document portal `/run/user/...` path?

PR #1020, opened 2023-05-10
Translate sandboxed path to host path on OpenFile

xhorak:
Opening the file by calling `org.freedesktop.portal.OpenURI` `OpenFile` method launches the target application with the `/run/user/...` path. This can confuse user to show the strange file location. In case of Libreoffice it opens the file as read only (despite it's not read only).

Resolves: #1011

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -662,8 +662,18 @@ handle_open_in_thread_func (GTask *task,
       g_autofree char *path = NULL;
       gboolean fd_is_writable;
       g_autoptr(GError) local_error = NULL;
+      g_autofree char *host_path = NULL;
 
       path = xdp_app_info_get_path_for_fd (request->app_info, fd, 0, NULL, &fd_is_writable, &local_error);
+
+      host_path = get_real_path_for_doc_path (path, request->app_info);
+      if (host_path)
+        {
+          g_debug ("OpenFile: translating path value '%s' to host path '%s'", path, host_path);
+          g_clear_pointer (&path, g_free);
+          path = g_steal_pointer (&host_path);
+        }
+
       if (path == NULL ||
           (writable && !fd_is_writable) ||
           (!xdp_app_info_is_host (request->app_info) && !writable && fd_is_writable))


xhorak:
@matthiasclasen Could you please have a look?

xhorak:
> A few general comments. These are mainly things that need testing, not necessarily problems with the suggested change:
> 
>     * There is no guarantee that there is any 'host file' behind the fd that is passed to OpenFile - this might just be some data that the caller has saved in a temp location inside its sandbox, and then exported to to the document store.
Is it the case? We're trying to find the host path in the document portal entries, so there should not be any temp files involved. If they are, they won't be found by the document portal and behavior stays (broken, but this PR does not trying to fix this).

> 
>     * The launched app may itself be sandboxed, and in that case, the portal will export the file in the document store anyway. That should work with the host path just as well, but needs testing
Yes, I've seen the export code. Question is if it's needed when when we have the `--file-forwarding` which should deal with passing the host files to the target application.

matthiasclasen:
> Is it the case? We're trying to find the host path in the document portal entries, so there should not be any temp files involved. If they are, they won't be found by the document portal and behavior stays (broken, but this PR does not trying to fix this).

What I meant here is that a sandboxed app can:
- Write a file to its filesystem area (ie ~/.var/app/...)
- Pass an fd to this file to org.freedesktop.portal.OpenURI.OpenFile
- Expect the portal to do all the right things to launch another sandboxed app in a way that gives it access to that file

But as you say, this case should continue to work as before: there's no host path, so we keep using the original.

matthiasclasen:
From my perspective, this is good to land.

src/open-uri.c
```diff
@@ -664,6 +664,15 @@ handle_open_in_thread_func (GTask *task,
...
       path = xdp_app_info_get_path_for_fd (request->app_info, fd, 0, NULL, &fd_is_writable, &local_error);
+
+      char *host_path = get_real_path_for_doc_path (path, request->app_info);
+      if (host_path)
+        {
+          g_debug ("OpenFile: translating path value '%s' to host path '%s'", path, host_path);
+          g_free(path);
+          path = host_path;
+        }
+
```
GeorgesStavracas, line 675: I think using `g_autofree` would be nice here. Something like this:

```c
      g_autofree char *host_path = NULL;

      path = xdp_app_info_get_path_for_fd (request->app_info, fd, 0, NULL, &fd_is_writable, &local_error);

      host_path = get_real_path_for_doc_path (path, request->app_info);
      if (host_path)
        {
          g_debug ("OpenFile: translating path value '%s' to host path '%s'", path, host_path);
          g_clear_pointer (&path, g_free);
          path = g_steal_pointer (&host_path);
        }
```

(In the future we'll be able to use `g_set_str()` here, which will be even nicer, but not now because we'd be depending on a GLib version that's too recent)

GeorgesStavracas (approved):
Thanks. Left only one cosmetic suggestion, but AFAICS this seems fine, and people are happy with this approach :)

xhorak:
Cool, could you please merge this one?

GeorgesStavracas (approved):
Code LGTM now. Please rebase against the `main` branch, and remove the merge commit. Also please change the commit's title to to `open-uri: Translate sandboxed path to host path on OpenFile` (the body of the commit message is fine, please leave it as is)

xhorak:
Huh, I hope I fixed everything now.

GeorgesStavracas:
Thank you very much :)

PR #1021, opened 2023-05-11
meson: expose the pytest suite as meson feature and require it in the CI

whot:
Previously the pytests were enabled if pytest and the required python modules were available. Make this more explicit by having a meson feature for that.

Suggested by @smcv in https://github.com/flatpak/xdg-desktop-portal/pull/1016#issuecomment-1542305566

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -154,18 +154,6 @@ endif
 
 enable_installed_tests = get_option('installed-tests')
 
-summary({
-    'Enable docbook documentation': build_docbook,
-    'Enable pipewire support': have_pipewire,
-    'Enable libsystemd support': have_libsystemd,
-    'Enable geoclue support': have_geoclue,
-    'Enable libportal support': have_libportal,
-    'Enable installed tests:': enable_installed_tests,
-  },
-  section: 'Optional builds',
-  bool_yn: true,
-)
-
 ###### systemd units, dbus service files, pkgconfig
 
 base_config = configuration_data()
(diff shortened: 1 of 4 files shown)


smcv (changes requested):

smcv (changes requested):
Looks good to me, other than a typo

meson.build
```diff
@@ -154,18 +154,6 @@ endif
...
-    'Enable docbook documentation': build_docbook,
-    'Enable pipewire support': have_pipewire,
-    'Enable libsystemd support': have_libsystemd,
-    'Enable geoclue support': have_geoclue,
-    'Enable libportal support': have_libportal,
-    'Enable installed tests:': enable_installed_tests,
-  },
-  section: 'Optional builds',
-  bool_yn: true,
-)
```
eli-schwartz, line 167: You do not strictly speaking need to move this around -- summary() can be called multiple times, and will merge identically named sections. Since the addition is defined in a subdir anyway, that might be fine to append there too.

whot: Thanks! I did have it that way at first but for some reason it didn't look/feel right so I moved it around instead. Happy to change it either way though.

tests/meson.build
```diff
@@ -265,15 +265,17 @@ test(
...
   'python3',
   modules: ['dbus', 'dbusmock', 'gi'],
-  required: false,
+  required: get_option('pytest'),
 )
 
-if pytest.found() and python.found()
+enable_pytest = python.found() and python.found()
+
+if enable_pytest
```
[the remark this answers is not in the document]
whot: sigh. I did test with invalid python modules to make sure it works but of course not with missing pytest. Fixed now, thanks for spotting this.

smcv (approved):

meson.build
smcv: I don't feel strongly about this in either direction, so let's go with what @whot preferred.

PR #1022, opened 2023-05-11
Uri validation

matthiasclasen:
We should not accept nonsense strings if the documentation says that we expect a uri.


GeorgesStavracas (approved):
Seems reasonable to me, thanks

PR #1023, opened 2023-05-12
Fix configuration group name in documentation

Sodivad:
As described above in the page the group name is [preferred]


ebassi:
Good catch

PR #1030, opened 2023-05-29
tests: sort the test .portal files and generate them from meson

whot:
Those files don't include all portals (on purpose? by accident?) and they're unsorted so finding whether a portal is there is harder than necessary. so let's at least sort them and generated them from meson so we have one portal per line.

Since `limited.portal` and `tests.portal` share the implementations we can share that bit in meson too.


GeorgesStavracas (approved):
This looks nice to me, but I think it's important for @ebassi to ACK or NACK this before landing

ebassi:
ACK-by: me

PR #1035, opened 2023-06-12
Draft: RFC: add inputcapture trigger for shortcuts and immediate capture

whot:
This is a draft only to start discussion, no implementation beyond the XML specification.

It sits on top of #714 (and #762) but intentionally kept separate for now, I don't want this to hold up either of those.

#714 adds the input capture portal but the only triggers are pointer barriers. This works for e.g. the [input-leap](https://github.com/input-leap/input-leap/) use-case.

This PR proposes two extra triggers:
- `SetEnableTrigger` to allow an application to effectively say "start capturing now"
- `SetGlobalShortcuts` to allow input capture to be triggered by (previously aggreed on) [global shortcuts](https://github.com/flatpak/xdg-desktop-portal/pull/711)

They're fairly identical in the end but the SetEnableTrigger approach would usually require focus, the SetGlobalShortcuts can work without focus.

These could be use by e.g. accessibility technologies that need to take over the input feed on-demand. The shortcuts approach could also be used by a generic macro daemon like autohotkeys that lives external to the compositor.

In both cases the overall architecture could look like this:

```
compositor | ------- xdg desktop portal ------|   application
    EIS    |                                  |
     |  \__|__->______ InputCapture ____->____| libei receiver
      \____|___<-_____ RemoteDesktop ____<-___| libei sender
```

In other words, the application sets up an InputCapture and RemoteDesktop session (and GlobalShortcuts, if applicable). Once input is received the application can send back input events to the compositor where it is processed as expected. Notably: because the real input is "detached" there is no conflict with normal input events.

For the macro daemon this would allow easy triggering of complex sequences, e.g. a tmux-like `shortcut + A` types out "Hello world". It would also allow for things like input forwarding to games - a game runtime manager would simply re-set the EnableTrigger on each focus out, thus (typically) getting the input re-captured whenever it gets back into focus. This would provide a similar effect as the wayland relative-pointer/shortcuts-inhibit/pointer-constraint set of protocols.

For accessibility technologies it's the same thing, the captured input can be converted into whatever other data and sent back to the compositor (e.g. keyboard-to-mouse or touchpad-to-keyboard or tablet-to-whatever conversion).

cc @jadahl, @ofourdan, @garnacho


orowith2os:
I'm guessing adding a dependency on libei for global keybinds wouldn't be worth it at all here, and would be a bit complicated to handle vs just having the portal dole them out?

Wanted to be sure, as it's something that's popped into my mind several times in the past few weeks.

whot:
yeah, it feels a bit wrong to send what is likely a single shortcut trigger via libei and libei isn't really optimised for that case.

swick:
@whot is this still something you want to push forward?

PR #1047, opened 2023-06-30
document-portal: Use correct signature for xdp_fuse_forget

arrowd:
The recent Clang compiler (LLVM 16) became stricter so compiling this code on 32 bit arch results in a error.

Relevant log: https://pkg-status.freebsd.org/beefy17/data/main-i386-default/p4785b313b958_se8efee297c/logs/xdg-desktop-portal-1.16.0_1.log


arrowd:
Can we get this in?

Reference to the upstream documentation, just in case: https://libfuse.github.io/doxygen/structfuse__lowlevel__ops.html

jadahl:
lgtm

PR #1048, opened 2023-07-11
global-shortcuts: Limit signals broadcast to sender

GeorgesStavracas:
Currently the Global Shortcuts portal emits the 3 signals of the D-Bus interface (Activated, Deactivated, and the ShortcutsChanged signals) using the auto-generated D-Bus functions (xdp_dbus_global_shortcuts_emit_*).

These functions end up calling their GObject signals, which call into the signal handlers, which then finally call g_dbus_connection_emit_signal().

The problem is that they call g_dbus_connection_emit_signal() passing NULL to the 'destination_bus_name' parameters, which GDBus documents it to "emit to all listeners". That means we are actively sending these shortcut events to all applications that are willing to listen to these signals!

Fix that by calling g_dbus_connection_emit_signal() directly, and passing the sender bus name properly. This is the same approach that the Inhibit and Location portals use to limit the broadcast of their own signals.

(This has only been compile tested)

### DIFF
--- a/src/global-shortcuts.c
+++ b/src/global-shortcuts.c
@@ -589,13 +589,22 @@ activated_cb (XdpDbusImplGlobalShortcuts *impl,
               GVariant *options,
               gpointer data)
 {
+  GDBusConnection *connection = g_dbus_proxy_get_connection (G_DBUS_PROXY (impl));
   g_autoptr(Session) session = lookup_session (session_id);
   GlobalShortcutsSession *global_shortcuts_session = (GlobalShortcutsSession *)session;
 
   g_debug ("Received activated %s for %s", session_id, shortcut_id);
 
   if (global_shortcuts_session && !global_shortcuts_session->closed)
-    xdp_dbus_global_shortcuts_emit_activated (data, session_id, shortcut_id, timestamp, options);
+    g_dbus_connection_emit_signal (connection,
+                                   session->sender,
+                                   "/org/freedesktop/portal/desktop",
+                                   "org.freedesktop.portal.GlobalShortcuts",
+                                   "Activated",
+                                   g_variant_new ("(ost@a{sv})",
+                                                  session_id, shortcut_id,
+                                                  timestamp, options),
+                                   NULL);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
@aleixpol could you please test this against the KDE portal? Nothing should effectively change for apps nor the KDE backend, I just want to make sure nothing regresses due to this change

hsitter:
I know this isn't really helpful for the PR at hand but I tried to run it on the KDE stack and all I got was assertion failures. I'm a bit out of my depth here, any help would be appreciated.

For one the type string seems wrong here:

```diff
diff --git a/src/global-shortcuts.c b/src/global-shortcuts.c
index 217aafb..31b8b55 100644
--- a/src/global-shortcuts.c
+++ b/src/global-shortcuts.c
@@ -333,7 +333,7 @@ xdp_verify_shortcuts (GVariant *shortcuts,
   g_autoptr(GVariantIter) iter = NULL;
 
   iter = g_variant_iter_new (shortcuts);
-  while (g_variant_iter_loop (iter, "sa{sv}", &shortcut_name, &values))
+  while (g_variant_iter_loop (iter, "(s@a{sv})", &shortcut_name, &values))
     {
       GVariantBuilder shortcut_builder;
 
```

But even with that fixed I get problems with the shortcut_builder and I can't seem to find what's wrong there. The values variable looks valid when I run `g_variant_print(values, TRUE));`

```
(/usr/libexec/xdg-desktop-portal:356073): GLib-CRITICAL **: 18:56:25.240: g_variant_builder_end: assertion 'valid_builder' failed

(/usr/libexec/xdg-desktop-portal:356073): GLib-CRITICAL **: 18:56:25.240: g_variant_get_type: assertion 'value != NULL' failed

(/usr/libexec/xdg-desktop-portal:356073): GLib-CRITICAL **: 18:56:25.240: g_variant_type_is_array: assertion 'g_variant_type_check (type)' failed

(/usr/libexec/xdg-desktop-portal:356073): GLib-CRITICAL **: 18:56:25.240: g_variant_get_type_string: assertion 'value != NULL' failed

(/usr/libexec/xdg-desktop-portal:356073): GLib-ERROR **: 18:56:25.240: g_variant_new: expected array GVariantBuilder but the built value has type '(null)'
Trace/breakpoint trap (core dumped)
```

jadahl:
> Which reminds me how badly this portal needs tests slightly_frowning_face

Wouldn't be too horribly hard to add tests that use the python-dbusmock helpers to mock the `.impl.portal.` backend.

GeorgesStavracas:
With the fixes from https://github.com/flatpak/xdg-desktop-portal/pull/1055, I think we're safe to land this

PR #1051, opened 2023-07-15
Make PipeWire dependency mandatory

GeorgesStavracas:
PipeWire has solidified its position in the Linux world over the last few years, and nowadays pretty much all relevant Linux distributions ship PipeWire 0.3.

Make the PipeWire dependency required.


ebassi (approved):
:+1: from me

PR #1053, opened 2023-07-18
build: Bump version to 1.17.99

jadahl:
This is to let users depend on > 1.17 when it needs portal APIs pending release.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'xdg-desktop-portal',
   'c',
-  version: '1.17.0',
+  version: '1.17.99',
   meson_version: '>= 0.56.2',
   license: 'LGPL-2.0-or-later',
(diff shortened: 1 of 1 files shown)


smcv:
1.17.0 doesn't yet exist as a release, so anything added post-1.16.x is not something that third-party projects can depend on yet. If there are development snapshots of the future 1.18.0 release, they should be versioned 1.17.0, 1.17.1, ...

If you want a milestone release to test against (like the way you might use mutter 45~alpha in gnome-shell) then the way to get that would be to release 1.17.0 (tagged as a prerelease), similar to what happened for 1.15.0.

jadahl:
Ah it was bumped in 686cd0eab4e68ed694b9e9f6368ec4afbaa53141 which is a bit confusing but ok.

Issue #1054, opened 2023-07-21
Global shortcuts timestamps incompatible with input timestamps


jadahl:
The `Activated` and `Deactivated` timestamps are specified to be "as seconds and microseconds since the Unix epoch". This has some problems:

  * Input timestamps are practically always using the monotonic clock. The only exception is Xorg from *very* long ago. The clock being different here means we can't pass them forward, thus have to make up new timestamps. That looses information about the actual time the action was triggered.
  * "Since the Unix epoch" means a bunch of things that since long is no longer used for anything other displaying a clock to the user, as it means it'd be affected by timezone changes, when the clock, etc.

Is the "base" for the timestamp really needed? If it really is, it should arguably be using the same as all other input like events on the system, meaning the monotonic clock.

Should we just call it a "bug" and change it? Otherwise we'll need a version bump.

Cc: @aleixpol @GeorgesStavracas

jadahl:
Attempted to see where `xdg-desktop-portal-kde` gets its timestamp from, and it appears to be using the X11 timestamp (monotonic clock in milliseconds), or hard code `0` in the non-X11 case: https://invent.kde.org/plasma/kglobalacceld/-/blob/11d65c8c35ff4d8f0f7a38bf2bf3889eb6973934/src/component.cpp#L199-211 so it's not following the spec. It also makes me believe the intention was to use a clock similar to the input timestamps, since that is what kglobalaccel roughly uses, only that it does not related to any unix epoch, meaning the wording as is is a bug, that we can probably fix without bumping any versions. I don't think there are any other backend implementations than the KDE one today.

[...]

PR #1055, opened 2023-07-21
Global shortcuts fixes and tests

jadahl:
While adding basic testing to global shortcuts, I ran into a bunch of issues that were fixed along the way (unnecessary options filtering, refcount issues, etc).

One of the tests will race twice a year, but I hope we can change the test if https://github.com/flatpak/xdg-desktop-portal/issues/1054 can be fixed by changing the meaning of the timestamps in the (de)activated signals.


jadahl:
Cc: @aleixpol mind reviewing?

tests/templates/meson.build
```diff
@@ -2,6 +2,7 @@ template_files = [
   '__init__.py',
   'email.py',
   'remotedesktop.py',
+  'globalshortcuts.py',
```
GeorgesStavracas, line 5: Might be better to sort alphabetically here

GeorgesStavracas (approved):
Just one minor nit. LGTM otherwise. A bit disappointing that so many `g_variant_ref_sink()`s had to be sprinkled in the file, but so is life.

jadahl:
> A bit disappointing that so many `g_variant_ref_sink()`s had to be sprinkled in the file, but so is life.

An alternative is to do `g_steal_pointer()` in the places that caused the double-frees (e.g. `xdp_dbus_impl_global_shortcuts_call_bind_shortcuts()`)

GeorgesStavracas:
> An alternative is to do g_steal_pointer() in the places that caused the double-frees (e.g. xdp_dbus_impl_global_shortcuts_call_bind_shortcuts())

Now that you mention; I think I'd prefer that :slightly_smiling_face:

aleixpol (approved):
That is very cool, thank you very much, @jadahl.

jadahl:
> Now that you mention; I think I'd prefer that slightly_smiling_face

Done.

> 23/23 xdg-desktop-portal:pytest / pytest                             OK             57.86s

The timeout seems to happen at 60s - this will start to be flaky I think. Would be nice if the pytest suite could be split up per portal.

GeorgesStavracas:
> The timeout seems to happen at 60s - this will start to be flaky I think. Would be nice if the pytest suite could be split up per portal.

Yes, but I wouldn't block this PR on that :)

jadahl:
> Yes, but I wouldn't block this PR on that :)

Pushed a commit that bumps the timeout to 120s. Will give us some breathing room.

Issue #1057, opened 2023-07-21
Run pytest tests individually


GeorgesStavracas:
https://github.com/flatpak/xdg-desktop-portal/pull/1055 introduced a commit to extend the timeout of pytests to 120s.

That's a temporary breathing room for the test not to get flacky, but ideally we'd run individual portal tests separatedly

ebassi:
Picking this up since I've already had to touch the test suite, but will likely have to wait a couple of weeks—GUADEC and scheduled PTO.

PR #1060, opened 2023-07-28
tests: run pytest with -n auto if pytest-xdist is available

whot:
All our tests must be invocable in parallel, so let's make use of that. This brings the pytest run down to around 8 seconds on my machine.

Unfortunately the -n auto command is available only if xdist is available too and pytest fails if we provide it otherwise so we need to check if we have that module.

This reduces the need for #1057

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -78,7 +78,7 @@ jobs:
     - name: Install dependencies for the pytest test suite
       run: |
         $RUN_CMD apt-get install --quiet -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


whot:
`23/23 xdg-desktop-portal:pytest / pytest                             OK              45.74s`

well, sadly that's less of an impact than I'd been hoping for...

GeorgesStavracas (approved):

GeorgesStavracas:
Not much but welcomed nonetheless

PR #1063, opened 2023-07-31
Add a pre-commit hook (and do so in the CI too)

whot:
May not be everyone's cup of beverage but i've started to like it quite a fair bit - running `pre-commit install` locally means I get sanity checks before committing, so things like whitespace errors (or in this case ruff errors) are a thing of the past.


.github/workflows/check.yml
```diff
@@ -7,6 +7,26 @@ env:
...
+
+    - name: Check out xdg-desktop-portal
+      uses: actions/checkout@v3
+
+    - name: Install pre-commit
+      run: pip install pre-commit
+
+    - name: Run pre-commit checks
+      run: pre-commit run --all
+
```
WhyNotHugo, line 29: Using https://pre-commit.ci/ would be a lot simpler, and it runs a lot faster.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: trailing-whitespace
+- repo: https://github.com/psf/black
+  rev: 22.10.0
+  hooks:
+  - id: black
+    args: ['--check', '--diff', '.']
```
WhyNotHugo, line 11: I'd remove these args. The make this hook substantially less useful. By default, this hook formats the code with black if necessary. With these args, the hook just warns that black needs to be run manually.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: v0.0.254
+  hooks:
+  - id: ruff
+    args: ['--ignore=E741,E501', '.']
```
WhyNotHugo, line 16: I'd put the ignored rules in `pyproject.toml` or `setup.cfg` so that IDEs and manual usages of `ruff` will use the same settings as the pre-commit hook.

.pre-commit-config.yaml
whot: ftr, I do a lot of `git add -p` style commits, so running `black` on commit means my code gets moved around while I still have the editor open - depending on what other work is going on that has the potential for losing work. Hence the specific choice of only doing `--check` here (and `--diff` to see what needs changing).

.github/workflows/check.yml
whot: Thanks, that sounds like a plan but ... how do we actually use that? I'm staring at that website but I can't find any guidance on how to enable this.

WhyNotHugo: You have to click on the "Sign in with GitHub" button. I guess it needs to be done by someone with permissions to enable it for this repository.

.pre-commit-config.yaml
WhyNotHugo: I always use `git add -p` and I've never seen any data loss due to using `pre-commit`. Usually it'll stash your changes before running black, and if your stash and its changes have conflicts, it just keeps your version around.

whot:
The two first commits were now merged in #1341, I'm going to drop the rest. If someone wants to pick this up please do so, I have too many things on my plate.

PR #1069, opened 2023-08-07
open-uri: Pass activation token to the FileManager1 interface

### LINKED ISSUE #876: XDG activation token is not passed to the FileManager1 interface
Empty string is send as the StartupID argument: https://github.com/flatpak/xdg-desktop-portal/blob/753fba165d2cbf3f8f04b96285903d76c0996c1c/src/open-uri.c#L715

On Wayland, file managers seem to interpret this argument as XDG activation token, so it could be passed here

ilya-fedin:
Fixes #876

I've also notices a typo in handle_open_directory, apparently because it was copied from handle_open_file...


whot (approved):
this LGTM

GeorgesStavracas (approved):
Thanks

PR #1073, opened 2023-08-16
Allow building man pages without docbook

WhyNotHugo:
Downstream distributions will want to build and package man pages, but generally won't need to build the docbook themselves.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -36,31 +36,30 @@ if build_docbook
         install_dir: docs_dir,
       )
   endforeach
+endif
 
-  rst2man = find_program('rst2man', 'rst2man.py', required: false)
-  if rst2man.found()
-    rst2man_flags = [
-      '--syntax-highlight=none',
-    ]
-    
-    man_pages = [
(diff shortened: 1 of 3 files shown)


doc/meson.build
```diff
@@ -36,31 +36,31 @@ if build_docbook
...
-  rst2man = find_program('rst2man', 'rst2man.py', required: false)
-  if rst2man.found()
-    rst2man_flags = [
-      '--syntax-highlight=none',
-    ]
-    
-    man_pages = [
-      { 'input': 'portals-conf.rst', 'output': 'portals.conf', 'section': '5' },
-    ]
+rst2man = find_program('rst2man', 'rst2man.py', required: get_option('man-pages'))
```
jadahl, line 41: Shouldn't need to do this twice.

WhyNotHugo: Sorry, I think I messed up with on my final revision.

Should be fixed now.

GeorgesStavracas (approved):

GeorgesStavracas:
Thanks!

PR #1079, opened 2023-08-23
Improve visuals of documentation

GeorgesStavracas:
Nothing fancy. I won't continue with this as we're eventually moving towards Sphynx, but there's no reason to not make things look less bad until then.


GeorgesStavracas:
This won't hurt, and won't be an obstacle to any future improvements or overhauls to the documentation system we have in place, so I'm tempted to just merge it already

PR #1080, opened 2023-08-24
Add mapping_id to streams

jadahl:
From the commit that introduces it:

This uuid is intended to be used for matching a PipeWire stream with a
libei region, so that one e.g. can emit absolute pointer events using
libei on a window screen cast. This was not possible before, since there
was no way to match a region with a stream. This also makes it easier to
pair streams and libei regions in general, as one doesn't need to look
at any sizes or positions.

See https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/246.

Cc: @whot

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -23,7 +23,7 @@
 
       The Screen cast portal allows to create screen cast sessions.
 
-      This documentation describes version 4 of this interface.
+      This documentation describes version 5 of this interface.
   -->
   <interface name="org.freedesktop.impl.portal.ScreenCast">
     <!--
@@ -235,6 +235,19 @@
               The type of the content which is being screen casted.
               For available source types, see the AvailableSourceTypes property.
               This property was added in version 3 of this interface.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>mapping_id s</term>
+            <listitem><para>
+              An optional identifier used to map different aspects of the
+              resource this stream corresponds to.
+            </para><para>
+              When used in a remote desktop session, the mapping_id can be used to
+              match a libei region of absolute libei devices. There may be
+              multiple absolute libei devices, and each device may have multiple
+              regions, but a mapping_id will only match one of these regions per
+              device. This property was added in version 5 of this interface.
             </para></listitem>
           </varlistentry>
         </variablelist>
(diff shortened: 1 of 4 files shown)


whot (approved):
FTR, I'm thinking some other name than uuid (to avoid confusing with real UUIDs) may be good but I have failed to come up with any so far. So this LGTM, thanks!

jadahl:
> FTR, I'm thinking some other name than uuid (to avoid confusing with real UUIDs) may be good but I have failed to come up with any so far. So this LGTM, thanks!

Not super happy with "uuid" either, I just blindly copied the term from libei. Other ideas are "resource-handle", "pairing-handle", "associate-with".

jadahl:
I was just suggested "mapping_id" which I kind of like. What do you think?

jadahl:
Changed "uuid" to "mapping_id".

whot:
libei PR is now [libei !249](https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/249) and has been merged

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -237,6 +237,15 @@
               This property was added in version 3 of this interface.
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>uuid s</term>
```
whot, line 241:
```suggestion
            <term>mapping_id s</term>
```

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -265,6 +265,19 @@
...
             </para></listitem>
           </varlistentry>
+          <varlistentry>
+            <term>mapping_id s</term>
+            <listitem><para>
+              An optional identifier used to map different aspects of the
+              resource this stream corresponds to.
+	    </para><para>
+	      When used in a remote desktop session, the mapping_id can be used to
+              match a libei region of absolute libei device devices. There may
```
whot, line 275:
```suggestion
              match a libei region of absolute libei devices. There may
```

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -265,6 +265,19 @@
...
+              be multiple absolute libei devices, and each device may have
+              multiple regions, but a mapping_id will only match one of these
+              regions. This property was added in version 5 of this interface.
```
whot, line 278:
```suggestion
              regions per device. This property was added in version 5 of this interface.
```
Because in theory we could have 2 absolute devices each with a region that maps to stream S1 while stream S2 may only have one (or none).

Issue #1081, opened 2023-08-24
Search path for portals.conf 


smcv:
Suppose for the sake of an example that we're running GNOME Classic, so XDG_CURRENT_DESKTOP is "GNOME-Classic;GNOME".

As of 1.17.0, the search path seems to be, most important first:

* `$XDG_CONFIG_HOME/xdg-desktop-portal/portals.conf`
* `$XDG_CONFIG_HOME/xdg-desktop-portal/gnome-classic-portals.conf`
* `$XDG_DESKTOP_PORTAL_DIR/gnome-classic-portals.conf`
* `$XDG_CONFIG_HOME/xdg-desktop-portal/gnome-portals.conf`
* `$XDG_DESKTOP_PORTAL_DIR/gnome-portals.conf`

where $XDG_CONFIG_HOME is done according to the XDG basedir spec as usual, and $XDG_DESKTOP_PORTAL_DIR defaults to /etc/xdg-desktop-portal.

I think before releasing this as a stable 1.18.0, we should seriously consider changing this around a bit.

`$XDG_DESKTOP_PORTAL_DIR` is undocumented and seems to be primarily for unit tests. I think if this is set, it should either disable all the other paths, or be inserted into the list as the highest priority - otherwise, the unit tests are going to fail for a user who has per-user customization.

Under normal circumstances, I think we should be searching at least one directory in `/usr`, to support the general effort to make running a system with a near-empty `/usr` a valid thing to do: desktop environments like GNOME should be installing their default choice of portals into `/usr`, not `/etc`. I think `/usr/share/xdg-desktop-portal/*.conf` would make sense as the `/usr` "mirror" of `/etc/xdg-desktop-portal`.

The XDG basedir spec is most useful if it's used consistently, so that for example a user can have `XDG_CONFIG_HOME=$HOME/.config-$hostname`, `XDG_CONFIG_DIRS=$HOME/.config:/etc/xdg` if they want to, but as currently implemented we don't search XDG_CONFIG_DIRS.

It also seems odd that the user can have either desktop-specific configuration or global configuration, but if they have global configuration, it "wins" - I would have expected that a desktop-specific configuration should "win".

I think we should seriously consider aligning this with the mimeapps spec, <https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html#file>, which would solve a lot of this immediately.

So my proposal is this:

[...]

PR #1082, opened 2023-08-24
Make the search order for portals.conf consistent with mimeapps.list

smcv:
Previously, we would search for a user-specific portals.conf as highest precedence, and then iterate through desktop environments, searching for a user-specific desktop-specific file followed by a system-wide desktop-specific file. Now, the precedence order is consistently that directories are most important, we only pay attention to desktop environments within a single directory, and desktop-specific configuration is always higher-precedence than generic within the same directory.

Previously, desktop environments and OS vendors had to install their per-desktop configuration into /etc, contradicting the general design principle of putting defaults in /usr and reserving /etc for sysadmin overrides. Now, we search both /etc and /usr/share (with typical environment variables and build-time settings).

Previously, we didn't fully follow the XDG Base Directory specification when looking up per-user configuration; now we do. This is potentially significant for users who have more than one per-user directory, for example a machine-specific directory within a home directory shared between machines.

Update documentation accordingly.

Resolves: https://github.com/flatpak/xdg-desktop-portal/issues/1081

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -387,61 +387,114 @@ load_portal_configuration_for_dir (gboolean    opt_verbose,
   return NULL;
 }
 
-void
-load_portal_configuration (gboolean opt_verbose)
+/*
+ * Returns: %TRUE if configuration was found in @dir
+ */
+static gboolean
+load_config_directory (const char *dir,
+                       const char **desktops,
+                       gboolean opt_verbose)
 {
   g_autoptr(PortalConfig) conf = NULL;
-  g_autofree char *user_portal_dir = NULL;
-  const char **desktops;
-  const char *portal_dir;
 
-  /* We need to override this in the tests */
-  portal_dir = g_getenv ("XDG_DESKTOP_PORTAL_DIR");
-  if (portal_dir == NULL)
-    portal_dir = SYSCONFDIR "/xdg-desktop-portal";
-
-  user_portal_dir = g_build_filename (g_get_user_config_dir (),
-                                      "xdg-desktop-portal",
-                                      NULL);
-
-  conf = load_portal_configuration_for_dir (opt_verbose, user_portal_dir, "portals.conf");
-  if (conf != NULL)
-    {
-      if (opt_verbose)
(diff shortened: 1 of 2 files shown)


src/portal-impl.c
```diff
@@ -387,61 +387,114 @@ load_portal_configuration_for_dir (gboolean    opt_verbose,
...
   g_autofree char *user_portal_dir = NULL;
+  const char * const *dirs;
+  const char * const *iter;
   const char **desktops;
   const char *portal_dir;
 
+  desktops = get_current_lowercase_desktops ();
+
   /* We need to override this in the tests */
   portal_dir = g_getenv ("XDG_DESKTOP_PORTAL_DIR");
```
smcv, line 446: I left this as an intentionally undocumented implementation detail, since it's only there for the tests as far as I can see.

src/portal-impl.c
```diff
@@ -387,61 +387,114 @@ load_portal_configuration_for_dir (gboolean    opt_verbose,
...
+    {
+      g_autofree char *dir = g_build_filename (*iter, XDP_SUBDIR, NULL);
+
+      if (load_config_directory (dir, desktops, opt_verbose))
+        return;
     }
+
+  /* ${datadir}/xdg-desktop-portal/(DESKTOP-)portals.conf */
+  if (load_config_directory (DATADIR "/" XDP_SUBDIR, desktops, opt_verbose))
+    return;
```
smcv, line 497: In practice this is redundant with `XDG_DATA_DIRS` but I thought we should probably search it anyway?

GeorgesStavracas (approved):
LGTM, it works. Thanks Simon, this is looking excellent

smcv:
It doesn't need to be immediately, but I'd like to get a 1.17.1 out with this change at some point, so that when we start asking desktop environment maintainers (upstream or downstream) to install their own `foo-portals.conf` we can ask them to install it as `/usr/share/xdg-desktop-portal/foo-portals.conf`, reserving `/etc` for the sysadmin (in the same spirit as how libeconf and systemd behave).

PR #1083, opened 2023-08-24
Only warn about UseIn if we actually pay attention to it

smcv:
There's nothing inherently wrong with UseIn, and it's valid for portal implementations to provide it for backwards compatibility with xdg-desktop-portal < 1.17 or desktop environments that don't yet provide a DESKTOP-portals.conf. Instead of warning whenever a portal has it, warn when we use it to choose a portal (which must mean that we didn't find a better configuration mechanism).


ebassi (approved):
LGTM

PR #1086, opened 2023-08-25
docs: Clarify InputCapture.Enable options

bilelmoussaoui:
Currently the documentation mentions the need of passing a handle token but it is not used as the method itself doesn't return an object path.

The Disable/Release methods don't take such option, so I assumed it is a copy/paste failure

cc @whot


whot (approved):
good catch, thanks. I think that's a double copy/paste, in an earlier version of the portal all methods used requests but I'm not sure Enable ever did anyway. Thanks!

GeorgesStavracas (approved):

PR #1087, opened 2023-08-25
Handle app_id's not havinge a matching .desktop file

swick:
Applications which are not flatpaks don't enforce that the app id has a
matching desktop file and in this case info is NULL and we crash. Fall
back to the app ap when we can't get the name from the desktop file.

Same issue in screenshot, wallpaper and location.

Also changing the code to be more similar.


TingPing:
Typo in all the commit messages: "havinge" -> "having".

Otherwise I think it seems reasonable.

swick:
Thanks, fixed.

GeorgesStavracas (approved):

PR #1088, opened 2023-08-28
docs: Clarify PrepareInstall response

bilelmoussaoui:
It was missing the icon field + mentioning it would always return a value even if the used doesn't set the editable icon / editable name options

### DIFF
--- a/data/org.freedesktop.portal.DynamicLauncher.xml
+++ b/data/org.freedesktop.portal.DynamicLauncher.xml
@@ -153,7 +153,13 @@
           <varlistentry>
             <term>name s</term>
             <listitem><para>
-              The name chosen by the user for the launcher.
+              The name chosen by the user for the launcher or the predefined one.
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>icon v</term>
+            <listitem><para>
+              The icon chosen by the user for the launcher or the predefined one.
             </para></listitem>
           </varlistentry>
           <varlistentry>


data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -153,7 +153,13 @@
           <varlistentry>
             <term>name s</term>
             <listitem><para>
-              The name chosen by the user for the launcher.
+              The name chosen by the user for the launcher or the predefined one.
```
ebassi, line 156:
```suggestion
              The name chosen by the user for the launcher, or the predefined one.
```

Also: who chooses the "predefined" name? The portal implementation? The host?

data/org.freedesktop.portal.DynamicLauncher.xml
```diff
@@ -153,7 +153,13 @@
...
+            </para></listitem>
+          </varlistentry>
+          <varlistentry>
+            <term>icon v</term>
+            <listitem><para>
+              The icon chosen by the user for the launcher or the predefined one.
```
ebassi, line 162:
```suggestion
              The icon chosen by the user for the launcher, or the predefined one.
```

Same as above: who chooses the predefined icon?

ebassi (changes requested):

data/org.freedesktop.portal.DynamicLauncher.xml
bilelmoussaoui: The caller of PrepareInstall can pass name/icon and they can edited by the user if the call to PrepareInstall allows them to

bilelmoussaoui: So not sure if predefined is the correct wording

swick:
Opened a new PR for this: https://github.com/flatpak/xdg-desktop-portal/pull/1798

PR #1089, opened 2023-08-28
ci: Exercise creating dist tarball

GeorgesStavracas:
Every release we have to cross our fingers that meson dist isn't broken, and that's not fantastic. Let's test it on every pull request, so we can catch PRs that break dist earlier.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -52,6 +52,7 @@ jobs:
           desktop-file-utils \
           fuse3 \
           gettext \
+          git \
           gnome-desktop-testing \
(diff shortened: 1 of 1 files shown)


ebassi:
On GitLab I usually upload the dist tarball as an artefact on a CI job I run for tags.

In theory, it should be possible to run a CI job on tags that creates a new GitHub release, and then attaches the tarball as an artefact: https://github.com/marketplace/actions/create-release

whot (approved):

GeorgesStavracas:
Removed dist tarball uploading for now

PR #1094, opened 2023-09-02
print: Bump interface version to 2

A6GibKm:
A new accept_label option was added, this should bump the version number.

See 65e625e63dfdbdaa288d8ee82091b8324117bb3e.


jadahl:
This lgtm

PR #1097, opened 2023-09-06
xdg-utils: Properly quote flatpak command

jsparber:
`xdp_app_info_rewrite_commandline` has to respect `quote_escape` also when no commandline is given.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -397,8 +397,10 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
           for (i = 1; commandline[i]; i++)
             g_ptr_array_add (args, maybe_quote (commandline[i], quote_escape));
         }
-      else
+      else if (quote_escape)
         g_ptr_array_add (args, g_shell_quote (app_info->id));
+      else
+        g_ptr_array_add (args, g_strdup (app_info->id));
       g_ptr_array_add (args, NULL);
 
       return (char **)g_ptr_array_free (g_steal_pointer (&args), FALSE);


src/xdp-utils.c
```diff
@@ -397,8 +397,11 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
           for (i = 1; commandline[i]; i++)
             g_ptr_array_add (args, maybe_quote (commandline[i], quote_escape));
         }
-      else
+      else if (quote_escape) {
         g_ptr_array_add (args, g_shell_quote (app_info->id));
+      } else {
+        g_ptr_array_add (args, g_strdup (app_info->id));
+      }
```
GeorgesStavracas, line 404: Please follow the code style of the file. Curly braces are in their own lines

jsparber: fixed it.

smcv:
> xdp_app_info_rewrite_commandline has to respect quote_escape also when no commandline is given

This looks right as far as I can tell, but what bug are you fixing? It would be much easier to assess whether this is correct with either a comment here or an issue report that briefly describes: steps to reproduce; expected result; actual result.

jangernert:
I tried to build an run this to see if it fixes a bug I am hitting. But I failed to make it work. So here are my observations in hope it helps to further review the merge request:

- install [ashpd demo](https://flathub.org/apps/com.belmoussaoui.ashpd.demo) and open it
- in the background section enable "Auto Start" and press "Request"
- `.config/autostart/` will have the file `com.belmoussaoui.ashpd.demo.desktop`
- the app id is escaped in the exec line and won't work: `Exec=flatpak run ''\\''com.belmoussaoui.ashpd.demo'\\'''`
- go back to ashpd demo and fill in something into the "Command" entry and press "Request" again
- the autostart desktop file will change with the following exec line `Exec=flatpak run --command=test123 com.belmoussaoui.ashpd.demo`
- the app id is not escaped and the autostart will work

smcv:
> to see if it fixes a bug I am hitting

Please open an issue that describes the bug you are hitting (version of x-d-p / what you did / expected result / actual result), while using an unpatched xdg-desktop-portal (without this PR applied).

jangernert:
> > to see if it fixes a bug I am hitting
> 
> Please open an issue that describes the bug you are hitting (version of x-d-p / what you did / expected result / actual result), while using an unpatched xdg-desktop-portal (without this PR applied).

#1118

jsparber:
Sorry for not responding for the last few weeks.

@smcv  I didn't think this would fix an actual bug, i just noticed that the code path was wrong

@jangernert Does this fix issue #1118 for you?

jangernert:
> @jangernert Does this fix issue https://github.com/flatpak/xdg-desktop-portal/issues/1118 for you?

I wasn't  able to test it. I was able to build and run the branch. But all the portal requests kept failing. Even the ones not related to this MR. So I must have done something wrong.

jsparber:
> > @jangernert Does this fix issue #1118 for you?
> 
> I wasn't able to test it. I was able to build and run the branch. But all the portal requests kept failing. Even the ones not related to this MR. So I must have done something wrong.

I had the same problem initially adding a symlink `ln -s /usr/share/xdg-desktop-portal /usr/local/share/xdg-desktop-portal` after installing to the `/usr/local/` prefix solves the issue.  I'm not sure if this is the best solution but it makes it work

jangernert:
> I had the same problem initially adding a symlink ln -s /usr/share/xdg-desktop-portal /usr/local/share/xdg-desktop-portal after installing to the /usr/local/ prefix solves the issue. I'm not sure if this is the best solution but it makes it work for Fedora

That did the trick for me as well. And I can confirm: #1118 was fixed running this branch compared to the fedora package. Didn't test master without this MR, though. So take it with a tiny bit of salt.

GeorgesStavracas (approved):

PR #1098, opened 2023-09-08
Draft background: Add ability to specify timers to start app

jsparber:
This is more a proof of concept therefore it's marked as draft.

Since autostart on system start and autostarting an app in ad a specific time are closely related it makes sense for me to not introduce a new portal but add an additional feature to the `Background` portal. 

I implemented the timer backend in `xdg-dekstop-portal-gnome` using systemd user timers. (https://gitlab.gnome.org/jsparber/xdg-desktop-portal-gnome/-/commits/add_timer_background)

And in gnome clocks https://gitlab.gnome.org/jsparber/gnome-clocks/-/commits/use_background_portal

### DIFF
--- a/data/org.freedesktop.impl.portal.Background.xml
+++ b/data/org.freedesktop.impl.portal.Background.xml
@@ -103,6 +103,30 @@
     </method>
 
     <!--
+      EnableTimer:
+      @app_id: App id of the application
+      @timer: List of times the app should be started
+      @commandline: The commandline to used to start the app at the specificied time
+      @flags: Flags influencing the details of the timer
+      @result: TRUE if timer was enabled, FALSE otherwise
+
+      Enables or disables autostart for an application.
+
+      The following flags are understood:
+      <simplelist>
+        <member>1: Wake system from suspend</member>
+        <member>2: Use D-Bus activation</member>
+      </simplelist>
+    -->
+    <method name='EnableTimer'>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="as" name="timer" direction="in"/>
+      <arg type="as" name="commandline" direction="in"/>
+      <arg type="u" name="flags" direction="in"/>
+      <arg type="b" name="result" direction="out"/>
+    </method>
+
+    <!--
       RunningApplicationsChanged:
 
       This signal gets emitted when applications change their state
--- a/src/background.c
+++ b/src/background.c
@@ -110,6 +110,12 @@ typedef enum {
   AUTOSTART_FLAGS_ACTIVATABLE = 1 << 0,
 } AutostartFlags;
 
+typedef enum {
+  TIMER_FLAGS_NONE = 0,
+  TIMER_FLAGS_ACTIVATABLE = 1 << 0,
+  TIMER_FLAGS_WAKE = 1 << 1,
+} TimerFlags;
+
 static GVariant *
 get_all_permissions (void)
 {
@@ -704,6 +710,10 @@ handle_request_background_in_thread_func (GTask *task,
   AutostartFlags autostart_flags = AUTOSTART_FLAGS_NONE;
   gboolean activatable = FALSE;
   g_auto(GStrv) commandline = NULL;
+  const char * const *run_times = { NULL };
+  gboolean wake_system = FALSE;
+  gboolean timer_enabled;
+  TimerFlags timer_flags = TIMER_FLAGS_NONE;
 
   REQUEST_AUTOLOCK (request);
 
@@ -712,9 +722,16 @@ handle_request_background_in_thread_func (GTask *task,
   g_variant_lookup (options, "autostart", "b", &autostart_requested);
   g_variant_lookup (options, "commandline", "^a&s", &autostart_exec);
   g_variant_lookup (options, "dbus-activatable", "b", &activatable);
+  g_variant_lookup (options, "timer", "^a&s", &run_times);
+  g_variant_lookup (options, "wake-system", "b", &wake_system);
 
-  if (activatable)
+  if (activatable) {
     autostart_flags |= AUTOSTART_FLAGS_ACTIVATABLE;
+    timer_flags |= TIMER_FLAGS_ACTIVATABLE;
+  }
+
+  if (wake_system)
+    timer_flags |= TIMER_FLAGS_WAKE;
 
   app_id = xdp_app_info_get_id (request->app_info);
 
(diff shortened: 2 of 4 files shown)


src/background.c
```diff
@@ -704,6 +710,10 @@ handle_request_background_in_thread_func (GTask *task,
   AutostartFlags autostart_flags = AUTOSTART_FLAGS_NONE;
   gboolean activatable = FALSE;
   g_auto(GStrv) commandline = NULL;
+  const char * const *run_times = { NULL };
```
hfiguiere, line 713:
```suggestion
  g_autofree const char * const *run_times = NULL;
```

`^a&s` return like `g_variant_get_strv()` it will leak otherwise.

razzeee:
As the context wasn't clear to me and I needed to dig a little bit.

This can be used to wake up the system and start an app with certain commands.

The example julian implemented is for gnome on mobile devices. Specifically the clocks/alarms app saving their alarms via the portal. And the system then waking and starting the app.

src/background.c
```diff
@@ -782,14 +799,16 @@ handle_request_background_in_thread_func (GTask *task,
...
 
   autostart_enabled = FALSE;
+  timer_enabled = FALSE;
 
   commandline = xdp_app_info_rewrite_commandline (request->app_info, autostart_exec,
                                                   FALSE /* don't quote escape */);
   if (commandline == NULL)
     {
-      g_debug ("Autostart not supported for: %s", app_id);
+      g_debug ("Autostart and schedualed autostart not supported for: %s", app_id);
```
razzeee, line 808:
```suggestion
      g_debug ("Autostart and scheduled autostart not supported for: %s", app_id);
```

Typo

src/background.c
```diff
@@ -873,11 +907,23 @@ validate_commandline (const char *key,
 
   return TRUE;
 }
+static gboolean
+validate_timer (const char *key,
+                      GVariant *value,
+                      GVariant *options,
+                      GError **error)
+{
+	// TODO: validate
```
razzeee, line 916: Marking this todo with a comment

data/org.freedesktop.impl.portal.Background.xml
```diff
@@ -102,6 +102,30 @@
       <arg type="b" name="result" direction="out"/>
     </method>
 
+    <!--
+      EnableTimer:
+      @app_id: App id of the application
+      @timer: List of times the app should be started
```
razzeee, line 108: I find it kind of weird, that this is a list but named with a singular. While `flags` is named plural. So I guess this should be plural too?

razzeee:
One thing I couldn't gather from this (but that might be due to me not knowing much about portals etc) is how the timers have to look, but my assumtion of that being a string might just be wrong.

Anyway, mostly I'm wondering if timezones are something this should care about.

jsparber:
> One thing I couldn't gather from this (but that might be due to me not knowing much about portals etc) is how the timers have to look, but my assumtion of that being a string might just be wrong.
> 
> Anyway, mostly I'm wondering if timezones are something this should care about.

This needs a proper format definition for the timers. I used simple systemds format which includes timezones see https://wiki.archlinux.org/title/Systemd/Timers


Thanks for the code review, but I'm looking actually more for feedback if this portal and implementation direction makes sense.

src/background.c
```diff
@@ -873,11 +907,23 @@ validate_commandline (const char *key,
 
   return TRUE;
 }
+static gboolean
```
hfiguiere, line 910: `nit
```suggestion

static gboolean
```

src/xdp-utils.c
```diff
@@ -397,8 +397,11 @@ xdp_app_info_rewrite_commandline (XdpAppInfo *app_info,
           for (i = 1; commandline[i]; i++)
             g_ptr_array_add (args, maybe_quote (commandline[i], quote_escape));
         }
-      else
+      else if (quote_escape) {
         g_ptr_array_add (args, g_shell_quote (app_info->id));
+      } else {
+        g_ptr_array_add (args, g_strdup (app_info->id));
+      }
```
hfiguiere, line 404: the indentation style doesn't match

src/background.c
```diff
@@ -782,14 +799,16 @@ handle_request_background_in_thread_func (GTask *task,
...
 
   commandline = xdp_app_info_rewrite_commandline (request->app_info, autostart_exec,
                                                   FALSE /* don't quote escape */);
   if (commandline == NULL)
     {
-      g_debug ("Autostart not supported for: %s", app_id);
+      g_debug ("Autostart and schedualed autostart not supported for: %s", app_id);
     }
-  else if (!xdp_dbus_impl_background_call_enable_autostart_sync (background_impl,
+  else {
```
hfiguiere, line 810: the indentaton style doesn't match

calebccff:
imho the design of this portal could have a huge knock-on effect for years to come as apps adopt it, I have a few concerns with the proposal here:

1. Requiring that apps provide precise times will make it impossible to later optimise them by grouping tasks together as we have no way to know how precise the specified time actually needs to be (e.g. if your email and RSS app both want to update every 30 minutes, but don't care when, we should be able to have the device wake up just once every 30 minutes to handle this rather than twice).
2. Not having a way to provide hints about what kind of work the app wants to do similarly makes future optimisation work harder to do without breaking apps.

Obviously a lot of these potential future optimisations require a lot of work across the entire userspace stack, but I think if we take some steps to set ourselves up for that it will help a lot as we move more in this direction.

Android has over a decades worth of theory-in-practise on this subject, so I think we should at the very least see what they're doing and make sure that we don't repeat the same mistakes.

The modern approach is WorkManager: https://developer.android.com/develop/background-work/background-tasks/persistent#workmanager-features

It has a lot of dynamic options for scheduling and prioritising work, as well as handling things like retrying work that failed, specifying requirements like network access, or that some tasks should only be run on unrestricted networks, or not run when on low battery, etc, etc...

The Doze overview page is also really useful here: https://developer.android.com/training/monitoring-device-state/doze-standby

If the scope of this draft PR is *just* to handle things like GNOME clocks to let them wake up the system, then I think we should be explicit in the API: it should allow scheduling one or more alarms that will wake up the device and launch the app. Apps that are not literally alarm apps should not be using this API.

If we want to handle stuff like email/rss sync, we should have an API that allows these apps to properly express their intent so that we can optimise them in the future (and be clear about what capabilities are offered, to-the-minute accuracy probably not being one of them). Otherwise we risk making it nearly impossible to design a system like Doze without breaking apps, as well as potentially ending up in a situation where your phone screen turns on every 15 minutes because the RSS reader wants to check for updates...

The Android "`setAlarmClock()`" API definitely tells a story here: https://developer.android.com/reference/android/app/AlarmManager#setAlarmClock(android.app.AlarmManager.AlarmClockInfo,%20android.app.PendingIntent)

swick:
@jsparber Are you still interested in pushing this forward?

Mikenux:
@swick: Didn't you forget https://github.com/flatpak/xdg-desktop-portal/discussions/1385 ?

@jsparber: I have a new concept (UX/UI), which can (logically) be applied generally through the use of indices (needed for the user interface and for handling exceptions related to the device state). However, the idea is recent and therefore not yet well developed, especially since I am currently writing about write access permission.

jsparber:
> @jsparber Are you still interested in pushing this forward?

I think we can close this for now, since I won't have time in the near future anyway.

Issue #1102, opened 2023-09-15
Mechanism to retain compatibility across x-d-p 1.16 -> 1.17


smcv:
With the move from x-d-p 1.16 to 1.17, users of non-integrated, self-assembled or non-portal-aware desktop environments will find that Flatpak/Snap apps can no longer find a suitable portal backend (#1077 is an example of this). Prior to 1.17, users of these environments were implicitly relying on the fact that distro package dependencies would pull in at least one portal backend (in practice x-d-p-gtk), and x-d-p would run it. As a temporary stopgap I've patched in a hard-coded fallback to x-d-p-gtk in Debian's x-d-p, at a lower priority than trying UseIn; but that isn't going to help other distros, and @ebassi indicated in #1077 that this is unlikely to be accepted upstream.

I am not happy about having that hard-coded fallback, but I'm also not happy about the prospect of knowingly making desktop environments and desktop-environment-like things regress, because I expect that as soon as I remove the fallback, I'll get high-severity bug reports asking "why can't I just", which are relatively likely to be accompanied by accusations of incompetence or malfeasance.

I have been trying to get desktop environments to (set `XDG_CURRENT_DESKTOP` and) add a `${DESKTOP}-portals.conf`, either upstream ([GNOME](https://gitlab.gnome.org/GNOME/gnome-session/-/issues/114), [GNOME Flashback](https://gitlab.gnome.org/GNOME/gnome-flashback/-/issues/91), [Mate](https://github.com/mate-desktop/mate-desktop/issues/574), [Cinnamon](https://github.com/linuxmint/cinnamon/issues/11857), [Xfce](https://gitlab.xfce.org/xfce/xfce4-session/-/issues/181)) or [downstream](https://udd.debian.org/cgi-bin/bts-usertags.cgi?user=xdg-desktop-portal%40packages.debian.org&tag=portals.conf). Actively advocating this in a selection of major desktop environments has already been quite time consuming, so I am not going to be able to do this for all of the desktop environments that I don't use, and certainly not for the long tail of window-managers-that-act-like-a-desktop-environment. I've also had some push-back from window manager maintainers who [don't want to be responsible for choosing portals for their users](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1050911).

So I'm asking here: can we do something better/more-compatible in x-d-p that will "do what I mean" for users of less-major desktop environments, without losing the benefits of `portals.conf(5)`?

Some options that spring to mind:

[...]

PR #1103, opened 2023-09-15
Add fallback configuration for what to do if no portal is found via UseIn

smcv:
See #1102. This is an untested implementation of option 2 (but with a slightly different filename, `fallback.conf` might be better).

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -541,6 +541,14 @@ portal_impl_matches_config (const PortalConfig *config,
   return FALSE;
 }
 
+static void
+warn_please_use_portals_conf (void)
+{
+  g_warning_once ("The preferred method to match portal implementations "
+                  "to desktop environments is to use the portals.conf(5) "
+                  "configuration file");
+}
+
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
@smcv how bad would it be to release xdg-desktop-portal without this? Should the issue be considered a release blocker?

smcv:
> how bad would it be to release xdg-desktop-portal without this? Should the issue be considered a release blocker?

This specific PR is known-broken (I thought it compiled successfully on my machine, but maybe it never did, and it certainly fails CI) so it's definitely not a release blocker. I pushed it here as a proof-of-concept to give maintainers an idea of how much code one of the options from #1102 would end up adding.

For #1102 I would like to have at least some sort of idea of what direction we're going in, so that I can use that as an input when making decisions in Debian. Please can we discuss that on #1102 rather than here, though?

GeorgesStavracas:
Obsoleted by https://github.com/flatpak/xdg-desktop-portal/pull/1199

PR #1106, opened 2023-09-18
Accept portals by default in find_all_portal

kennylevinsen:
Support for configuring the Settings interface accidentally made configuration required. Restore the default accept behavior for the settings portal by distinguishing between not being configured, and not being allowed by the configuration.

See comments in https://github.com/flatpak/xdg-desktop-portal/issues/1017.

### DIFF
--- a/doc/portals-conf.rst
+++ b/doc/portals-conf.rst
@@ -100,6 +100,23 @@ special values ``none`` and ``*`` can be used:
 
   Use the first portal implementation found, in lexicographical order.
 
+PORTAL LOADING
+--------------
+
+There are two types of interfaces: Exclusive interfaces, and unified
+interfaces. Exclusive interfaces are provided by a single portal. Unified
+interfaces are provided my multiple portals at once.
+
+At the current time, the only unified interface is the Settings portal.
+
+A portal is only loaded for an exclusive interface if configured, by either
+being a member of the default list or the particular interface list.
+
+A portal is loaded for a unified interface if the portal is a member of the
+particular interface list, or by default if no configuration is provided for
+that particular interface. As such, portals do not need to be in the default
+list to be used for unified interfaces.
+
 EXAMPLE
 -------
 
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -29,6 +29,13 @@
 #include <glib.h>
 #include <gio/gio.h>
 
+typedef enum
+{
+  MATCH_POSITIVE,
+  MATCH_NEGATIVE,
+  NO_MATCH,
+} MatchResult;
+
 typedef struct _PortalInterface {
   /* dbus_name is NULL if this is the default */
   char *dbus_name;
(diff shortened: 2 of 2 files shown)


smcv:
cc @ebassi (author of #1018)

I think this would be a lot easier to review if it came with a change to `doc/portals-conf.rst` describing what the intended behaviour is. An issue report with "steps to reproduce / expected result / actual result" would probably also be helpful to understand what it is that you are intending to change here.

Is the intended behaviour change here to say that as a special exception, `default=...` does not configure Settings, and instead, Settings implicitly defaults to combining the settings provided by every backend that has a Settings implementation - so it will load (GNOME|KDE|Cinnamon|...) settings, regardless of whether you are on GNOME, KDE, Cinnamon or something else, unless the desktop environment has explicitly configured `org.freedesktop.impl.portal.Settings=foo;bar;baz;`?

If that's what is happening here, it seems to me that it's going to reintroduce the issue described in https://github.com/flatpak/xdg-desktop-portal/issues/1017#issuecomment-1633972386, which is a genuine problem: we have seen x-d-p-gnome breaking apps (making them take ~ 30 seconds to start up) in non-GNOME desktops like Cinnamon, and we have also seen x-d-p-xapp breaking apps (in the equivalent way) in non-(Cinnamon|MATE|Xfce) desktops like GNOME. The fact that those two implementations independently had an equivalent problem suggests to me that other portal implementations have that issue or will have it in future. Do you have a plan for how to avoid this PR re-introducing that problem?

kennylevinsen:
> I think this would be a lot easier to review if it came with a change to doc/portals-conf.rst describing what the intended behaviour is. 

I added something, but it required me to name the different portal behaviors - let the bikeshedding begin!

---

A bit of elaboration: Before #1018 no configuration had any effect on portals for the settings interface. The intention was to allow explicit configuration when desired, but not to break any existing behaviors.

The change unfortunately ended up making some configuration (either default= or explicit) required. While I do not find this bad on its own (more consistent in my opinion), breaking changes were not at all intended. This change makes it so that explicit configuration is possible, but making it so that no explicit configuration is the same as `*` and thus behaving as it did before.

To clarify what behavior is expected:

## #1018  behavior
### Expected positive matches for gtk portal

```
# Test 1: Default
[preferred]
default=gtk

# Test 2: Explicit config
[preferred]
org.freedesktop.impl.portal.Settings=gtk
```

### Expected negative matches for gtk portal

Well, anything else.

## #1106  behavior
### Expected positive matches for gtk portal

```
# Test 1: Explicit config
[preferred]
org.freedesktop.impl.portal.Settings=gtk

# Test 2: Explicit wildcard
[preferred]
org.freedesktop.impl.portal.Settings=*

# Test 3: File with no Settings interface config
[preferred]
default=kde

# Test 4: No file
# <crickets>
```

### Expected negative matches for gtk portal

```
# Test 1: Explicit empty list
[preferred]
org.freedesktop.impl.portal.Settings=

# Test 2: Explicit none
[preferred]
org.freedesktop.impl.portal.Settings=none

# Test 3: Explicit other portal
[preferred]
org.freedesktop.impl.portal.Settings=kde
```

smcv:
> An issue report with "steps to reproduce / expected result / actual result" would probably also be helpful to understand what it is that you are intending to change here.

Is #1112 that missing issue report?

smcv:
See https://github.com/flatpak/xdg-desktop-portal/issues/1112#issuecomment-1733683797 for some of the options for resolving #1112.

> it seems to me that it's going to reintroduce the issue described in https://github.com/flatpak/xdg-desktop-portal/issues/1017#issuecomment-1633972386, which is a genuine problem

I am still concerned that this solution to #1112 risks re-introducing that. Perhaps this is the right route, I don't know - but we should look at the options before merging anything.

GeorgesStavracas:
As per https://github.com/flatpak/xdg-desktop-portal/issues/1112#issuecomment-1867173539, let's close this one

Issue #1041, opened 2023-06-20
Return value for org.freedesktop.impl.portal.Access.AccessDialog


WhyNotHugo:
I've written my own implementation of an Access Portal, but want to ensure that my return values are correct.

The documentation for the return values merely repeats the data types, but does not address the _meaning_ of these arguments:

https://github.com/flatpak/xdg-desktop-portal/blob/dea5cac995300e98e7509188f2a7c6011c894f41/data/org.freedesktop.impl.portal.Access.xml#L40-L41

As for the `response` argument, I've been able to figure out that `0` is "grant access" and non-zero is "deny access". Is this correct? Or do specific non-zero responses have a specific meaning? Does any value represent "grant access only once"?

Regarding the `results` argument: is it just one (or more) of the provided `options/choices`?

I can send a PR with improvements to the documentation with these details, but I'd rather confirm that my guesses are correct first.

Marukesu:
> As for the response argument, I've been able to figure out that 0 is "grant access" and non-zero is "deny access". Is this correct? Or do specific non-zero responses have a specific meaning? Does any value represent "grant access only once"?

there's a enum with possible/known values in request.h that matches what the gtk backend does. AFAIK, the portals calling access only check if 0 was returned.

> Regarding the results argument: is it just one (or more) of the provided options/choices?

it's all passed choices, with the final values.

WhyNotHugo:
https://github.com/flatpak/xdg-desktop-portal/pull/1114

PR #1114, opened 2023-09-25
Document the return value of AccessDialog()

WhyNotHugo:
Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1041

### DIFF
--- a/data/org.freedesktop.impl.portal.Access.xml
+++ b/data/org.freedesktop.impl.portal.Access.xml
@@ -37,7 +37,7 @@
         @subtitle: Subtitle for the dialog
         @body: Body text, may be ""
         @options: Vardict with optional further information
-        @response: Numeric response
+        @response: Numeric response. The values allowed match the values allowed for #org.freedesktop.portal.Request::Response signal.
         @results: Vardict with the results of the call
 
         Presents a "deny/grant" question to the user.


data/org.freedesktop.impl.portal.Access.xml
```diff
@@ -37,7 +37,8 @@
         @subtitle: Subtitle for the dialog
         @body: Body text, may be ""
         @options: Vardict with optional further information
-        @response: Numeric response
+        @response: Numeric response. A value of 0 indicates that the user has
+          agreed to grant permission. Any other value is treated as a refusal.
```
whot, line 41: (almost) all the other `impl.portal` just state `Numeric response` too - it'd be nice to update all those, and ideally to something like: "The value returned as the response argument in the #org.freedesktop.portal.Request::Response for the corresponding Request". This way you get the automatic link to where the documentation for the return values is.

whot (changes requested):

data/org.freedesktop.impl.portal.Access.xml
WhyNotHugo: I'm not sure I understand. What is _"the #org.freedesktop.portal.Request::Response for the corresponding Request"_ in this case?

whot: Huh, indeed, this is one (of the few?) that doesn't have an immediately corresponding request in the client API. I think in that case something like "The value range of the response matches the values allowed for #org.freedesktop.portal.Request::Response" for this one. Still would be nice to update the others though.

WhyNotHugo: Updated.

WhyNotHugo:
It's not rendering correctly; I'm not entirely sure why.

WhyNotHugo:
Looks like the `#org.freedesktop.portal.Request::Response` doesn't actually add a link tho.

I'm not entirely sure what format this is, so I'm not sure where to look for its documentation. How do I add an actual link to the anchor?

WhyNotHugo:
Oh, there's an example right above it :facepalm: 

I tried this:

```
diff --git a/data/org.freedesktop.impl.portal.Access.xml b/data/org.freedesktop.impl.portal.Access.xml
index 29a9f70..5759e02 100644
--- a/data/org.freedesktop.impl.portal.Access.xml
+++ b/data/org.freedesktop.impl.portal.Access.xml
@@ -37,7 +37,7 @@
         @subtitle: Subtitle for the dialog
         @body: Body text, may be ""
-        @response: Numeric response. The values allowed match the values allowed for #org.freedesktop.portal.Request::Response signal.
+        @response: Numeric response. The values allowed match the values allowed for <link linkend="#gdbus-signal-org-freedesktop-portal-Request.Response">#org.freedesktop.portal.Request::Response</link> signal.
         @results: Vardict with the results of the call
 
         Presents a "deny/grant" question to the user.
```

But it fails to build:

```
ninja: job failed: /usr/bin/xmlto xhtml-nochunks -o doc -m ../doc/xmlto-config.xsl --searchpath /home/hugo/src/github.com/flatpak/xdg-desktop-portal/build/src doc/portal-docs.xml
xmlto: /home/hugo/src/github.com/flatpak/xdg-desktop-portal/build/doc/portal-docs.xml does not validate (status 3)
xmlto: Fix document syntax or use --skip-validation option
/home/hugo/src/github.com/flatpak/xdg-desktop-portal/build/doc/portal-docs.xml:133: element link: validity error : Syntax of value for attribute linkend of link is not valid
/home/hugo/src/github.com/flatpak/xdg-desktop-portal/build/doc/portal-docs.xml:133: element link: validity error : IDREF attribute linkend references an unknown ID "#gdbus-signal-org-freedesktop-portal-Request.Response"
Document /home/hugo/src/github.com/flatpak/xdg-desktop-portal/build/doc/portal-docs.xml does not validate
ninja: subcommand failed
```

I'm not sure how to work around this. Any insight is welcome.

GeorgesStavracas:
We're reworking the docs system on #1157 but this is good to have already

WhyNotHugo:
Sphinx sound like a great choice :+1:

Issue #1124, opened 2023-10-04
Transient screen cast stream restoration no longer works


aiddya:
Screen cast stream restoration with persist mode as transient no longer works with xdg-desktop-portal 1.18. It works if the persist mode is until explicitly revoked.

### Steps to reproduce
1. Open this [WebRTC test page](https://mozilla.github.io/webrtc-landing/gum_test.html) with Chromium on a distro with xdg-desktop-portal 1.18
2. Click on "screen capture" and switch to "Window" or "Entire screen" tabs.
3. Select a source on portal dialog and wait for the thumbnail to appear
4. Click on "Share"

### Expected behavior
Shared screen/window is displayed.

### Actual behavior
The portal dialog appears again.

DBus logs show that the application is passing restore token correctly and the portal implementation is returning restore data. The bug is either in storing the restore token along with restore data or while retrieving it.

cc @jadahl

aiddya:
It appears that the bug is a result of deleting all transient restore tokens once the screen cast session is closed. However, the documentation for transient persist mode says that "Permissions persist as long as the application is running".

grulja:
I can reproduce. I expect this is related to https://github.com/flatpak/xdg-desktop-portal/pull/1004.

GeorgesStavracas:
This is not reproducible on Firefox. Are you sure this is not a Chromium issue?

GeorgesStavracas:
Now that I test this, I think I've seen that on Chromium for a long time... certainly before #1004.

grulja:
> This is not reproducible on Firefox. Are you sure this is not a Chromium issue?

Firefox works differently and doesn't rely on stream restoration. I've seen this issue before with Google Meet, which might be a bug in Chromium (on my TODO list), but for example the WebRTC test page has been working all the time. There was no Chromium change to screen sharing for a while.

aiddya:
The flow in Chromium is that it starts a screen cast session to show thumbnails on the screen/window picker and it restores the stream in a new session to show the stream on the web page.  Firefox does not show thumbnails, so it doesn't use stream restoration.

[...]

PR #1125, opened 2023-10-04
Delete transient permissions when peer dies

GeorgesStavracas:
Not when the session is closed. See patches.

Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1124

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -24,15 +24,6 @@
 
 enum
 {
-  INTERNAL_CLOSED,
-
-  N_SIGNALS
-};
-
-static guint signals[N_SIGNALS];
-
-enum
-{
   PROP_0,
 
   PROP_SENDER,
@@ -182,8 +173,6 @@ session_close (Session *session,
     return;
 
   SESSION_GET_CLASS (session)->close (session);
-
-  g_signal_emit (session, signals[INTERNAL_CLOSED], 0);
 
   if (notify_closed)
     {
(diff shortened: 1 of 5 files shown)


src/session.c
```diff
@@ -270,6 +271,8 @@ close_sessions_in_thread_func (GTask *task,
       session_close (session, FALSE);
     }
```
jadahl, line 261: Perhaps this function (`close_sessions_in_thread_func()`) should get a new name now that it does more things than closing sessions. Alternatively, add a dedicated step in `peer_died_cb()` after the other two.

GeorgesStavracas: Went with the latter (the first version of this patch did exactly than before I moved it to `close_sessions_in_thread_func()`, it felt natural to move it back to `peer_died_cb()`)

src/xdg-desktop-portal.c
```diff
@@ -229,6 +230,7 @@ peer_died_cb (const char *name)
 {
   close_requests_for_sender (name);
   close_sessions_for_sender (name);
+  xdp_session_persistence_delete_transient_permissions_for_sender (sender);
```
jadahl, line 233: s/sender/name/

jadahl (changes requested):

src/xdg-desktop-portal.c
GeorgesStavracas: Of course

PR #1132, opened 2023-10-09
Fix two issues with the spelling of PipeWire

whot:
One is just a documentation issue, the second one *should*'ve produced warnings on every use of  `Camera.OpenPipeWireRemote` so I'm a bit worried why this hasn't shown up anywhere.


jadahl:
I checked, and it's `method_needs_request()` which correctly returns `FALSE` for `OpenPipeWireRemote` (spelled correctly). When that function returns `FALSE` we never get into the "which-arg-is-the-token" function.

PR #1136, opened 2023-10-10
doc: Fix the InputCapture barrier examples

whot:
A barrier is either at the top or at the left edge, not both. This means the actual barriers need to go from zone->x to zone->width - 1, the next pixel (zone->x + zone->width) is the first pixel of the adjacent zone if any.

We could allow this in the spec but it makes implementations a lot harder: on a two-screen setup the edge pixel would need special casing, the barrier 0-1920 and 1920-3840 are both valid on either screen despite having a single-pixel overlap. An easier approach is to make sure each barrier only covers the pixels intended, i.e. 0-1919 and 1920 to 3839 in this case.

cc @jadahl

### DIFF
--- a/data/org.freedesktop.portal.InputCapture.xml
+++ b/data/org.freedesktop.portal.InputCapture.xml
@@ -205,8 +205,11 @@
         (from compositor-determined input devices) are sent to the application
         via the transport layer.
 
-        Pointer barriers are situated on the top or left of their respective pixels and a
-        pointer barrier must be situated at the outside boundary of the union of all zones.
+        Pointer barriers are situated on the top (for horizontal barriers) or left
+        (for vertical barriers) edge of their respective pixels and their width or height
+        is inclusive each pixel's width or height. In other words, a barrier spanning
+        x1=0, x2=1 is exactly two pixels wide. A pointer barrier must be situated at the
+        outside boundary of the union of all zones.
         A pointer barrier must be fully contained within one zone.
 
         For example, consider two zones of size 1920x1080 with offsets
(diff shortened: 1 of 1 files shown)


data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -214,12 +215,12 @@
...
-        <listitem><para>top edge of right screen: <literal>x1=1920,y1=0,x2=3840,y1=0</literal></para></listitem>
-        <listitem><para>bottom edge of right screen: <literal>x1=1920,y1=1080,x2=3840,y1=1080</literal></para></listitem>
-        <listitem><para>left edge of left screen: <literal>x1=0,y1=0,x2=0,y1=1920</literal></para></listitem>
-        <listitem><para>right edge of right screen: <literal>x1=3840,y1=0,x2=3840,y1=1920</literal></para></listitem>
+        <listitem><para>top edge of left screen: <literal>x1=0,y1=0,x2=1919,y1=0</literal></para></listitem>
+        <listitem><para>bottom edge of left screen: <literal>x1=0,y1=1080,x2=1919,y1=1080</literal></para></listitem>
+        <listitem><para>top edge of right screen: <literal>x1=1920,y1=0,x2=3839,y1=0</literal></para></listitem>
+        <listitem><para>bottom edge of right screen: <literal>x1=1920,y1=1080,x2=3839,y1=1080</literal></para></listitem>
+        <listitem><para>left edge of left screen: <literal>x1=0,y1=0,x2=0,y1=1919</literal></para></listitem>
+        <listitem><para>right edge of right screen: <literal>x1=3840,y1=0,x2=3840,y1=1919</literal></para></listitem>
```
jadahl, line 223: Trying to put this into ascii graphics to understand what you mean...

Each pixel looks like this:

```
X---+
|   |
+---+
```

where the X corresponds to the x/y position, and any barrier will be *inclusive* of the pixel width or height depending on the orientation of the barrier in question.

So if we have the following area between two monitors:

```
A---B---+
|   |   |
+---+---+
```

If we assume the monitors both pixel A and B are on are 1920x1080, and we're on the top pixel row, this change means that the example changes the A from being the same as B, i.e. 1920-1 = 1919. Right?

Do we have that detail spelled out somewhere?

whot: > where the X corresponds to the x/y position, and any barrier will be inclusive of the pixel width or height depending on the orientation of the barrier in question.

correct

> If we assume the monitors both pixel A and B are on are 1920x1080, and we're on the top pixel row, this change means that the example changes the A from being the same as B, i.e. 1920-1 = 1919. Right?

I'm struggling to parse this sentence but: a barrier is inclusive all pixel widths, so a barrier of `x1=0,x2=1919` is exactly 1920 pixels wide. A barrier going `x1=0,x2=1920` would be one pixel into the zone B and is thus invalid.

So I think we agree.

I've fixed the example (had a y=1920 copy/paste) and added an example, I hope that's a bit better now, it's not spelled out anywhere else though.

whot: addendum just to make this more clear. If we zoom into your `B` pixel and illustrate how wide a barrier `b1` with `x2=1919` vs  `b2` with `x2=1920` is:
```
==b1============
==b2====================
-------+--------B-------+-------
 1918  |  1919  |  1920 | 1921
```

`b2` overlaps into the next zone and is thus invalid, `b1` is the correct one.

jadahl: The new docs makes it very clear now, thanks!

PR #1138, opened 2023-10-11
document-portal-fuse: reduce the scope of a pointer

hfiguiere:
This will make sure the pointer isn't used improperly.


jadahl:
lgtm

PR #1146, opened 2023-10-11
docs: Document Since for FileChooser

A6GibKm:
These options were added since version 3 of this interface.


GeorgesStavracas:
Why are you bumping the interface version to 4?

A6GibKm:
Because they are new methods.

GeorgesStavracas:
Please document it in the commit message in details then

A6GibKm:
Done.

GeorgesStavracas (approved):

Issue #689, opened 2022-01-13
xdg-document-portal keeps files open indefinitely, preventing unmounting of removable media


nurupo:
I'm trying out [Chromium flatpak](https://github.com/flathub/org.chromium.Chromium). I downloaded a few files using Chromium to a removable drive and then attempted to unmount it, but the system kept telling me that it can't do that, that the device is busy. I `lsof` the mount point to see what is using the fs and lo-and-behold:

```
$ sudo lsof /media/[REDACTED]
lsof: WARNING: can't stat() fuse.portal file system /run/user/1000/doc
      Output information may be incomplete.
COMMAND       PID   USER   FD   TYPE DEVICE  SIZE/OFF    NODE NAME
xdg-docum 1564418 nurupo   19u   REG  254,1    151987 8912913 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   20u   REG  254,1    826201 8912914 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   21u   REG  254,1   1873749 8912915 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   22u   REG  254,1 170303175 8912916 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   23u   REG  254,1   2276729 8912917 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   24u   REG  254,1 170130747 8912918 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   27u   REG  254,1    791113 8912921 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   28u   REG  254,1    308858 8912922 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   29u   REG  254,1   9789825 8912925 /media/[REDACTED] (deleted)
xdg-docum 1564418 nurupo   31u   REG  254,1    324365 8912924 /media/[REDACTED] (deleted)
```

The files were downloaded hours ago, deleted hours ago too (notice the "deleted" after each file), and Chromium was closed long time ago, but xdg-document-portal still has them open and is preventing me from unmounting the drive, forcing me to kill it. What gives? I don't want having to fish its PID and kill it every time I want to unmount my drive. That's 0/10 usability.

[...]

PR #1159, opened 2023-10-18
xdg-document-portal: Close files as needed

hfiguiere:
- Issue #689

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -1,3 +1,24 @@
+/*
+ * Copyright © 2018 Red Hat, Inc
+ * Copyright © 2023 GNOME Foundation Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Contributors:
+ *      Hubert Figuière <hub@figuiere.net>
+ */
+
 #include "config.h"
 
 #define FUSE_USE_VERSION 35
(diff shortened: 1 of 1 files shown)


hfiguiere:
This needs much more testing. (and maybe a few more bits of cleanup)

document-portal/document-portal-fuse.c
```diff
@@ -455,6 +459,68 @@ xdp_physical_inode_ref (XdpPhysicalInode *inode)
...
+  else
+    {
+//      G_LOCK (physical_inodes);
+
+      if (!g_atomic_int_compare_and_exchange ((int *) &inode->use_count, old_use, old_use + 1))
+        {
+//          G_UNLOCK (physical_inodes);
+          goto retry_atomic_decrement1;
+        }
+      inode->fd = fd;
```
GeorgesStavracas, line 485: Doesn't it leak ` inode->fd` here?

document-portal/document-portal-fuse.c
```diff
@@ -167,12 +167,15 @@ G_LOCK_DEFINE (domain_inodes);
 
 typedef struct {
   gint ref_count; /* atomic */
+  gint use_count; /* atomic */
```
GeorgesStavracas, line 170: How's the use count any different from the refcount? Can you please clarify why is it necessary to track them separatedly?

document-portal/document-portal-fuse.c
hfiguiere: we could add a check that it's -1 and if not close it. but this code path it is supposed to be closed.

document-portal/document-portal-fuse.c
hfiguiere: because it gets kept alive and the fd gets "revived". the _use and _release calls are not matching the _ref and _unref.

hfiguiere: see https://github.com/flatpak/xdg-desktop-portal/pull/1159/files#diff-2f05bdfc1f7bed1308fa7848357777aad2cadd1d6133d92678296ef2697c4e49R2122

(I have an updated version of that patch with more consistent formatting but the same semantics)

hfiguiere:
My biggest fear with this PR is to have missed something and create conditions that would cause data loss. 

@alexlarsson I'm sorry to ping you, but it feel like you would be the best person to help reviewing this change. Or even guidance to thoroughly test the corner cases.

document-portal/document-portal-fuse.c
```diff
@@ -165,14 +165,27 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (XdpDomain, xdp_domain_unref)
 
 G_LOCK_DEFINE (domain_inodes);
 
+/* The reference count and use count are different. The latter still
+ * allow keeping track of the file without keeping it open, which otherwise
+ * keep the files around indefinitely.
+ * See https://github.com/flatpak/xdg-desktop-portal/issues/689
+ */
```
orowith2os, line 172: Why not make a weak reference where you'd use the `use_count`? Would that solve anything? Ideally it wouldn't keep the file open, so it can still be freed up when nothing else references it.

alexlarsson:
This change seems very wrong to me. I mean, what is even the point of having a XdgPhysicalInode in memory if the fd is closed, as that is all it tracks? 

Anyway, this seems non-threadsafe, as there is no guarantee that after a call to ensure_docdir_inode() that physical->fd is open. The atomics guarantee that the memory lives, sure, but the fd can be closed at any time by some other thread, so you can't use the PhysicalInode for anything.

In addition, the entire reason we're keeping the O_PATH fd open in the XdgPhysicalInode is to avoid issues if an external party modifies the backing filesystem (i.e. not via the fuse fs). If we close the fd when the kernel drops the inode from the cache, then the entire point of this feature (as documented in the comment at the top of the file) is lost.

As for the underlying issue, the portal fuse implementation should never keep around PhysicalInode references to something that is not in use. Like for instance, a XdpPhysicalInode is only being held open by an XdpInode, and those should only be kept open if the kernel is referencing it via the fuse upcall APIs.

So, given that we regularly flush the dcaches for the fuse mount, we should only keep a inode open if there is some long-term kernel reference keeping an inode in memory, such as a process keeping a file open or having it as current working directoy. As that seems to not be the case here, we must have some kind of leak, and the real fix is to find the cause of the leak and fix that. What we need to do is make a minimal reproducer and track what reference is keeping the inode alive.

alexlarsson:
I tried the reproducer in https://github.com/celluloid-player/celluloid/issues/778, and even when the movie player has exited, xdg-document-portal keeps the video open, as seen by lsof. 

However, when I flush the kernel inode cache (sudo sh -c 'echo 2 > /proc/sys/vm/drop_caches') it is released, because the kernel sends a bunch of FORGET events for the inodes.

So, it seems like the ownership tracking in the kernel is working, but maybe our workaround that is supposed to keep the kernel from caching these inodes forever is not working?

alexlarsson:
This seems like a better fix: https://github.com/flatpak/xdg-desktop-portal/pull/1190

PR #1160, opened 2023-10-18
global-shortcuts: Convert timestamp to

### LINKED ISSUE #1054: Global shortcuts timestamps incompatible with input timestamps

davidedmundson:
On X11 and wayland timestamps are using the monotonic clock for input events, global shortcuts tie in closely with this. This patch brings it in line using the same wording as wayland events.

This is technically a breakage in the spec, but pragmatically brings it in line with the current real world implementations.

Fixes #1054

### DIFF
--- a/data/org.freedesktop.impl.portal.GlobalShortcuts.xml
+++ b/data/org.freedesktop.impl.portal.GlobalShortcuts.xml
@@ -124,7 +124,7 @@
         Activated:
         @session_handle: Session that requested the shortcut
         @shortcut_id: the application-provided ID for the notification
-        @timestamp: The timestamp, as seconds and microseconds since the Unix epoch.
+        @timestamp: the time of activation with millisecond granularity, with an undefined base.
         @options: Vardict with optional further information
 
         Emitted when a shortcut is activated.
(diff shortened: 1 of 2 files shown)


GeorgesStavracas (approved):
Thanks! LGTM, but perhaps @jadahl wants to ACK as well since the original diagnose of this spec issue came from them?

davidedmundson:
> xdg-desktop-portal-kde uses microseconds, not milliseconds, so might as well fix that too while at it. At least that was the conclusion I made in #1054.

xdg-desktop-portal-kde uses milliseconds, that was your conclusion in the spec. It was out of line with the spec at the time, but your proposal was to match the timestamps used by X11 and wayland, which makes the existing implementation in line again.

PR #1161, opened 2023-10-18
Updated website sources URL

kacperpaczos:
The new link is more accurate, I was inspired by the fact that I couldn't find the page at first,


GeorgesStavracas (approved):
Perfect, thank you

PR #1162, opened 2023-10-18
Added information about an application

kacperpaczos:
that allows you to check the compatibility of portals in your system.


GeorgesStavracas:
Well, I don't think this is worthy adding, IMHO.

Not because the app isn't good or anything, but because the portals project is supposed to be slightly more agnostic than that, and not promote individual apps over others. Sorry!

Keep the patches coming though, these small adjustments are fantastic

PR #1168, opened 2023-10-23
global-shortcuts: Emit signals to session correctly

GeorgesStavracas:
The appropriate object path is the session's object path, represented by its id.


GeorgesStavracas:
@whot @jadahl something that would be good to figure out, is why the Python test didn't catch this, and how can it be adapted to test for this issue?

GeorgesStavracas:
Welp, that explains it - the Python test is apparently testing the wrong behaviour (or at least, as far as I understood, this is the wrong behaviour...)

whot:
I can't find where I discussed this with @jadahl but this behaviour seems to be correct, it's a legacy of the portal that things happen not on the session object itself but always on the xdg-desktop-portal object. The session only ever has `Close`, all other signals are sent on the portal but have the session handle as argument.

Switching this will break things because everything else expects the signals to happen on the main portal object.

Oh, and the `Session` shouldn't even have that interface.

jadahl:
It's a legacy that e.g. the remote desktop isn't also a an interface on the session, because one first creates the session, and only later associates "things" with it (e.g. remote desktop, screen cast, clipboard).

The session itself is still a D-Bus object on some path, since it has a method (`Close()`), so with that in mind, the sane thing would be that the `Closed` signal happens on that object as well.

jadahl:
I was confusing myself, I somehow thought this was about the `org.freedesktop.Session.Closed` signal. In the case of `GlobalShortcuts` and any other signal that is similar to it, the signal must be emitted on the same object that the interface is on, meaning `/org/freedesktop/porta/desktop`.

PR #1169, opened 2023-10-23
utils: Handle host apps in app_info_map_pids

swick:
Mapping pids/tids is only necessary when apps are running inside another pid namespace. Apps on the host usually do not. If they put parts of themselves into one, it's their responsibility to do the mapping.

Instead of having to make sure that the app is not a host app before calling this function, handle host apps trivially by not mapping anything and returning success.

---

This change should only affect the gamemode portal when called from a host app.

### DIFF
--- a/src/realtime.c
+++ b/src/realtime.c
@@ -56,23 +56,21 @@ G_DEFINE_TYPE_WITH_CODE (Realtime, realtime, XDP_DBUS_TYPE_REALTIME_SKELETON,
                                                 realtime_iface_init));
 
 static gboolean
-map_pid_if_needed (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
+map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
 {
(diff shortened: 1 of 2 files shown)


src/realtime.c
```diff
@@ -56,22 +56,20 @@ G_DEFINE_TYPE_WITH_CODE (Realtime, realtime, XDP_DBUS_TYPE_REALTIME_SKELETON,
                                                 realtime_iface_init));
 
 static gboolean
-map_pid_if_needed (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
+map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
 {
-  if (!xdp_app_info_is_host (app_info))
+  if (!xdp_app_info_map_pids (app_info, pid, 1, error))
   {
```
jadahl, line 62: Could clean up the messed up indentation while at it.

swick: done

GeorgesStavracas (approved):
LGTM, thanks

PR #815, opened 2022-06-17
Accent color on settings portal

lleyton:
# Accent Colors

## Introduction

Accent colors provide a way for users to personalize their desktop in a simple, developer-friendly, and effective way. Throughout the community there has been a general interest in the inclusion of accent colors within apps and desktop environments. This proposal aims to standardize an accent color key on the Settings portal.

## Details

A new key on the Settings portal, `accent-color`, would be defined under the `org.freedesktop.appearance` namespace. The value of this key is a variant. In the case that an accent color is requested, the underlying value would be a struct containing three doubles, each being an RGB value in the range of [0,1]. Whereas, if the user has no preference or wants the application to use their own branding colors, the type will be a uint32 with the value 0, representing no preference. Clients may consume this value in order to style their user interface using the requested accent color. Clients should handle an unknown value in the same way they would handle a "no-preference" value.

## Implementation

A reference implementation of the client can be found in tauOS's [libhelium](https://github.com/tau-OS/tau-helium). I have also forked [xdg-desktop-portal-gnome](https://github.com/tau-OS/xdg-desktop-portal-gnome) adding support for the accent-color key.

## Signatories

- Lleyton "Lea" Gray, Fyra Labs
- Paulo "Lains" Galardi, Fyra Labs

## Acks

- [x] @alice-mkh (GNOME)
- [x] @JoshStrobl (Budgie)
- [x] @Sodivad (KDE)
- [x] @jackpot51 (Cosmic)
- [x] @danirabbit (Elementary)

PR #1175, opened 2023-10-27
High Contrast hint on settings portal

dominichayesferen:
# Introduction
High Contrast is a feature where, much as the name implies, the contrast of applications that support the feature is increased, whether it be adding more opacity to lines in applications, or in the rare case changing the entire palette of an application to be of the maximum possible colour contrast. Irregardless of implementation, this hint has existed for quite a long time in the UNIX space, as well as existing in most other Operating Systems, and this pull request gives this long-standing accessibility hint a standardised location for both existing and future uses of high contrast to come.

# Details
A new key on the settings portal, `high-contrast`, would be introduced into the `org.freedesktop.appearance` namescape. This key will use a boolean value.

The design of this key is based on every existing instance of a high contrast switch - in all platforms I know of, including those in the UNIX space who provide this value in a non-standardised manner, it has always been a simple boolean value which, if True, enables high contrast, but otherwise restores normal application styling if False.

This serves as a hint, just like its non-standardised variations, informing applications of the desire for contrast to be increased - applications are absolutely free to ignore it if, for example, they believe their default application appearance is of high enough contrast.

The hope for a standardised high contrast switch is that it will allow toolkits' existing high contrast options to be easily accessible from other Desktop Environments, such as GTK's high contrast option outside of GNOME, as well as encouraging programs, natively supported on XDG-compliant platforms, that technically have high contrast support programmed in, such as Mozilla Firefox and the Chromium Project, to expose this functionality on XDG-compliant platforms too.

# Acks

- [x] @snwh (GNOME)
- [x] @sonnyp (GNOME)
- [x] @JoshStrobl (Budgie)
- [x] @CarlSchwan (KDE)
- [x] @mmstick (Cosmic)
- [x] @danirabbit (Elementary)
- [x] @emilio (Gecko (Firefox's) Browser Engine)
- [x] @nothingneko (Helium)

# Implementations of High Contrast

- [ ] KDE Plasma: https://invent.kde.org/plasma/breeze/-/merge_requests/397
- GNOME, and all GTK-powered Desktop Environments: High Contrast boolean-switch in Settings
- Firefox: Currently only available on GNOME

Technically also exists in Chromium, although is currently inaccessible on Linux.

# Implementations using this standard

None yet.

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -46,6 +46,16 @@
       in the sRGB color space, in the range [0,1].
       Out-of-range RGB values should be treated as an unset accent color.
 
+    * ``org.freedesktop.appearance``  ``contrast`` (``u``)
+
+      Indicates the system's preferred contrast level.
+      Supported values are:
+
+      - ``0``: No preference (normal contrast)
+      - ``1``: Higher contrast
+
+      Unknown values should be treated as 0 (no preference).
+
     Implementations can provide other keys; they are entirely
     implementation details that are undocumented. If you are a
     toolkit and want to use this please open an issue.
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -46,6 +46,16 @@
       the sRGB color space, in the range [0,1]. Out-of-range RGB values should
       be treated as an unset accent color.
 
+    * ``org.freedesktop.appearance``  ``contrast`` (``u``)
+
+      Indicates the system's preferred contrast level.
+      Supported values are:
+
+        * ``0``: No preference (normal contrast)
+        * ``1``: Higher contrast
+
+      Unknown values should be treated as ``0`` (no preference).
+
     Implementations can provide other keys; they are entirely
     implementation details that are undocumented. If you are a
     toolkit and want to use this please open an issue.


GeorgesStavracas:
I think this is pretty uncontroversial as is, thanks for proposing it. Let's just check if anyone objects to this proposal.

GeorgesStavracas:
Just taking the list of ACKs from #815:

- [ ] @alice-mkh (GNOME)
- [ ] @JoshStrobl (Budgie)
- [ ] @Sodivad (KDE)
- [ ] @jackpot51 (Cosmic)
- [ ] @danirabbit (Elementary)

JoshStrobl:
Ack on the Budgie side. Seems like a pretty obvious key to me.

dominichayesferen:
I'll also ping @emilio in here in advance, as I remember them participating in the accent colour portal MR discussion, as a pseudo-rep of sorts for Gecko and thus presumably Firefox - would you be able to, by chance, be a, or alternatively get a, representative from Mozilla Firefox's maintainers in here by chance to also see if they suggest anything to add as part of a standardised high contrast hint?

As for Chromium, not sure about how to get in touch with getting a representative for their side, for the record.

lleyton:
I wonder if the namespace for the key should be something like `org.freedesktop.accessibility` instead of `org.freedesktop.appearance`. That's a minor detail though, everything else seems fine to me :p

dominichayesferen:
Follow up from lleyton's above comment:
If the general consensus is in favour of Lleyton's idea, should it be called "org.freedesktop.accessibility" or "org.freedesktop.a11y"?

JoshStrobl:
@dominichayesferen Personally prefer `accessibility` over `a11y` but not fussed either way.

GeorgesStavracas:
I have a philosophical appreciation of Cassidy's ["accessibility features are just features"](https://blog.elementary.io/accessibility-features-are-just-features/) assertion, so, in my humble opinion, `org.freedesktop.appearance` is fine. But it's not a strong opinion.

GeorgesStavracas:
I'm going to leave it up to you all to agree on the naming scheme, I'm fine with all the options presented so far, and my preference for `org.freedesktop.appearance` over others is very faint.

danirabbit:
I don't have super strong opinions here. We haven't historically supported a separate high contrast mode because our goal is to meet [WCAG AA by default](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) and my understanding is that folks with significant vision loss past this level would likely need other types of assistance like magnification or a screen reader.

WCAG also talks about some folks needing low contrast, so I'm not sure if it makes sense to go ahead and add a three-state setting for contrast if there's going to be a contrast setting? Or if it would be better assumed that a low contrast setting would be implemented by a compositor shader or something

orowith2os:
Ack on `appearance`.

> Similarly, as not all Desktop Environments currently employ any form of High Contrast switches, this hint should also be considered optional for the Portal implementers - if you do not want a high contrast option in your Desktop Environment, simply hardcode returning False into your Portal implementation (remember, however, that by doing so third-party applications supporting this feature will never expose their high contrast implementations if this preference is merged and they start following it).

Did any app ever expose it in their UI? GNOME's and KDE's sure didn't. It's probably in the same realm as dark mode; it should be controlled in the system menus, not per-app. 

But that would be getting into design discussions unrelated to this change. Nothing to block this on. I'd like to merge this as-is.

dominichayesferen:
> Ack on `appearance` (as a third party).
> 
> > Similarly, as not all Desktop Environments currently employ any form of High Contrast switches, this hint should also be considered optional for the Portal implementers - if you do not want a high contrast option in your Desktop Environment, simply hardcode returning False into your Portal implementation (remember, however, that by doing so third-party applications supporting this feature will never expose their high contrast implementations if this preference is merged and they start following it).
> 
> Did any app ever expose it in their UI? GNOME's and KDE's sure didn't. It's probably in the same realm as dark mode; it should be controlled in the system menus, not per-app.
> 
> But that would be getting into design discussions unrelated to this change. Nothing to block this on. I'd like to merge this as-is.

I'm referring to the following:

- Applications: They can choose to read whatever the preference is, and if they want to and the preference returns True they can switch into a state of higher contrast
- Desktops: GNOME, Cinnamon, MATE, Cinnamon, maybe Budgie, and possibly XFCE (not sure though) all expose high contrast, in a non-standardised form, in their System Settings applications (currently under Accessibility).

dominichayesferen:
> I don't have super strong opinions here. We haven't historically supported a separate high contrast mode because our goal is to meet [WCAG AA by default](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) and my understanding is that folks with significant vision loss past this level would likely need other types of assistance like magnification or a screen reader.
> 
> WCAG also talks about some folks needing low contrast, so I'm not sure if it makes sense to go ahead and add a three-state setting for contrast if there's going to be a contrast setting? Or if it would be better assumed that a low contrast setting would be implemented by a compositor shader or something

I'm more than willing to convert this to a tri-state like theme-preference is, but while a part of me says we should just so that we don't need to make a V2 to add it in the future, another part of me says that might be too niche, what with literally no platform I know of having a 'low contrast' option except ancient GTK3 and maybe GTK2.

dominichayesferen:
Actually, given that you could argue theme-preference's "Prefer Light" is of similar niche... probably best for it to have it just on the off chance - although I REALLY don't expect anyone to ever implement it anywhere, though.

I suppose if nobody NACKs that idea, I'll implement it tomorrow or the next day.

GeorgesStavracas:
Then I suggest turning this into a `contrast` string property or somesuch, with 2 documented values: `"default"` and `"high-contrast"`. This will allow documenting a third `"low-contrast"` mode in the future without breaking API.

I don't think it's useful to add low-contrast mode right now. It seems like an unnecessary burden right now, given that apparently nobody implements it.

dominichayesferen:
I'd probably make it the following:
0 - normal
1 - high
2 - low (undocumented until needed?)

GeorgesStavracas:
That would work as well. Just don't document low-contrast right now, we should evaluate adding that later.

dominichayesferen:
*dualstate
ah whatever doesn't matter 'cos I'll probably need to manually squeeze the commits anyway later.

dominichayesferen:
Anyway, how's that for a structure to accommodate lower contrast?

dominichayesferen:
Also just in case there's another person on the off chance that doesn't realise High Contrast exists in GNOME:
image

orowith2os:
> Anyway, how's that for a structure to accommodate lower contrast?

Can't be any better, I like it. 

>  Think of it as similar logic to the reasoning for having a potential colour blindness hint in Portal in the future - letting applications, especially games come into mind, account themselves for any filters in use.

This, and many other settings hidden behind the `org.gnome` namespace, would also be great candidates for including in the Settings portal next, since color blindness might take a bit to design and work on on the toolkit side of things.

lleyton:
> I have a philosophical appreciation of Cassidy's ["accessibility features are just features"](https://blog.elementary.io/accessibility-features-are-just-features/) assertion, so, in my humble opinion, `org.freedesktop.appearance` is fine. But it's not a strong opinion.

I wasn't aware of that article, but I'll admit the argument is quite compelling. `org.freedesktop.appearance` makes sense then.

> Anyway, how's that for a structure to accommodate lower contrast?

Seems good with me :)

> This, and many other settings hidden behind the org.gnome namespace, would also be great candidates for including in the Settings portal next, since color blindness might take a bit to design and work on on the toolkit side of things.

Agreed.

dominichayesferen:
I feel that y'all will be able to handle discussion about this merge request in particular while I'm absent from it in lieu of IRL stuff and focusing on college, so to allow for that to potentially happen, it's time I undraft this.

emilio:
I had missed the ping above, but this seems fine to me as well. Thanks for working on this!

dominichayesferen:
> I had missed the ping above, but this seems fine to me as well. Thanks for working on this!

Should I count that as an ack on Mozilla's side, or...?

emilio:
Sure

snwh:
I would concur with @GeorgesStavracas that overall this seems pretty uncontroversial. I'm reminded of the [prefers-contrast](https://www.w3.org/TR/mediaqueries-5/#prefers-contrast) media query in CSS, which has already been mentioned, and would also agree on doing a 3 state key for `contrast` that somewhat mirrors the WCAG spec:

- `default`: no preference, or a deference to the current state
- `high`: hint for an active option that increases contrast 
- `low`: hint for an active option that reduces contrast

I do think even if not all the stakeholders fully implement something like low contrast, it'd be good to have it in and documented just so it can be forward-looking, especially if this is an accommodation that could be made in future for a subset of users.

dominichayesferen:
@sonnyp @snwh Can I confirm that you two are ACKing this MR, given your messages and who you may be representing?

dominichayesferen:
Alright, with that, there's only a few ACKs left I would think.

CarlSchwan:
The proposal looks sensible to me.

ack from KDE

snwh:
> @sonnyp @snwh Can I confirm that you two are ACKing this MR, given your messages and who you may be representing?

Yeah, weighing in from GNOME. :+1:

mmstick:
ack from COSMIC

danirabbit:
Ack 😊 thanks everyone

dominichayesferen:
Wow, this actually reached 'full house' in terms of its ACKs, thanks everyone from basically all of the corners of FreeDesktop and beyond for providing your concerns, and eventually ACKs! 💙 

I suppose this means this is ready for merging, then, but that's on those who can merge. In the meantime, if there are any MRs made in toolkits or desktop environments, with this proposal in mind, please drop them in this discussion so I can add them to the main message.

GeorgesStavracas (approved):
Please fix the commit history to have a single, properly formatted commit (see the main branch for commit message guidelines).

nothingneko:
Ack from Helium. Let's do this

dominichayesferen:
Alright, gotta remember how to flatten commits properly, so that I don't mess up and accidentally auto-close this PR.

nothingneko:
I think you can do it from the GitHub UI. Should be a button for a maintainer to squash and merge.

dominichayesferen:
@GeorgesStavracas Ready!

dominichayesferen:
Guess that means I'm now officially a FreeDesktop standard author, nice.

dominichayesferen:
oh yeah, @emilio, now it's merged you might be interested in the bug report made in advance: https://bugzilla.mozilla.org/show_bug.cgi?id=1861775

hfiguiere:
Shall we implement it in the -gtk portal as well?

PR #1176, opened 2023-10-28
Ensure FileChooser dialog only sends file:// URIs to apps

GeorgesStavracas:
xdg-desktop-portal does not do the normalization, this is up to backends, and it's now documented that they have to do this. But we can validate the received URIs. Do that.


ebassi (approved):
:+1:  from me.

As far as I know, all portal implementations already normalise the URIs to `file://` scheme (outside of the GNOME portal), so we don't need to notify people about this change.

GeorgesStavracas:
Last push just improved the new docs commit a bit more, no functional changes

GeorgesStavracas:
Matthias LGTMed on Matrix \o/

PR #1177, opened 2023-10-28
Draft: Colour filter hint in Settings Portal

### LINKED ISSUE #647: High contrast and color blindness settings
Similar to the org.freedesktop.appearance.color-scheme key in the settings portal it might be a good idea to expose

1. preference for high contrast UIs for people with bad contrast perception
2. type of color blindness to avoid certain colors and color combinations in UIs

dominichayesferen:
Unlike the high contrast one, this one is a draft because I know Desktop Environments like GNOME don't yet have a concrete idea about the design they would prefer for this, the colour blindness filters they want to expose, etc.

As such, I expect this MR to thoroughly change before it exits Draft status, and for it to take forever to even exit Draft status.

Fixes #647, #1151 

# Introduction
Colour blindness is a prevalent problem, and Operating Systems in the XDG-compliant space are starting to realise this and accommodate to those suffering from such issues with their sight. So far, elementary OS and KDE Plasma have implemented colour filters for the colour blind users, by means of compositor shaders. However, this implementation has a design limitation that this pull request sets sight on allowing a fix to be made for.

# Details
A new key on the settings portal, `color-filter`, would be introduced into the `org.freedesktop.appearance` namescape. This key currently consists of all colour blindness filters implemented by elementary OS, as well as accommodating for greyscale, however more filters could be added at a later date to this pool of values.

As aforementioned, this is largely inspired by elementary OS's implementation of their colour shaders, although this design may always be changed to accommodate for the requests of other Desktop Environments.

This hint is designed with being paired with colour shaders in mind - applications should only use this hint to adjust their user interface for the current filter.

Additionally, greyscale is made available here to allow implementations of greyscale shaders, such as elementary OS's, to be officially recognised as shaders that applications should account for.

Brown Monotone was originally considered as a shader, here, however this is easily supplemented by allowing greyscale to be redshifted.

# Cases for colour filter hints for applications

## Libadwaita

Libadwaita has a high reliance on colours to denote certain types of buttons, however if colour blindness filters are not accounted for, on certain colour blindness filters the denotation can end up being too similar to be uniquely identifiable.

image

This would obviously especially be the case on greyscale filters, where colours are just not usable for denotation, period.

image

If Libadwaita applications were able to know a colour filter is in use, they could then adjust themselves accordingly to better take advantage of the limited, or non-existent, colour pool.

## Games, such as Geometry Dash

While not native so not able to yet leverage this standard if merged, I'd still like to use Geometry Dash to illustrate a gaming context for why applications would want to know about the current colour filter.

Before 2.2, one of the biggest aspects of Geometry Dash's gameplay, the portals and orbs, were similarly denoted by colours only. However, after concerns for colourblind players, the upcoming 2.2 update adds the option for colourblind players to add symbols to portals and orbs to allow denotion without colours (_this would obviously be useful on greyscale filters, too_).

image

If games are able to know which colour filter is in use, such as via this hint, they could technically automatically enable such optimisations, in their gameplay, for audiences using these colour filters.

# Acks

This may be a draft currently, but to save George the effort, using the acks list from #815 
- [ ] @alice-mkh (GNOME)
- [ ] @JoshStrobl (Budgie)
- [ ] @Sodivad (KDE)
- [ ] @jackpot51 (Cosmic)
- [ ] @danirabbit (Elementary)

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -56,7 +56,7 @@
         <listitem><para>
           Indicates the system's preferred color filter.
           Applications should only use this to adjust themselves for the current filter.
-          It is the compositor's job to apply the colour filter globally.
+          It is the compositor's job to apply the color filter globally.
           Supported values are:
           <simplelist>
             <member>0: No preference</member>
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -51,6 +51,25 @@
           Out-of-range RGB values should be treated as an unset accent color.
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>org.freedesktop.appearance color-filter u</term>
+        <listitem><para>
+          Indicates the system's preferred color filter.
+          Applications should only use this to adjust themselves for the current filter.
+          It is the compositor's job to apply the colour filter globally.
+          Supported values are:
+          <simplelist>
+            <member>0: No preference</member>
+            <member>1: Protanopia</member>
+            <member>2: Protanopia (higher color contrast)</member>
+            <member>1: Deuteranopia</member>
+            <member>2: Deuteranopia (higher color contrast)</member>
+            <member>1: Tritanopia</member>
+            <member>1: Grayscale</member>
+          </simplelist>
+          Unknown values should be treated as 0 (no preference).
+        </para></listitem>
+      </varlistentry>
     </variablelist>
 
     Implementations can provide other keys; they are entirely
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -62,10 +62,10 @@
             <member>0: No preference</member>
             <member>1: Protanopia</member>
             <member>2: Protanopia (higher color contrast)</member>
-            <member>1: Deuteranopia</member>
-            <member>2: Deuteranopia (higher color contrast)</member>
-            <member>1: Tritanopia</member>
-            <member>1: Grayscale</member>
+            <member>3: Deuteranopia</member>
+            <member>4: Deuteranopia (higher color contrast)</member>
+            <member>5: Tritanopia</member>
+            <member>6: Grayscale</member>
           </simplelist>
           Unknown values should be treated as 0 (no preference).
         </para></listitem>
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -56,7 +56,7 @@
         <listitem><para>
           Indicates the system's preferred color filter.
           Applications should only use this to adjust themselves for the current filter.
-          It is the compositor's job to apply the colour filter globally.
+          It is the compositor's job to apply the color filter globally.
           Supported values are:
           <simplelist>
             <member>0: No preference</member>
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -51,6 +51,25 @@
           Out-of-range RGB values should be treated as an unset accent color.
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>org.freedesktop.appearance color-filter u</term>
+        <listitem><para>
+          Indicates the system's preferred color filter.
+          Applications should only use this to adjust themselves for the current filter.
+          It is the compositor's job to apply the colour filter globally.
+          Supported values are:
+          <simplelist>
+            <member>0: No preference</member>
+            <member>1: Protanopia</member>
+            <member>2: Protanopia (higher color contrast)</member>
+            <member>1: Deuteranopia</member>
+            <member>2: Deuteranopia (higher color contrast)</member>
+            <member>1: Tritanopia</member>
+            <member>1: Grayscale</member>
+          </simplelist>
+          Unknown values should be treated as 0 (no preference).
+        </para></listitem>
+      </varlistentry>
     </variablelist>
 
     Implementations can provide other keys; they are entirely
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -62,10 +62,10 @@
             <member>0: No preference</member>
             <member>1: Protanopia</member>
             <member>2: Protanopia (higher color contrast)</member>
-            <member>1: Deuteranopia</member>
-            <member>2: Deuteranopia (higher color contrast)</member>
-            <member>1: Tritanopia</member>
-            <member>1: Grayscale</member>
+            <member>3: Deuteranopia</member>
+            <member>4: Deuteranopia (higher color contrast)</member>
+            <member>5: Tritanopia</member>
+            <member>6: Grayscale</member>
           </simplelist>
           Unknown values should be treated as 0 (no preference).
         </para></listitem>


dominichayesferen:
Side-note: For LibAdwaita + greyscale, I have a few design ideas that I'll share in the coming days when I have time to design them.

danirabbit:
In my research here what I heard was the folks probably want this to be a separate setting and not tied into whether they use assistive filters. Especially in collaborative or support environments this can make things actually worse because what they see on screen is not what everyone else is seeing.

For example a person with protanopia may have a concept of what they consider to be green and what they consider to be red and if they're using an assistive filter they can communicate with team mates that "this is a red button" even if they're not seeing exactly what someone else is. But if their interface is changed completely they won't be able to have that common communication with others.

I'm not sure about other platforms but iOS has a separate "different without color" binary setting that is completely independent from color assistance filters

dominichayesferen:
I mean, my idea for application implementation was more-so to add subtle changes to accomodate for the shaders, such as destructive action maybe getting a subtle texture up top, akin to the unstable application header texture, in LibAdwaita and/or getting icons for additional denotation, but yeah fair point.

danirabbit:
I think generally we should be pushing the defaults to not rely solely on color to communicate information. If something like adding an icon or a texture is beneficial to folks with color deficiency maybe we should just do that for everyone all the time?

dominichayesferen:
I mean, yes, but there again there might be edge-cases, such as the aforementioned games context, that may still provide merit for providing a hint irregardless.

It all comes down to implementation on the application's side, at the end of the day, and I really don't see 100% of applications willing to accommodate deciding to accommodate anyway without knowing a colour filter is in use.

swick:
-1 for exposing color filter settings to clients
+1 for exposing color vision deficiency to clients

* Apps should always be usable without any color information
* Using color can increase the usability of apps
* Apps should always chose their colors in a way that they work with various color vision deficiencies
* Knowing about the color vision deficiency of a user increases the amount of different colors one can chose without impacting usability
* The color vision deficiency filters are a fallback to be used when the apps fail accommodating to the users color vision deficiency

danirabbit:
> such as the aforementioned games context, that may still provide merit for providing a hint irregardless.

Right so this is exactly the kind of potentially collaborative case where the folks I spoke said they explicitly didn't use options that were designed as color assistance because it made their experience worse. Opting someone into a different interface just because they're using an assistive tool is not what they're expecting. So a hint here should be explicit about its purpose both so it feels deterministic for the people using it and so that app developers know exactly what it should do

danirabbit:
> The color vision deficiency filters are a fallback to be used when the apps fail accommodating to the users color vision deficiency

From what I've heard, I think it's more the other way around. Folks with color vision disabilities seem to largely be more interested in assistance that meets them at the severity of their disability and not like immediately jumping to segregated experiences unless their disability absolutely requires it to function

GeorgesStavracas:
I think this needs a lot of discussion and it's not entirely clear to me how desktop should handle it, so I'm not assigning any particular milestone for this.

dominichayesferen:
Just want to be clear here:
I plan on updating this MR on Friday at the earliest to accomodate for current feedback - I'm not ignoring any feedback, just haven't had time to look in here due to assignments.

In advance, @swick how best would a redesigned version of this draft be changed to accomodate for your feedback? Besides a rename, would it remain the same but without the "higher contrast" options, or alternatively become a generalised "I am colourblind" boolean? I'm assuming the former, but thought I'd ask first.
Also, @danirabbit, would you be up for the suggestion at https://github.com/flatpak/xdg-desktop-portal/pull/1177#issuecomment-1783867854 being applied to this draft spec idea?

swick:
> In advance, @swick how best would a redesigned version of this draft be changed to accomodate for your feedback? Besides a rename, would it remain the same but without the "higher contrast" options, or alternatively become a generalised "I am colourblind" boolean?

It should describe the kind of color vision deficiency that a user has: no red, no green, no blue, no color or reduced contrast.

One could argue that a strength would also be helpful but I think we should start with a minimum and if there is need for it we can add it later.

swick:
> From what I've heard, I think it's more the other way around. Folks with color vision disabilities seem to largely be more interested in assistance that meets them at the severity of their disability and not like immediately jumping to segregated experiences unless their disability absolutely requires it to function

Interesting. I guess this is also a problem with apps switching to a very different mode instead of being designed with CVD in mind and then only having to adjust slightly.

Either way, keeping them separate and not exposing the active filters gives the user more control while keeping the implementation complexity at a minimum. I think we agree here.

Mikenux:
The problem is that if an app has its own things to adapt to the user's vision, then it should logically know if a filter is active in order to disable it, but then the user indeed has no control . Perhaps the app should report back to the system whether it has its own adaptations, so that the user can choose to enable or disable a color filter for a specific application?

Besides, shouldn't the application also know the severity (since it was mentioned) to know if it has the necessary adaptations?

orowith2os:
For a reference on games using this setting, DOOM Eternal provides colorblind filters in-game. It would be a perfect example of a slight possibility of games using this setting.

dominichayesferen:
I'll admit - I forgot about this PR entirely since my last message.

Unfortunately I'm going to be busy until Friday, because gotta do more assignments, so won't incorporate the feedback just yet, BUT if there's more feedback to be given here, do provide it in advance... and I'll try not to forget this time when I'm free again to rebase it and incorporate everyones' feedback.

swick:
Accent colors have landed which is great, but it doesn't change signal colors such as red for no and green for yes. The CVD information could be very helpful for this. Is this something we could do on the libadwaita side @alice-mkh?

alice-mkh:
We heavily discourage use of colors for that without secondary indicator. So if you see something like that in apps - e.g. a red for error without an icon - file bugs.

swick:
Sure, and that's a good thing! But wouldn't it be better if we even re-colored the red/green indicators to orange/blue instead for people with red-green blindness?

alice-mkh:
Orange/blue make no sense semantically. They don't just need to be different, they also need to make sense. Would a colorblind person even recognize blue as an error, rather than, dunno, a hyperlink?

And then you also have icons and illustrations, which wouldn't match those colors anymore.

If you really want to recolor things, you'd be better off with a compositor wide filter IMO. Then it's also clear in context what blue means - as opposed to randomly blue messages in some apps.

swick:
The filters are far from perfect and from what I've learned, most people only turn them on occasionally. I'm also not asking random apps to re-color this but rather libadwaita, so that the colors can be learned.

alice-mkh:
Would be nice to have some data about how many people use blue/orange colors in other OS too. Surely those filters aren't perfect - but what would make a toolkit doing the same thing suddenly better?

swick:
> Surely those filters aren't perfect - but what would make a toolkit doing the same thing suddenly better?

That the toolkit can choose where to apply it.

The compositor will re-color everything, including images and videos etc. They will look differently, even to people with CVD (and the same kind of CVD) because each person has slightly different sensitivity curves which implies different strength and different perception. So when the CVD isn't extreme, one might prefer to experience images and videos without the extreme re-coloring of the filter and only use it when something is not legible.

swick:
Sorry, but this is about recoloring very specific UI elements and color is an amazing indicator making it much easier to see what's going on at a glance (but one which should not be relied upon alone).

alice-mkh:
While still being a huge headache for app developers. Unless there's clear data that people need/use an option to change color, I don't think it's happening at least in libadwaita, sorry.

swick:
How is that a headache for app developers?

alice-mkh:
The same way as not knowing what any other color is gonna be is a problem. Except instead of 1 color (accent color) you now have an additional 4 (error/warning/success/destructive).

This also encourages people to use color-only for differentiation, instead of being forced to add other indicators.

Anyway, I'm gonna stop replying here - this is unproductive. I will only consider further input from colorblind people.

swick:
There are filters to simulate color blindness and there are gnome shell extensions which do have them. Try it out.

alice-mkh:
That tells me nothing, I'm not colorblind.

Forgot to unsubscribe, sorry.

swick:
> That tells me nothing, I'm not colorblind.

That's why you should use the filter to *simulate* color blindness.

Mikenux:
@swick: I agree that decisions for GNOME components need to be discussed in the GNOME space and come back here if there is anything to add here.  I believe you should consider that it will also be difficult to manage an interaction between accent color and color blindness colors (maybe that's one problem raised by Alice?) compared to having to style the UI with only color blindness colors (i.e. no different accent colors for people with color blindness).

Nobody-Special:
Hi! Got here form the libadwaita:gnome.org matrix room.
I would like to point out that the pine64.org has released a pine-note,
so with the future in mind where some devices might operate solely in grey-scale
I think that the approach that the danirabbit has proposed would be the best, that is:

> we should be pushing the defaults to not rely solely on color to communicate information

swick:
That one should not rely on color alone as indicator is obvious and consensus but doesn't preclude further improvements.

GeorgesStavracas:
Okay so as far as I can see, there isn't general consensus on this yet, at least not enough to justify it being a pull request. If people are still interested in discussing and pushing this forward, please open a discussion in the [Discussions tab](https://github.com/flatpak/xdg-desktop-portal/discussions) and link to this PR.

I'll go ahead and close this, but this is not a rejection of the idea. At any time, if more consensus is reached, please feel free to reopen or open a new PR.

PR #1178, opened 2023-10-28
Typo fix. Move the alt text attribute to an element where it is allowed and works

rthrejheytjyrtj545:
According to the HTML specification, `alt` is specified by `img`, and there is no additional workaround in the web platform for the authors' errors in calculating the accessible name for the `picture` case.


TingPing:
Change looks good but could you choose a more readable commit message.

Maybe like "Move incorrect alt attribute from picture to img"

PR #1182, opened 2023-11-02
tests: Add a devenv for our most common test environment

whot:
Trying to figure out the environment variables to set to run the tests
manually is annoying, so let's use technology to solve that and pretend
we're living in the future where technology actually helps us.
And this glorious future now allows us to do:
```
  $ meson devenv -Cbuilddir
  $ pytest
  $ ./tests/test-portals
```
and so on.


GeorgesStavracas (approved):
Looks helpful!

PR #1186, opened 2023-11-03
inputcapture: add support for persist_mode and restore_token

whot:
Identical to the existing RemoteDesktop/ScreenCast persist modes, just this time for InputCapture. This allows an application to keep the permissions past things like screen savers activating and whatnot.

Keeping this as draft until we have the mutter side for real-world testing beyond the test suite. libportal obviously will need updates as well.

cc @jadahl

### DIFF
--- a/data/org.freedesktop.impl.portal.InputCapture.xml
+++ b/data/org.freedesktop.impl.portal.InputCapture.xml
@@ -26,7 +26,7 @@
       #org.freedesktop.portal.InputCapture portal, see that portal's
       documentation for details on methods, signals and arguments.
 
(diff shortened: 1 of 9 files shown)


sithlord48:
@whot should this be finished before #1683  ?

whot:
> should this be finished before https://github.com/flatpak/xdg-desktop-portal/pull/1683 ?

Not technically a requirement, can be added at the same time, or afterwards (possibly merging into the Init bit) but right now I have zero implementation for it either way and it doesn't look like I'll find time anytime soon.

swick:
@whot what's the status of this?

PR #1189, opened 2023-11-06
Fix a few issues with the portal tests

whot:
With this patchset, I can finally get openuri to pass here *and* run the tests in parallel in a fraction of the time. Totally rad, or whatever the current cool kids' preferred phrasing is.


tests/test-portals.c
```diff
@@ -173,6 +173,20 @@ global_setup (void)
   session_bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &error);
   g_assert_no_error (error);
 
+  if (g_getenv("XDP_DBUS_MONITOR"))
```
GeorgesStavracas, line 176: Nit: missing space before `(`

GeorgesStavracas (approved):
One small nit, LGTM otherwise

tests/test-portals.c
whot: sigh, there's always one that sneaks through. Fixed, thanks!

PR #1190, opened 2023-11-07
Fix lead of file descriptors for PhysicalInode:s

alexlarsson:
This fixes https://github.com/flatpak/xdg-desktop-portal/issues/689

The mechanism we're using to flush the dcache and thus avoiding long-term caching of physical inodes which keep a file descriptor open doesn't seem to work quite right.

A trivial test of cat /run/user/1000/doc/$docid/$filename properly leads to the PhysicalFile being freed. However, an application that keeps the file open for a longer time (longer than the delay we have before the current dcache flush workaround is run), ends up with the dcache entry for the inode to be kept in cache. This means that we when the last other kernel reference (like an open fd) to the inode is gone we don't get a FORGET event from fuse, and thus we never free the PhysicalInode.

This is clearly visible, because if you end up with the file descriptor leak like in the issue, then flushing the dcache (`sudo sh -c 'echo 2 > /proc/sys/vm/drop_caches'`) fixes the leak.

It seems the current workaround of just invalidating the toplevel directory entry doesn't work, so we replace it with tracking each individual lookup that ends up with a physical inode and flush all of them. This seems to fix the issue for me.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -153,8 +153,6 @@ struct _XdpDomain {
   guint64 doc_dir_inode;
   guint32 doc_flags;
 
-  int doc_queued_invalidate; /* Access atomically, 1 if queued invalidate */
-
   /* Below is mutable, protected by mutex */
   GMutex  tempfile_mutex;
   GHashTable *tempfiles; /* Name -> physical */
@@ -223,6 +221,14 @@ typedef struct {
 XdpInode *root_inode;
 XdpInode *by_app_inode;
 
+typedef struct {
+  ino_t parent_ino;
+  char name[0];
+} XdpInvalidateData;
+
+static GList *invalidate_list;
+G_LOCK_DEFINE (invalidate_list);
+
 static XdpInode *xdp_inode_ref (XdpInode *inode);
 static void xdp_inode_unref (XdpInode *inode);
 
@@ -237,6 +243,8 @@ static int ensure_docdir_inode (XdpInode *parent,
                                 int o_path_fd_in, /* Takes ownership */
                                 struct fuse_entry_param *e,
                                 XdpInode **inode_out);
+
+static void queue_invalidate_dentry (XdpInode *parent, const char *name);
 
 static gboolean
 app_can_write_doc (PermissionDbEntry *entry, const char *app_id)
@@ -1061,6 +1069,8 @@ get_tempfile_for (XdpInode *parent,
   /* This is atomic, because we're called with the lock held */
   g_hash_table_replace (domain->tempfiles, tempfile->name, xdp_tempfile_ref (tempfile));
 
+  queue_invalidate_dentry (parent, tempfile->name);
+
   if (tempfile_out)
     *tempfile_out = g_steal_pointer (&tempfile);
   return 0;
@@ -1107,6 +1117,8 @@ create_tempfile (XdpInode *parent,
 
   /* This is atomic, because we're called with the lock held */
   g_hash_table_replace (domain->tempfiles, tempfile->name, xdp_tempfile_ref (tempfile));
+
+  queue_invalidate_dentry (parent, tempfile->name);
 
   if (tempfile_out)
     *tempfile_out = g_steal_pointer (&tempfile);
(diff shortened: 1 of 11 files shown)


alexlarsson:
@hfiguiere Please take a look at this. I did local testing for that this fixed the particular issues I ran into, but it needs more global testing.

alexlarsson:
Hmm, seems like the tests are falling in `xdg-desktop-portal/test-document-fuse.sh.test`.

alexlarsson:
Yeah, seems like when the tests re-rexports a path from the fuse mountpoint it sometimes thinks the file descriptor is deleted...

alexlarsson:
I don't understand these CI errors. They look like timeout, but I bumped the timeout to 20 min, and none of the runs that fail are near 20 min.

GeorgesStavracas:
@whot do you happen to know or have any guesses of what's going on with the testsuite here?

alexlarsson:
Hmm, that broke to. Wth is going on?

document-portal/document-portal-fuse.c
```diff
@@ -1707,37 +1719,53 @@ ensure_doc_inode (XdpInode *parent,
...
-  if (session && g_atomic_int_get (&doc_domain->parent_inode->kernel_ref_count) > 0)
-    fuse_lowlevel_notify_inval_entry (session, parent_ino, doc_domain->doc_id,
-                                      strlen (doc_domain->doc_id));
+  for (GList *l = to_invalidate; l != NULL; l = l->next)
+    {
+      XdpInvalidateData *data = l->data;
+      if (session)
+        fuse_lowlevel_notify_inval_entry (session, data->parent_ino, data->name,
+                                          strlen (data->name));
+      g_free (data);
```
hfiguiere, line 1738: doesn't this leak `data->name` since it is g_strdup() ?

alexlarsson: Yeah.

alexlarsson:
Maybe its a resource issue, i fixes the pointed out leak and some others, lets see now.

alexlarsson:
Disabling the document portal test in the "meson test" seems to have made CI pass again. I have no idea why this is, i mean it passes as part of the installed tests, and it passes locally here. And when It did run, that particular test always passed, the tests just got a SIGKILL at some other random point.

WTF?

document-portal/permission-db.c
```diff
@@ -1023,7 +1022,7 @@ permission_db_entry_get_permissions_variant (PermissionDbEntry *entry,
     {
       m = (start + end) / 2;
 
-      child = g_variant_get_child_value (app_array, m);
+      g_autoptr(GVariant) child = g_variant_get_child_value (app_array, m);
```
GeorgesStavracas, line 1025: Nit: mind moving the variable declaration to the start of the `while` block?

src/xdp-utils.c
```diff
@@ -1094,17 +1096,28 @@ verify_proc_self_fd (XdpAppInfo *app_info,
      but there is not much to do about this. */
   if (g_str_has_suffix (path_buffer, " (deleted)"))
     {
-      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_FILENAME,
-                   "Cannot share deleted file: %s", path_buffer);
-      return NULL;
+      if (documents_mountpoint != NULL &&
+          g_str_has_prefix(path_buffer, documents_mountpoint))
```
GeorgesStavracas, line 1100: Style nit: please add a space before `(`

src/xdp-utils.c
```diff
@@ -1094,17 +1096,28 @@ verify_proc_self_fd (XdpAppInfo *app_info,
...
+        {
+          /* Unfortunately our workaround for dcache purging triggers
+             o_path file descriptors on the fuse filesystem being
+             marked as deleted, so we have to allow these here and
+             rewrite them. This is safe, becase we will stat the file
+             and compare to make sure we end up on the right file. */
+          path_buffer[symlink_size - strlen(" (deleted)")] = 0;
```
GeorgesStavracas, line 1107: Perhaps it's worth asserting `g_assert (symlink_size > strlen (" (deleted)")`?

Also, perhaps it's clearer to pass an explicit character here: `\0`

document-portal/permission-db.c
```diff
@@ -734,7 +734,7 @@ permission_db_update (PermissionDb *self)
 
   g_return_if_fail (PERMISSION_IS_DB (self));
 
-  root = gvdb_hash_table_new (NULL, NULL);
+  g_autoptr(GHashTable) root = root = gvdb_hash_table_new (NULL, NULL);
```
GeorgesStavracas, line 737: The double self-assignment doesn't seem right here. Also, minor nit, mind moving the `g_autoptr(GHashTable) root = NULL` declaration to the beginning of the function?

GeorgesStavracas (changes requested):
Minor nits only

src/xdp-utils.c
hfiguiere: Can't we make `strlen(" (deleted)")` a constant instead? I don't think the compile will optimize that

alexlarsson: We just verified that it has " (deleted)" as a suffix. Not sure why it would not fit in there?

@hfiguiere GCC will in fact optiize strlen of constant strings.

tests/account.c
```diff
@@ -40,16 +40,19 @@ account_cb (GObject *obj,
       res = g_variant_lookup (ret, "id", "&s", &s); 
       g_assert (res == (t != NULL));
       if (t) g_assert_cmpstr (s, ==, t);
+      free(t);
```
hfiguiere, line 43:
```suggestion
      free (t);
```

tests/account.c
```diff
@@ -40,16 +40,19 @@ account_cb (GObject *obj,
...
 
       t = g_key_file_get_string (keyfile, "account", "name", NULL);
       res = g_variant_lookup (ret, "name", "&s", &s); 
       g_assert (res == (t != NULL));
       if (t) g_assert_cmpstr (s, ==, t);
+      free(t);
```
hfiguiere, line 49:
```suggestion
      free (t);
```

tests/account.c
```diff
@@ -40,16 +40,19 @@ account_cb (GObject *obj,
...
 
       t = g_key_file_get_string (keyfile, "account", "image", NULL);
       res = g_variant_lookup (ret, "image", "&s", &s); 
       g_assert (res == (t != NULL));
       if (t) g_assert_cmpstr (s, ==, t);
+      free(t);
```
hfiguiere, line 55:
```suggestion
      free (t);
```

document-portal/document-portal-fuse.c
```diff
@@ -1707,37 +1719,53 @@ ensure_doc_inode (XdpInode *parent,
...
+
+  XdpInvalidateData *data = g_malloc0(sizeof(XdpInvalidateData) + strlen(name) + 1);
+  data->parent_ino = parent->ino;
+  strcpy(data->name, name);
 
-  g_timeout_add (1000, invalidate_doc_domain, xdp_domain_ref (doc_domain));
+  if (invalidate_list == NULL)
+    g_timeout_add (10, invalidate_dentry_cb, NULL);
+
+  invalidate_list = g_list_append(invalidate_list, data);
```
hfiguiere, line 1768:
```suggestion
  invalidate_list = g_list_append (invalidate_list, data);
```

document-portal/document-portal-fuse.c
```diff
@@ -1707,37 +1719,53 @@ ensure_doc_inode (XdpInode *parent,
...
-  if (!g_atomic_int_compare_and_exchange (&doc_domain->doc_queued_invalidate, old, 1))
-    return; // Someone else set it to 1, return
+  for (GList *l = invalidate_list; l != NULL; l = l->next)
+    {
+      XdpInvalidateData *data = l->data;
+      if (data->parent_ino == parent->ino && strcmp (name, data->name) == 0)
+        return;
+    }
+
+  XdpInvalidateData *data = g_malloc0(sizeof(XdpInvalidateData) + strlen(name) + 1);
```
hfiguiere, line 1761:
```suggestion
  XdpInvalidateData *data = g_malloc0 (sizeof (XdpInvalidateData) + strlen (name) + 1);
```

src/xdp-utils.c
hfiguiere:
```suggestion
          path_buffer[symlink_size - strlen (" (deleted)")] = 0;
```

hfiguiere (approved):
Minor formatting nits.

LGTM

hfiguiere:
@GeorgesStavracas let's merge this and I'll fix the smmall nits.

GeorgesStavracas:
@hfiguiere good idea. I've edited a pair of commits, you can fix the coding style now :)

Issue #1077, opened 2023-08-22
No more OpenURI when using XDG_CURRENT_DESKTOP=sway and gtk portal


stacyharper:
When using Sway, and having -gtk and -wlr installed:

With this last 1.17.0, there is no more available implementation to handle OpenURI requests.

If I edit `/usr/share/xdg-desktop-portal/portals/gtk.portal`, and add `sway` in the `UseIn`, then it works.

The 1.17.0 release note mention some change in how implementations are loaded, is there some required configuration now?

smcv:
Yes, there is some required configuration now: with x-d-p 1.17.0, maintainers of desktop environments are expected to set up the portals that they want their desktop environment to use. See the portals.conf man page or its source code <https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals-conf.rst>.

If you're using a complete desktop environment like GNOME or KDE Plasma, it should (eventually) set this up for you. For example, if `XDG_CURRENT_DESKTOP` is set to `gnome`, `gnome-portals.conf` will be used.

If you're piecing together a desktop environment out of individual components, you'll need to provide your own configuration. For example, if `XDG_CURRENT_DESKTOP` is set to `sway`, you can create `portals.conf` or `sway-portals.conf` in `$XDG_CONFIG_HOME/xdg-desktop-portal` or `/etc/xdg-desktop-portal`.

(I don't know which of those models Sway is closest to.)

If no `portals.conf` is found, there is a fallback which reads the old `UseIn` fields; but x-d-p-gtk only declares `UseIn=gnome`. When there is no portal that declares that it should be used for a particular desktop environment, as an intentional change in 1.17.0, x-d-p intentionally does not have the behaviour of older versions that would arbitrarily choose whatever portal happens to be first in alphabetical order and hope that it might work.

stacyharper:
Thanks for this clarification. I opened up this corresponding MR on Sway on Alpine Linux:

https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/50395/diffs

smcv:
I'm reopening this because I think we could have a better backwards compatibility story here. `-gtk` is the closest thing we have to a reference implementation of a portal backend, and it has historically been used (by OS vendors, desktop environments, and users of self-assembled or otherwise unsupported desktop environments) as the portal implementation of last resort. With xdp 1.17.x, that no longer works.

[...]

PR #1192, opened 2023-11-08
Set gtk as the default fallback portals backend

TingPing:
Some users don't actually use a desktop so it makes sense to fallback to *something* and gtk is a good choice.

```
XDP: Preferred portals for interface 'default': gtk
XDP: Using portal configuration file '/home/tingping/jhbuild/install/share/xdg-desktop-portal/portals.conf' for non-specific desktop
```

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -167,6 +167,11 @@ if bwrap.found()
   helper_def = '-DHELPER="@0@"'.format(bwrap.full_path())
 endif
 
+install_data(
+  'portals.conf',
+  install_dir: datadir / 'xdg-desktop-portal',
+)
+
 executable(
   'xdg-desktop-portal-validate-icon',
   'validate-icon.c',
--- a/src/portals.conf
+++ b/src/portals.conf
@@ -1,0 +1,5 @@
+[preferred]
+# This is the ultimate fallback for when nothing else is configured.
+# xdg-desktop-portal-gtk is fairly desktop neutral and implements many interfaces so it is a reasonable default.
+# See `man portals.conf` for more information on configuration.
+default=gtk


smcv:
This is an implementation of option 1 from #1102.

As I mentioned on #1102, the reason I didn't do this in Debian is that `/usr/share/xdg-desktop-portal/portals.conf` is higher-precedence than `UseIn`, so until every desktop environment with its own portal backend (including wlroots-based compositors) has a `portals.conf`, shipping a `/usr/share/xdg-desktop-portal/portals.conf` will cause those environments to regress: they would only use `x-d-p-gtk` as selected by `portals.conf`, and would not use `x-d-p-wlr`.

For instance, this would disable use of [x-d-p-wlr](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/wlr.portal) for screenshots and screencasts under Wayfire, until/unless Wayfire [ships a `wayfire-portals.conf` which would reinstate that](https://github.com/WayfireWM/wayfire/issues/1995).

Instead, what I did in Debian was to hard-code x-d-p to try x-d-p-gtk *after* checking `UseIn`: https://salsa.debian.org/debian/xdg-desktop-portal/-/blob/debian/1.18.1-1/debian/patches/debian/portal-impl-Hard-code-x-d-p-gtk-as-a-last-resort-fallback.patch

See also #1103 which is a non-working proof-of-concept for a version of this that has semantics closer to what I did in Debian.

TingPing:
> Instead, what I did in Debian was to hard-code x-d-p to try x-d-p-gtk after checking UseIn

This sounds like an acceptable solution to me. Anything is better than the current situation of some users just having no portal impl for common interfaces.

smcv:
Based on earlier discussions on #1077 and #1102, I'd assumed that the patch I used in Debian would be summarily rejected upstream, so I didn't send it as a merge request; but if it's something that people want, it's a straightforward change.

ebassi:
I don't have a *super* strong opinion on this; it seems a bunch of people using Arch have been caught unprepared, probably because of the hands-off nature of the downstream, and because nobody expects new configuration files. In all honesty, while I empathise with this, I also don't think a lot of effort upstream should be devoted to cater to the long tail of people building their own environment; they made the bed, they get to sleep in it. At most, we can document and explicitly tell integrators about the new requirements, but this is a case of end users doing the integration work.

The fact that the GTK portal is now less tied to GNOME makes it a good candidate for an unconditional callback **but** I can easily see complaints from people arguing that they shouldn't be forced to install an additional portal if they don't explicitly opt into it.

So, I'm neutral on any change that adds the GTK portal as a fallback *after* the configuration and the `UseIn` key.

TingPing:
>  I also don't think a lot of effort upstream should be devoted to cater to the long tail of people building their own environment; they made the bed, they get to sleep in it. At most, we can document and explicitly tell integrators about the new requirements, but this is a case of end users doing the integration work.

The result will honestly just be upset users because of a regression xdg-desktop-portal caused. There is no "integrators" so it will never be resolved.

smcv:
> There is no "integrators" so it will never be resolved.

To be fair, there definitely are integrators (the users of assemble-it-yourself desktop environments) - they just don't think of themselves that way, because they were hoping to get the benefits (fine-grained control) while imposing the costs (responsibility for making the overall system work) on someone else.

GeorgesStavracas:
So, @ebassi says:

> I'm neutral on any change that adds the GTK portal as a fallback after the configuration and the UseIn key

I personally think that's a fine solution for now.

@smcv would you like to propose such patch? If we incorporate this change, will that allow dropping your patches from the Debian package?

smcv:
> I can easily see complaints from people arguing that they shouldn't be forced to install an additional portal if they don't explicitly opt into it

So don't install it? Just because x-d-p has a last-resort fallback to x-d-p-gtk if it happens to be installed, that doesn't necessarily mean that packagers should add a hard dependency on x-d-p-gtk, or even a weak dependency on it.

In Debian, x-d-p does not depend on any backends. Packages that particularly benefit from x-d-p (like Flatpak) are currently expected to have dependencies on both x-d-p and `xdg-desktop-portal-gtk | xdg-desktop-portal-backend`, where `xdg-desktop-portal-backend` is a virtual package jointly provided by all backends - although I'm thinking of changing that to something more like `xdg-desktop-portal-gtk | xdg-desktop-portal-portals-conf`, where `xdg-desktop-portal-portals-conf` would be a virtual package jointly provided by `gnome-session` and other desktop environments that provide a `portals.conf(5)`.

smcv:
> If we incorporate this change, will that allow dropping your patches from the Debian package?

Yes, #1199 is functionally equivalent to our only Debian-specific patch at the moment.

smcv:
> it seems a bunch of people using Arch have been caught unprepared, probably because of the hands-off nature of the downstream, and because nobody expects new configuration files

I think if I hadn't included the equivalent of #1199 in Debian, we would have had various complaints from Debian users using less-common or less-complete desktop environments, too. The major desktop environments like GNOME and KDE Plasma are by far the most popular, but there's a very long tail of "simple window managers" whose users expect that everything will (somehow) continue to work without any particular action from them.

We can't have it both ways: if we want to ensure that app authors can rely on at least basic functionality of x-d-p (file chooser, etc.) on all semi-reasonable Linux desktops, then unfortunately we do need to think about some sort of compatibility story for fvwm users; or conversely, if we want a working x-d-p to require desktop-environment-specific functionality, then we can't claim that it's something that can be universally relied on by app authors.

TingPing:
Closing in favor of #1199

PR #1195, opened 2023-11-11
xdg-document-portal: Fix sqlite3 usage

### LINKED ISSUE #1117: SQLite does not work on the document portal's filesystem
Trying to open an SQLite database using libsqlite or the command line utility does fail both from within the sandbox, and from the host system. Reading from, writing to and creating files works as expected from inside and outside the sandbox. Only SQLite does not work.

The following steps can be used to replicate the issue:

1. Optionally, replace the running instance of xdg-document-portal to observe logs:
    ```
    /usr/lib/xdg-document-portal --replace --verbose
    ```
2. Make xdg-desktop-portal share a folder containing an SQLite database file with a sandboxed application.
3. Using the sqlite3 command, try to open this database, in my case: 
    ```
    ~ $ sqlite3 /run/user/1000/doc/e1b555fb/Musik/test.db
     ```

SQLite fails with the non-descriptive error message:

```
Error: unable to open database "/run/user/1000/doc/e1b555fb/Musik/test.db": unable to open database file
```

The logs from xdg-document-portal contain not very much information besides `error 40`:

```
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: ACCESS b4ee787c774c261d
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: ACCESS b4ee787c774c261d
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: OPEN b4ee787c774c261d RDONLY
XDP: GETATTR b4ee787c774c261d
XDP: GETATTR b4ee787c774c261d
XDP: GETATTR b4ee787c774c261d
XDP: GETATTR b4ee787c774c261d
XDP: FLUSH b4ee787c774c261d
XDP: RELEASE b4ee787c774c261d
XDP: GETATTR 1
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: GETATTR b0cfa07c30fc868b
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: GETATTR b4ee787c774c261d
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: OPEN b4ee787c774c261d RDWR
XDP: OPEN -> error 40
XDP: LOOKUP 5:Musik
XDP: LOOKUP 5:Musik => b0cfa07c30fc868b
XDP: LOOKUP b0cfa07c30fc868b:test.db
XDP: LOOKUP b0cfa07c30fc868b:test.db => b4ee787c774c261d
XDP: OPEN b4ee787c774c261d RDONLY
XDP: OPEN -> error 40
```

I also tried to get more information from libfuse by setting a custom log function using `fuse_set_log_func ()`, but the library doesn't seem to log anything (also, logs should appear in stderr by default).

This issue is loosely related to #463, which does not apply here, because write access to the whole directory is already present. Also, there are a few mentions of the problem with other applications that need to access SQLite databases, like for example here sqlitebrowser/sqlitebrowser#3400.

hfiguiere:
- Fix open with O_NOFOLLOW: Resolve the symlink from /proc before opening
- Implement setlk, getlk

Close #1117

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -1,5 +1,6 @@
 /*
  * Copyright © 2018 Red Hat, Inc
+ * Copyright © 2023 GNOME Foundation Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,6 +17,7 @@
  *
  * Authors:
  *       Alexander Larsson <alexl@redhat.com>
+ *       Hubert Figuière <hub@figuiere.net>
  */
 
 #include "config.h"
@@ -1897,6 +1899,28 @@ xdp_fuse_open (fuse_req_t req,
     return;
 
   path = fd_to_path (inode->physical->fd);
+
+  /*
+   * `path` is a path to the fd entry in `/proc`, which is a symlink
+   * to the actual file. Opening it directly with `O_NOFOLLOW` will
+   * fail. So we should resolve it first then we can honour the no
+   * follow flag.
+   */
+  if (open_flags & O_NOFOLLOW)
+    {
+      ssize_t res;
+      char resolved_path[PATH_MAX] = {0, };
+      res = readlink (path, resolved_path, sizeof (resolved_path));
+
+      if (res == sizeof (resolved_path))
+        return xdp_reply_err (op, req, ENAMETOOLONG);
+      if (res < 0)
+        return xdp_reply_err (op, req, errno);
+
+      g_clear_pointer (&path, g_free);
+      path = g_strdup (resolved_path);
+    }
+
   fd = open (path, open_flags, 0);
   if (fd == -1)
     return xdp_reply_err (op, req, errno);
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -1,3 +1,23 @@
+/*
+ * Copyright © 2018 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
+ */
+
 #include "config.h"
 
 #define FUSE_USE_VERSION 35
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -3161,10 +3161,16 @@ xdp_fuse_setlk (fuse_req_t req,
                 int sleep)
 {
   const char *op = "SETLK";
+  XdpFile *file = (XdpFile *)fi->fh;
+  int res;
 
   g_debug ("SETLK %lx", ino);
 
-  xdp_reply_err (op, req, ENOSYS);
+  res = fcntl (file->fd, F_SETLK, lock);
+  if (res < 0)
+    return xdp_reply_err (op, req, errno);
+
+  return xdp_reply_err (op, req, 0);
 }
 
 static void
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -3147,10 +3147,16 @@ xdp_fuse_getlk (fuse_req_t req,
                 struct flock *lock)
 {
   const char *op = "GETLK";
+  XdpFile *file = (XdpFile *)fi->fh;
+  int res;
 
   g_debug ("GETLK %lx", ino);
 
-  xdp_reply_err (op, req, ENOSYS);
+  res = fcntl (file->fd, F_GETLK, lock);
+  if (res < 0)
+    return xdp_reply_err (op, req, errno);
+
+  return xdp_reply_err (op, req, 0);
 }
 
 static void


hfiguiere:
rebased. all test passed locally.

hfiguiere:
It the first patch that fail the test. But I can't see from the logs what the error is.

hfiguiere:
The modified approach is to special case the `O_NOFOLLOW`

GeorgesStavracas:
@alexlarsson I wonder if you could review this one?

document-portal/document-portal-fuse.c
```diff
@@ -1897,7 +1899,25 @@ xdp_fuse_open (fuse_req_t req,
...
+       */
+      res = readlink (path, actual_path, sizeof (actual_path));
+      if (res == -1)
+        return xdp_reply_err (op, req, errno);
+      actual_path[res] = '\0';
+
+      fd = open (actual_path, open_flags, 0);
+    }
+  else
+    fd = open (path, open_flags, 0);
```
GeorgesStavracas, line 1920: I wonder if, instead of looking for the `open_flags` first, this function should do something like:

```c
  fd = open (path, open_flags, 0);
  if (errno == ELOOP && (open_flags & O_NOFOLLOW))
    {
      char actual_path[PATH_MAX + 1];
      ssize_t res;

      res = readlink (path, actual_path, sizeof (actual_path));

      ...

      fd = open (actual_path, open_flags, 0);
    }
```

The downside is that it would call `open()` twice.

document-portal/document-portal-fuse.c
```diff
@@ -1897,7 +1899,25 @@ xdp_fuse_open (fuse_req_t req,
     return;
 
   path = fd_to_path (inode->physical->fd);
-  fd = open (path, open_flags, 0);
+  if (open_flags & O_NOFOLLOW)
+    {
+      int res;
```
GeorgesStavracas, line 1904: I think this should be an `ssize_t`?

document-portal/document-portal-fuse.c
```diff
@@ -1897,7 +1899,25 @@ xdp_fuse_open (fuse_req_t req,
...
+      /*
+       * `path` is a path to the fd entry in `/proc`, which is a symlink
+       * to the actual file. Opening it directly with `O_NOFOLLOW` will
+       * fail. So we should resolve it first then we can honour the no
+       * follow flag.
+       */
+      res = readlink (path, actual_path, sizeof (actual_path));
+      if (res == -1)
+        return xdp_reply_err (op, req, errno);
+      actual_path[res] = '\0';
```
GeorgesStavracas, line 1915: I think this should perform more validation checks for the output of `readlink`. As a reference, [this is what the BPF kernel tool does](https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/common.c#L387-L395).

document-portal/document-portal-fuse.c
```diff
@@ -3157,10 +3157,16 @@ xdp_fuse_setlk (fuse_req_t req,
                 int sleep)
 {
   const char *op = "SETLK";
+  XdpFile *file = (XdpFile *)fi->fh;
+  int res;
 
   g_debug ("SETLK %lx", ino);
 
-  xdp_reply_err (op, req, ENOSYS);
+  res = fcntl (file->fd, F_SETLK, lock);
```
GeorgesStavracas, line 3165: Is it necessary to also implement `xdp_fuse_getlk`?

GeorgesStavracas (changes requested):

GeorgesStavracas:
Is there any way to actually test this in the testsuite?

hfiguiere:
> Is there any way to actually test this in the testsuite?

Not sure.

But the way I tested it is export a well know slqite3 database (a copy of course), and inside the flatpak run sqlite3 and perform updates and such.
Patch 1: failed at opening
Patch 2: failed when doing a `.dump`

I have run a few more things including dropping tables without hitch.

document-portal/document-portal-fuse.c
hfiguiere: We know that here `path` points to the `/proc`, this is what `fd_to_path ()` returns. So we must resolve this.

Originally I was just resolving it unconditionally but then the tests started failing so I made the just to only make the exception for `O_NOFOLLOW` because we know this will break and kept the existing code path intact.

document-portal/document-portal-fuse.c
hfiguiere: Actually the fix is to pass `sizeof (actual_path) - 1` as we make sure it's null terminated.

hfiguiere: ... and still checking res for the length because if it truncated then it's wrong. and return `ENOENT` as errno.

document-portal/document-portal-fuse.c
hfiguiere: done

document-portal/document-portal-fuse.c
hfiguiere: fixed a bit more thoroughly

document-portal/document-portal-fuse.c
hfiguiere: this is done now.

document-portal/document-portal-fuse.c
```diff
@@ -1897,7 +1899,26 @@ xdp_fuse_open (fuse_req_t req,
...
+      res = readlink (path, actual_path, sizeof (actual_path));
+      if (res == sizeof (actual_path))
+        return xdp_reply_err (op, req, ENAMETOOLONG);
+      if (res < 0)
+        return xdp_reply_err (op, req, errno);
+
+      fd = open (actual_path, open_flags, 0);
+    }
+  else
+    fd = open (path, open_flags, 0);
```
GeorgesStavracas, line 1921: Okay, I think code-wise this is fine now. I'll propose the following style change though:

```c
  path = fd_to_path (inode->physical->fd);

  /* `path` is a path to the fd entry in `/proc`, which is a symlink
   * to the actual file. Opening it directly with `O_NOFOLLOW` will
   * fail. So we should resolve it first then we can honour the no
   * follow flag. */
  if (open_flags & O_NOFOLLOW)
    {
      char resolved_path[PATH_MAX] = { 0, };
      ssize_t res;

      res = readlink (path, resolved_path, sizeof (resolved_path));
      if (res == sizeof (resolved_path))
        return xdp_reply_err (op, req, ENAMETOOLONG);
      if (res < 0)
        return xdp_reply_err (op, req, errno);

      g_clear_pointer (&path, g_free);
      path = g_strdup (resolved_path);
    }

  fd = open (path, open_flags, 0);
```

document-portal/document-portal-fuse.c
```diff
@@ -3158,10 +3158,16 @@ xdp_fuse_setlk (fuse_req_t req,
...
+  int res;
 
   g_debug ("SETLK %lx", ino);
 
-  xdp_reply_err (op, req, ENOSYS);
+  res = fcntl (file->fd, F_SETLK, lock);
+  if (res == 0)
+    return xdp_reply_err (op, req, 0);
+  else
+    return xdp_reply_err (op, req, errno);
```
GeorgesStavracas, line 3176: Seems fine to me. I have looked at other functions and the style seems all over the place. So I'll suggest the following:

```c
  res = fcntl (file->fd, F_SETLK, lock);
  if (res != 0)
    return xdp_reply_err (op, req, errno);

  return xdp_reply_err (op, req, 0);
```

document-portal/document-portal-fuse.c
```diff
@@ -3144,10 +3144,16 @@ xdp_fuse_getlk (fuse_req_t req,
...
+  int res;
 
   g_debug ("GETLK %lx", ino);
 
-  xdp_reply_err (op, req, ENOSYS);
+  res = fcntl (file->fd, F_GETLK, lock);
+  if (res == 0)
+    return xdp_reply_err (op, req, 0);
+  else
+    return xdp_reply_err (op, req, errno);
```
GeorgesStavracas, line 3156: Same here, please use the following code style:

```c
  res = fcntl (file->fd, F_GETLK, lock);
  if (res != 0)
    return xdp_reply_err (op, req, errno);

  return xdp_reply_err (op, req, 0);
```

GeorgesStavracas (changes requested):
Minor comments only. Please apply the style suggestions and we can merge this :)

document-portal/document-portal-fuse.c
hfiguiere: I was trying to avoid an extra allocation.

hfiguiere: done

GeorgesStavracas (approved):

swick:
```python
import errno
import fcntl
import multiprocessing
import os
import struct
from pathlib import Path

import pytest

import tests.xdp_doc_utils as xdp_doc
import tests.xdp_utils as xdp


@pytest.fixture
def xdp_app_info() -> xdp.AppInfo:
    return xdp.AppInfoHost(app_id="")


def lock_region(fd, lock_type, start=0, length=0):
    lockdata = struct.pack("hhllll", lock_type, os.SEEK_SET, start, length, 0, 0)
    try:
        fcntl.fcntl(fd, fcntl.F_SETLK, lockdata)
        return True
    except OSError as e:
        if e.errno in (errno.EAGAIN, errno.EACCES):
            return False
        raise


def child_try_lock(path, result_queue):
    fd = os.open(path, os.O_RDWR)
    try:
        got_lock = lock_region(fd, fcntl.F_WRLCK)
        result_queue.put(got_lock)
        result_queue.get()
    finally:
        os.close(fd)


def child_lock_then_signal(path, locked_event, release_event):
    fd = os.open(path, os.O_RDWR)
    try:
        assert lock_region(fd, fcntl.F_WRLCK)
        locked_event.set()
        release_event.wait()
    finally:
        os.close(fd)


def query_lock(fd, start=0, length=0):
    lockdata = struct.pack("hhllll", fcntl.F_WRLCK, os.SEEK_SET, start, length, 0, 0)
    result = fcntl.fcntl(fd, fcntl.F_GETLK, lockdata)
    return struct.unpack("hhllll", result)[0]


@pytest.fixture
def tmpfile():
    file_path = Path(os.environ["TMPDIR"]) / "lock-test-tmpfile"
    file_path.write_bytes(b"data")
    path = str(file_path)
    return path, path


@pytest.fixture
def docfile(xdg_document_portal, dbus_con):
    documents_intf = xdp.get_document_portal_iface(dbus_con)
    mountpoint = xdp_doc.get_mountpoint(documents_intf)

    file_path = Path(os.environ["TMPDIR"]) / "lock-test-docfile"
    xdp_doc.write_bytes_atomic(file_path, b"data")
    doc_id = xdp_doc.export_file(documents_intf, file_path)
    fuse_path = str(mountpoint / doc_id / "lock-test-docfile")
    return fuse_path, str(file_path)


@pytest.fixture(params=["tmpfile", "docfile"])
def test_file(request):
    return request.getfixturevalue(request.param)


class TestDocumentLocking:
    def test_exclusive_locks_conflict(self, test_file):
        lock_path, _ = test_file

        fd_a = os.open(lock_path, os.O_RDWR)
        try:
            assert lock_region(fd_a, fcntl.F_WRLCK)

            result_queue = multiprocessing.Queue()
            child = multiprocessing.Process(
                target=child_try_lock, args=(lock_path, result_queue)
            )
            child.start()
            try:
                got_lock = result_queue.get(timeout=5)
                assert not got_lock, (
                    "Second process should not acquire a conflicting exclusive lock"
                )
            finally:
                result_queue.put("done")
                child.join(timeout=5)
        finally:
            os.close(fd_a)

    def test_closing_one_fd_does_not_release_other_locks(self, test_file):
        lock_path, real_path = test_file

        locked_event = multiprocessing.Event()
        release_event = multiprocessing.Event()

        proc_a = multiprocessing.Process(
            target=child_lock_then_signal,
            args=(lock_path, locked_event, release_event),
        )
        proc_a.start()
        try:
            assert locked_event.wait(timeout=5), "Process A failed to acquire lock"

            fd_b = os.open(lock_path, os.O_RDWR)
            os.close(fd_b)

            # Query the lock on the real underlying file since the FUSE
            # mount does not implement F_GETLK.
            check_fd = os.open(real_path, os.O_RDWR)
            try:
                lock_type = query_lock(check_fd)
                assert lock_type != fcntl.F_UNLCK, (
                    "Lock should still be held after unrelated fd close"
                )
            finally:
                os.close(check_fd)
        finally:
            release_event.set()
            proc_a.join(timeout=5)


try:
    xdp.ensure_fuse_supported()
except xdp.FuseNotSupportedException as e:
    pytest.skip(f"No fuse support: {e}", allow_module_level=True)
```

swick:
This shows that locking does not work as intended with callers not having a lock when they expect to have one and the reverse. The test uses a tmpfs inode as a baseline to ensure that tests are correct and the failures in the document fuse are legit.

Fixed in https://github.com/flatpak/xdg-desktop-portal/pull/2152 which also adds flock.

PR #1196, opened 2023-11-12
file-transfer: Handle directories

### LINKED ISSUE #911: file transfer does not handle folders
This was reported here: https://gitlab.gnome.org/GNOME/gtk/-/issues/5348

The underlying issue is that AddFiles validates the fds by passing VALIDATE_FD_FILE_TYPE_REGULAR to validate_fd().

I think we should probably add an AddFolders method to the DBus interface.

hfiguiere:
~~Currently On top of #1195 for conveniance.~~

This need further testing but drag and drop of a folder from Nautilus to Amberol works.

Close #911


document-portal/document-portal.h
```diff
@@ -42,7 +42,7 @@ char ** document_add_full (int                      *fd,
                            int                      *parent_dev,
                            int                      *parent_ino,
                            int                       n_args,
-                           DocumentAddFullFlags      flags,
+                           DocumentAddFullFlags     *flags,
```
GeorgesStavracas, line 45: Small suggestion, since flags is now an array with size n_args, it might be clearer to move it to above the int n_args argument.

document-portal/file-transfer.c
```diff
@@ -228,6 +229,7 @@ file_transfer_add_file (FileTransfer *transfer,
 
   file = g_new (ExportedFile, 1);
   file->path = g_strdup (path);
+  file->is_dir = S_ISDIR(parent_st_buf->st_mode) != 0;
```
GeorgesStavracas, line 232: Nit: space before `(`

Comment: I see in other places that `S_ISDIR` is used as a boolean already, maybe you don't need the `!= 0` after all?

document-portal/file-transfer.c
```diff
@@ -281,6 +284,7 @@ file_transfer_execute (FileTransfer *transfer,
   fds = g_new (int, n_fds);
   parent_devs = g_new (int, n_fds);
   parent_inos = g_new (int, n_fds);
+  the_flags = g_new (guint32, n_fds);
```
GeorgesStavracas, line 287: `guint32` → `uint32_t`

document-portal/document-portal.c
```diff
@@ -775,7 +780,7 @@ document_add_full (int                      *fd,
                    int                      *parent_dev,
                    int                      *parent_ino,
                    int                       n_args,
-                   DocumentAddFullFlags      flags,
+                   DocumentAddFullFlags     *the_flags,
```
GeorgesStavracas, line 783: I think I'd prefer if this variable was called `documents_flags`, or `add_full_flags`. Thoughts?

document-portal/file-transfer.c
```diff
@@ -294,11 +298,12 @@ file_transfer_execute (FileTransfer *transfer,
           return NULL;
         }
 
+      the_flags[i] = flags | (file->is_dir ? DOCUMENT_ADD_FLAGS_DIRECTORY : 0);
```
GeorgesStavracas, line 301: Here, I think it would be slightly better to rename the `flags` variable to `common_flags`, and name the array `documents_flags`

GeorgesStavracas (changes requested):

document-portal/document-portal.c
hfiguiere: `documents_flags` is a good idea.

document-portal/file-transfer.c
hfiguiere: `DocumentAddFullFlags` is the actual type. Will use that instead.

document-portal/file-transfer.c
hfiguiere: good catch. it came from code that changed. This is definitely a boolean.

hfiguiere:
Dropping single file now work (was broken).

Also fixed a memory leak found while this error was happening

hfiguiere:
Fixed the tests. It was a stupid mistake, note a fundamental issue with the patch.

To be noted that the FileTransfer D-Bus API documentation extends to directories. I don't thinks it is necessary to change the actual API. It will error on older version like it did before.

Demo:

[Screencast from 2023-11-15 21-42-01.webm](https://github.com/flatpak/xdg-desktop-portal/assets/114441/fa514b50-0e13-474e-ab93-46294d0bee24)

hfiguiere:
In the above screencast, the only code change is this PR. Nautilus is from Fedora and Amberol from flathub.

GeorgesStavracas (commented):
This looks reasonable to me now. However, I think it would be good to add at least 2 tests for this either in `test-doc-portal.c` and/or in `test-document-fuse.h`

document-portal/document-portal.c
```diff
@@ -704,6 +727,7 @@ portal_add_full (GDBusMethodInvocation *invocation,
 
   n_args = g_variant_n_children (array);
   fd = g_new (int, n_args);
+  documents_flags = g_new (DocumentAddFullFlags, n_args);
```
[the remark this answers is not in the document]
hfiguiere: Given that n_args comes from the DBus arguments, I have mixed feelings here.

document-portal/file-transfer.c
```diff
@@ -451,7 +457,7 @@ retrieve_files (GDBusMethodInvocation *invocation,
   const char *key;
   FileTransfer *transfer;
   g_auto(GStrv) files = NULL;
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
```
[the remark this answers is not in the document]
hfiguiere: done

document-portal/document-portal.c
```diff
@@ -697,6 +699,7 @@ portal_add_full (GDBusMethodInvocation *invocation,
   GDBusMessage *message;
   GUnixFDList *fd_list;
   g_autofree int *fd = NULL;
+  g_autofree DocumentAddFullFlags *documents_flags;
```
GeorgesStavracas, line 702: Please initialize this to NULL

document-portal/document-portal.c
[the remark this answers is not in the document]
GeorgesStavracas: I'm fine with either, but I'm leaning slightly towards @hfiguiere's reasoning. It's not like this is a hot path or a memory intensive path anyway.

document-portal/document-portal.c
hfiguiere: oops. my bad. done.

hfiguiere:
> This looks reasonable to me now. However, I think it would be good to add at least 2 tests for this either in `test-doc-portal.c` and/or in `test-document-fuse.h`

Some of this code is exercised already as it found some bugs in the new code.

There is no FileTransfer test. Working on this.

document-portal/document-portal.c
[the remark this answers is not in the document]
matthiasclasen: Yeah, seems fine to defer that optimization until it shows up on profiles.

document-portal/file-transfer.c
```diff
@@ -426,7 +432,7 @@ add_files (GDBusMethodInvocation *invocation,
           return;
         }
 
-      if (!validate_fd (fd, app_info, VALIDATE_FD_FILE_TYPE_REGULAR, &st_buf, &parent_st_buf, &path, &fd_is_writable, NULL) ||
+      if (!validate_fd (fd, app_info, VALIDATE_FD_FILE_TYPE_ANY, &st_buf, &parent_st_buf, &path, &fd_is_writable, NULL) ||
```
matthiasclasen, line 435: Should we keep using REGULAR here for fds that are not referring to directories?

document-portal/document-portal.h
```diff
@@ -41,8 +41,8 @@ gboolean validate_fd (int           fd,
 char ** document_add_full (int                      *fd,
                            int                      *parent_dev,
                            int                      *parent_ino,
+                           DocumentAddFullFlags     *documents_flags,
                            int                       n_args,
-                           DocumentAddFullFlags      flags,
                            XdpAppInfo               *app_info,
```
matthiasclasen, line 46: This change is so that each fd can get its own set of flags?

An alternative to this flag array would be to let the caller split up their fd list into files and directories, and do separate calls (with one set of flags each) to add them.

Not sure that is better. It would let us avoid this array of flags, but it is certainly more work for the callers.

document-portal/file-transfer.c
hfiguiere: ANY mean either DIR or REGULAR, checked against S_ISREG () or S_ISDIR (). So this does exactly what we want.

document-portal/document-portal.h
hfiguiere: I felt having `document_add_full()` handle it was really simpler on every aspects. Settings the array of flags is easier than splitting the fd list.

hfiguiere:
> This looks reasonable to me now. However, I think it would be good to add at least 2 tests for this either in `test-doc-portal.c` and/or in `test-document-fuse.h`

#1206 implement a lot of testing for the FileTransfer portal prior to the MR.

hfiguiere:
On top of #1208 so we get the tests.

GeorgesStavracas (approved):
LGTM now, and the tests are :pinched_fingers:

PR #1203, opened 2023-11-16
Add SchemeSupported to the OpenURI portal

xhorak:
To check whenever specific scheme is supported by the host.

### DIFF
--- a/data/org.freedesktop.portal.OpenURI.xml
+++ b/data/org.freedesktop.portal.OpenURI.xml
@@ -29,7 +29,7 @@
        URIs (e.g. a http: link to the applications homepage)
        under the control of the user.
 
-       This documentation describes version 4 of this interface.
+       This documentation describes version 5 of this interface.
   -->
   <interface name="org.freedesktop.portal.OpenURI">
     <!--
@@ -161,6 +161,24 @@
       <arg type="h" name="fd" direction="in"/>
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        SchemeSupported:
+        @scheme: The URI protocol scheme to check
+        @options: Vardict with optional further information
+        @supported: True if the URI scheme is supported
+
+        Checks if OpenURI supports the protocol scheme @scheme.
+
+        The @options vardict currently has no supported entries.
+
+        The SchemeSupported method was introduced in version 5 of the OpenURI portal API.
+    -->
+    <method name="SchemeSupported">
+      <arg type="s" name="scheme" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="b" name="supported" direction="out"/>
     </method>
 
     <property name="version" type="u" access="read"/>
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -926,6 +926,31 @@ handle_open_in_thread_func (GTask *task,
 }
 
 static gboolean
+handle_scheme_supported (XdpDbusOpenURI *object,
+                         GDBusMethodInvocation *invocation,
+                         const gchar *arg_scheme,
+                         GVariant *arg_options)
+{
+  g_autoptr(GAppInfo) app_info = NULL;
+
+  if (arg_scheme == NULL || *arg_scheme == '\0')
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR,
+                                             XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             "Scheme not specified");
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  app_info = g_app_info_get_default_for_uri_scheme (arg_scheme);
+
+  g_debug ("Handler for scheme: %s%s found.", arg_scheme, app_info ? "" : " not");
+  g_dbus_method_invocation_return_value (invocation, g_variant_new ("(b)", app_info != NULL));
+
+  return G_DBUS_METHOD_INVOCATION_HANDLED;
+}
+
+static gboolean
 handle_open_uri (XdpDbusOpenURI *object,
                  GDBusMethodInvocation *invocation,
                  const gchar *arg_parent_window,
@@ -1094,12 +1119,13 @@ open_uri_iface_init (XdpDbusOpenURIIface *iface)
   iface->handle_open_uri = handle_open_uri;
   iface->handle_open_file = handle_open_file;
   iface->handle_open_directory = handle_open_directory;
+  iface->handle_scheme_supported = handle_scheme_supported;
 }
 
 static void
 open_uri_init (OpenURI *openuri)
 {
-  xdp_dbus_open_uri_set_version (XDP_DBUS_OPEN_URI (openuri), 4);
+  xdp_dbus_open_uri_set_version (XDP_DBUS_OPEN_URI (openuri), 5);
 }
 
 static void
--- a/tests/openuri.c
+++ b/tests/openuri.c
@@ -5,6 +5,7 @@
 #include <libportal/portal.h>
 #include "xdp-utils.h"
 #include "xdp-impl-dbus.h"
+#include "xdp-dbus.h"
 
 #include "utils.h"
 
@@ -477,4 +478,57 @@ test_open_directory (void)
 
   while (!got_info)
     g_main_context_iteration (NULL, TRUE);
+}
+
+void
+test_scheme_supported (void)
+{
+  gboolean supported;
+  GVariantBuilder builder;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) options = NULL;
+  g_autoptr(GDBusConnection) session_bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &error);
+
+  XdpDbusOpenURI *proxy = xdp_dbus_open_uri_proxy_new_sync (session_bus,
+                                                            G_DBUS_PROXY_FLAGS_NONE,
+                                                            "org.freedesktop.portal.Desktop",
+                                                            "/org/freedesktop/portal/desktop",
+                                                            NULL,
+                                                            &error);
+
+  g_assert_no_error (error);
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE_VARDICT);
+  options = g_variant_ref_sink (g_variant_builder_end (&builder));
+  /* Existing scheme */
+  xdp_dbus_open_uri_call_scheme_supported_sync (proxy,
+                                                "https",
+                                                options,
+                                                &supported,
+                                                NULL,
+                                                &error);
+  g_assert_no_error (error);
+  g_assert_true (supported);
+
+  /* Non existing scheme */
+  xdp_dbus_open_uri_call_scheme_supported_sync (proxy,
+                                                "bogusnonexistanthandler",
+                                                options,
+                                                &supported,
+                                                NULL,
+                                                &error);
+  g_assert_no_error (error);
+  g_assert_false (supported);
+
+  /* Missing scheme name */
+  xdp_dbus_open_uri_call_scheme_supported_sync (proxy,
+                                                "",
+                                                options,
+                                                &supported,
+                                                NULL,
+                                                &error);
+  g_assert_error (error,
+                  XDG_DESKTOP_PORTAL_ERROR,
+                  XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT);
+  g_object_unref (proxy);
 }
--- a/tests/test-portals.c
+++ b/tests/test-portals.c
@@ -448,7 +448,7 @@ DEFINE_TEST_EXISTS(inhibit, INHIBIT, 3)
 DEFINE_TEST_EXISTS(location, LOCATION, 1)
 DEFINE_TEST_EXISTS(network_monitor, NETWORK_MONITOR, 3)
 DEFINE_TEST_EXISTS(notification, NOTIFICATION, 1)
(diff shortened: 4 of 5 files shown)


grulja:
I think this change as it is, is a bit incomplete. If an app/client is supposed to call this API, it needs to be added to the `org.freedesktop.portal.OpenURI` instead, because your addition is what backend implementations (xdp-kde/xdp-gnome) are supposed to implement, but these APIs are not called directly by clients and possibly not even visible outside the sandbox by default. With an additional API, you should also bump a version of the `org.freedesktop.portal.OpenURI` so clients can check whether the version of xdp that is running on the host supports it. I'm not sure a backend counterpart is needed as you can probably tell the host can handle given scheme using the `g_app_*` APIs (I'm not an expert in this regard). That would mean that also the implementation of this additional API is missing in this MR.

tests/test-portals.c
```diff
@@ -444,7 +444,7 @@ DEFINE_TEST_EXISTS(inhibit, INHIBIT, 3)
 DEFINE_TEST_EXISTS(location, LOCATION, 1)
 DEFINE_TEST_EXISTS(network_monitor, NETWORK_MONITOR, 3)
 DEFINE_TEST_EXISTS(notification, NOTIFICATION, 1)
-DEFINE_TEST_EXISTS(open_uri, OPEN_URI, 3)
+DEFINE_TEST_EXISTS(open_uri, OPEN_URI, 4)
```
grulja, line 447: This feels like saying we have test coverage for OpenURI in version 4, but we don't. Would it be possible to add a test for this? Since it's just a simple call, it probably should be fairly trivial.

src/open-uri.c
```diff
@@ -919,6 +919,27 @@ handle_open_in_thread_func (GTask *task,
...
+{
+  if (arg_scheme == NULL || *arg_scheme == '\0')
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR,
+                                             XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             "Scheme not specified");
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+  GAppInfo* app_info = g_app_info_get_default_for_uri_scheme (arg_scheme);
```
grulja, line 935: You are leaking `GAppInfo`. 
The docs says: The caller of the function takes ownership of the data, and is responsible for freeing it.

grulja (changes requested):

grulja:
@xhorak also please squash your commits or split it to reasonable parts and remove the merge commit.

tests/test-portals.c
xhorak: Currently I'm not completely sure what to test. Maybe someone could direct me to the similar test?

xhorak:
@matthiasclasen Could you please check this one?

grulja:
Also, maybe this can be permission based? To address @matthiasclasen concerns? 

Like, first ask will popup a permission dialog, to ask whether you want to allow given application access to this information, store it for this app (app_id) and have all the following ones without any permission dialog needed in case the permission has been granted already? 

That would mean using the generic Access portal for this.

data/org.freedesktop.portal.OpenURI.xml
```diff
@@ -161,6 +161,20 @@
       <arg type="o" name="handle" direction="out"/>
     </method>
 
+    <!--
+        SchemeSupported:
+        @scheme: The URI protocol scheme to check
+        @supported: True when URI scheme is supported on the host system
+
+        Asks whenever host system supports the protocol scheme.
```
swick, line 169:
```suggestion
        @scheme: The URI protocol scheme to check
        @supported: True if the URI scheme is supported

        Checks if OpenURI supports the protocol scheme @scheme.
```

swick:
Seems perfectly reasonable.

data/org.freedesktop.portal.OpenURI.xml
```diff
@@ -163,6 +163,20 @@
...
     </method>
 
+    <!--
+        SchemeSupported:
+        @scheme: The URI protocol scheme to check
+        @supported: True if the URI scheme is supported
+
+        Checks if OpenURI supports the protocol scheme @scheme.
+
+        The SchemeSupported method was introduced in version 4 of the OpenURI portal API.
```
grulja, line 173: Version 5.

grulja (commented):
You also need to squash all the commits into one.

grulja:
I think the test is failing because you need to also bump [the version in the test file](https://github.com/flatpak/xdg-desktop-portal/blob/main/tests/test-portals.c#L451). Still better would be to not just bump it, but also wrote a simple test.

xhorak:
> Seems perfectly reasonable. What are the concerns about exposing this by default (i.e. not gated behind some permission)?
> 
> The xml still has to be adjusted and a test would be nice, as pointed out by @grulja.

Looking at the tests, they seems to be mostly implemented by libportal. Maybe we could add more when the libportal adds SchemeSupported implementation?

grulja:
> > Seems perfectly reasonable. What are the concerns about exposing this by default (i.e. not gated behind some permission)?
> > The xml still has to be adjusted and a test would be nice, as pointed out by @grulja.
> 
> Looking at the tests, they seems to be mostly implemented by libportal. Maybe we could add more when the libportal adds SchemeSupported implementation?

You can make a simple sync DBus call there.

data/org.freedesktop.portal.OpenURI.xml
grulja: Still remains to be changed to version 5.

grulja (commented):
LGTM. I'm not an expert to GLib, but it looks sane to me.

NIT: still missing to bump the version in the xml file

tests/openuri.c
```diff
@@ -478,3 +478,61 @@ test_open_directory (void)
...
+      "org.freedesktop.portal.OpenURI",
+      NULL,
+      &error);
+
+    g_assert_no_error (error);
+
+    GVariant *result;
+    gboolean supported;
+
+    // Existing scheme
```
swick, line 501: This kind of comment isn't technically in the C spec, so we always use `/* */`

data/org.freedesktop.portal.OpenURI.xml
```diff
@@ -163,6 +163,20 @@
...
+        @supported: True if the URI scheme is supported
+
+        Checks if OpenURI supports the protocol scheme @scheme.
+
+        The SchemeSupported method was introduced in version 5 of the OpenURI portal API.
+    -->
+    <method name="SchemeSupported">
+      <arg type="s" name="scheme" direction="in"/>
+      <arg type="b" name="supported" direction="out"/>
+    </method>
```
swick, line 178: You might want to add an options `a{sv}` parameter (with no supported keys for now). More often than not, we notice that we'd like to extend a method at some point.

src/open-uri.c
```diff
@@ -925,6 +925,28 @@ handle_open_in_thread_func (GTask *task,
...
+  if (arg_scheme == NULL || *arg_scheme == '\0')
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR,
+                                             XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             "Scheme not specified");
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  g_autoptr(GAppInfo) app_info = g_app_info_get_default_for_uri_scheme (arg_scheme);
```
swick, line 942: Declarations at the top, so `g_autoptr(GAppInfo) app_info = NULL;` at the top and `app_info = g_app_info_get_default_for_uri_scheme (arg_scheme);` here.

tests/openuri.c
```diff
@@ -478,3 +478,61 @@ test_open_directory (void)
   while (!got_info)
     g_main_context_iteration (NULL, TRUE);
 }
+
+void
+test_scheme_supported(void) {
+  g_autoptr(GError) error = NULL;
+
+    g_autoptr(GDBusProxy) proxy = g_dbus_proxy_new_for_bus_sync (
```
swick, line 486:
```suggestion
void
test_scheme_supported(void)
{
  g_autoptr(GError) error = NULL;
  g_autoptr(GDBusProxy) proxy = g_dbus_proxy_new_for_bus_sync (
```

swick, line 486: Instead of using the generic `GDBusProxy` you can use the generated `XdpDbusOpenURI *xdp_dbus_open_uri_proxy_new_sync()` and related functions.

tests/openuri.c
```diff
@@ -478,3 +479,59 @@ test_open_directory (void)
...
 }
+
+void
+test_scheme_supported(void) 
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) result = NULL;
+  g_autoptr(GDBusConnection) session_bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &error);
+
+  XdpDbusOpenURI *proxy = xdp_dbus_open_uri_proxy_new_sync (session_bus,
```
swick, line 490:
```suggestion
  g_autoptr(XdpDbusOpenURI) proxy = xdp_dbus_open_uri_proxy_new_sync (session_bus,
```

tests/openuri.c
```diff
@@ -478,3 +479,59 @@ test_open_directory (void)
...
+                                                            "org.freedesktop.portal.Desktop",
+                                                            "/org/freedesktop/portal/desktop",
+                                                            NULL,
+                                                            &error);
+
+  g_assert_no_error (error);
+
+  gboolean supported;
+
+  GVariantBuilder builder;
```
swick, line 501: declaration on the top

tests/openuri.c
```diff
@@ -478,3 +479,59 @@ test_open_directory (void)
...
+  g_variant_builder_init (&builder, G_VARIANT_TYPE_VARDICT);
+  
+  /* Existing scheme */
+  xdp_dbus_open_uri_call_scheme_supported_sync (proxy,
+                                                "https",
+                                                g_variant_builder_end (&builder),
```
swick, line 507: can reuse the empty vardict:
```c
g_autoptr(GVariant) options = g_variant_ref_sink (g_variant_builder_end (&builder));
```
and pass that to the calls.

tests/openuri.c
```diff
@@ -478,3 +479,59 @@ test_open_directory (void)
   while (!got_info)
     g_main_context_iteration (NULL, TRUE);
 }
+
+void
+test_scheme_supported(void)
```
swick, line 484:
```suggestion
test_scheme_supported (void) 
```

data/org.freedesktop.portal.OpenURI.xml
```diff
@@ -163,6 +163,22 @@
       <arg type="o" name="handle" direction="out"/>
     </method>
 
+    <!--
+        SchemeSupported:
+        @scheme: The URI protocol scheme to check
+        @options: Vardict with optional further information - currently unused 
+        @supported: True if the URI scheme is supported
+
+        Checks if OpenURI supports the protocol scheme @scheme.
```
swick, line 172:
```suggestion
        @options: Vardict with optional further information 
        @supported: True if the URI scheme is supported

        Checks if OpenURI supports the protocol scheme @scheme.

        The @options vardict currently has no supported entries.
```

tests/openuri.c
xhorak: Hm, looks like it does not work fine:
```
../tests/openuri.c: In function ‘test_scheme_supported’:
../tests/openuri.c:493:13: error: unknown type name ‘XdpDbusOpenURI_autoptr’; did you mean ‘XdpDbusOpenURIProxy’?
  493 |   g_autoptr(XdpDbusOpenURI) proxy = xdp_dbus_open_uri_proxy_new_sync (session_bus,
      |             ^~~~~~~~~~~~~~
../tests/openuri.c:493:3: error: cleanup argument not a function
  493 |   g_autoptr(XdpDbusOpenURI) proxy = xdp_dbus_open_uri_proxy_new_sync (session_bus,
      |   ^~~~~~~~~
```
Looks like XdpDbusOpenURI is not g_autoptr compatible.

swick: Huh, indeed. We turn off autocleanup for most GDbus generated sources. I'll try to figure out if we can enable it but for this PR manually unref'ing is okay then.

swick:
In the file `org.freedesktop.portal.OpenURI.xml` at line 32 it still says version 4.

swick:
Otherwise this LGTM!

tests/openuri.c
swick: When https://github.com/flatpak/xdg-desktop-portal/pull/1468 gets merged this shouldn't be an issue anymore.

xhorak:
> In the file `org.freedesktop.portal.OpenURI.xml` at line 32 it still says version 4.

Oh sorry, I've somehow only focused on the line 176.

tests/openuri.c
swick: It has already been merged so you could rebase and update the code to use g_autoptr again.

src/open-uri.c
```diff
@@ -925,6 +925,31 @@ handle_open_in_thread_func (GTask *task,
...
 }
 
+static gboolean
+handle_scheme_supported (XdpDbusOpenURI *object,
+                         GDBusMethodInvocation *invocation,
+                         const gchar *arg_scheme,
+                         GVariant *arg_options)
+{
+  g_autoptr(GAppInfo) app_info = NULL;
+
```
matthiasclasen, line 935: I think if we add options, we should check them and filter out unsupported options (ie all of them, currently)

matthiasclasen: Apart from that, +1 from me.

grulja:
@swick since 1.19.0 is not out yet, can we get this change in?

jadahl:
> can we get this change in?

This merge request was merged 3 weeks ago.

PR #1206, opened 2023-11-18
Add FileTransfer portal test

hfiguiere:
This should land before #1196 or require changes.

Probably more testing of the output should be done.

### DIFF
--- a/tests/test-document-fuse.py
+++ b/tests/test-document-fuse.py
@@ -1,5 +1,25 @@
 #!/usr/bin/env python3
 
+# Copyright © 2020 Red Hat, Inc
+# Copyright © 2023 GNOME Foundation Inc.
+#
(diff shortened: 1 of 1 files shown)


hfiguiere:
One part of the file transfer portal that is not tested is the part where the files are sent to a different application.

TODO.

hfiguiere:
> One part of the file transfer portal that is not tested is the part where the files are sent to a different application.
> 
> TODO.

The only way to test this is to run the test from inside a flaptak. I don't have a simple strategy here.

In `file_transfer_execute`:

```C
  /* if the target is unsandboxed, just return the files as-is */
  if (xdp_app_info_is_host (target_app_info))
```

hfiguiere:
Still better than last time and increasing the coverage.

GeorgesStavracas (approved):

PR #1223, opened 2023-11-24
docs: properly detect dependencies for documentation

hfiguiere:
I tried to build the documentation and I was missing several dependencies.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -64,6 +64,7 @@ src_includes = include_directories('src')
 i18n = import('i18n')
 gnome = import('gnome')
 pkgconfig = import('pkgconfig')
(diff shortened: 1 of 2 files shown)


GeorgesStavracas (approved):

JakobDev:
sphinxext.opengraph and sphinx_copybutton should be optional

hfiguiere:
if you don't explicitly request to build the documentation it's fine. If you want to build the documentation, with the current setup it's necessary. This doesn't change the requirements, it just make them explicit.

JakobDev:
We could make it optional for building the documentation

PR #1226, opened 2023-11-25
Fix social card

razzeee:
This should fix the social card, as the image is now copied correctly


GeorgesStavracas (approved):

razzeee:
Confirmed working on live system now :tada:

PR #1231, opened 2023-11-29
restore-token: Store data with a single D-Bus call

GeorgesStavracas:
Replace the combination of set_permission_sync() + the manual call to SetValue() with a single call to Set().

This removes the unnecessary extra D-Bus call, and makes the world an tiny little bit of a better place.


jadahl:
> This removes the unnecessary extra D-Bus call, and makes the world an tiny little bit of a better place.

It sure does!

PR #1239, opened 2023-12-09
doc: Fix broken rST refs

tytan652:
Unfortunately relies on https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3751 to have fixed anchors generation to fix them all.

Fix all broken rST refs  (red links) in the documentation.


GeorgesStavracas (approved):

GeorgesStavracas:
I'm going to merge it regardless of the GLib fix, because this is already broken and will continue to be as long as GLib updates.

Issue #1111, opened 2023-09-23
The portal config file does not actually specify the order of portals to be tried


q234rty:
The [documentation](https://github.com/flatpak/xdg-desktop-portal/blob/1.18.0/doc/portals-conf.rst#file-format) suggests that

> Each key in the group contains a semi-colon separated list of portal backend
implementation, to be searched for an implementation of the requested interface,
**in the same order as specified in the configuration file**.

However, this does not seems to match the actual behavior: In KDE with
```
[preferred]
org.freedesktop.impl.portal.FileChooser=gtk;kde
```
The kde file chooser is still opened despite `gtk` being specified first.

Looking at the code, it seems that the implementation are sorted based on UseIn and name in https://github.com/flatpak/xdg-desktop-portal/blob/1.18.0/src/portal-impl.c#L317 , then tried in that order in https://github.com/flatpak/xdg-desktop-portal/blob/1.18.0/src/portal-impl.c#L550 , so instead of searching in the same order as in the config and returning the first portal implementing the interface, which is what the documentation suggests, it searches in the "old" order based on UseIn and name, using the config only as an allowed list of portals.

GeorgesStavracas:
hm, no, probably not

alyssais:
I've opened https://github.com/flatpak/xdg-desktop-portal/pull/1240 to try to fix this.

GeorgesStavracas:
Thanks, very much appreciated

PR #1240, opened 2023-12-12
portal-impl: fix config ordering

alyssais:
Quoting portals.conf(5):

> Each key in the group contains a semi-colon separated list of portal backend
> implementation, to be searched for an implementation of the requested interface,
> in the same order as specified in the configuration file.

But this wasn't actually true.  If the portals were set to z;y and z and y both implemented the same portal, y would be used.

Fixing this requires reworking how portals are selected — going through the config file and selecting the first available configured portal, rather than going through each known portal and checking whether it's allowed in the config file.

find_all_portal_implementations() is unchanged.  The portal-first approach is fine for it, and it would be difficult for it to share as much code with find_portal_implementation() now that it's written to stop as soon as a configured portal is found.

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/1111

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -548,6 +548,77 @@ warn_please_use_portals_conf (void)
                   "configuration file");
 }
 
+static PortalImplementation *
+find_fallback_portal_implementation (const char *interface)
+{
+  /* x-d-p-gtk has historically been the portal UI backend used by desktop
+   * environments with no backend of their own.
+   * If it isn't installed, that is not an error: we just don't use it. */
+  for (const GList *l = implementations; l != NULL; l = l->next)
+    {
+      PortalImplementation *impl = l->data;
+
+      if (!g_str_equal (impl->dbus_name, "org.freedesktop.impl.portal.desktop.gtk"))
+        continue;
+
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
+        continue;
+
+      g_warning ("Choosing %s.portal for %s as a last-resort fallback",
+                 impl->source, interface);
+      warn_please_use_portals_conf ();
+      return impl;
+    }
+
+  return NULL;
+}
+
+static PortalImplementation *
+find_any_portal_implementation (const char *interface)
+{
+  for (const GList *l = implementations; l != NULL; l = l->next)
+    {
+      PortalImplementation *impl = l->data;
+
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
+        continue;
+
+      g_debug ("Falling back to %s.portal for %s", impl->source, interface);
+      return impl;
+    }
+
+  return find_fallback_portal_implementation (interface);
+}
+
+static PortalImplementation *
+find_portal_implementation_iface (const PortalInterface *iface)
+{
+  for (size_t i = 0; iface->portals && iface->portals[i]; i++)
+    {
+      const char *portal = iface->portals[i];
+
+      g_debug ("Found '%s' in configuration for %s", portal, iface->dbus_name);
+
+      if (g_str_equal (portal, "none"))
+        return NULL;
+
+      if (g_str_equal (portal, "*"))
+        return find_any_portal_implementation (iface->dbus_name);
+
+      for (const GList *l = implementations; l != NULL; l = l->next)
+        {
+          PortalImplementation *impl = l->data;
+
+          if (g_str_equal (portal, impl->source))
+            return impl;
+        }
+
+      g_debug ("%s.portal is unrecognized", portal);
+    }
+
+  return find_any_portal_implementation (iface->dbus_name);
+}
+
 PortalImplementation *
 find_portal_implementation (const char *interface)
 {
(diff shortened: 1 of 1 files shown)


src/portal-impl.c
```diff
@@ -548,25 +548,107 @@ warn_please_use_portals_conf (void)
...
+       * to allow for specific interfaces to override the default
+       */
+      for (i = 0; i < config->n_ifaces; i++)
+        {
+          const PortalInterface *iface = config->interfaces[i];
+
+          if (g_strcmp0 (iface->dbus_name, interface) == 0)
+	    {
+	      impl = find_portal_implementation_iface (iface);
+	      goto skip_default;
```
GeorgesStavracas, line 640: Hm, I'd rather avoid this `goto`. I think this whole block can be written as:

```c
  if (config)
    {
      PortalImplementation *impl = NULL;

      /* Interfaces have precedence over the "default" catch all,
       * to allow for specific interfaces to override the default
       */
      for (i = 0; i < config->n_ifaces; i++)
        {
          const PortalInterface *iface = config->interfaces[i];

          if (g_strcmp0 (iface->dbus_name, interface) == 0)
            {
              impl = find_portal_implementation_iface (iface);
              break;
            }
        }

      if (impl == NULL)
        impl = find_portal_implementation_iface (config->default_portal);

      if (impl != NULL)
        {
          g_debug ("Using %s.portal for %s (config)", impl->source, interface);
          return impl;
        }
    }
```

src/portal-impl.c
```diff
@@ -548,25 +548,107 @@ warn_please_use_portals_conf (void)
...
+        }
+
+      impl = find_portal_implementation_iface (config->default_portal);
+
+    skip_default:
+      if (impl != NULL)
+	{
+	  g_debug ("Using %s.portal for %s (config)", impl->source, interface);
+	  return impl;
+	}
```
GeorgesStavracas, line 651: No tabs please

GeorgesStavracas (changes requested):

src/portal-impl.c
alyssais: That wouldn't have the same behaviour.  If somebody sets a specific interface to none, your code would have it fall back to the default implementation, which isn't what the configuration asked for.  There are two different null-ish states here — a portal can be specifically disabled, or it can not be explicitly configured, and both cases need to be handled differently.

GeorgesStavracas: Use a `gboolean` variable then

alyssais: Done.

src/portal-impl.c
```diff
@@ -548,21 +548,106 @@ warn_please_use_portals_conf (void)
...
+      g_debug ("%s.portal is unrecognized", *portal);
+    }
+
+  return find_any_portal_implementation (iface->dbus_name);
+}
+
 PortalImplementation *
 find_portal_implementation (const char *interface)
 {
+  PortalImplementation *impl = NULL;
```
GeorgesStavracas, line 623: I didn't notice this in the first round - any particular reason to be moving this variable here? I think I'd prefer if it's redeclared inside each scope its used

src/portal-impl.c
```diff
@@ -548,21 +548,106 @@ warn_please_use_portals_conf (void)
...
+      return impl;
+    }
+
+  return find_fallback_portal_implementation (interface);
+}
+
+static PortalImplementation *
+find_portal_implementation_iface (const PortalInterface *iface)
+{
+  for (char *const *portal = iface->portals; *portal; portal++)
```
GeorgesStavracas, line 596:
```c
for (size_t i = 0; iface->portals && iface->portals[i]; i++)
  {
    const char *portal = iface->portals[i];
```

GeorgesStavracas (changes requested):
Just two minor changes I didn't spot before, LGTM other than that

src/portal-impl.c
```diff
@@ -576,7 +663,7 @@ find_portal_implementation (const char *interface)
     {
      for (l = implementations; l != NULL; l = l->next)
         {
-          PortalImplementation *impl = l->data;
+	  PortalImplementation *impl = l->data;
```
GeorgesStavracas, line 666: Sorry to be a pain, but no tabs please

GeorgesStavracas (approved):
LGTM with the final comment above, but I'd like to have @ebassi's review before landing

swick:
ping @ebassi

alyssais:
(I'm not sure to what extent https://github.com/flatpak/xdg-desktop-portal/commit/1394bb2817254eaac20648db7235b01bc8c69666 affected this — there's definitely some overlap, because I also fixed that bug.  But given it's caused complicated merge conflicts, and it was merged instead of this despite my PR being open way longer, I want to feel confident that's not going to happen again before I put any more time in.)

SoumyaRanjanPatnaik:
> (I'm not sure to what extent [1394bb2](https://github.com/flatpak/xdg-desktop-portal/commit/1394bb2817254eaac20648db7235b01bc8c69666) affected this — there's definitely some overlap, because I also fixed that bug. But given it's caused complicated merge conflicts, and it was merged instead of this despite my PR being open way longer, I want to feel confident that's not going to happen again before I put any more time in.)

Hi @alyssais, I'm the author of [1394bb2](https://github.com/flatpak/xdg-desktop-portal/commit/1394bb2817254eaac20648db7235b01bc8c69666). I was aware of this PR when I wrote that patch, and I tried it out see if the bug was already fixed (it didn't). 

I think the only reason that my PR was merged faster than yours was that it was bumped up by several users from the Arch community. 

I think this is a very important PR that I think should definitely get merged since it has been the cause of bugs like [this](https://github.com/regolith-linux/regolith-desktop/issues/1003#issuecomment-2091623531) in the distribution I maintain.

SoumyaRanjanPatnaik:
LMK if you need any help with testing or resolving conflicts.

alyssais:
@SoumyaRanjanPatnaik if you want to resolve the conflicts that'd be great!  I had a go but I found it difficult because it's been so long I've forgotten a lot of the nuances I was trying to get right here.

smcv:
@SoumyaRanjanPatnaik, if you can provide a separate PR that supersedes this one, using `Co-authored-by` to credit both yourself and @alyssais, that would probably be the way forward here.

SoumyaRanjanPatnaik:
> @SoumyaRanjanPatnaik, if you can provide a separate PR that supersedes this one, using `Co-authored-by` to credit both yourself and @alyssais, that would probably be the way forward here.

@smcv I've created a PR that supersedes this one.

GeorgesStavracas:
#1378 was merged. Apologies for this mess.

Issue #1234, opened 2023-12-04
Potential regression in the document portal


sonnyp:
### Operating System

Fedora

### XDG Desktop Portal version

1.18

### XDG Desktop Portal version (Other)

1.18.2

### Desktop Environment

GNOME

### Expected Behavior

rust works fine in the document portal

### Current Behavior

rustc throws fs related errors such as

```
error: couldn't get the current directory of the process

Caused by:
  No such file or directory (os error 2)
```

or

```
error: Current directory is invalid: No such file or directory (os error 2)

error: could not compile `demo` (lib) due to previous error
```

### Steps to Reproduce

1. Install Workbench
2. Open the Library
3. Select the "Welcome" entry in Rust
4. Make a change to a string or something
5. Close window, it will prompt to save the project, do so
6. In Workbench menu, open the project with "Open Project"
7. Press Run

If you reproduce the issue, it will happen until reboot.
A reboot might solve it

### Anything else we should know?

Reproducible with: 1.18.2
Not reproducible with 1.18.0

Mikenux:
I installed Workbench (via Software) and launched it, but had to install the Rust extensions. I installed the extensions, then closed Workbench, restarted it, and no reproduction. I uninstalled by deleting data (via Software), except Rust extensions. I reinstalled it and no reproduction.

hfiguiere:
I can't reproduce with these steps.

I run xdg-document-portal 1.18.2 from Fedora.

hfiguiere:
Above I saved the project. 

When I re-open it and press "Run" I can reproduce.

hfiguiere:
The error:
```
  Downloaded 166 crates (10.5 MB) in 1.79s (largest was `linux-raw-sys` at 1013.8 KB)
   Compiling demo v1.0.0 (/run/flatpak/doc/265c72bf/TestWorkbenchRust)
error: Current directory is invalid: No such file or directory (os error 2)

error: could not compile `demo` (lib) due to previous error
```

Mikenux:
I can also reproduce by saving the project. Notice that the path to the folder is `/run/flatpak/doc/...` which does not exist, while the same path (after the `/doc`) exists in `/run/user/1000 /doc`.

hfiguiere:
The funny thing is that while I get the error, if I flatpak enter I can access all the files in question.

hfiguiere:
The fatal call is in cargo:

```
[...]

PR #1243, opened 2023-12-20
document-portal: Don't invalidate directories

hfiguiere:
This work around issue #1234

fuse_lowlevel_notify_inval_entry() will invalidate directories too, which is a problem as it invalidate the current workdir directory. So now we don't invalidate them and will let the kernel deal with it.

This is consistent with the passthrough_hp example from fuse.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -1847,17 +1847,21 @@ xdp_fuse_lookup (fuse_req_t req,
     }
   else
     {
+      g_autoptr(XdpInode) inode = NULL;
+      struct stat buf;
+
       g_assert (parent_domain->type == XDP_DOMAIN_DOCUMENT);
 
       fd = xdp_document_inode_open_child_fd (parent, name, open_flags, 0);
       if (fd < 0)
         return xdp_reply_err (op, req, -fd);
 
-      res = ensure_docdir_inode (parent, fd, &e, NULL); /* Takes ownership of fd */
+      res = ensure_docdir_inode (parent, fd, &e, &inode); /* Takes ownership of fd */
       if (res != 0)
         return xdp_reply_err (op, req, -res);
 
-      queue_invalidate_dentry (parent, name);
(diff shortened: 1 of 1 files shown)


hfiguiere:
Turns out it prevents the filesystem to be ejected.

So this is unlikely to be the right fix.

hfiguiere:
Right now the situation is as follow:

- without that a directory accessed through the portals, that fail with cwd being invalid.
- with this patch the directory might be held too long like it was before #1190 - see #689.

I haven't found a different solution that would not hold the directories too long.

hfiguiere:
My take is as follow: after applying this patch the use case that is broken in #1234 is fixed. But this make the directories being held too much as in #689.

I don't know how to solve both properly but I think this patch as a workaround present a better tradeoff as a work around fix.

nurupo:
Does it even make sense for anything to be using a directory under `/run/user/1000/doc/` as its CWD?

<details><summary>(A little rant about applications using `/run/user/1000/doc/`)
</summary>

---

While on this topic, something that has been bothering me a lot is that if I save a file in a flatpak Chromium that uses xdg-desktop-portal to a remote drive at `/media/my-drive/foobar/`, the next time I save a file, the directory that the save dialog window opens up in is something like `/run/user/1000/doc/db0809ab/` instead of the previous `/media/my-drive/foobar/`. While `/run/user/1000/doc/db0809ab/` does list my previously saved file, if I save the 2nd file into this directory, the 2nd file doesn't appear in `/media/my-drive/foobar/` alongside the 1st file, the 1st file is in there by itself, the 2nd file only appears in `/run/user/1000/doc/db0809ab/` alongside the 1st file. So when saving multiple files into the same directory, I have to fight the save file dialog window resetting to `/run/user/1000/doc/db0809ab/` and manually navigate to `/media/my-drive/foobar/`. EVERY. SINGLE. TIME. I save a file. It's so annoying. I feel like Chomium using `/run/user/1000/doc/` is some sort of a bug and it shouldn't be used.

---
</details>

Issue #1241, opened 2023-12-18
xdg-desktop-portal tries to access non-existing org.freedesktop.impl.portal.Request


Sodivad:
### Operating System

Neon 22.04

### XDG Desktop Portal version

Git

### XDG Desktop Portal version (Other)

main

### Desktop Environment

KDE

### Expected Behavior

`xdg-desktop-portal` should not try to access non-existent object/give the impl a chance to create them

### Current Behavior

I can see that for  requests an app makes xdg-desktop-portal  tries to read properties of a non-existing Request. The event flow is as follows with `org.freedesktop.portal.FileChooser.OpenFile` as an example

-  `org.freedesktop.portal.FileChooser.OpenFile` is called
- `xdg-desktop-portal` makes a call to the  impl  at  `/org/freedesktop/portal/desktop/request/connection/token` `org.freedesktop.DBus.Properties.GetAll('org.freedesktop.impl.portal.Request',)`
- `xdg-desktop-portal` calls org.freedesktop.impl.portal.FileChooser.OpenFile
- `xdg-desktop-portal` send the method return of the original `org.freedesktop.portal.FileChooser.OpenFile`

### Steps to Reproduce

- Use a flatpaked app or make a call manually 

### Anything else we should know?

I can see this order of event flows with multiple interfaces. At least FileChooser, Screenshot and Screencast

GeorgesStavracas:
Is this causing any problem? GDBusProxy initialization calls `org.freedesktop.DBus.Properties.GetAll()` on the object during construction, but the `org.freedesktop.impl.portal.Request` interface does not have any properties, so this ends up never returning nothing of interest.

Sodivad:
> Is this causing any problem? 

Not to my knowledge. I  just keep spotting this in bustle 

>  but the org.freedesktop.impl.portal.Request interface does not have any properties, so this ends up never returning nothing of interest.

Even if it it had properties, It would fail as the Request object does not exist yet.

GeorgesStavracas:
So it's not a critical issue or anything. Cool. In any case, it would be good to figure it out.

GeorgesStavracas:
Could you please verify if https://github.com/flatpak/xdg-desktop-portal/pull/1244 fixes that for you, without introducing any regressions?

PR #1244, opened 2023-12-20
impl-request: Do not load properties on construction

GeorgesStavracas:
The org.freedesktop.impl.portal.Request interface has no properties, and is constructed against a different object path. We don't need to fetch properties for that.

Don't do that.

Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1241


whot (approved):

Sodivad:
Fixes the issue for me, thanks!

PR #1247, opened 2023-12-25
test: The test were leaving '2' files behind

hfiguiere:
The syntax to output to stderr was incorrect

### DIFF
--- a/tests/test-document-fuse.sh
+++ b/tests/test-document-fuse.sh
@@ -85,12 +85,12 @@ if [ -n "${XDP_UNINSTALLED:-}" ]; then
 fi
 
 # First run a basic single-thread test
-echo Testing single-threaded &>2
+echo Testing single-threaded >&2
 "${test_srcdir}/test-document-fuse.py" --iterations ${ITERATIONS} -v
 echo "ok single-threaded"
 
 # Then a bunch of copies in parallel to stress-test
-echo Testing in parallel &>2
+echo Testing in parallel >&2
 PIDS=()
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
Hm. My bash redirection knowledge is quite limited, but according to [this documentation page](https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Redirecting-Standard-Output-and-Standard-Error), redirecting `stdout` to `stderr` is documented as `1>&2`. This branch changes it to `>&2`, without the `1`.

Is this bashism? If yes, is it a harmless bashism, or something to avoid? Is `1` always the default?

hfiguiere:
It was doing `&>2` which redirect into `2` (a file name). `>&2` mean stdout to stderr (the stdout is implicit).

GeorgesStavracas (approved):

smcv:
FYI:

`>&2` or equivalently `1>&2` is standardized POSIX sh syntax for "make fd 1 into a duplicate of fd 2". The 1 does not need to be given explicitly, because `>` or `>>` defaults to acting on fd 1 unless otherwise stated.

`&>2` is a bashism (i.e. documented by bash, but with no specified meaning in arbitrary POSIX shells). In bash, it means send both stdout and stderr to a file named `2`. The POSIX equivalent of `&>foo` is `>foo 2>&1`, or equivalently and more explicitly `1>foo 2>&1`.

Confusingly, bash also allows writing `>foo 2>&1` as `>&foo`, but only if the filename `foo` is neither a number nor `-`, each of which means something different.

PR #1252, opened 2023-12-27
Update the D-Bus docs URL in the help text

Isti115:


src/xdg-desktop-portal.c
```diff
@@ -1,4 +1,4 @@
-/*
+https://flatpak.github.io/xdg-desktop-portal/docs//*
```
GeorgesStavracas, line 1: I don't think you meant to do this?

Isti115: Ouch, sorry, I don't know how I managed to mess that up... :man_facepalming: 
I used the online GitHub editor, I guessed that it will be fine for such a simple change, but I must've misclicked somewhere, sorry! I'll update the patch in a few minutes!

GeorgesStavracas: No problem, it looked suspiciously like an erroneous  copy-paste :slightly_smiling_face:

GeorgesStavracas (approved):

PR #1253, opened 2023-12-28
More documentation improvements

GeorgesStavracas:
This is a relatively big reorganization of the documentation pages, hopefulle for the better.

This PR:
 * Adds an Introduction page
 * Adds a Contributing page (and updates CONTRIBUTING.md to point to the online docs)
 * Adds a Portal Backends page, and moves quite a few things there
 * Documents the `.portal` file format
 * Documents a bit more the `portals.config` file format
 * Documents what's needed for XDG Desktop Portal to discover new backends
 * Moves the Background Apps Monitoring API page into the Portal Backends page
 * Adds a little Python script that fixes RST titles (while GLib 2.78.4 doesn't hit CI) and replaces the reverse domain name titles of the generated pages

Overall I'm quite happy with the end result. You can see how it looks here: https://georgesstavracas.github.io/xdg-desktop-portal/docs/index.html


hfiguiere (approved):
LGTM

doc/introduction.rst
```diff
@@ -0,0 +1,43 @@
+Introduction
+============
+
+XDG Desktop Portal is a session service that provides D-Bus interfaces for
```
swick, line 4: This sounds like the service is available only to sandboxed apps.

doc/introduction.rst
```diff
@@ -0,0 +1,43 @@
...
+sandboxed apps. These D-Bus interfaces can be used by sandboxed apps without
+granting special permissions on their sandbox manifest.
```
swick, line 6: I know what you're referring to here but I think this will be confusing to people unaware that e.g flatpak has a manifest that can grant coarse access to host resources.

Maybe it's better to focus on the fact that with portals, permissions are granted or denied at runtime by the user?

doc/contributing.rst
```diff
@@ -0,0 +1,90 @@
...
+-----------
+
+XDG Desktop Portal usually runs as a user session service, initialized on
+demand through D-Bus activation. It usually starts with the session though,
+as many desktop environments try to talk to XDG Desktop Portal on startup.
+XDG Desktop Portal initializes specific backends through D-Bus activation
+as well.
+
+
+Building
```
swick, line 27: Advising people to install development code into `/usr` isn't exactly a safe thing to do.

Can we come up with a scheme where the developers don't have to modify the host system too much?

There will be cases where you probably want to install to the host `/usr` but in that case at least tell the developer that this is potentially dangerous and maybe even that there are ways to install to the system and then being able to roll back (`ostree admin unlock`, `systemd-sysext`, ...).

swick (commented):
Looks great. There are a few things I think we could improve.

ebassi (approved):
For reference, I've opened [this MR for GLib](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3793) to generate shorter titles for the D-Bus interface pages with gdbus-codegen.

doc/contributing.rst
GeorgesStavracas: I merely moved the contents of `CONTRIBUTING.md` into this page. I just moved it here because @hfiguiere suggested in last month's discussion (https://github.com/flatpak/xdg-desktop-portal/discussions/1232#discussioncomment-7708916). I'm not particularly interested in reviewing this page myself right now so patches would be welcomed.

GeorgesStavracas:
> For reference, I've opened [this MR for GLib](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3793) to generate shorter titles for the D-Bus interface pages with gdbus-codegen.

Thank you, very much appreciated. I'll update the Python script when this MR reaches the CI image.

GeorgesStavracas:
In the spirit of keeping things moving, I'll merge this PR. If you have further improvements, please open up new PRs. Thanks!

PR #1269, opened 2024-01-24
Add GetRealPath to document portal

### LINKED ISSUE #475: Real directory name not accessible from file opened via document portal

JakobDev:
This new function allows Apps to get the Real Path from a Document ID. IF you are using the FileChooser Portal, you get the a `/run/user/1000/doc` path. That's a Problem for Apps who display the Path in the UI e.g. a recent files menu.This new function allows Apps to get the real Path, so it be displayed probably in the UI.

Sandboxed Apps can only get the real path for those document IDs they have access to.

This PR is currently missing tests, as I don't know if this new function gets accepted. Please also note that I don't have much experience in C.

Fixes #475


document-portal/document-portal.c
```diff
@@ -1416,6 +1416,68 @@ portal_list (GDBusMethodInvocation *invocation,
...
+                                             XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             "Invalid ID passed");
+      return TRUE;
+    }
+
+  if (!xdp_app_info_is_host (app_info))
+    {
+      g_autofree const char **apps = NULL;
+      g_autofree const char *app_id = NULL;
+      static gboolean id_found = FALSE;
```
grulja, line 1445: Why is this variable static? Wouldn't it make it always return true in case we already successfully found one?

grulja:
Also, should there be some sanity checks, like only accept the path with the `file://` scheme + mention it in the docs? It's always passed like that in the `FileChooser` portal so we should make sure there is some consistency.

JakobDev:
>Also, should there be some sanity checks, like only accept the path with the `file://` scheme (or not?)

You just pass the ID, not the path. If you path is e.g. `/run/user/1000/doc/123` the ID is `123`. It's the same usage as the` Info` function.

> and mention it in the docs?

This should be generated automatically from the XML file.

grulja:
> > Also, should there be some sanity checks, like only accept the path with the `file://` scheme (or not?)
> 
> You just pass the ID, not the path. If you path is e.g. `/run/user/1000/doc/123` the ID is `123`. It's the same usage as the` Info` function.
> 

I see, didn't realize that and it makes sense.

> > and mention it in the docs?
> 
> This should be generated automatically from the XML file.

I meant in case we were going to pass path with `file://` suffix, that it should be mentioned in the docs, but as you pointed out it's just the document id and not the path. No action needed then.

data/org.freedesktop.portal.Documents.xml
```diff
@@ -273,5 +273,17 @@
       <arg type='s' name='app_id' direction='in'/>
       <arg type='a{say}' name='docs' direction='out'/>
     </method>
+
+    <!--
+        GetRealPath:
+        @doc_id: the ID of the file in the document store
+        @path: the path for the file in the host filesystem
+
+        Gets the filesystem path for a document store entry.
```
grulja, line 282: Following some other methods described here in this XML file, I think it should mention that this call is available inside the sandbox when the app has access to the given file. Also a version bump for this portal might be needed so clients can check whether this method is available or not.

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,68 @@ portal_list (GDBusMethodInvocation *invocation,
...
+              id_found = TRUE;
+              break;
+            }
+        }
+
+      if (!id_found)
+        {
+          g_dbus_method_invocation_return_error (invocation,
+                                                 XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                                 "Invalid ID passed");
```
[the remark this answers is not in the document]
JakobDev: Is it possible to have a ID part of a Flatpak, but with no permission?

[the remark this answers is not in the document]
grulja: What I mean is, that you already found a database entry for given ID, right? At this point (in the code) you are looking whether the database entry has entry for your application (based on app_id). Maybe the variable name `id_found` is a bit confusing and it should be `app_found` perhaps? If the app is not found in the database entry, I think it means you don't have permission to this document and you shouldn't be able to get the real path. That's my understanding of this code, therefore the error shouldn't be about a wrong argument, because that's what you already checked at the beginning of this function, but  a missing entry for given `app_id` is a problem and means you shouldn't be able to get the real path for missing permission.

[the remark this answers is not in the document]
JakobDev: I use the same error on both places, so a sandboxed App don't know if a ID does not exists or if the App don't have Permission for the ID.  I don't think it's a security issue if the App knows a ID exists, but I think Sandboxed Apps don't need to know about that.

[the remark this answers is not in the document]
grulja: As a client you should get the real reason why your request failed. I don't want to debug why my request failed when providing an existing document ID, while the reason for failure is that I no longer have permission to this document. There are [other calls](https://github.com/flatpak/xdg-desktop-portal/blob/main/document-portal/document-portal.c#L176) in this API giving you information about you not having enough permission.

data/org.freedesktop.portal.Documents.xml
grulja: Please, also mention that this method was introduced in version 5 of the document portal. Similar to [other methods](https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.portal.Documents.xml#L130).

JakobDev:
I have now added a test

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,68 @@ portal_list (GDBusMethodInvocation *invocation,
...
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             "Invalid ID passed");
+      return TRUE;
+    }
+
+  if (!xdp_app_info_is_host (app_info))
+    {
+      g_autofree const char **apps = NULL;
+      g_autofree const char *app_id = NULL;
```
grulja, line 1444: This will cause double-free. You don't have to free `app_id` in this case.

grulja: I haven't checked, but this is most likely the reason why the test fails. It was also not working for me as it crashed on double-free and removing the `g_autofree` made it work.

grulja: I also don't see the full log, but running the test locally with asan, it doesn't pass, because there is a memory leak.

This works for me:
```
diff --git a/tests/test-doc-portal.c b/tests/test-doc-portal.c
index e2d9f0c..cd39912 100644
--- a/tests/test-doc-portal.c
+++ b/tests/test-doc-portal.c
@@ -687,16 +687,11 @@ test_add_named (void)
   assert_doc_not_exist (id1, basename1, "com.test.App2");
 }
 
-static char *
-doc_get_real_path (const char *basename)
-{
-  return g_build_filename (outdir, basename, NULL);
-}
-
 static void
 test_get_real_path (void)
 {
   g_autofree char *doc_id = NULL;
+  g_autofree char *expected_real_path = NULL;
   g_autofree char *real_path = NULL;
   const char *basename = "real-path";
   g_autoptr(GVariant) reply = NULL;
@@ -720,7 +715,8 @@ test_get_real_path (void)
 
   g_assert_no_error (error);
   g_variant_get (reply, "(s)", &real_path);
-  g_assert_cmpstr (real_path, ==, doc_get_real_path(basename));
+  expected_real_path = g_build_filename (outdir, basename, NULL);
+  g_assert_cmpstr (real_path, ==, expected_real_path);
 }
 
 static void
```

grulja: Can you please remove the static function? I don't think there is a need for it as it's just one-liner and it's not used anywhere else. Also please squash your commits into one.

grulja: Aha, sorry, you did that already, I was just looking at the last commit.

grulja:
Can you please squash your commits into one?

data/org.freedesktop.portal.Documents.xml
```diff
@@ -273,5 +273,21 @@
...
+
+        This call is available inside the sandbox, if the application has the permission for the document.
+
+        This method was added in version 5 of this interface.
+    -->
+    <method name="GetRealPath">
```
grulja, line 288: I'll leave this to others to decide, but I think better naming would be just `HostPath` instead `GetRealPath`.

swick:
So, we are not going for extended file attributes then? Would be nice to have some rationale for why this path has been chosen. I'm worrying a bit about the performance here because every single file lookup will incur multiple dbus messages between 4 different processes.

swick:
> Maybe a stupid question as this is something I know nothing about. Wouldn't extended file attributes as proposed in the original bug be GIO specific?

No, extended file attributes are a standard linux feature. They are exposed in GIO via GFileAttribute currently, so this is only about how Glib/GIO would provide a nice API for it.

> Why multiple DBus messages between 4 different processes?

App, xdg-dbus-proxy, dbus broker, portal.

grulja:
> > Maybe a stupid question as this is something I know nothing about. Wouldn't extended file attributes as proposed in the original bug be GIO specific?
> 
> No, extended file attributes are a standard linux feature. They are exposed in GIO via GFileAttribute currently, so this is only about how Glib/GIO would provide a nice API for it.

It would still require all the other counterparts (e.g. Qt) to implement support for it, right? I also think this would be less discoverable, even though it's probably a better solution. Maybe this solution can be a temporary one? Most importantly because it would take way more time to get this available for clients. You would need this to land in GLib first in order to support it in xdg-desktop-portal and it would also require the most recent versions of Glib for clients to use it.

> 
> > Why multiple DBus messages between 4 different processes?
> 
> App, xdg-dbus-proxy, dbus broker, portal.

That's 4 processes involved, but I think it's still one message for `App → Portal` and one message back for `Portal → App`.  It's also not like this is going to be used frequently.

swick:
One could write code to read the extended attribute just like they could write code to call the portal. There is no need to wait for GIO or anything else really. It would just be more convenient.

That being said, I'm not against this portal in principle.

grulja:
> One could write code to read the extended attribute just like they could write code to call the portal. There is no need to wait for GIO or anything else really. It would just be more convenient.
> 

Right, but it's more low-level and not something clients will adopt as easily as simple DBus call.  

> We could however support both: This Function and a file attribute,. But adding a file attribute is outside the scope of this PR.

I agree, making this one as first option, while having file attributes as something to use in the future along this one.

Mikenux:
As written (one doc ID), does this mean the app should send multiple requests for multiple files? If that's the case, wouldn't it be better to be able to request multiple document IDs at once so there is only one request instead of multiple ones? (also have only one response instead of multiple ones)

Also, is it possible to handle this with a file chooser option? (returning files and their paths at the same time, or is it the extended file attributes?)

grulja:
> As written (one doc ID), does this mean the app should send multiple requests for multiple files? If that's the case, wouldn't it be better to be able to request multiple document IDs at once so there is only one request instead of multiple ones? (also have only one response instead of multiple ones)
> 

I think one will be the most common use-case, but I guess this PR can be changed to support more at once to be future-proof.

> Also, is it possible to handle this with a file chooser option? (returning files and their paths at the same time, or is it the extended file attributes?)

This would not work in case when you are not opening the file first using file chooser portal. The path from the document portal can be already saved in an app as part of settings.  In such case you just need to get the host path without going through file chooser portal first.

Mikenux:
> Most Apps don't call the file chooser portal directly. They just use the function the Toolkit provides and the Toolkit uses the Portal. So most Apps can't take advantage of such a Option.

Yes, but it is expected that the toolkits will implement the function (as you said, they already integrate portals).

grulja:
> > but I guess this PR can be changed to support more at once
> 
> I could add a second function which takes an array. Many D_Bus libraries are not in a good state, so I think we should also provide a solution which just takes a string.

It wouldn't make sense to have two same calls, where one does the same thing just for more document IDs. It should not be a problem to make this accept an array of strings and return something like `map<docId, hostPath>` as a result.

That said, I believe most of the cases will be using this to translate just one path and that's even my use case so I don't really insist on this.

+1 from me on this MR, with just a suggestion to change the naming of that method

swick:
We already have much more complicated method signatures and things work fine. This is really not a good argument at all.

hfiguiere:
If the bindings are the problem the bindings should be fixed instead. Otherwise you are just pushing for more technical debt.

grulja:
@JakobDev are you going to update this change to take more document IDs? I can look into it myself if you don't have time to do so. I really want this change landed as soon as possible.

grulja (commented):
I wrote some suggestions how I think this should be changed to correctly document the method and to make the method return map<string, bytearray>. I tested it and it works. I haven't tried to update the test.

grulja:
> > return map<string, bytearray>.
> 
> You can implemnt that if you want. I ave little to no motivation in messing with glib further. The last one already took a few hours.

I can take over if you don't mind. I wrote suggestions for the changes, but it still requires to finish the tests.

grulja:
I have submitted https://github.com/flatpak/xdg-desktop-portal/pull/1364 to continue with the changes I suggested here.

Issue #1278, opened 2024-02-05
tests: several failures when running in gnome from non gnome terminal


pierre-labastie:
### Operating System

Ubuntu 23.10, linux from scratch development

### XDG Desktop Portal version

1.18

### XDG Desktop Portal version (Other)

1.18.2

### Desktop Environment

GNOME

### Expected Behavior

All tests pass

### Current Behavior

Five tests fail:
```
test-portals-camera
test-portals-inhibit
test-portals-wallpaper
pytest test_remotedesktop
pytest test_globalshortcuts
```

### Steps to Reproduce

1. Open an uxterm from menu (may have to install it first)
2. build with `mkdir build && cd build && meson setup && ninja`
3. run `ninja test`
...

### Anything else we should know?

Same issue also with the konsole terminal. Note that the unit associated to uxterm is of the form `app-gnome-uxterm-xxxx.scope` (or `app-gnome-debian\x2duxterm-xxxx.scope` on ubuntu), so that the app_id is non empty, and I guess all the failures come from that, because the app_id is supposed to be empty by all tests accessing the permission store.

I have a working patch for the C tests, but I don't know how to access the app_id in python. On the other hand, I am not sure what your policy is for tests: I think there is no problem when running them in the CI set up.

Note also that the scope unit for gnome-terminal starts with `vte-spawn`, hence it generates an empty app_id, and the tests pass in that terminal...

whot:
Confirmed, I get the same in kitty and thus run the test-suite in gnome-terminal or gnome-console. I haven't figured out yet how to guess whether appid is set from the tests yet, the easiest would be to just remove that particular set of asserts though.

GeorgesStavracas:
Fixed by https://github.com/flatpak/xdg-desktop-portal/pull/1279

PR #1279, opened 2024-02-07
Fix test failures on non gnome-terminal in GNOME

pierre-labastie:
Because programs running in a non gnome-terminal belong to a .scope unit starting with app, xdg-desktop-portal assings a non empty appid to them. On the other hand, for setting permissions into the permission store, tests assume that appid is the empty string. To fix this, retrieve the appid during global setup a,d store it in a global variable. It is then available to tests, and can be passed when setting permissions.

Fix 3 test failures out of 5 reported in #1278

The other failures are in python tests.

### DIFF
--- a/tests/camera.c
+++ b/tests/camera.c
@@ -14,6 +14,7 @@ static int got_info;
 
 extern XdpDbusImplPermissionStore *permission_store;
 extern XdpDbusImplLockdown *lockdown;
+extern gchar *appid;
 
 static void
 set_camera_permissions (const char *permission)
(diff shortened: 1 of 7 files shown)


whot:
thanks heaps, I've used your first patch in #1280 and used it as inspiration for the "proper" fix for the pytest failures (albeit with complications...)

whot (approved):
this LGTM. I don't know when I'll get to finish #1280 so I reckon we should merge this one in the meantime to stop the test case failures.

GeorgesStavracas:
Thanks for the review @whot

PR #1280, opened 2024-02-08
Fix pytest failures if an appid is set

whot:
This incorporates #1279 (the first patch is from there and from @pierre-labastie).

The rest addresses the Python failings of the test if an appid is set: we make xdg-utils a shared library we can use via GI in python and then fetch the appid.

The frustrating thing: if we run the Python tests with a blank XDG_DATA_DIR our proxy resolver portal fails (see the last two commits). Which means our meson devenv has a different environment that cannot be used for the pytest run without manually changing the environment variables.


swick:
I'm trying to get rid of the whole pid->AppInfo lookup. Can't we instead introduce a way to force a specific AppInfo for tests?

whot:
This would be the best solution but tbh I don't know where to start looking on how to do that. Any hints?

swick:
Something as simple at this https://github.com/swick/xdg-desktop-portal/commit/ca900048de57e9ec2aab7ea6e1d8607902ce4df7 seems to work already. Probably needs a bit more work.

GeorgesStavracas:
Should this be closed now that https://github.com/flatpak/xdg-desktop-portal/pull/1309 was merged

whot:
yep, thanks for the reminder.

PR #1281, opened 2024-02-08
AppInfo cleanups, pidfd based pid mapping, device portal removal

swick:
Pulled out from https://github.com/flatpak/xdg-desktop-portal/pull/1268. All the parts that can be merged right now without requiring the new dbus Containers1 interface.


swick:
Ping. Anyone up for a review?

swick:
Unfortunately had to revert a commit from https://github.com/flatpak/xdg-desktop-portal/pull/1279 because it depends on `xdp_get_app_info_from_pid`. https://github.com/flatpak/xdg-desktop-portal/pull/1280 also uses it, so we need to figure out how to deal with this.

swick:
Rebased onto https://github.com/flatpak/xdg-desktop-portal/pull/1281 which makes the tests work again.

/cc @GeorgesStavracas @smcv

src/xdp-utils.c
```diff
@@ -881,31 +881,21 @@ xdp_get_app_info_from_pid (GDBusConnection  *connection,
...
 
-  app_info = parse_app_info_from_flatpak_info (pid, &local_error);
-  if (app_info == NULL && local_error)
-    {
-      g_propagate_error (error, g_steal_pointer (&local_error));
-      return NULL;
-    }
+  app_info = parse_app_info_from_flatpak_info (pid, error);
+
+  if (app_info == NULL && *error)
```
GeorgesStavracas: Hm, not a fan of this. Usually dereferencing `error` is avoided, using a local GError and propagating it to the function argument is preferred.

src/xdp-utils.c
```diff
@@ -847,6 +848,135 @@ xdp_app_info_from_snap (int          pid,
   return TRUE;
 }
 
+static gboolean
+xdp_connection_get_pid_legacy (GDBusConnection  *connection,
+                               const char       *sender,
+                               GCancellable     *cancellable,
+                               int              *out_pidfd,
+                               guint32          *out_pid,
```
GeorgesStavracas, line 866: Since you're touching this: `uint32_t`?

src/xdp-utils.c
```diff
@@ -847,6 +848,135 @@ xdp_app_info_from_snap (int          pid,
...
+  g_variant_get (reply, "(u)", out_pid);
+  return TRUE;
+}
+
+static gboolean
+xdp_connection_get_pidfd (GDBusConnection  *connection,
+                          const char       *sender,
+                          GCancellable     *cancellable,
+                          int              *out_pidfd,
+                          guint32          *out_pid,
```
GeorgesStavracas, line 895: Ditto

src/xdp-utils.c
```diff
@@ -978,7 +978,7 @@ xdp_connection_get_pidfd (GDBusConnection  *connection,
 }
 
 static XdpAppInfo *
-lookup_cached_app_info_by_sender (const char *sender)
+cache_lookup_app_info_by_sender (const char *sender)
```
GeorgesStavracas, line 991: I don't see a need to rename this function, actually

src/xdp-utils.c
GeorgesStavracas: Nevermind, in the end this is gone anyway

src/xdp-utils.c
swick: I did it so that  `cache_lookup_app_info_by_sender` and `cache_insert_app_info` both have the same prefix. In general we call things "object_verb" so this seems fitting, no?

GeorgesStavracas: I'd say that applies if the function was acting on an object, e.g. ` cache_lookup_app_info_by_sender (cache, sender);`

swick: the object is "global" but if you insist I can change it back

GeorgesStavracas: No hard opinions on my side

piotrdrag:
You need to update POTFILES.in as well: https://l10n.gnome.org/module/xdg-desktop-portal/#main

GeorgesStavracas:
@piotrdrag do you think it's possible to add such check as a CI step to this project? I think it's better if we make POTFILES.in being up-to-date a pre-condition to merging PRs

piotrdrag:
Various projects in and around GNOME already do that, usually with a custom script or the long-abandoned intltool-update. I don’t know how the former works out, but `intltool-update -m` is unreliable in various ways.

swick:
Opened https://github.com/flatpak/xdg-desktop-portal/pull/1331 but if anyone has pointers at how to handle this well in CI I might add that as well...

PR #1283, opened 2024-02-14
i18n: Add missing translatable file

sabriunal:
Add missing dynamic-launcher.c to the POTFILES.in


GeorgesStavracas (approved):

piotrdrag:
Are the strings in this file ever visible to the end user? They look like they shouldn’t have been marked for translation in the first place.

GeorgesStavracas:
That may be the case. PR welcome if you conclude they should be translated;

PR #1296, opened 2024-03-08
docs: clarify guarantees of the File Chooser API

alyssais:
The application should not depend on where the URL points, or how access to the file is granted, just that access has been granted at that URL.

@matthiasclasen, I've set you as the commit author since I used the wording you suggested on Matrix.

### DIFF
--- a/data/org.freedesktop.portal.FileChooser.xml
+++ b/data/org.freedesktop.portal.FileChooser.xml
@@ -28,8 +28,8 @@
       backend will present the user with a file chooser dialog.
 
       The selected files will be made accessible to the application
-      via the document portal, and the returned URI will point
-      into the document portal fuse filesystem in ``/run/user/$UID/doc/``.
+      (which may involve adding it to the :ref:`Documents
+      portal<org.freedesktop.portal.Documents>`).
 
       This documentation describes version 4 of this interface.
   -->


data/org.freedesktop.portal.FileChooser.xml
```diff
@@ -28,8 +28,7 @@
       backend will present the user with a file chooser dialog.
 
       The selected files will be made accessible to the application
-      via the document portal, and the returned URI will point
-      into the document portal fuse filesystem in ``/run/user/$UID/doc/``.
+      (which may involve adding it to the document portal).
```
GeorgesStavracas, line 31: Maybe add a link to the Document portal here?

alyssais: How would I do that?  If I write ``:ref:`Document portal<org.freedesktop.portal.Document>` `` in this section, it just generates `<span class="xref std std-ref">Document portal</span>`, with no hyperlink.

alyssais: Ah, got it — it's the Document**s** portal, in the plural.

alyssais: Done.

GeorgesStavracas (approved):

PR #1299, opened 2024-03-09
docs: clarify guarantees of the Screenshot API

alyssais:
Same as c7dce4e ("docs: clarify guarantees of the File Chooser API").

(I've checked, and no other portal documentation mentions /run/user apart from the document portal, so this should be the last of these fixes.)


GeorgesStavracas (approved):

GeorgesStavracas:
Thanks!

PR #1303, opened 2024-03-15
notification: Don't fail requests with unsupported properties

jsparber:
This filters unsupported properties from a notification request without
failing it. This allows backwards compatibility after adding new
properties. This should had been the case from the beginning so that we
don't have to break API when extending the notification interface. See
[1].

[1] https://dbus.freedesktop.org/doc/dbus-api-design.html#extensibility

Also improve tests and add icon tests to make sure that this doesn't break icons.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -234,33 +234,41 @@ check_value_type (const char *key,
 }
 
 static gboolean
-check_priority (GVariant *value,
-                GError **error)
+parse_priority (GVariantBuilder  *builder,
+                GVariant         *value,
+                GError          **error)
 {
   const char *priorities[] = { "low", "normal", "high", "urgent", NULL };
+  const char *priority;
 
   if (!check_value_type ("priority", value, G_VARIANT_TYPE_STRING, error))
     return FALSE;
 
-  if (!g_strv_contains (priorities, g_variant_get_string (value, NULL)))
+  priority = g_variant_get_string (value, NULL);
+
+  if (!g_strv_contains (priorities, priority))
     {
       g_set_error (error,
                    XDG_DESKTOP_PORTAL_ERROR,
                    XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                   "%s not a priority", g_variant_get_string (value, NULL));
+                   "%s not a priority", priority);
       return FALSE;
     }
+
+  g_variant_builder_add (builder, "{sv}", "priority", value);
 
   return TRUE;
 }
 
 static gboolean
-check_button (GVariant *button,
-              GError **error)
+parse_button (GVariantBuilder  *builder,
+              GVariant         *button,
+              GError          **error)
 {
   int i;
-  gboolean has_label = FALSE;
-  gboolean has_action = FALSE;
+  g_autoptr(GVariant) label = NULL;
+  g_autoptr(GVariant) action = NULL;
+  g_autoptr(GVariant) target = NULL;
 
   for (i = 0; i < g_variant_n_children (button); i++)
     {
@@ -268,23 +276,35 @@ check_button (GVariant *button,
       g_autoptr(GVariant) value = NULL;
 
       g_variant_get_child (button, i, "{&sv}", &key, &value);
+
       if (strcmp (key, "label") == 0)
-        has_label = TRUE;
+        {
+          if (!check_value_type (key, value, G_VARIANT_TYPE_STRING, error))
+            return FALSE;
+
+          if (!label)
+            label = g_steal_pointer (&value);
+        }
       else if (strcmp (key, "action") == 0)
-        has_action = TRUE;
+        {
+          if (!check_value_type (key, value, G_VARIANT_TYPE_STRING, error))
+            return FALSE;
+
+          if (!action)
+            action = g_steal_pointer (&value);
+        }
       else if (strcmp (key, "target") == 0)
-        ;
+        {
+          if (!target)
+            target = g_steal_pointer (&value);
+        }
       else
         {
-          g_set_error (error,
-                       XDG_DESKTOP_PORTAL_ERROR,
-                       XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                       "%s not valid key", key);
-          return FALSE;
+          g_debug ("Unsupported button property %s filtered from notification", key);
         }
     }
 
-  if (!has_label)
+  if (!label)
     {
       g_set_error_literal (error,
                            XDG_DESKTOP_PORTAL_ERROR,
@@ -293,7 +313,7 @@ check_button (GVariant *button,
       return FALSE;
     }
 
-  if (!has_action)
+  if (!action)
     {
       g_set_error_literal (error,
                            XDG_DESKTOP_PORTAL_ERROR,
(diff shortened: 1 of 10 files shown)


src/notification.c
```diff
@@ -234,129 +234,184 @@ check_value_type (const char *key,
...
+
+      if (G_IS_THEMED_ICON (deserialized_icon))
+        {
+          g_autoptr(GVariant) serialized_icon = NULL;
+
+          serialized_icon = g_icon_serialize (deserialized_icon);
+
+          g_variant_builder_add (builder, "{sv}", "icon", serialized_icon);
+          return TRUE;
+        }
```
swick, line 374: this falls through to `check_value_type` and then fails there with a bit misleading error. might want to return here and get the error message right.

src/notification.c
```diff
@@ -234,129 +234,184 @@ check_value_type (const char *key,
...
       g_autoptr(GVariant) value = NULL;
 
       g_variant_get_child (button, i, "{&sv}", &key, &value);
+
       if (strcmp (key, "label") == 0)
-        has_label = TRUE;
+        {
+          if (!check_value_type (key, value, G_VARIANT_TYPE_STRING, error))
+            return FALSE;
+          label = g_steal_pointer (&value);
```
swick, line 278: this is leaking when there are multiple "label" entries. some for the other keys. either `g_clear_pointer (&label, g_variant_unref)` or change the if before to check for `label == NULL` depending on if you want to use the first or last occurrence of the key.

src/notification.c
```diff
@@ -234,129 +234,184 @@ check_value_type (const char *key,
...
+      g_variant_builder_close (builder);
+    }
+  else
     {
-      g_set_error_literal (error,
-                           XDG_DESKTOP_PORTAL_ERROR,
-                           XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                           "action key is missing");
-      return FALSE;
+      g_debug ("Notification button needs a label and action property");
```
swick, line 309: shouldn't this `return FALSE` here? after all, there was an array entry for a button that didn't result in a button.

src/notification.c
```diff
@@ -234,129 +234,184 @@ check_value_type (const char *key,
...
-                   XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                   "%s not a priority", g_variant_get_string (value, NULL));
-      return FALSE;
-    }
+  priority = g_variant_get_string (value, NULL);
+
+  if (g_strv_contains (priorities, priority))
+    g_variant_builder_add (builder, "{sv}", "priority", value);
+  else
+    g_debug ("Unsupported priority %s filtered from notification", priority);
```
swick, line 252: same question here: ignoring unsupported properties is fine and all, but when the data is bogus, it should result in an error, no?

AdrianVovk: well if we want to later add more possible priorities...

swick: That's why we the portals are versioned.

jsparber:
@swick thanks for the review :)

src/notification.c
swick: Seeing the changed code: is it documented what happens with multiple keys? Should this maybe even return an error?

swick: The documentation of the portal doesn't mention anything so this seems to be undefined and the code should be fine. Either way, we should update the spec to either throw an error when a key exists multiple times (which might not be possible if there are actual apps which already do this), or at least update if the first or last occurrence of a key is used.

jsparber: not sure, the docs don't say anything about it, and the old code just hands what ever keys are present to the backend. The gtk backend simple passes everything to GNOME Shell or uses `g_variant_lookup_value()` when using FDO. The docs for ` g_variant_lookup_value()`  don't specify a behavior, but a quick look at the code it seams that it just returns the first found value.

swick: Alright, this is fine then.

It would still be good to define what happens with multiple keys in a dict for all portals (first or last occurrence wins) but let's not block this MR on it.

swick:
LGTM

jsparber:
> LGTM. Even though portals are versioned and this isn't really required, this makes it much nicer for apps to use.
> 
> e: I'll redact this. The notifications test is crashing. This needs fixing first.

The tests fails on CI because, `xdg-desktop-portal-validate-icon` isn't installed. But all tests are run `XDP_UNINSTALLED=1`. I need to figure out what's the best solution.

swick:
Probably just some environment variable to override the icon_validator executable path that you set in test-portals.c and the python test framework.

jsparber:
> Probably just some environment variable to override the icon_validator executable path that you set in test-portals.c and the python test framework.

Not sure if that's so easy, we don't want to add the possibility to set the icon_validator path via env variable.

swick:
Might want to take a look at https://github.com/flatpak/xdg-desktop-portal/pull/1309/commits.

If the env var is only for testing then we probably want the names to be consistent (`XDG_DESKTOP_PORTAL_TEST_VALIDATE_ICON` or something like that).

Also, let's set the environment from within the test harnesses and not the meson environment.

jsparber:
> If the env var is only for testing then we probably want the names to be consistent (`XDG_DESKTOP_PORTAL_TEST_VALIDATE_ICON` or something like that).

I think the env var can be generally useful, not just for testing.

jsparber:
the icon validator doesn't work, but the CI logs don't show any reason. Locally it works just fine. It could have to do something with the sandbox.

jsparber:
> It's running into a timeout (`TIMEOUT 30.01s killed by signal 15 SIGTERM`) which suggests that it's either actually taking a really long time or your new test is getting stuck somehow.

that's normal, if a test fails. Since the mocked backend just stops talking back. The real issue is in the meson log.

```
(/src/builddir/tests/../src/xdg-desktop-portal:4770): xdg-desktop-portal-WARNING **: 15:51:47.056: : Child process exited with code 1
error: **
```

swick:
*Sight* that should be fixed in the test harness...

Anyway, the test portal crashes in response to `test_bytes_icon`:
```
0x00007ffff7ae8834 in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install glib2-2.78.3-1.fc39.x86_64 glibc-2.38-16.fc39.x86_64 libblkid-2.39.3-6.fc39.x86_64 libffi-3.4.4-4.fc39.x86_64 libmount-2.39.3-6.fc39.x86_64 libselinux-3.5-5.fc39.x86_64 pcre2-10.42-1.fc39.2.x86_64 zlib-1.2.13-4.fc39.x86_64
(gdb) bt
#0  0x00007ffff7ae8834 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff7a968ee in raise () at /lib64/libc.so.6
#2  0x00007ffff7a7e8ff in abort () at /lib64/libc.so.6
#3  0x00007ffff7e90056 in g_assertion_message[cold] () at /lib64/libglib-2.0.so.0
#4  0x000000000045056e in handle_add_notification (object=0x4a3060, invocation=0x7fffe40059d0, arg_app_id=0x7fffe40038a0 "org.example.App", arg_id=0x4abbc0 "test7", arg_notification=0x7fffe40042b0)
    at ../tests/backend/notification.c:65
#5  0x000000000040fdde in _g_dbus_codegen_marshal_BOOLEAN__OBJECT_STRING_STRING_VARIANT
    (closure=0x4a30f0, return_value=0x7fffffffd640, n_param_values=5, param_values=0x4abc20, invocation_hint=0x7fffffffd620, marshal_data=0x0) at src/xdp-impl-dbus.c:1322
#6  0x000000000042e185 in xdp_dbus_impl_notification_method_marshal_add_notification
    (closure=0x4a30f0, return_value=0x7fffffffd640, n_param_values=5, param_values=0x4abc20, invocation_hint=0x7fffffffd620, marshal_data=0x0) at src/xdp-impl-dbus.c:23474
#7  0x00007ffff7c4e52a in g_closure_invoke () at /lib64/libgobject-2.0.so.0
#8  0x00007ffff7c7cfec in signal_emit_unlocked_R.isra.0 () at /lib64/libgobject-2.0.so.0
#9  0x00007ffff7c6ad3b in g_signal_emitv () at /lib64/libgobject-2.0.so.0
#10 0x000000000042f51c in _xdp_dbus_impl_notification_skeleton_handle_method_call
    (connection=0x4914c0, sender=0x7fffe4003bb0 ":1.3", object_path=0x7fffe40054f0 "/org/freedesktop/portal/desktop", interface_name=0x499d20 "org.freedesktop.impl.portal.Notification", method_name=0x7fffe4002a80 "AddNotification", parameters=0x7fffe4003c00, invocation=0x7fffe40059d0, user_data=0x4a3060) at src/xdp-impl-dbus.c:24283
#11 0x00007ffff7dc55d3 in g_dbus_interface_method_dispatch_helper () at /lib64/libgio-2.0.so.0
#12 0x00007ffff7da76e8 in call_in_idle_cb.lto_priv () at /lib64/libgio-2.0.so.0
#13 0x00007ffff7ec878d in g_idle_dispatch () at /lib64/libglib-2.0.so.0
#14 0x00007ffff7ecbe5c in g_main_context_dispatch_unlocked.lto_priv () at /lib64/libglib-2.0.so.0
#15 0x00007ffff7f26f18 in g_main_context_iterate_unlocked.isra () at /lib64/libglib-2.0.so.0
#16 0x00007ffff7ecd447 in g_main_loop_run () at /lib64/libglib-2.0.so.0
#17 0x000000000044af0d in main (argc=1, argv=0x7fffffffdcf8) at ../tests/backend/test-backends.c:151
```

jsparber:
Fixing the CI took way to long. :face_exhaling:

swick:
Alright, this looks all good to me!

jsparber:
@GeorgesStavracas This was the merge request that included pretty much the same commits as the once you cherry picked in #1386.  Closing.

Issue #983, opened 2023-03-08
Notifications Portal v2 proposal


jonas2515:
_Quick introduction for those who don't know me: I'm Jonas (aka [verdre](https://gitlab.gnome.org/verdre)) from the GNOME community, mostly working on gnome-shell/mutter but also across the stack to get things ready for phones._

The notifications portal is currently of very limited use for apps with more advanced notification needs, especially with smartphone use-cases in mind.

Here's a proposal for a new version of the notifications portal, trying to kickstart development and open room for discussion.

#### Things we want

 - Playing sound or vibrating on notifications
 - Persistent notifications that remain at the top of the tray and can't be dismissed (see also https://gitlab.gnome.org/Teams/Design/os-mockups/-/issues/193)
 - Standardized hints about the content of the notification, inspired by the ideas in [Desktop Notification Specication](https://specifications.freedesktop.org/notification-spec/latest/)
 - Standardized actions that are useful for a range of similar apps, eg. for inline replies to chat messages
 - Grouping similar notifications, eg. for multiple chat messages from one contact
 - Presenting rich content in notifications, eg. progress bars or images

#### Things we don't want

- Notifications are always presented immediately to the user, we don't want API to defer the presentation to a later point.
- We don't want API for showing "n unseen notifications" badges on app icons.
- We don't want persistent notifications to be abused as app indicators, that's why they are removed when app closes.
- We don't want to expose notification urgency as it is right now to avoid abuse by apps. 

#### Permission requests
- Every app has to request permission to send notifications using the `RequestNotifications()` method.
- Access to notifications may later be revoked by the user, so `AddNotification()` can fail.

#### Content Hints
Used for all notifications that need specific handling, eg.

- giving a low-battery notification priority
- overriding do not disturb state for extreme weather warnings
- ringing alarm clocks with a special and recurring ringtone

There are three different namespaces that any hint must adhere to:

 - `class`: What type of content the notification is. Possible hints are standardized with the protocol. Only a single `class` hint may be used per notification.
   - `class.im.message`
   - `class.alarm.ringing`
   - `class.call.incoming`
   - `class.call.ongoing`
   - `class.call.missed`
   - `class.weather.warning.extreme`
   - `class.cellbroadcast.danger.extreme`
   - `class.cellbroadcast.danger.severe`
   - `class.cellbroadcast.amberalert`
   - `class.cellbroadcast.test`
   - `class.os.battery.low`
   - `class.browser.webNotification`

[...]

PR #1298, opened 2024-03-08
Implement v2 notification spec

jsparber:
After lot of consideration I started implementing parts of the [proposed and discussed notification API](https://github.com/flatpak/xdg-desktop-portal/issues/983 )

For now this includes:

- [x] Set icon additionally via file descriptor 
- [x] Set sound via bytes, file descriptor
- [x]  `desktop-file-id` property
- [x]  `markup-body`  property
- [x]  `display-hint` property
- [x]   ~~new actions, similar to buttons (the name may be a little confusing, open for suggestions)~~ Part of buttons now
- [x]  ~~content-type~~  `category` property
- [x]  Button purpose
- [x]  Supported options for ~~content-type~~ `category`, ~~action purpose~~ and button purpose API

This needs changes in libportal https://github.com/flatpak/libportal/pull/147 for tests.

I also started writing the changes needed in xdg-desktop-portal-gtk: https://github.com/jsparber/xdg-desktop-portal-gtk/tree/implement_notification_v2 and other portal backends need to do the same thing.

Parts that didn't made it from #1304 into this MR may be added it a later revision of the portal. 

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/485

Issue #1026, opened 2023-05-13
Event, Task, and Simple Reminder Notifications


Mikenux:
## Goal

The goal is to register the properties of events, simple reminders, and tasks ("To Do") in order to identify them. Such identification would make it possible to:
- Have a notification type that can be toggled on and off for events, simple reminders, and tasks;
- Show a standard notification with standard actions;
- Display these different items in standard widgets.

## Required
- A Task Store to register any "task", namely events, simple reminders, and tasks in this case. Registered "tasks" are categorized by app.

## App Registration
- An app with events, simple reminders, and/or tasks is registered when it is installed or updated. The fact that the app uses such items must be detectable. Also, this registration is updated when the app is updated (if there is a change in what it supports).
- The app communicates which properties of the events, simple reminders and/or tasks, as well as those of the notification actions it supports (see the registrations below).
- An app with tasks communicates whether it supports setting a task's level of completion − called "Completion", by registering the unique completion format it uses: a "format" property where the possible values are:
	- "binary" for defining a task as completed or not, and nothing else.
	- "percentage" for the use of percentage.
	- "fraction" for the use of fractions. The app also indicates the number of levels (e.g. 10).
	- "labels" for the use of labels. The app lists the labels to use in ascending order of completion (e.g. Not Started, In Progress, Half Done, Almost Done, Done).

## General
- Simple reminders and tasks that are undated are not registered, so not available in standard widgets.
- Apps with events, simple reminders, and/or tasks don't need to be launched or run in the background. Updating items according to the triggered notification action is performed on the system side. 
- Apps share their event / simple reminder / task registrations with the Task Store. This allows apps to update the display of the items within their interfaces when they are launched or running.
- For events and tasks that include "Participants" and "Location" properties (see the registrations below), the app must expose attendees (i.e. contacts) and use an app-side service for the location. Location service is app-side if one service is preferred over another; the user selects the location on the app side, and the app returns a name (e.g. place, company name) and coordinates.

## Widgets
- An event widget displays all events.
- A simple reminder widget only displays simple reminders with a date.
- A task widget only displays dated tasks.
- Widgets can open the relevant app.
- All of these items can have actions to reschedule and cancel them (see Notification Actions below for each item).
- Tasks can have an action to set their level of completion (see Notification Actions below for Tasks).

## Portal and App User Interfaces

The portal UI is a standard dialog for adding and updating events, simple reminders, and tasks.

[...]

Issue #1025, opened 2023-05-13
Alarm Notifications (with Task Store)


Mikenux:
## Goal

The goal is to register the properties of alarms in order to identify them. Such identification would make it possible to:
- Have an "alarm" notification type that can be toggled on and off;
- Show a standard notification with standard actions;
- Display alarms in a standard widget.

## Required
- A Task Store to register any "task", namely alarms in this case. Registered "tasks" are categorized by app.

## App Registration
- An app with alarms is registered when it is installed or updated. The fact that the app uses alarms must be detectable. Also, this registration is updated when the app is updated (if there is a change in what it supports).
- The app communicates whether it supports sound playback, vibration patterns, and/or LED patterns. If it doesn't support one of these, the system should ideally be able to play a default sound, vibration pattern, or LED pattern. This way, when in a specific feedback mode (i.e. sound, vibration, silent, or do not disturb), the system sends a signal to tell the app what can be played.
- The app communicates which properties it supports (see Alarm Registration below).

## General
- An app with alarms is launched in the background (i.e. without displaying a window), either at the opening of the user session and if there are alarms registered for the day, or by waking it up at alarm time or a bit earlier if needed.
- When an alarm is added, it is automatically enabled. If an alarm is disabled, it is unregistered. To be registered again, it must be re-enabled.

## Permissions
- When an alarm is enabled and must ring on the day of its addition, the app is allowed to run in the background. Otherwise, it can't.
- Permission to play a sound, vibration pattern or LED pattern is given to the app only at the time of the alarm or at the time the alarm repeats.

## Alarm Widget

An alarm widget only displays active alarms (i.e. those that have been enabled) and can open the app with the alarms. The widget can have an action to cancel (disable) an alarm.

## Portal and App User Interfaces

The portal UI is a standard dialog for adding and updating alarms.

Deleting an alarm can be done using a delete button in the standard dialog or in the app UI. If the delete button is in the app UI, it triggers the display of a standard delete confirmation dialog. The minimum number of relevant properties appear in this dialog to indicate to the user which alarm is deleted.

Alarms can be enabled and disabled, but this is usually done in the app UI. In this case, the ideal would be to use a confirmation dialog box (presenting the name of the alarm and/or the ringing duration). To enable and disable several alarms, the app could have an "Apply" action which triggers a dialog box listing all the alarms to be enabled and disabled.

## Alarm Registration

The properties for registering an alarm are:
- Name. Optional.
- Description. Optional.
[...]

PR #1304, opened 2024-03-15
Draft: docs/notification: Write XML interface definition for version 2

jsparber:
This is to be considered the direction I will move the notification portal to. 
**But it's implementation won't happen in this pull request.**

This takes the discussion in
https://github.com/flatpak/xdg-desktop-portal/issues/983 and turns it into an actual possible v2 of the notification interface.

The permission part was intentionally left out since it's tangential to the new properties and features proposed here.

### DIFF
--- a/data/org.freedesktop.portal.Notification.xml
+++ b/data/org.freedesktop.portal.Notification.xml
@@ -41,7 +41,7 @@
       #org.freedesktop.portal.Notification::ActionInvoked signal
       to the application.
 
-      This documentation describes version 1 of this interface.
+      This documentation describes version 2 of this interface.
   -->
   <interface name="org.freedesktop.portal.Notification">
     <!--
@@ -53,37 +53,111 @@
 
         The ID can be used to later withdraw the notification.
         If the application reuses the same ID without withdrawing,
-        the notification is replaced by the new one.
+        the notification is updated with the new one. It's possible
+        to set a ``display-hint`` to animate replacing the notification
+        instead of updating it.
 
         The format of the serialized notification is a vardict, with
         the following supported keys, all of which are optional:
 
+        * ``application-id`` (``s``)
+
+          An optional application ID for Flatpak applications, otherwise it's mandatory.
+          This has to be the Flatpak application ID or a desktop file ID, with optionally a postfix.
+          There is no reason to set this property if the application uses Flatpak and no postfix is set.
+          E.g ``org.libreoffice.LibreOffice.Writer`` where
+          ``org.libreoffice.LibreOffice`` is the flatpak application ID.
+          If this property isn't set it will be picked up automatically, if possible.
+
+          See `Application ID <https://flatpak-docs.readthedocs.io/en/latest/conventions.html>`_ and
+          `Desktop File ID <https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id>`_
+
         * ``title`` (``s``)
 
           User-visible string to display as the title.
+          The title should be a short string, if is too
+          long it may not be fully displayed.
 
         * ``body`` (``s``)
 
-          User-visible string to display as the body.
+          User-visible string to display as the body of the notification.
+          This can be a long description but note that if it is too long
+          it may not be fully displayed.
+
+          Any markup is removed from the string including new lines.
+
+        * ``markup-body`` (``s``)
+
+          The same as ``body`` but with support for markup formatting.
+          The markup is XML-based and supports a small subset of HTML
+          ``<b>...</b>``, ``<i>...</i>`` and ``<a href="...">...</a>``.
+
+          Any markup not supported, e.g. new lines, will be removed from
+          the string. In the future, the set of supported markup may be extended.
+
+          This can be a long string but note that it may not be fully displayed.
 
         * ``icon`` (``v``)
 
-          Serialized icon (see `g_icon_serialize()
-          <https://docs.gtk.org/gio/method.Icon.serialize.html>`_).
+          A serialized icon to add to the notification.
+          Supported image formats are png, jpeg or svg.
 
-          The portal only accepts serialized GThemedIcon and GBytesIcons. Both of
-          these have the form (sv). For themed icons, the string is "themed",
-          and the value is an array of strings containing the icon names.
-          For bytes icons, the string is "bytes", and the value is a bytestring
-          containing the icon data in png, jpeg or svg form. For historical
-          reasons, it is also possible to send a simple string for themed
-          icons with a single icon name.
+          The format for
+          serialized icon is a tuple (sv) with the following supported keys:
+
+          * ``themed`` (``as``)
+
+            A themed icon containing an array of strings with the icon names.
+
+            This is the same format as a serialized `GThemedIcon
+            <https://docs.gtk.org/gio/class.ThemedIcon.html>`_.
+
+          * ``bytes`` (``ay``)
+
+            A bytes icon containing a bytestring with the icon data in
+            png, jpeg or svg form.
+
+            This is the same format as a serialized `GBytesIcon
+            <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
+
+          * ``file-descriptor`` (``h``)
+
+            A file descriptor to an image file in png, jpeg or svg form.
+
+          For historical reasons, it is also possible to send a simple string
+          for themed icons with a single icon name.
 
           There may be further restrictions on the supported kinds of icons.
 
+        * ``image`` (``a{sv}``)
+
+          A serialized image to add to the body of the notification. The format for
+          serialized ``image`` is the same as for ``icon``.
+          Supported image formats are png, jpeg or svg.
+
+          For accessibility, an alternative textual description should be set via ``image-alt``.
+
+        * ``image-alt`` (``s``)
+
+          A textual description of the image inside the body.
+          This is used for accessibility, screen readers will read out this description.
+
+        * ``progress`` (``a{sv}``)
+
+          If set a progress bar will be shown in the notification.
+
+          For now the the supported properties for a progress are:
+
+          * ``value`` (``d``)
+
+            The ``value`` of the progress in range between ``0.0`` and ``1.0``
+
         * ``priority`` (``s``)
 
-          The priority for the notification. Supported values:
+          The priority for the notification. The notification server may
+          ignore this property if it has it's own policy.
+
+          Supported values:
 
           - ``low``
           - ``normal``
@@ -106,21 +180,361 @@
 
           * ``label`` (``s``)
 
-            User-visible label for the button. Mandatory.
+            User-visible label for the button.
+            If this property is missing the action may not be displayed.
 
           * ``action`` (``s``)
 
             Name of an action that is exported by the application. The action
-            will be activated when the user clicks on the button. Mandatory.
+            will be activated when the user clicks on the button.
+            If this property is missing the action may not be displayed.
 
           * ``target`` (``v``)
 
             Target parameter to send along when activating the action.
+
+          * ``icon`` (``v``)
+
+            A serialized icon to be added to a button. The format for
+            is the same as for the notification ``icon``.
+
+          * ``purpose`` (``s``)
+
+            The ``purpose`` of the button. This information may be used by the
+            notification server to style the button specially.
+
+            The standardized hints are defined as part of a ``content-type``.
+            Applications are allowed to use any string as a ``purpose`` but
+            should stick to the form ``class.specific``.
+            If the server doesn't understand the ``purpose` it will be
+            ignored and the button will be shown as a normal button.
+
+            Additional purposes may be defined by notification servers.
+
+          Any of the above button properties may be omitted, if the notification server
+          can't display the button with the given information it won't display it at all.
+
+        * ``actions`` (``aa{sv}``)
+
+          Array of actions that the notification server will invoke
+          automatically in specific cases. These actions may be called never or even multiple times.
+
+          Invoked actions will contain as parameter a vardict ``a{sv}`` containing the
+          ``target` and the response of the action.
+
+          The format for a serialized action is a vardict with the following supported keys:
+
+          * ``action`` (``s``)
+
+            Name of the action. This action needs to be exported by the application.
+
+          * ``purpose``
+            The purpose of the action.
+
+            * ``system.custom-alert``:
+
+              This action will be called by the notification server
+              whenever the notification is shown.
+
+              The purpose of this action is to allow apps to use custom methods of
+              notifying the user, for example, to play audio from a special
+              source like a streaming service or a radio station.
+
+            * ``user.replay-with-text``:
+
+              Inline replies for instant messaging apps.
+              The user-provided text will be added to the response vardict as ``reply-text`` (``s``);
+
+            Additional purposes may be defined by notification servers.
+
+          * ``target`` (``v``)
+
+            Target parameter to send along when activating the action.
+
+            This target will be added to the response vardict as ``target`` (``v``);
+
+        * ``content-type`` (``s``)
+
+          The ``content-type`` describes the type of content of a notification.
+          A notification server may use this information to display the
+          notification specially. Some content types also include
+          button purposes that can be set for a button so that the notification
+          can know the purpose of the button.
+
+          The following types are standarized so far:
+
+          * ``im.message``
+
+            Intended for instant messaging apps displaying notifications for new messages.
+
+          * ``alarm.ringing``
+
+            Intended for alarm clock apps
+
+          * ``call.incoming``
+
+            Intended for call apps to notify the user about an incoming call.
+
+            This type has the following button purposes:
+
+            * ``call.accept``:
+
+              Accept the incoming call.
+
+            * ``call.decline``:
+
+              Decline the incoming call.
+
+          * ``call.ongoing``
+
+            Intended for call apps while a call is ongoing.
+
+            This type has the following button purposes:
+
+            * ``call.decline``:
+
+              Hang up the ongoing call.
+
+            * ``call.enable-speakerphone``:
+
+              Enable the speakerphone for the ongoing call.
+
+            * ``call.disable-speakerphone``:
+
+              Disable the speakerphone for the ongoing call.
+
+          * ``call.missed``
+
+            Intended to be used by call apps when a call was missed.
+
+          * ``weather.warning.extreme``
+
+            Intended to be used to display an extreme weather warning.
+
+          * ``cellbroadcast.danger.extreme``
+
+            Intended to be used to display extreme danger warnings broadcasted by the cell network.
+
+          * ``cellbroadcast.danger.severe``
+
+            Intended to be used to display severe danger warnings broadcasted by the cell network.
+
+          * ``cellbroadcast.amberalert``
+
+            Intended to be used to display amber alerts broadcasted by the cell network.
+
+          * ``cellbroadcast.test``
+
+            Intended to be used to display tests broadcasted by the cell network.
+
+          * ``os.battery.low``
+
+            Intended to be used to indicate that the system is low on battery.
+
+          * ``browser.web-notification``
+
+            Intended to be used by browsers to mark notifications as web notifications
+
+          Additional types may be defined by notification servers.
+
+        * ``sound`` (``v``)
+
+        A serialized sound to add to the notification.
+        Supported image formats are oga and wav.
+
+        The format for
+        serialized sound is a tuple (sv) with the following supported keys:
+
+        * ``bytes`` (``ay``)
+
+          A bytes sound containing a bytestring with the sound data.
+
+          This is the same format as a serialized `GBytesIcon
+          <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
+
+        * ``file-descriptor`` (``h``)
+
+          A file descriptor to a sound file.
+
+        * ``silent``
+
+          A "silent" sound indicates that the system setting should be ignored and
+          no sound should be played for the notification.
+
+        There may be further restrictions on the supported kinds of sounds.
+
+        * ``vibration`` (``au``)
+
+          Vibration pattern is an array of uint32 to describe
+          alternating periods in which the device should vibrate and not vibrate in milliseconds.
+          This array can contain an odd number of items since the last non-vibrating period can be omitted.
+
+          An empty array indicates to ignore system settings and don't vibrate.
+
+          This is a similar API to `Web Vibration API <https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API>`_.
+
+        * ``led-color`` (``(yyyy)``)
+
+          Notification LED color as 4 bytes value ``(red, green, blue, alpha)``.
+
+          A tuple of ``(0000)`` indicates to ignore system settings and don't turn on the LED.
+
+        * ``display-hint`` (``as``)
+
+          An array of ways to display the notification. If none are set, or the notification server has its own policy, it is
+          free to decide how and where to display the notification.
+
+          * ``transient``
+
+            The notification is displayed only as a banner and won't be kept
+            by the server in a tray.
+
+            It's a programmer error to specify ``tray`` at the same time.
+
+          * ``tray``
+
+            No banner for the notification will be displayed and
+            the notification is placed in the tray.
+
+            It's a programmer error to specify ``transient`` at the same time.
+
+          * ``persistent``
+
+            Make the notification persistent in the notification tray.
+            The user can’t dismiss it using the usual close button or gesture.
+
+            Apps are only allowed to display persistent notifications
+            as long as they have a window. Once the last window of an app
+            is closed the persistent notification will be removed.
+
+          * ``hide-on-lockscreen``
+
+            Don't show the notification on the lockscreen.
+
+          * ``hide-content-on-lockscreen``
+
+            Only show the notification ``title`` and ``icon`` on the lockscreen.
+
+          * ``show-as-new``
+
+            If a notification with the same ``id`` of the app exists already
+            replace the previous notification, by removing the old notification
+            (including animation, etc) and adding a new notification.
+
+            If this hint isn't specified the notification is updated without any flickering.
+
+          Additional purposes may be defined by notification servers.
+
+        * ``grouping`` (``a{sv}``)
+          Notifications can be grouped. This behavior is often called threading.
+          For example, this is useful for multiple messages in the same chat.
+
+          * ``id`` (``s``)
+            Notifications with the same ID will be grouped.
+
+          * ``title`` (``s``)
+            The title of the notification group.
+
+          * ``icon`` (``v``)
+            The icon of the notification group.
+            This is the same format as the notification ``icon`` property.
       -->
     <method name="AddNotification">
       <arg type="s" name="id" direction="in"/>
       <arg type="a{sv}" name="notification" direction="in"/>
     </method>
+
+    <!--
+        ListSupportedProperties:
+        @supported_properties: A list of supported properties and their supported values, if applicable
+
+        Returns a list of supported properties with
+        specific values if applicable for the property.
+
+        Applications only need to check this list for support of a property
+        if they want to implement a fallback for it to ensure that the user is
+        correctly notified.
+
+        The following list contains all possible properties:
+
+        * ``application-id``
+        * ``title``
+        * ``body``
+        * ``icon`` (``as``)
+
+          - ``themed``
+          - ``bytes``
+          - ``file-descriptor``
+
+        * ``image`` (``as``)
+
+          - ``themed``
+          - ``bytes``
+          - ``file-descriptor``
+
+        * ``image-alt``
+
+        * ``priority`` (``as``)
+
+          - ``low``
+          - ``normal``
+          - ``high``
+          - ``urgent``
+
+        * ``default-action``
+        * ``default-action-target``
+        * ``buttons`` (``a{sv}``)
+
+          - ``label``
+          - ``icon`` (``as``)
+          - ``action``
+          - ``target``
+          - ``purpose`` (``as``)
+
+        * ``actions`` (``a{sv}``)
+
+          - ``action``
+          - ``target``
+          - ``purpose`` (``as``)
+
+        * ``content-type`` (``as``)
+
+          - ``im.message``
+          - ``alarm.ringing``
+          - ``call.incoming``
+          - ``call.ongoing``
+          - ``call.missed``
+          - ``weather.warning.extreme``
+          - ``cellbroadcast.danger.extreme``
+          - ``cellbroadcast.danger.severe``
+          - ``cellbroadcast.amberalert``
+          - ``cellbroadcast.test``
+          - ``os.battery.low``
+          - ``browser.webNotification``
+
+          If the notification server uses custom content types the server advertises them in this list.
+
+        * ``sound`` (``as``)
+
+          - ``bytes``
+          - ``file-descriptor``
+          - ``silent``
+
+        * ``vibration``
+        * ``led-color``
+        * ``display-hint`` (``as``)
+
+          - ``transient``
+          - ``tray``
+          - ``persistent``
+          - ``hide-on-lockscreen``
+          - ``hide-content-on-lockscreen``
+          - ``show-as-new``
+    -->
+    <method name="ListSupportedProperties">
+      <arg type="a{sv}" name="supported_properties" direction="out"/>
+    </method>
+
     <!--
         RemoveNotification:
         @id: Application-provided ID for this notification
--- a/data/org.freedesktop.impl.portal.Notification.xml
+++ b/data/org.freedesktop.impl.portal.Notification.xml
@@ -48,6 +48,20 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
       <arg type="a{sv}" name="notification" direction="in"/>
     </method>
+
+    <!--
+        ListSupportedProperties:
+        @supported_properties: A list of supported properties and their supported values, if applicable
+
+        Returns a list of supported properties with
+        specific values if applicable for the property.
+
+        The format of the @supported_properties is the same as for
+        :ref:`org.freedesktop.portal.Notification.ListSupportedProperties`.
+    -->
+    <method name="ListSupportedProperties">
+      <arg type="a{sv}" name="supported_properties" direction="out"/>
+    </method>
     <!--
         RemoveNotification:
         @app_id: App id of the application


data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+            * ``system.custom-alert``:
+
+              This action will be called by the notification server
+              whenever the notification is shown.
+
+              The purpose of this action is to allow apps to use custom methods of
+              notifying the user, for example, to play audio from a special
+              source like a streaming service or a radio station.
+
+            * ``user.replay-with-text``:
```
camelCaseNick, line 243: There is a typo:

```suggestion
            * ``user.reply-with-text``:
```

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+            Intended to be used to display severe danger warnings broadcasted by the cell network.
+
+          * ``cellbroadcast.amberalert``
+
+            Intended to be used to display amber alerts broadcasted by the cell network.
+
+          * ``cellbroadcast.test``
+
+            Intended to be used to display tests broadcasted by the cell network.
```
camelCaseNick, line 328: Is it specifically designed not to support push notification based warn apps, like EUwarn, Katwarn, or NINA? Wouldn't it make sense to broaden the scope from `cellbroadcast.*` to any type of `broadcast.*`?

jonas2515: Cell broadcast is standardized, so it was easy to find the 5 kinds of alerts that can be sent via cell broadcast, and it made sense to add them here as they are unlikely to change.

For all those warn-apps, afaik there's no standard, and I'm not sure which different kinds of alerts they are sending. Possibly the alerts are grouped into kinds similar to the cell broadcasts ones, but here someone would have to do the research.

Mikenux: @jonas2515: So what is their relationship with their use by apps?

camelCaseNick: Cell broadcast itself doesn't have those, doesn't it? It is used for all kinds of things, including public warning messages. AfaIk, those specified entirely elsewhere, incl. in the Wireless Emergency Alerts, or EU-Alert specifications. They can be serialised as XML with the Common Alerting Protocol. All those share similar urgency and importance definitions. And at least all those specifications listed above are used outside cell-broadcasted, as well. (Maybe I'm missing something. I tried to find relevant specifications and protocols, but it feels like the big picture is complicated, with essentially compatible, but in different jurisdictions re-defined systems.)

data/org.freedesktop.portal.Notification.xml
jsparber: thanks :)

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
         If the application reuses the same ID without withdrawing,
-        the notification is replaced by the new one.
+        the notification is updated with the new one. It's possible
+        to set a ``display-hint`` to animate replacing the notification
+        instead of updating it.
 
         The format of the serialized notification is a vardict, with
         the following supported keys, all of which are optional:
 
+        * ``application-id`` (``s``)
```
Mikenux, line 63: Why have this property? Isn't it already picked up for flatpak? Or it is not retrieved in case of postfix? What is the goal?

The description is also confusing: mandatory, optional, with optionally a postfix...

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
         * ``body`` (``s``)
 
-          User-visible string to display as the body.
+          User-visible string to display as the body of the notification.
+          This can be a long description but note that if it is too long
+          it may not be fully displayed.
+
+          Any markup is removed from the string including new lines.
+
+        * ``markup-body`` (``s``)
```
Mikenux, line 89: This is the same as in the notification specification. But I'm wondering why there is this, instead of just querying if the server supports markup, thus only having the 'body' property instead of two properties.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+          For historical reasons, it is also possible to send a simple string
+          for themed icons with a single icon name.
 
           There may be further restrictions on the supported kinds of icons.
 
+        * ``image`` (``a{sv}``)
+
+          A serialized image to add to the body of the notification. The format for
+          serialized ``image`` is the same as for ``icon``.
+          Supported image formats are png, jpeg or svg.
```
Mikenux, line 136: What to expect in terms of notification appearance when the icon and image are used? I assume the icon is on the left and the image at the bottom, but I think it's good to be clear.

Mikenux, line 136: What to expect in terms of notification appearance when the icon and image are used? I assume the icon is on the left and the image at the bottom, but I think it's good to be clear.

data/org.freedesktop.portal.Notification.xml
AdrianVovk: Because even if the server supports markup, some clients may want to send notifications without markup

If you have an app that you know won't use markup, you can send normal `body` and avoid having to sanitize the contents

data/org.freedesktop.portal.Notification.xml
AdrianVovk: Flatpaks: We can get an app-id from Flatpak, so it's optional to specify it here. But if you want to ship a app-id w/ postfix, we can't automatically determine it so you must specify it

Non-flatpaks: We can't get an app-id from anywhere so it's required

IMO it is a bit complicated seemingly pointlessly. We can probably just mark it required in the spec and just error out if a Flatpak tries to lie about who it is

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+            * ``call.decline``:
+
+              Hang up the ongoing call.
+
+            * ``call.enable-speakerphone``:
+
+              Enable the speakerphone for the ongoing call.
+
+            * ``call.disable-speakerphone``:
```
AdrianVovk, line 302: As mentioned in the other thread, we should be able to pause/unpause the call

I also think it makes sense to give a mute/unmute button too

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+          * ``im.message``
+
+            Intended for instant messaging apps displaying notifications for new messages.
+
+          * ``alarm.ringing``
+
+            Intended for alarm clock apps
+
+          * ``call.incoming``
```
AdrianVovk, line 274: call.incoming.video?

I guess my video call proposals would require more work in the spec

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+        A serialized sound to add to the notification.
+        Supported image formats are oga and wav.
+
+        The format for
+        serialized sound is a tuple (sv) with the following supported keys:
+
+        * ``bytes`` (``ay``)
+
+          A bytes sound containing a bytestring with the sound data.
```
AdrianVovk, line 350: sound data in what format? MP3? OGG? Raw PCM?

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+          This is the same format as a serialized `GBytesIcon
+          <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
```
AdrianVovk, line 353: this is unclear. The docs you're linking here talk about GBytesIcon holding a "common image file format like PNG". Doesn't make sense for sound bytes...

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+            Intended to be used to indicate that the system is low on battery.
+
+          * ``browser.web-notification``
+
+            Intended to be used by browsers to mark notifications as web notifications
+
+          Additional types may be defined by notification servers.
+
+        * ``sound`` (``v``)
```
AdrianVovk, line 340: Did you mean `(sv)` here? rather than just `v` and then saying it's an `(sv)` below?

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+          long it may not be fully displayed.
 
         * ``body`` (``s``)
 
-          User-visible string to display as the body.
+          User-visible string to display as the body of the notification.
+          This can be a long description but note that if it is too long
+          it may not be fully displayed.
+
+          Any markup is removed from the string including new lines.
```
AdrianVovk, line 87: This makes it sound like if the client gives a string like `<i>foobar</i>` the notification server will just render `foobar` (just without making it italics). That's not what we want, though. We want an app to say `<i>foobar</i>` and the server should render `<i>foobar</i>`

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
         * ``priority`` (``s``)
 
-          The priority for the notification. Supported values:
+          The priority for the notification. The notification server may
+          ignore this property if it has it's own policy.
+
+          Supported values:
 
           - ``low``
           - ``normal``
```
AdrianVovk, line 163: (GitHub isn't letting me add a comment on `default-action` so I'll add it here)

I think we should specify that incoming call notifications aren't allowed to have `default-action`s. Same with incoming alarms. That way we'll be able to do the whole "clicking a notification opens the full-screen incoming call/alarm UI"

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+            The standardized hints are defined as part of a ``content-type``.
+            Applications are allowed to use any string as a ``purpose`` but
+            should stick to the form ``class.specific``.
+            If the server doesn't understand the ``purpose` it will be
+            ignored and the button will be shown as a normal button.
+
+            Additional purposes may be defined by notification servers.
+
+          Any of the above button properties may be omitted, if the notification server
+          can't display the button with the given information it won't display it at all.
```
AdrianVovk, line 215: Just to double-check: An ongoing call notification could add an action to hold the call w/ an action and target set, but no label or icon set, so that the notification server knows it's possible to place the call on hold but doesn't ever render the button. Right?

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+          * ``target`` (``v``)
+
+            Target parameter to send along when activating the action.
+
+          * ``icon`` (``v``)
+
+            A serialized icon to be added to a button. The format for
+            is the same as for the notification ``icon``.
+
+          * ``purpose`` (``s``)
```
AdrianVovk, line 201: Should notification servers be allowed to imply `label/icon` based on `purpose`?

In other words, if you have an incoming call should the app be allowed to say just `purpose: call.accept` and omit label or icon, and safely assume that the notification server will have a default string to answer / decline the call? Or should the app always set a label if it really wants the action to show up?

AdrianVovk, line 201: We should specify explicitly that the `purpose` allows the notification server to completely ignore any specified `label` or `icon`, or any other thing about presentation.

For example, in my call-waiting proposal sometimes `call.accept` will be displayed as two distinct actions: "Hold and Accept" or "End and Accept"

data/org.freedesktop.portal.Notification.xml
AdrianVovk: Ah, never mind. I see `actions` below

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+          * ``alarm.ringing``
+
+            Intended for alarm clock apps
+
+          * ``call.incoming``
+
+            Intended for call apps to notify the user about an incoming call.
+
+            This type has the following button purposes:
```
AdrianVovk, line 278: These can be used as action purposes too, right? We should specify that

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+        * ``silent``
+
+          A "silent" sound indicates that the system setting should be ignored and
+          no sound should be played for the notification.
+
+        There may be further restrictions on the supported kinds of sounds.
+
+        * ``vibration`` (``au``)
+
+          Vibration pattern is an array of uint32 to describe
```
AdrianVovk, line 368: Hmm maybe an example here could be useful? Or is the link to the web vibration API enough?

data/org.freedesktop.portal.Notification.xml
Mikenux: OGA and WAV, in "Supported image formats" (wrong word). :wink:

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+
+          * ``value`` (``d``)
+
+            The ``value`` of the progress in range between ``0.0`` and ``1.0``
+
         * ``priority`` (``s``)
 
-          The priority for the notification. Supported values:
+          The priority for the notification. The notification server may
+          ignore this property if it has it's own policy.
```
Mikenux, line 158: Although this is not a new property and is perhaps a clarification of the documentation, I think ignoring this property would not encourage app developers to use it in case of a cross-desktop UX is wanted. For this, it makes it useless.

Regarding this version of the portal, a priority does not seem necessary or is inherent to the content type, e.g. alarms. If you agree on that, then maybe ignore it or say not to use it when using a content type.

data/org.freedesktop.portal.Notification.xml
Mikenux: What about desktop files if an app is lying (assuming the app isn't sandboxed and can perhaps guess installed apps)? Furthermore, shouldn't such identification be elsewhere than in a specific portal?

data/org.freedesktop.portal.Notification.xml
Mikenux: Ok, thanks.

data/org.freedesktop.portal.Notification.xml
jsparber: the idea is to filter out markup like [FDO notifications](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#markup)  requires servers to do.

> That's not what we want, though.

@AdrianVovk Any reason why you say that we don't want that?

data/org.freedesktop.portal.Notification.xml
jsparber: > What about desktop files if an app is lying (assuming the app isn't sandboxed and can perhaps guess installed apps)? Furthermore, shouldn't such identification be elsewhere than in a specific portal?

Host applications don't need to guess they can just look up the desktop files. We can't force anything on them.

I think renaming this to `desktop-file-id` makes the distinction between app id and this property much better. We can have something similar to already existing portals https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.DynamicLauncher.html

data/org.freedesktop.portal.Notification.xml
jsparber: that's not part of the specs, this will depend on the notification server.

data/org.freedesktop.portal.Notification.xml
jsparber: that's the reason why i removed the mandatory part of the `label` and `action`  property. In general the app should still set a `label` because the server may not understand the `purpose` or need it even when the button has a `purpose`.

data/org.freedesktop.portal.Notification.xml
jsparber: The idea is to make it the same structure as `icon` just to reduce confusion, but the wording cold be better especially if we really go with the `slient` sound thing.

data/org.freedesktop.portal.Notification.xml
jsparber: The sound property inspired by `GIcon` and has mostly the same serialization format. `GBytesIcon` don't actually care about the content they don't even know about it. But yes reading this sounds odd.

data/org.freedesktop.portal.Notification.xml
Mikenux: What is the purpose of this? If it's about getting the app name, this might be problematic if the app is lying, because the app name is useful for users to attribute a problem to the correct app. However, I don't know what the current status of notifications is with non-sandboxed apps.

data/org.freedesktop.portal.Notification.xml
Mikenux: In https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#icons-and-images "icon" is deprecated and "image" is used instead. However, there is "app_icon" which is clearly the app icon. So here, given that there is "image", what is the purpose of "icon" vs. "image"? The goal is more to clarify their purposes when both are used.

data/org.freedesktop.portal.Notification.xml
Mikenux: I don't know if having a way to fallback to a generic notification (i.e. presenting a normal button) is a good idea. One of the purposes of having specific types of notifications is to provide the user with the option to turn them off. Therefore, notifications must be identifiable as such thanks to their style. If a notification server doesn't support it, then adds their support, then that server should remove their normal display - probably not nice in terms of transition.

However, it is worth noting, in light of @AdrianVovk's proposal regarding apps on the lock screen, that it might not be possible to display the app itself if we think about ensuring privacy (e.g. speaker and microphone toggle, maybe don't show name of important contacts). I mention this because app developers in this case may want to customize the system-styled notification (e.g. using colors), which may or may not impact identification.

data/org.freedesktop.portal.Notification.xml
jsparber: i actually implemented it already, but this sounds like it makes a lot of sense to pass `body` unseen.

data/org.freedesktop.portal.Notification.xml
jsparber: > Action property should still be mandatory, no?

Yeah probably.

Mikenux: > Hmm? No? If a notification server supports incoming call notifications, I don't think it makes much sense to disable support for incoming call notifications. The server either supports them or not
> 
> Also Jonas wasn't interested in trying to filter groups of notifications from apps. Apps will either be able to send all notifications or no notifications

I thought of that at first, as this was in my mockups to offer a global experience of notifications (https://gitlab.gnome.org/Teams/Design/whiteboards/uploads/cd2437eaa7f0bb486adc9fb67fb4064a/NotificationProfilesv3.png). However, if an app supports multiple types of notifications, users may still want to not receive some of them. In case of calls, maybe someone wants to never have video call (or not activate the camera accidentaly) and just receive audio calls? Also: what about other desktops? what if the app has other notification types (don't know which, but who knows...) [here for managing]?

> This is fine. Software and UI is updated all the time. Presumably updates made by the notification server will improve the UI, and make the notifications better to use, not worsen it.
> 
> Notifications servers that don't support a custom UI yet can trivially fall back to generic notification with generic action API. That's simply the best we can do while ensuring the notification is delivered

That does not resolve the privacy issue if apps can do what they want: with a no-constrained type they have no obligation to offer speaker and microphone enabling/disabling.

> Sorry I'm confused... Where is the conversation about showing apps themselves coming from?

I re-read, and misinterpretation, sorry. I mixed that with the other apps (calculator, etc.).

Mikenux: As far as I know, there are currently no such notifications; they can therefore be implemented. Additionally, if we can communicate with a stylized notification (how to give permission is currently unknown) and with specific (non-custom) buttons, permission (purpose of portals) can be granted for audio (speaker and microphone) and the camera when the relevant buttons are pressed. The permission can be revoked by clicking on it again and automatically at the end of the call. By having them as generic notifications because support for their content type is lacking, how can we reliably grant permission due to this lack of support? Also, if an app wants to be sandboxed by supporting the content types, it cannot have such dynamic permission, or it means giving the app permission to always access the speaker and the microphone when launched or to have differentiated management of permissions (dynamic if content type supported, general if content type not supported).

Mikenux:
For buttons, if custom buttons are ultimately allowed in addition to those added due to content type, perhaps have a limit on their total number.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+          * ``cellbroadcast.amberalert``
+
+            Intended to be used to display amber alerts broadcasted by the cell network.
+
+          * ``cellbroadcast.test``
+
+            Intended to be used to display tests broadcasted by the cell network.
+
+          * ``os.battery.low``
```
Mikenux, line 330: For applications? If yes, case examples.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+          A bytes sound containing a bytestring with the sound data.
+
+          This is the same format as a serialized `GBytesIcon
+          <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
+
+        * ``file-descriptor`` (``h``)
+
+          A file descriptor to a sound file.
+
+        * ``silent``
```
Mikenux, line 359: I assume it's for alarms, but, referring to the alarm example mentioned in the issue, it doesn't make sense to me to have one or more alarm notifications in the tray. This risks overloading it or there is wasted space if grouped.

If there are no other cases, then this property should not exist. This is an implicit property for an alarm.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+          Vibration pattern is an array of uint32 to describe
+          alternating periods in which the device should vibrate and not vibrate in milliseconds.
+          This array can contain an odd number of items since the last non-vibrating period can be omitted.
+
+          An empty array indicates to ignore system settings and don't vibrate.
+
+          This is a similar API to `Web Vibration API <https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API>`_.
+
+        * ``led-color`` (``(yyyy)``)
```
Mikenux, line 376: Maybe add a pattern for a silent mode. Colors + alternation between increasing and decreasing brightness, with durations. Brightness intensity (if applicable) should be done on the system side.

Mikenux:
Additional purposes and types. As far as I know there is nothing in the portals that cannot be evaluated. I believe this idea will be rejected.

For vibration and LED color, how can apps present users with an interface to set a custom pattern/color, choose one, or test it?

data/org.freedesktop.portal.Notification.xml
jsparber: no, actions have a different set of purposes. Maybe action's `purpose` field should have a different name to make the distinction clear.

data/org.freedesktop.portal.Notification.xml
Mikenux: I mean, still using the example of alarms, it's up to the system to decide where to draw a pending alarm. This is not something an app should control.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+            as long as they have a window. Once the last window of an app
+            is closed the persistent notification will be removed.
+
+          * ``hide-on-lockscreen``
+
+            Don't show the notification on the lockscreen.
+
+          * ``hide-content-on-lockscreen``
+
+            Only show the notification ``title`` and ``icon`` on the lockscreen.
```
Mikenux, line 416: I'm really wondering how this is user defined, since it's the user's decision whether to show a notification on the lock screen and whether or not to hide the body (and some buttons?). I think this is something that should be controlled by the system through content type or with hints (e.g. calendar type for event notifications - just an example here as these notifications are not covered here).

If you have another idea on how to handle this, please explain. Indeed, when it comes to privacy, I think we need reliability: a mistake or bad behavior is not tolerated.

AdrianVovk: Nack. Apps should be able to say that their content is sensitive no matter what the notification server thinks about it, and if the user configures their notifications to appear on the lock screen an app should be able to say "no no, this is actually really sensitive"

data/org.freedesktop.portal.Notification.xml
AdrianVovk: Silent notifications can be used for lots of other things...

"Hey you should reboot your device for an update" should be a silent notification

"hey your ebike battery is running low, you should make sure to charge it" from your bike's companion app is a non-time-sensitive notification that doesn't warrant a noise but still should be a notification just in case the user really needs that reminder

"you have data left over from 15 uninstalled apps, click here to manage" should be silent too.

I can come up with infinitely more use-cases for silent notifications. They're necessary...

data/org.freedesktop.portal.Notification.xml
AdrianVovk: ok then for the call-waiting thing we'd need to define an action for ongoing-call notifications to put the call on hold

AdrianVovk: But I'm not really sure why we should have necessarily separated purposes for actions and buttons. Buttons are just actions that we try to show in the notification's UI (especially if the notification server doesn't know about the notification's content type).

jsparber: they are pretty different. `buttons`  are always be buttons in the ui, but the placement, styling or labeling may be different then a "normal" button. `actions` on the other hand are ways to allow the notification server to communicate to the app. The two options we have right now can't be displayed as buttons.

data/org.freedesktop.portal.Notification.xml
Mikenux: Adrian, that's the example I have in mind because I worked on that, but they aren't an object of discussion here. Besides you already have the response on how I manage that in the mockup I linked.

> If we tie show-on-lockscreen to the content type, we're painting ourselves into a corner: apps sending similar notification types in different situations won't be able to define different behavior

With a global only configuration of notifications, yes we have a problem. However, this problem is resolved by making this setting available per app (which makes sense as it is a per app setting).

Also, for me, sensitive information need reliability, I think we need an interface to agree to let the information be displayed or not. Having notification content type is then needed to have a revocable setting (knowing that we need to define theses types).

You have examples on how I define importance (priority) and sensitivity in the following issues: https://github.com/flatpak/xdg-desktop-portal/issues/1025 and https://github.com/flatpak/xdg-desktop-portal/issues/1026. If you have comments, please comment on those instead to avoid a possible huge flow of comments.

> an app should be able to say "no no, this is actually really sensitive"

I'm not against having apps defining such setting if there are cases for that, but examples are needed then. The thing is that a fully controlled app setting also need to be user-controlled (relative to what I already said on sensitivity).

data/org.freedesktop.portal.Notification.xml
agx: > Cell broadcast is standardized, so it was easy to find the 5 kinds of alerts that can be sent via cell broadcast, and it made sense to add them here as they are unlikely to change.

Cell broadcasts has a lot of possible channels, also additional information on whether this should be a popup or not. As the apps that provide that information will be rather a few than many woudln't it be better to have this in a separate portal that is then also easier to tweak rather than in a very generic portal used by many apps that will be harder to change / agree on changes?

data/org.freedesktop.portal.Notification.xml
agx, line 376: Different devices have different leds. Wouldn't it be better to specify the led type (e.g. the notification led) and have the device pick the right color? Also in most cases it would likely be better to have the notification server pick the right feedback for an event and to use this only for overrides which then would be good to spell out here.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+
+            * ``call.decline``:
+
+              Decline the incoming call.
+
+          * ``call.ongoing``
+
+            Intended for call apps while a call is ongoing.
+
+            This type has the following button purposes:
```
agx, line 292: Instead of trying to munge the call interaction with all it's subtleties into a generic notification API wouldn't it be better to have a specific interface/portal for that? The receiving side is then free to display things in the same way as a notification but that shouldn't be a requirement. Things like switching ongoing calls between different "call providers" (say matrix, signal and regular phone calls) will otherwise be more limited than they need to be. Same goes for audio device selection, etc.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+          If this property isn't set it will be picked up automatically, if possible.
+
+          See `Application ID <https://flatpak-docs.readthedocs.io/en/latest/conventions.html>`_ and
+          `Desktop File ID <https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id>`_
+
         * ``title`` (``s``)
 
           User-visible string to display as the title.
+          The title should be a short string, if is too
+          long it may not be fully displayed.
```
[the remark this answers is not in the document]
AdrianVovk: I think that's entirely up to the notification server that will be displaying the notification. So we can't specify what "too long" means here. It means however much fits in the UI of the server

App developer needs to understand that they should avoid sending lots of text. They'll probably want to test the notification on the common desktops

Android doesn't have a defined limit either. It just says that
> By default, the notification's text content is truncated to fit one line

We could probably specify something similar

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+
+            This is the same format as a serialized `GBytesIcon
+            <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
+
+          * ``file-descriptor`` (``h``)
+
+            A file descriptor to an image file in png, jpeg or svg form.
+
+          For historical reasons, it is also possible to send a simple string
+          for themed icons with a single icon name.
```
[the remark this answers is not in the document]
AdrianVovk: v2 is backwards compatible w/ v1 I'm pretty sure

data/org.freedesktop.portal.Notification.xml
```diff
@@ -53,37 +53,111 @@
...
+
+          * ``bytes`` (``ay``)
+
+            A bytes icon containing a bytestring with the icon data in
+            png, jpeg or svg form.
+
+            This is the same format as a serialized `GBytesIcon
+            <https://docs.gtk.org/gio/class.BytesIcon.html>`_.
+
+          * ``file-descriptor`` (``h``)
```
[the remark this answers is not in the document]
AdrianVovk: When sending a file handle over dbus it is always the responsibility of the receiver to close the file handle because you get a duplicate. This is true for at least gdbus and sd-bus

Wdym by lifetime guarantees? As long as the FD is open it's possible to read from it - enforced by the kernel

data/org.freedesktop.portal.Notification.xml
AdrianVovk: This notification portal (w/ the small changes I'm asking for) is pretty much a superset of CallKit, which is Apple's special API for dealing with call interaction. In other words, this portal can do everything CallKit can do, and conversely any API we come up with trying to emulate CallKit will end up mostly being a duplicate of this notifications portal. So I see no reason to try and separate the two

data/org.freedesktop.portal.Notification.xml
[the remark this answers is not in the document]
agx: > When sending a file handle over dbus it is always the responsibility of the receiver to close the file handle 

I'm aware but would have spelled it out. Looking through other protocols that doesn't seem to be the case there either though so I guess it's fine.

> Wdym by lifetime guarantees? As long as the FD is open it's possible to read from it - enforced by the kernel

I was thinking about pipes or similar but rereading the spec it says "to an image file" so that's likely fine too. Thanks for pointing me at it.

data/org.freedesktop.portal.Notification.xml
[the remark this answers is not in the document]
agx: :+1: That would at least give an idea.

data/org.freedesktop.portal.Notification.xml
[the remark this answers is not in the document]
AdrianVovk: Probably works for pipes too. If you open an anonymous pipe, write into the write end, and hand the read end over dbus, then exit the program, I think the kernel will keep the written data in a buffer in memory until the FD is either closed or read from...

data/org.freedesktop.portal.Notification.xml
agx: Are we talking about https://developer.apple.com/documentation/callkit ? That has a much wider scope like blocking incoming calls, etc.

data/org.freedesktop.portal.Notification.xml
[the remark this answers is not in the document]
Mikenux: > They'll probably want to test the notification on the common desktops

For app developers, I think it would be better for them to avoid testing on different desktops, so taking into account different ways of presenting. Remember that Android has the advantage of just having one thing, not multiple. Also, the developers needs to know what a common desktop is.

> Android doesn't have a defined limit either. It just says that
> 
> > By default, the notification's text content is truncated to fit one line

That's a start. The other concern, if we want to avoid app developers having to test their notifications on multiple DEs, is if there is a way for all text to be displayed (i.e. will all text be displayed?). In GNOME, Julian implemented a button to expand a notification.

data/org.freedesktop.portal.Notification.xml
Mikenux: @agx: You can read discussion from https://github.com/flatpak/xdg-desktop-portal/pull/1304/files#r1532019682. We have a different understanding of what this portal is supposed to do.

data/org.freedesktop.portal.Notification.xml
Mikenux: Your examples make sense, but the user may want to be notified with a sound when the battery is low, depending on what is important to them or when they is away from their device. Also don't forget one of the keys: user control.

data/org.freedesktop.portal.Notification.xml
Mikenux: > Also Jonas wasn't interested in trying to filter groups of notifications from apps. Apps will either be able to send all notifications or no notifications

(I seem to have misread)

It makes sense to let apps send notifications if they have them. But, regardless of whether the type of content is used or not, this means that the app must first have permission to use the microphone (do we have a portal for this?) and the appropriate camera.

However, having a notification that allows to dynamically manage permissions (audio, video) is better in my opinion because more things can be offered (privacy, buttons do what they say, simply take advantage of dynamic permissions). Is this intended if content type is used?

data/org.freedesktop.portal.Notification.xml
Mikenux: > Wouldn't it be better to specify the led type (e.g. the notification led)

Isn’t it up to the device to use the right LED?

agx: But what if the right LED if you just give a color if it's not a RGB LED but e.g. a single color status LED. Also you can't give a pattern (as you note). You can also have multiple LEDS in different positions (front/back of the device). That's why I think it's better to say: I'd like to have feedback for this type of notification/event. 

Maybe:

If the device has an status LED this is meant as an override for the default LED feedback picked by the notification server. The color pattern might not be representable by the device. The notification server will pick an appropriate, user visible LED if possible.

Mikenux: Sorry, I'm not sure I understood what you said correctly.

For me, it's :
- Let the server choose the correct LED for feedback.
- For a pattern, let the app specify the colors, then let the server ignore the colors if applicable. If a LED can't handle brightness (like screen brightness), then turn the LED on and off (note: I don't know much about all the LEDs on a smartphone). The use of pattern is reserved for alarms and calls (for a generic notification, logically it is perhaps leaving the LED on?).

Does this correspond to what you want?

agx: I mean:

1. if the app doesn't set `led-color`: let the server pick a suitable led, color and pattern
2. if the app does specify a LED color and the device has a matching LED (either RGB or a fixed color LED matching the color) use that
3. if there's no matching LED do the same as in 1.

> The use of pattern is reserved for alarms and calls

Is that in the protocol? What about different patterns for SMS and IMs which is e.g. something I use to see at a glance what needs attention without even unblanking the phone's display.

data/org.freedesktop.portal.Notification.xml
AdrianVovk: Yes I'm talking about CallKit

and no, the scope isn't really that much wider. In theory sure it's a dedicated API to do all these things with calls. But if you look at how it's actually implemented: the app tells CallKit when a call is incoming, or when the app is in an ongoing call. It gives the server a list of actions that the app knows how to handle. The server, in response, implements everything by triggering these actions through the app. This setup just so happens to be extremely similar to this notifications API.

The server side can then implement whatever it wants: cross-app call waiting, do-not-disturb but allowing repeated callers, special incoming and ongoing call UI (that's what content-type is for, after all), etc. If the server side doesn't care then you'll still get basic notifications and bare-minimum functionality for call apps.

agx: This is the incoming call side. As far as I can tell callkit also handles the outgoing side as well, handles group calls dtmf, etc.  

That's why I think it makes more sense to have a  call specific portal rather than trying to match it to "notifications" with varying results depending on what the notification server actually implements. If we ever want sandboxed apps to do phone calls without exposing MMs API we need a call specific portal (which is even more desirable if that calls app is e.g. allowed to display content on/above the lock screen).

Initially this "calls" portal might only cover the call specific bits from this MR but it would give us the possibility to add more call specific features that don't map to notifications (like outgoing calls, joining calls together, emergency calling, etc).

Mikenux: I believe the problem is about how things are structured. When doing a portal, it is generally around a case (or multiple cases that can be grouped) and we have the following structure: method + method properties - which will translates here as, e.g.:
- AddIncomingCall (the method)
- @contact_id
- @contact_image
- @sound: system, ring/song, custom [radio, stream]
- @vibration_pattern
- @led_pattern
- @buttons (supported standardized buttons - perhaps not needed if using standardized buttons with dynamic permissions)
- @custom_buttons (additional non-standardized buttons - if needed, can only open the app, not available on lockscreen; e.g. opening contact list or just the app)
- All relevant signals depending on what button is activated in order to call another method, e.g. AddOngoingCall, etc.

This also offers clarity when reading the documentation. Here, we have multiple properties that seems structurally disorganized.

AdrianVovk: > This is the incoming call side. As far as I can tell callkit also handles the outgoing side as well, handles group calls dtmf, etc.

This handles the ongoing call side too! Note the ongoing call content type

Also, CallKit handles DTMF because CallKit is actually what renders the in-call UI for the Phone app on iOS. It's really strange. We're not interested in doing it that way. So we don't need to handle those things

> That's why I think it makes more sense to have a call specific portal rather than trying to match it to "notifications" with varying results depending on what the notification server actually implements.

The varying results is exactly what we want, though. We want Skype / Viber / whatever apps to be able to say "hey I'm in an ongoing call" and expect that at least some UI will appear, even if that UI is just a basic notification that everyone has supported forever. Desktop environments that don't care should retain basic functionality 


> If we ever want sandboxed apps to do phone calls without exposing MMs API we need a call specific portal (which is even more desirable if that calls app is e.g. allowed to display content on/above the lock screen).

The calls app cannot be sandboxed, especially on systems using home directory encryption. It'll be baked into the OS image.

I don't think having a portal to make actual phone calls makes sense. I can't think of any use-case for 3rd party phone apps that aren't solved by the fact that our OS is open source. On proprietary OSs people want third party phone apps to handle missing features from the built-in one

I don't think CallKit lets your app make actual phone calls via the SIM card. CallKit is only a framework for letting apps report to the OS that there's an incoming/ongoing (VOIP/whatever) call and expose some basic controls, no more.

> Initially this "calls" portal might only cover the call specific bits from this MR but it would give us the possibility to add more call specific features that don't map to notifications (like outgoing calls, joining calls together, emergency calling, etc).

Ongoing calls do map to notifications. You can't join calls cross-app, and if you're within an app I see no reason for the OS to even know about it. Not sure what you mean by emergency calling - how is that a portal?

agx: Thanks for the details. I'm aware it does `ongoing`. I'm after *outgoing*. This is needed for e.g. quick dialers (which people [already implemented](https://gitlab.com/Alaraajavamma/fastcalls)) or a separate UI on the car's dashboard when doing Android Auto. 

> I don't think CallKit lets your app make actual phone calls via the SIM card.

I don't think so either. What I mean is that we currently need to expose all these things to the sandbox if we want to initiate a call. That's why I think a call related portal (which would then also handle the outgoing / ongoing signalling in this portal would make sense.

cajus: Based on the **Inhibit** flag exported by KDE via DBus in _org.freedesktop.notifications_, it would be nice to be able to see in the current proposal whether DND is currently active or not.

data/org.freedesktop.portal.Notification.xml
```diff
@@ -106,21 +180,361 @@
...
+            It's a programmer error to specify ``transient`` at the same time.
+
+          * ``persistent``
+
+            Make the notification persistent in the notification tray.
+            The user can’t dismiss it using the usual close button or gesture.
+
+            Apps are only allowed to display persistent notifications
+            as long as they have a window. Once the last window of an app
+            is closed the persistent notification will be removed.
```
Mikenux, line 408: - "tray" and "persistent": Apps shouldn't have control over notification UX. If you don't agree, please give example cases for those.
- "transient": Also part of notification UX. However, it seems to be a content type to display, e.g., an action like paste has been done (here equivalent of in-app notification). I don't know what really to think about this, so examples are also welcome here.

jsparber:
this can be closed since https://github.com/flatpak/xdg-desktop-portal/pull/1298 landed. And we have a discussion for features that didn't land https://github.com/flatpak/xdg-desktop-portal/discussions/1495

PR #1306, opened 2024-03-17
Fix segfault when opening non-existant file

penn5:
When OpenURI was called with a file-descriptor to a file whose path was not available (whether due to sandboxing or race conditions) this caused an obscure crash, which is resolved by this commit.

This fixes https://retrace.fedoraproject.org/faf/problems/603686/


penn5:
Tests seem broken on `main`

penn5:
Rebased on main, hopefully that will fix tests

swick:
also LGTM

PR #1309, opened 2024-03-18
Allow tests to set the AppInfo to type HOST with a specific app id

swick:
This inverts the problem that https://github.com/flatpak/xdg-desktop-portal/pull/1279 and https://github.com/flatpak/xdg-desktop-portal/pull/1280 are trying to solve by passing in the app id that will be used for all client connections instead of querying for the app id that x-d-p found.

This is in particular useful because it removes the dependency on `xdp_get_app_info_from_pid` which I'm removing in https://github.com/flatpak/xdg-desktop-portal/pull/1281


whot (approved):
LGTM, thanks.

note that snakes don't have feet so they can't wear thongs ;)
(typo in commit message: `pythong`)

swick:
Fixed :)

GeorgesStavracas:
I originally thought that the test suite was borked, but actually I think the tests are up to something. Need to figure out why this PR is crashing some of the test runs.

whot:
@swick maybe try to rebase and force-push to retrigger the CI. I had issues in libwacom with ubuntu 22.04 and clang+asan for a while, that resolved itself last week and now it's all hunky and dory again. Could be the same issue here.

swick:
Thanks, that seems to have worked.

GeorgesStavracas (approved):

PR #1318, opened 2024-03-31
document-portal: Make possible to use on systems without splice support

arrowd:


document-portal/document-portal-fuse.c
```diff
@@ -2010,7 +2012,7 @@ xdp_fuse_read (fuse_req_t req,
   buf.buf[0].fd = file->fd;
   buf.buf[0].pos = off;
 
-  fuse_reply_data (req, &buf, FUSE_BUF_SPLICE_MOVE);
```
swick, line 2013: From the docs it sounds like `FUSE_BUF_SPLICE_MOVE` should be able to fall back to not using splice. Why is this needed?

document-portal/document-portal-fuse.c
```diff
@@ -3202,16 +3206,21 @@ static void
...
+  if (want_splice_support)
+    {
+      /* splice_read: use splice() to read from fuse pipe */
+      conn->want |= FUSE_CAP_SPLICE_READ;
+      /* splice_write: use splice() to write to fuse pipe */
+      conn->want |= FUSE_CAP_SPLICE_WRITE;
+      /* splice_move: move buffers from writing app to kernel during splice write */
+      conn->want |= FUSE_CAP_SPLICE_MOVE;
+      /* atomic_o_trunc: We handle O_TRUNC in create() */
+      conn->want |= FUSE_CAP_ATOMIC_O_TRUNC;
```
swick, line 3222: Why are you also getting rid of this?

swick (commented):
What problem is this trying to solve?

document-portal/document-portal-fuse.c
arrowd: Right, that was silly of me.

arrowd:
> What problem is this trying to solve?

A FUSE client should not assume that the server supports some features. The FreeBSD implementation of FUSE doesn't implement that splice stuff which makes the code fail at the early initialization step.

document-portal/document-portal-fuse.c
arrowd: From what I understand, the docs mean that a splice move **may** decay to copying in some circumstances, but it still requires that the splice feature is supported by the FUSE server.

document-portal/document-portal-fuse.c
```diff
@@ -3326,13 +3351,23 @@ xdp_fuse_thread (gpointer data)
       return NULL;
     }
 
+  fuse_opts = g_malloc0(sizeof (XdpFuseOptions));
+  fuse_opts->use_splice = 1;
```
swick, line 3355:
```suggestion
  fuse_opts->use_splice = TRUE;
```

document-portal/document-portal-fuse.c
```diff
@@ -3326,13 +3351,23 @@ xdp_fuse_thread (gpointer data)
...
+
   se = fuse_session_new (&args, &xdp_fuse_oper,
-                         sizeof (xdp_fuse_oper), NULL);
+                         sizeof (xdp_fuse_oper), fuse_opts);
+  if (se == NULL)
+    {
+      /* the second time do not request splice support */
+      fuse_opts->use_splice = 0;
```
swick, line 3362:
```suggestion
      fuse_opts->use_splice = FALSE;
```

document-portal/document-portal-fuse.c
```diff
@@ -3326,13 +3351,23 @@ xdp_fuse_thread (gpointer data)
...
+                             sizeof (xdp_fuse_oper), fuse_opts);
+    }
   if (se == NULL)
     {
       g_set_error (&thread_data->error, XDG_DESKTOP_PORTAL_ERROR,
-                   XDG_DESKTOP_PORTAL_ERROR_FAILED,
-                   "Can't create fuse session");
+                    XDG_DESKTOP_PORTAL_ERROR_FAILED,
+                    "Can't create fuse session");
       return NULL;
```
swick, line 3371: Need to free `fuse_opts`.

document-portal/document-portal-fuse.c
```diff
@@ -3202,27 +3216,37 @@ static void
...
+      conn->want |= FUSE_CAP_SPLICE_MOVE;
+    }
 }
 
 extern gboolean on_fuse_unmount (void *);
 
 static void
 xdp_fuse_destroy_cb (void *userdata)
 {
+  XdpFuseOptions* fuse_opts = (XdpFuseOptions*)userdata;
```
swick, line 3242: Don't cast void pointers.

document-portal/document-portal-fuse.c
```diff
@@ -3202,27 +3216,37 @@ static void
...
+
   g_debug ("DESTROY");
 
   /* Ensure we call this on the main thread */
   g_idle_add ((GSourceFunc) on_fuse_unmount, NULL);
+
+  g_free(fuse_opts);
```
swick, line 3249:
```suggestion
  g_free (fuse_opts);
```

document-portal/document-portal-fuse.c
```diff
@@ -3326,13 +3351,23 @@ xdp_fuse_thread (gpointer data)
       return NULL;
     }
 
+  fuse_opts = g_malloc0(sizeof (XdpFuseOptions));
```
swick, line 3354:
```suggestion
  fuse_opts = g_new0 (XdpFuseOptions, 1);
```

document-portal/document-portal-fuse.c
```diff
@@ -3202,27 +3216,37 @@ static void
 xdp_fuse_init_cb (void                  *userdata,
                   struct fuse_conn_info *conn)
 {
+  XdpFuseOptions* fuse_opts = (XdpFuseOptions*)userdata;
```
swick, line 3219:
```suggestion
  XdpFuseOptions* fuse_opts = userdata;
```

document-portal/document-portal-fuse.c
```diff
@@ -1994,6 +1994,10 @@ xdp_fuse_create (fuse_req_t             req,
   queue_invalidate_dentry (parent, filename);
 }
 
+typedef struct {
+  gboolean use_splice;
+} XdpFuseOptions;
```
swick, line 1999: Move to the other struct declarations.

smcv:
> The FreeBSD implementation of FUSE doesn't implement that splice stuff which makes the code fail at the early initialization step.

That seems like information that should be in the commit message and PR description, and/or in an issue report of the form "document-portal: Doesn't start on FreeBSD" that's closed by this PR.

swick:
Code LGTM but I can't judge the details of the FUSE changes.

arrowd:
Turns out that `xdp_fuse_init_cb ` is called not during `fuse_session_new` but later so this approach doesn't really work.

I'm thinking of a meson option that will set this compile time. Would that be acceptable?

arrowd:
I made Meson check for `splice` function to enable `WITH_SPLICE` CPP define. This is the same as what libfuse does: https://github.com/libfuse/libfuse/blob/8a9e2dc20c82259c9ded075f39383432c90d9786/meson.build#L67

arrowd:
Bump. Can we get this in?

document-portal/document-portal-fuse.c
```diff
@@ -3326,13 +3351,19 @@ xdp_fuse_thread (gpointer data)
...
+
   se = fuse_session_new (&args, &xdp_fuse_oper,
-                         sizeof (xdp_fuse_oper), NULL);
+                         sizeof (xdp_fuse_oper), fuse_opts);
   if (se == NULL)
     {
       g_set_error (&thread_data->error, XDG_DESKTOP_PORTAL_ERROR,
                    XDG_DESKTOP_PORTAL_ERROR_FAILED,
                    "Can't create fuse session");
+      g_free (fuse_opts);
```
GeorgesStavracas, line 3366: Nitpick: `g_clear_pointer (&fuse_opts, g_free);`

document-portal/document-portal-fuse.c
```diff
@@ -3202,27 +3216,37 @@ static void
...
 xdp_fuse_destroy_cb (void *userdata)
 {
+  XdpFuseOptions* fuse_opts = userdata;
+
   g_debug ("DESTROY");
 
   /* Ensure we call this on the main thread */
   g_idle_add ((GSourceFunc) on_fuse_unmount, NULL);
+
+  g_free (fuse_opts);
```
GeorgesStavracas, line 3249: Nitpick: `g_clear_pointer (&fuse_opts, g_free);`

GeorgesStavracas (changes requested):
Two small nitpicks, but generally this looks fine to me

document-portal/document-portal-fuse.c
```diff
@@ -260,6 +260,10 @@ typedef struct {
   char name[0];
 } XdpInvalidateData;
 
+typedef struct {
+  gboolean use_splice;
```
swick, line 264: Is there any need for this now? Could just replace opt->use_splice with #ifdef WITH_SPLICE ?

arrowd: This would spill `#ifdef`s across multiple places, which I find less readable.

And we might add more options in future, so I think it is better this way.

arrowd:
Can we get this in?

swick:
^ @GeorgesStavracas

GeorgesStavracas (approved):
Looks mostly good to me, I'll follow up with some cleanups

PR #1322, opened 2024-04-03
print: Add "supported_output_file_formats" option

GeorgesStavracas:
This allows applications tell portal backends which file formats they
support. In turn, portal backends can then limit the choices presented
to the user.


TingPing (approved):
Seems very reasonable to me.

matthiasclasen:
I assume this will need a corresponding gtk change to say that we only do pdf ?

matthiasclasen:
Looks fine to me too

GeorgesStavracas:
> I assume this will need a corresponding gtk change to say that we only do pdf ?

No. I actually thought so, and started looking at GtkPrintOperation - but realized soon enough that, actually, GtkPrintOperation is good and respect the output file format, it uses cairo-pdf for PDF, cairo-ps for PostScript, and plain cairo for SVG.

grulja:
I think this is fine from Qt POV, my understanding is that it can only print to PDF.

PR #1319, opened 2024-03-31
pipewire: Destroy registry object with remote

rmader:
This practically reintroduces commit fbe26534e14 which got reverted in ff8a8de6aedb5, which I only discovered after coming up with this solution on its own. The only difference is that
`discover_node_factory_sync()` gets removed as it doesn't really make much sense to keep around.

The commit message of fbe26534e14 (by msizanoen1
<qtmlabs@protonmail.com>)>
```
Currently, the PipeWire registry object is destroyed when
discover_node_factory_sync finishes its functionality. This causes
registry events to not be delivered to the global_removed_cb and
global_added_cb callbacks after discover_node_factory_sync, breaking
the functionality of the
org.freedesktop.portal.Camera.IsCameraPresent property in case of
camera device hot(un)plug.

This changes the discover_node_factory_sync function to store the
registry in the PipeWireRemote structure and destroy it in
pipewire_remote_destroy.
```

This holds true to this day and causes issues for the camera portal. Whatever crashes it caused at some point I can't reproduce - and Gnome-Shell, which has a very similar implementation for its camera monitor doesn't seem to suffer from crashes as well.

P.S.: the originally described issue:
> breaking the functionality of the org.freedesktop.portal.Camera.IsCameraPresent property in case of camera device hot(un)plug.

does not describe the main problem that many people are facing: if Pipewire initializes cameras a little too late, the portal won't see them.

---

cc: @GeorgesStavracas, @msizanoen1

PR #1325, opened 2024-04-04
camera: Stop forcing a remote roundtrip when creating PW FD

rmader:
For not yet understood reason it causes camera globals to be send twice to clients, resulting in duplicated camera entries. To be investigated before this should be merged.

This only happens since the commit mentioned below, thus: Fixes: 24d29bf (pipewire: Destroy registry object with remote)

---

See https://github.com/flatpak/xdg-desktop-portal/pull/1319

### DIFF
--- a/src/camera.c
+++ b/src/camera.c
@@ -166,8 +166,6 @@ open_pipewire_camera_remote (const char *app_id,
                                 G_N_ELEMENTS (permission_items),
                                 permission_items);
 
-  pipewire_remote_roundtrip (remote);
-
   return remote;
 }
 


rmader:
Note: this is not a real fix as it breaks permissions somehow, showing devices that shouldn't be there:

image

Also, for OBS things seem to work fine *without* this, no duplication happening.

rmader:
Closing, I'm pretty sure this has to be fixed on the PW side.

rmader:
No, haven't looked further nor filed an issue yet.

PR #1332, opened 2024-04-08
launch-context: Let GAppLaunchContext handle xdg-activation

swick:
g_app_info_launch* can activate an app via exec or dbus and it handles activation/startup notification correctly for both but requires the id/token from somewhere. Introduces a new XdpAppInfoLaunch where this token can be set via a method and use it in the OpenURI portal.


GeorgesStavracas (approved):
LGTM, thanks. Does the dynamic launcher portal also need to use this helper class?

swick:
Good point. Fixed that one as well.

PR #1335, opened 2024-04-11
launch-context: Return a copy of the token string

GeorgesStavracas:
GAppLaunchContext.get_startup_notify_id() expects a string copy, not a const string. Call g_strdup() to copy the string.


swick:
LGTM

Issue #1338, opened 2024-04-11
camera: When we fail to show the Access dialog permissions are set to `no`


A6GibKm:
### Operating System

Fedora Silverblue 40

### XDG Desktop Portal version

1.18

### Desktop Environment

GNOME

### Expected Behavior

If for some reason, it is not possible the display the dialog by org.freedesktop.impl.portal.Access.AccessDialog the permission should be left in an unset state rather than setting it to `no`

### Current Behavior

The permission is set to `no` immediately after the access dialog call fails.

### Steps to Reproduce

This is kinda racy, but snapshot 46.0 or 46.1 in certain setups could fail to display the permission dialog due to `Only the focused app is allowed to show a system access dialog` from GNOME Shell. My laptop can reproduce this reliably, my desktop cannot do it at all. This then can be verified with `flatpak permission-show`.

### Anything else we should know?

Failing to get focus on the shell in https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/accessDialog.js?ref_type=heads#L155 can cause an error, this goes through https://github.com/flatpak/xdg-desktop-portal/blob/main/src/camera.c#L125 and could be catch at that stage.

PR #1339, opened 2024-04-11
camera: Do not set permission=no on errors

A6GibKm:
The permission should be set to NO or YES only if the choice came from the user.

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/1338

### DIFF
--- a/src/camera.c
+++ b/src/camera.c
@@ -136,6 +136,10 @@ query_permission_sync (Request *request)
                                                          &error))
         {
           g_warning ("A backend call failed: %s", error->message);
+          /* We do not want to set the permission if there was an error and the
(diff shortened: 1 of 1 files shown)


A6GibKm:
I am not sure the if the `if` should be removed to also cover the `PERMISSION_ASK` case.

A6GibKm:
> I am not sure the if the `if` should be removed to also cover the `PERMISSION_ASK` case. From the commit:
> 
> > The permission should be set to NO or YES only if the choice came from
> > the user.
> 
> my guess is that the `if` should be removed.

Edited to always return early.

matthiasclasen:
Makes sense to me

PR #1342, opened 2024-04-13
device: Do not set permission=no on errors

A6GibKm:
The permission should be set to NO or YES only if the choice came from the user.

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/1338

(cherry picked from commit 8049324038cd9fd63bf6c13aca004f85f4bf0255)


A6GibKm:
cc @GeorgesStavracas backported :).

GeorgesStavracas (approved):
Looks good, but can you change the prefix to `device:`?

PR #1345, opened 2024-04-18
background: Improve validation of commandline option

GeorgesStavracas:
Check that the first commandline item doesn't start with whitespaces or
a hyphen.

Also sneakily plug a memory leak, g_variant_get_strv() is transfer-
container. Switch to g_autofree on the variable.

Mitigates: CVE-2024-32462

### DIFF
--- a/src/background.c
+++ b/src/background.c
@@ -854,7 +854,7 @@ validate_commandline (const char *key,
                       GError **error)
 {
   gsize length;
-  const char **strv = g_variant_get_strv (value, &length);
+  g_autofree const char **strv = g_variant_get_strv (value, &length);
 
   if (strv[0] == NULL)
     {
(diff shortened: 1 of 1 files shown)


src/background.c
```diff
@@ -854,7 +854,7 @@ validate_commandline (const char *key,
                       GError **error)
 {
   gsize length;
-  const char **strv = g_variant_get_strv (value, &length);
+  g_autofree const char **strv = g_variant_get_strv (value, &length);
```
ebassi, line 857: `g_autofree const` sounds like a copy-paste error.

GeorgesStavracas: This confused me to no end too, but it is not. What happens is that `g_variant_get_strv` is `transfer container`, it's an owned array of unowned strings. The function signature is `const char **` but it must be freed! I didn't know that until reading the documentation.

ebassi: The commit message says "use dup_strv()" but you didn't. Either the commit message needs to be fixed, or the commit.

ebassi (approved):

smcv:
> Related: CVE-2024-32462

I usually write "Mitigates: CVE-2024-32462" if one project is working around or mitigating a CVE in another project. (In this case I think it's Flatpak that has the vulnerability, because only Flatpak can completely fix it, but x-d-p can help to prevent it from being exploited.)

GeorgesStavracas:
Thanks @smcv, switched to "Mitigates" now

PR #1347, opened 2024-04-22
doc: Add ScreenCast source type meaning

tytan652:
The name "virtual" is too ambiguous by itself so a blurb is added for each type.

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -239,9 +239,9 @@
 
         A bitmask of available source types. Currently defined types are:
 
-        - ``1``: MONITOR
-        - ``2``: WINDOW
-        - ``4``: VIRTUAL
+        - ``1``: MONITOR: Share existing monitors
+        - ``2``: WINDOW: Share application windows
+        - ``4``: VIRTUAL: Extend with new virtual monitor
     -->
     <property name="AvailableSourceTypes" type="u" access="read"/>
     <!--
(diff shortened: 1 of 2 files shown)


swick:
Does `virtual` really mean virtual monitors here? What's the use case? @jadahl

jadahl:
> Does `virtual` really mean virtual monitors here? What's the use case? @jadahl

Yes it does. The use case to get a "virtual monitor" that extends your current session. E.g. Miracast as a separate output.

data/org.freedesktop.impl.portal.ScreenCast.xml
```diff
@@ -239,9 +239,9 @@
 
         A bitmask of available source types. Currently defined types are:
 
-        - ``1``: MONITOR
-        - ``2``: WINDOW
-        - ``4``: VIRTUAL
+        - ``1``: MONITOR: Monitors (non-virtual)
+        - ``2``: WINDOW: Application windows
+        - ``4``: VIRTUAL: Virtual monitors
```
jadahl, line 244: Perhaps we should be slightly more verbose:

```
        - ``1``: MONITOR: Share existing monitors
        - ``2``: WINDOW: Share application windows
        - ``4``: VIRTUAL: Extend with new virtual monitor
```

or something

Because `1` can still mean "virtual" monitor, just that something else created it.

tytan652: I applied the suggestion.

GeorgesStavracas (approved):

PR #1354, opened 2024-04-30
USB portal (cont.)

hfiguiere:
This supercede #1238 

Close #1238

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 20241024-1
+  IMAGE_TAG: 20241203-1
 
 on:
   workflow_call:
--- a/data/org.freedesktop.impl.portal.Usb.xml
+++ b/data/org.freedesktop.impl.portal.Usb.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2023 GNOME Foundation Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.Usb:
+      @short_description: USB portal backend interface
+
+      This portal lets applications register global shortcuts so they can
+      act regardless of the system state upon an input event.
+
+      This documentation describes version 1 of this interface.
+  -->
+  <interface name="org.freedesktop.impl.portal.Usb">
+
+    <!--
+        AcquireDevices:
+        @handle: Object path for the :ref:`org.freedesktop.impl.portal.Request` object representing this call
+        @parent_window: Identifier for the application window, see :ref:`Common Conventions <window-identifiers>`.
+        @app_id: App id of the application
+        @devices: Array of identifier, device information, and access options for each device
+        @options: Vardict with optional further information
+        @response: Numeric Request response
+        @results: Vardict with the results of the call
+
+        Selects the devices to open.
+
+        Each element of the @devices array contains the ID, information and
+        access options for each device the app wants to acquire.
+        The supported keys in the information vardict are the same as the
+        keys in the @devices vardict of
+        :ref:`org.freedesktop.portal.Usb.EnumerateDevices`.
+
+        The supported keys in the access options vardict are:
+
+        * ``writable`` (``b``)
+
+          Whether the device will be opened in read-write or read-only mode.
+          Default: False
+
+        There are no supported keys in the @options vardict.
+
+        The following results get returned via the @results vardict:
+
+        * ``devices`` (``a(sa{sv})``)
+
+          An array of identifier and access options which is a subset of
+          what was passed in via @devices.
+     -->
+    <method name="AcquireDevices">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="a(sa{sv}a{sv})" name="devices" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In4" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <property name="version" type="u" access="read"/>
+  </interface>
+</node>
--- a/data/org.freedesktop.portal.Usb.xml
+++ b/data/org.freedesktop.portal.Usb.xml
@@ -1,0 +1,243 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2020 Endless OS Foundation LLC
+               2023 GNOME Foundation Inc.
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+         Ryan Gonzalez <rymg19+github@gmail.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.Usb:
+      @short_description: Portal for USB device access
+
+      This interface lets sandboxed applications monitor and request
+      access to connected USB devices.
+
+      Applications should prefer specialized portals for specific
+      device types, such as the Camera portal for cameras.
+
+      This documentation describes version 1 of this interface.
+    -->
+  <interface name="org.freedesktop.portal.Usb">
+
+    <!--
+        CreateSession:
+        @options: Vardict with optional further information
+        @session_handle: Object path for the session.
+
+        Creates an USB monitoring session. This is only necessary to
+        receive device events, like device being plugged or unplugged.
+
+        Supported keys in the @options vardict include:
+
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the #org.freedesktop.portal.Session documentation for
+          more information about the session handle.
+    -->
+    <method name="CreateSession">
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
+    </method>
+
+    <!--
+        EnumerateDevices:
+        @options: Vardict with optional further information. There is currently
+        no option supported.
+
+        Enumerates all connected USB devices that this application has permission
+        to see.
+
+        The following results are returned in the @device vardict:
+
+        * ``id`` (``s``)
+
+          Each element of the @devices array contains the device ID, and a
+          device information vardict with the following keys:
+
+        * ``parent`` (``s``)
+
+          Device ID of the parent device.
+
+        * ``readable`` (``b``)
+
+          Whether the device can be opened for reading with
+          org.freedesktop.portal.Usb.AcquireDevices(). If not present, then
+          it should be assumed to be false.
+
+        * ``writable`` (``b``)
+
+          Whether the device can be opened for writing with
+          org.freedesktop.portal.Usb.AcquireDevices(). If not present, then
+          it should be assumed to be false.
+
+        * ``device-file`` (``s``)
+
+          A string path to the device node inside the /dev filesystem.
+
+        * ``properties`` (``a{sv}``)
+
+          A list of udev properties that this device has. These properties
+          are not parsed in any way by the portal, it is up to apps to parse
+          them.
+     -->
+    <method name="EnumerateDevices">
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a(sa{sv})" name="devices" direction="out"/>
+    </method>
+
+    <!--
+        AcquireDevices:
+        @parent_window: Identifier for the application window, see :ref:`Common Conventions <window-identifiers>`.
+        @devices: Array of device identifiers and access options
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+
+        Request to acquires (i.e. open) the given device nodes.  The
+        process of acquiring is finished by calling
+        `FinishAcquireDevices` after the request emitted a `Success`
+        response.
+
+        Each element of the @devices array contains the device ID, and an
+        access option vardict with the following keys:
+
+        * ``writable`` (``b``)
+
+          Whether the device will be opened in read-write or read-only mode.
+          Default: False
+
+        Supported keys in the @options vardict include:
+
+        * ``handle_token`` (``s``)
+
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation
+          for more information about the @handle.
+
+        The #org.freedesktop.portal.Request::Response signal is emitted without
+        any extra information.
+     -->
+    <method name="AcquireDevices">
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="a(sa{sv})" name="devices" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        FinishAcquireDevices:
+        @handle: The request object path for the acquisition to finish
+        @options: Vardict with optional further information
+        @results: Array of device ids, and the result of the access
+        @finished: Whether all device results were reported
+
+        Retrieves the file descriptors of the devices requested during
+        org.freedesktop.portal.Usb.AcquireDevices(). The file
+        descriptors remain usable until the underlying device is
+        removed, they are released with `ReleaseDevices`, the D-Bus
+        connections is closed, or the portal revokes the file
+        descriptor which can happen at any point. Devices which are
+        not needed anymore should be passed to `ReleaseDevices`.
+
+        If not all devices could be send, @finished will be false and
+        org.freedesktop.portal.Usb.FinishAcquireDevices must be called again
+        until @finished is true, before calling
+        org.freedesktop.portal.Usb.AcquireDevices again.
+
+        This method can only be called once for a given token, and only
+        after calling org.freedesktop.portal.Usb.AcquireDevices().
+
+        Each element of the @results array contains the device ID, and
+        a vardict with the following keys:
+
+        * ``success`` (``b``)
+
+          Whether the device access was successful or not.
+
+        * ``fd`` (``h``)
+
+          The file descriptor representing the device. The file
+          descriptor is meant to be passed on to the USB library.
+
+          Only present if this was a successful device access.
+
+        * ``error`` (``s``)
+
+          Error message describing why accessing the device was not
+          successful. Only present if this was an failed device access.
+
+        There are no supported keys in the @options vardict.
+     -->
+    <method name="FinishAcquireDevices">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a(sa{sv})" name="results" direction="out"/>
+      <arg type="b" name="finished" direction="out"/>
+    </method>
+
+    <!--
+        ReleaseDevices:
+        @devices: Array of device identifiers
+        @options: Vardict with optional further information
+
+        Releases previously acquired devices. The file descriptors of
+        thoses devices might become unusable as a result of this.
+
+        Each element of the @devices array contains the device ID of the device.
+
+        There are no supported keys in the @options vardict.
+     -->
+    <method name="ReleaseDevices">
+      <arg type="as" name="devices" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+    </method>
+
+    <!--
+        DeviceEvents:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @events: A list of events which occured.
+
+        The DeviceEvents signal is emitted when one or more USB devices have
+        been added, changed, or removed. This signal is only emitted for active
+        sessions created with org.freedesktop.portal.Usb.CreateSession().
+
+        Each element of the @events array is composed of the following fields:
+
+        * ``action`` (``s``)
+
+          Type of event that occurred. One of "add", "change", or "remove".
+
+        * ``id`` (``s``)
+
+          Device ID that the event occurred on.
+
+        * ``device`` (``a{sv}``)
+
+          Device properties attached to the ID. See
+          org.freedesktop.portal.Usb.EnumerateDevices() for a list of all
+          the properties that may be present in the vardict.
+      -->
+    <signal name="DeviceEvents">
+      <arg type="o" name="session_handle" direction="out"/>
+      <arg type="a(ssa{sv})" name="events" direction="out"/>
+    </signal>
+
+    <property name="version" type="u" access="read"/>
+  </interface>
+</node>
--- a/src/usb.c
+++ b/src/usb.c
@@ -1,0 +1,1567 @@
+/*
+ * Copyright © 2023-2024 GNOME Foundation Inc.
+ *             2020 Endless OS Foundation LLC
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+ *       Hubert Figuière <hub@figuiere.net>
+ *       Ryan Gonzalez <rymg19+github@gmail.com>
+ */
+
+#include "config.h"
+
+#include <ctype.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include <glib-unix.h>
+#include <glib/gi18n.h>
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+#include <gio/gdesktopappinfo.h>
+
+#include <gudev/gudev.h>
+
+#include "usb.h"
+#include "xdp-request.h"
+#include "xdp-permissions.h"
+#include "xdp-session.h"
+#include "xdp-dbus.h"
+#include "xdp-impl-dbus.h"
+#include "xdp-utils.h"
+#include "xdp-usb-query.h"
+
+#define PERMISSION_TABLE "usb"
+#define PERMISSION_ID "usb"
+#define MAX_DEVICES 8
+
+/* TODO:
+ *
+ * AccessDevices()
+ *  - Check if backend is returning appropriate device ids
+ *  - Check if backend is not increasing permissions
+ *  - Save allowed devices in the permission store
+ */
+
+struct _XdpUsb
+{
+  XdpDbusUsbSkeleton parent_instance;
+
+  GHashTable *ids_to_devices;
+  GHashTable *syspaths_to_ids;
+
+  GHashTable *sessions;
+  GHashTable *sender_infos;
+
+  GUdevClient *gudev_client;
+};
+
+#define XDP_TYPE_USB (xdp_usb_get_type ())
+G_DECLARE_FINAL_TYPE (XdpUsb, xdp_usb, XDP, USB, XdpDbusUsbSkeleton)
+
+static void xdp_usb_iface_init (XdpDbusUsbIface *iface);
+
+G_DEFINE_FINAL_TYPE_WITH_CODE (XdpUsb, xdp_usb, XDP_DBUS_TYPE_USB_SKELETON,
+                               G_IMPLEMENT_INTERFACE (XDP_DBUS_TYPE_USB, xdp_usb_iface_init));
+
+struct _XdpUsbSession
+{
+  XdpSession parent;
+
+  GHashTable *available_devices;
+};
+
+#define XDP_TYPE_USB_SESSION (xdp_usb_session_get_type ())
+G_DECLARE_FINAL_TYPE (XdpUsbSession,
+                      xdp_usb_session,
+                      XDP, USB_SESSION,
+                      XdpSession)
+
+G_DEFINE_TYPE (XdpUsbSession, xdp_usb_session, xdp_session_get_type ())
+
+typedef struct
+{
+  char *device_id;
+  gboolean writable;
+} UsbDeviceAcquireData;
+
+typedef struct _UsbOwnedDevice
+{
+  gatomicrefcount ref_count;
+
+  char *sender_name;
+  char *device_id;
+  int fd;
+} UsbOwnedDevice;
+
+typedef struct _UsbSenderInfo
+{
+  gatomicrefcount ref_count;
+
+  char *sender_name;
+  XdpAppInfo *app_info;
+
+  GHashTable *pending_devices; /* object_path → GPtrArray */
+
+  GHashTable *owned_devices; /* device id → UsbOwnedDevices */
+} UsbSenderInfo;
+
+static XdpDbusImplUsb *usb_impl;
+static XdpUsb *usb;
+
+static void usb_device_acquire_data_free (UsbDeviceAcquireData *acquire_data);
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (UsbDeviceAcquireData, usb_device_acquire_data_free)
+
+static void usb_owned_device_unref (UsbOwnedDevice *owned_device);
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (UsbOwnedDevice, usb_owned_device_unref)
+
+static void usb_sender_info_unref (UsbSenderInfo *sender_info);
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (UsbSenderInfo, usb_sender_info_unref)
+
+static gboolean
+is_gudev_device_suitable (GUdevDevice *device)
+{
+  const char *device_file = NULL;
+  const char *devtype = NULL;
+
+  g_assert (G_UDEV_IS_DEVICE (device));
+  g_return_val_if_fail (g_strcmp0 (g_udev_device_get_subsystem (device), "usb") == 0,
+                        FALSE);
+
+  device_file = g_udev_device_get_device_file (device);
+  if (!device_file)
+    return FALSE;
+
+  devtype = g_udev_device_get_property (device, "DEVTYPE");
+  if (!devtype || g_strcmp0 (devtype, "usb_device") != 0)
+    return FALSE;
+
+  return TRUE;
+}
+
+static char *
+unique_permission_id_for_device (GUdevDevice *device)
+{
+  g_autoptr(GString) permission_id = NULL;
+  const char *property;
+
+  permission_id = g_string_new ("usb:");
+
+  property = g_udev_device_get_property (device, "ID_VENDOR_ID");
+  if (property)
+    g_string_append (permission_id, property);
+
+  property = g_udev_device_get_property (device, "ID_MODEL_ID");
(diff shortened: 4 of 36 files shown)


hfiguiere:
After a painful rebasing.

hfiguiere:
Feature is linked to https://github.com/flatpak/flatpak/pull/5620 

I think this is ready for review.

src/xdg-desktop-portal.c
```diff
@@ -359,6 +363,15 @@ on_bus_acquired (GDBusConnection *connection,
...
     export_portal_implementation (connection,
                                   input_capture_create (connection, implementation->dbus_name));
+
+#ifdef HAVE_GUDEV
+  implementation = find_portal_implementation ("org.freedesktop.impl.portal.Usb");
+  if (implementation != NULL)
+    export_portal_implementation (connection,
+                                  usb_create (connection, implementation->dbus_name));
+  else
+    g_warning ("couldn't find implementation for USB");
```
swick, line 373: We're not warning in that case for any other portal.

src/usb.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#pragma once
+
+#include <gio/gio.h>
+
+GDBusInterfaceSkeleton * usb_create (GDBusConnection *connection,
+                                     const char      *dbus_name);
+
+void revoke_usb_devices_from_sender (const char *sender);
```
swick, line 30: Let's namespace this properly, so `usb_revoke_devices_from_sender`

It also doesn't revoke anything, does it? It just cleans up things when a client disconnects.

src/xdg-desktop-portal.c
```diff
@@ -184,6 +185,9 @@ export_portal_implementation (GDBusConnection *connection,
 static void
 peer_died_cb (const char *name)
 {
+#ifdef HAVE_GUDEV
+  revoke_usb_devices_from_sender (name);
```
swick, line 189: You can call `xdp_connection_track_name_owners` to connect to a callback all in `usb.c` without leaking this detail here.

src/xdp-utils.c
```diff
@@ -117,6 +121,198 @@ needs_quoting (const char *arg)
   return FALSE;
 }
 
+static void
+xdp_usb_rule_free (XdpUsbRule *rule)
+{
+  if (!rule)
```
swick, line 127: Same comment as on the flatpak MR

src/xdp-utils.c
```diff
@@ -117,6 +121,198 @@ needs_quoting (const char *arg)
...
+  g_assert (value != NULL);
+  g_assert (expected_length > 0 && expected_length <= 4);
+
+  len = strlen (value);
+  if (len != expected_length)
+    return FALSE;
+
+  n = strtol (value, &end, 16);
+
+  if (end - value != len)
```
swick, line 153: Not sure if this is UB or not, given that those are pointers you're doing arithmetic with.

src/xdp-utils.c
```diff
@@ -117,6 +121,198 @@ needs_quoting (const char *arg)
...
+  if (!parsed)
+    return NULL;
+
+  return g_steal_pointer (&usb_rule);
+}
+
+void
+xdp_usb_query_free (XdpUsbQuery *query)
+{
+  if (!query)
```
swick, line 277: Same here

src/xdp-utils.c
```diff
@@ -117,6 +121,198 @@ needs_quoting (const char *arg)
...
+    return;
+
+  g_clear_pointer (&query->rules, g_ptr_array_unref);
+  g_free (query);
+}
+
+XdpUsbQuery *
+xdp_usb_query_from_string (XdpUsbQueryType  query_type,
+                       const char      *string)
```
swick, line 286: indentation is off

src/xdp-utils.c
```diff
@@ -117,6 +121,198 @@ needs_quoting (const char *arg)
...
+      const char *rule = split[i];
+
+      usb_rule = xdp_usb_rule_from_string (rule);
+      if (!usb_rule)
+        return NULL;
+
+      g_ptr_array_add (usb_query->rules, g_steal_pointer (&usb_rule));
+    }
+
+  g_assert (usb_query->rules->len > 0);
```
swick, line 311: `g_return_val_if_fail (... NULL)`

swick (commented):
Checked the bits outside of `usb.c` so far.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+          more information about the session handle.
+    -->
+    <method name="CreateSession">
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
+    </method>
+
+    <!--
+        EnumerateDevices:
+        @options: Vardict with optional further information
```
swick, line 60: Should document that no options are currently supported. There are a few more of those cases in here.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+          org.freedesktop.portal.Usb.AcquireDevices(). If not present, then
+          it should be assumed to be false.
+
+        * ``device-file`` (``s``)
+
+          A string path to the device node inside the /dev filesystem.
+
+        * ``properties`` (``a{sv}``)
+
+          A list of udev properties that this device has. These properties
```
swick, line 93: I'm really worried about this. It essentially makes udev a stable API even though it really is not.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+        FinishAcquireDevices:
+        @options: Vardict with optional further information
+        @results: Array of device ids, and the result of the access
+        @finished: Whether all device results were reported
+
+        Retrieves the file descriptors of the devices requested during
+        org.freedesktop.portal.Usb.AcquireDevices().
+
+        This method can only be called once, and only after calling
+        org.freedesktop.portal.Usb.AcquireDevices().
```
swick, line 147: Should mention the `#org.freedesktop.portal.Request::Response` signal

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+
+        The DeviceEvents signal is emitted when one or more USB devices have
+        been added, changed, or removed. This signal is only emitted for active
+        sessions created with org.freedesktop.portal.Usb.CreateSession().
+
+        Each element of the @events array is composed of the following fields:
+
+        * ``action`` (``s``)
+
+          Type of event that occurred. One of "add", "change", or "remove".
```
swick, line 203: What happens when one has acquired a device and it got removed? Is the fd still usable? Only sometimes?

We don't have a mechanism for revoking the fd so a real device unplug will result in a remove with the fd becoming unusable but when the user wants to revoke a device, we can't do anything with the existing fd on the client and so it won't be.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+      <arg type="a(sa{sv})" name="results" direction="out"/>
+      <arg type="b" name="finished" direction="out"/>
+    </method>
+
+    <!--
+        ReleaseDevices:
+        @devices: Array of device identifiers
+        @options: Vardict with optional further information
+
+        Releases previously acquired devices.
```
swick, line 179: What does it mean to release a device? Is the fd still usable?

swick (commented):
Bunch of comments on the interface definition. Mostly about documentation but I'm a bit worried about what device removal means and how we should deal with udev properties.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  sender_info->sender_state = USB_SENDER_STATE_ACQUIRING_DEVICES;
+
+  xdp_dbus_impl_usb_call_acquire_devices (usb_impl,
+                                          request->id,
+                                          arg_parent_window,
+                                          xdp_app_info_get_id (request->app_info),
+                                          g_steal_pointer (&filtered_devices),
+                                          g_steal_pointer (&options),
+                                          NULL,
+                                          usb_access_devices_cb,
```
swick, line 1282:
```suggestion
                                          usb_devices_acquire_cb,
```
or `on_usb_devices_acquired`

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  if (!xdp_filter_options (arg_options,
+                           &options_builder,
+                           usb_access_devices_options,
+                           G_N_ELEMENTS (usb_access_devices_options),
+                           &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+  options = g_variant_builder_end (&options_builder);
```
swick, line 1248: options is an autoptr but `g_variant_builder_end` returns a floating reference which is not a proper reference and calling g_object_unref via the autoptr is an error.

I see this pattern everywhere in x-d-p though and I'm not sure if I'm missing something. /cc @GeorgesStavracas

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+
+  sender_info = usb_sender_info_from_request (self, request);
+  g_assert (sender_info != NULL);
+
+  if (sender_info->sender_state != USB_SENDER_STATE_DEFAULT)
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR,
+                                             XDG_DESKTOP_PORTAL_ERROR_FAILED,
+                                             "Cannot call AcquireDevices() with an unfinished "
```
swick, line 1258: Why not? The interface description doesn't mention this restriction at all.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  sender_info = usb_sender_info_from_request (usb, request);
+
+  g_assert (sender_info != NULL);
+  g_assert (sender_info->sender_state == USB_SENDER_STATE_ACQUIRING_DEVICES);
+
+  g_variant_builder_init (&results_builder, G_VARIANT_TYPE_VARDICT);
+
+  xdp_dbus_impl_usb_call_acquire_devices_finish (usb_impl, &response, &results, result, &error);
+
+  if (error)
```
swick, line 1014:
```suggestion
  if (!xdp_dbus_impl_usb_call_acquire_devices_finish (usb_impl, &response, &results, result, &error))
```

data/org.freedesktop.portal.Usb.xml
swick: I think the response signal should carry an id that should be passed to `FinishAcquireDevices`. That way there is no problem with calling `AcquireDevices` multiple times before calling `FinishAcquireDevices`.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+                                 access_data->device_id,
+                                 g_variant_dict_end (&dict));
+          continue;
+        }
+
+      /* Check we haven't already acquired the device */
+      owned_device = g_hash_table_lookup (sender_info->owned_devices, access_data->device_id);
+      if (!owned_device)
+        {
+          fd = open (device_file, access_data->writable ? O_RDWR : O_RDONLY);
```
swick, line 1401: Why would opening the file fail? Is udev already telling us that we should be able to open the device node (i.e. we have permission to do it) or is do we discover about bad permissions just here?

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+          g_variant_dict_insert (&dict, "error", "s", _("Not allowed"));
+          g_variant_builder_add (&results_builder, "(s@a{sv})",
+                                 access_data->device_id,
+                                 g_variant_dict_end (&dict));
+          continue;
+        }
+
+      /* Check we haven't already acquired the device */
+      owned_device = g_hash_table_lookup (sender_info->owned_devices, access_data->device_id);
+      if (!owned_device)
```
swick, line 1399: Why not first ensure that there is an owned device and then add it to the vardict and to the hashtable?

Then you could also move the code to actually open the device to the OwnedDevice

swick (commented):
Bunch of comments on the interface definition. Mostly about documentation but I'm a bit worried about what device removal means and how we should deal with udev properties.

src/xdp-utils.c
hfiguiere: https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Pointer-Arithmetic.html

> C defines pointer subtraction to be consistent with pointer-integer addition, so that (p3 - p1) + p1 equals p3, as in ordinary algebra.

we should be good.

src/xdp-utils.h
```diff
@@ -34,6 +35,55 @@
...
+} XdpUsbQueryType;
+
+typedef struct
+{
+  XdpUsbQueryType query_type;
+  GPtrArray *rules;
+} XdpUsbQuery;
+
+void xdp_usb_query_free (XdpUsbQuery *query);
+XdpUsbQuery *xdp_usb_query_from_string (XdpUsbQueryType  query_type,
```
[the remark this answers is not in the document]
hfiguiere: done

src/usb.h
hfiguiere: renamed for now.

src/xdp-utils.c
hfiguiere: no, we dereference query.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  char *device_id;
+  gboolean writable;
+} UsbDeviceAcquireData;
+
+static void
+usb_device_acquire_data_free (gpointer data)
+{
+  UsbDeviceAcquireData *access_data = (UsbDeviceAcquireData *) data;
+
+  if (!access_data)
```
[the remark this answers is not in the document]
hfiguiere: can't remove it: it is dereferenced

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (UsbDeviceAcquireData, usb_device_acquire_data_free)
+
+/* UsbOwnedDevice */
+
+typedef struct _UsbOwnedDevice
+{
+  gatomicrefcount ref_count;
+
+  char *sender_name;
```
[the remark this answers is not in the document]
hfiguiere: it's used in `usb_sender_info_release_device()`

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+
+  self->ids_to_devices = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_object_unref);
+  self->syspaths_to_ids = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  self->sessions = g_hash_table_new (g_direct_hash, g_direct_equal);
+  self->sender_infos = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, usb_sender_info_unref);
+
+  self->gudev_client = g_udev_client_new (subsystems);
+  g_signal_connect (self->gudev_client,
+                    "uevent",
+                    G_CALLBACK (on_gudev_client_uevent_cb),
```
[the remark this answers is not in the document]
hfiguiere: I think `_cb` has been the convention. Using that.

hfiguiere:
(I haven't addressed everything yet)

src/xdp-utils.c
swick: Don't pass in invalid pointers then. Functions can have pre-conditions.

src/usb.c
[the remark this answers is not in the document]
swick: We currently use both and that's why I'm asking.

src/usb.c
hfiguiere: many reason. It's IO nothing is guaranteed. The device could have disappear, could be unitialized. If it can go wrong it will go wrong.

src/xdp-app-info-flatpak.c
```diff
@@ -234,6 +234,56 @@ xdp_app_info_flatpak_validate_autostart (XdpAppInfo          *app_info,
   return TRUE;
 }
 
+static GPtrArray *
+xdp_app_info_flaptak_get_usb_queries (XdpAppInfo *app_info)
```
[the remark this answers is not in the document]
hfiguiere: implemented as lazy loaded.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+    {
+      g_autoptr(GVariant) device_variant = NULL;
+      GVariantBuilder devices_builder;
+      UsbSenderInfo *sender_info;
+      Session *session;
+
+      g_assert (G_UDEV_IS_DEVICE (device));
+      g_assert (g_strcmp0 (g_udev_device_get_subsystem (device), "usb") == 0);
+
+      session = (Session *) usb_session;
```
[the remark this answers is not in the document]
hfiguiere: this macro doesn't exist. And it's out of scope to fix in this PR.

src/xdp-utils.c
hfiguiere: when pre-conditions are NOT checked with any tool then they are just footguns. I know the tooling suck here, it's not a justification for dangerous practices. It should be safe with a NULL pointer.

src/xdp-app-info-flatpak.c
```diff
@@ -234,6 +234,56 @@ xdp_app_info_flatpak_validate_autostart (XdpAppInfo          *app_info,
...
+  usb_queries = g_ptr_array_new_with_free_func ((GDestroyNotify) xdp_usb_query_free);
+
+  if (XDP_IS_APP_INFO_FLATPAK (app_info))
+    {
+      g_auto(GStrv) enumerable_devices = NULL;
+      g_auto(GStrv) hidden_devices = NULL;
+
+      enumerable_devices = g_key_file_get_string_list (app_info_flatpak->flatpak_info,
+                                                       "USB Devices",
+                                                       "allowed-devices",
```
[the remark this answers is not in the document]
hfiguiere: fixed the literal.

src/usb.c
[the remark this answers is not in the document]
swick: You don't need a check for every possible programmer error and every time you dereference a pointer. In fact, your check is for exactly one out of all possible pointer values.

src/usb.c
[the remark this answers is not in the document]
swick: I wouldn't consider it out of scope to add this for Session and Request:

```c
G_GNUC_UNUSED static inline Session *SESSION (gpointer ptr)
{
  return G_TYPE_CHECK_INSTANCE_CAST (ptr, session_get_type (), Session);
}
```

src/usb.c
hfiguiere: I have added a sink in this specific location.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  XdpDbusUsbSkeletonClass parent_class;
+} UsbClass;
+
+static XdpDbusImplUsb *usb_impl;
+static Usb *usb;
+
+GType usb_get_type (void) G_GNUC_CONST;
+static void usb_iface_init (XdpDbusUsbIface *iface);
+
+G_DEFINE_TYPE_WITH_CODE (Usb, usb, XDP_DBUS_TYPE_USB_SKELETON,
```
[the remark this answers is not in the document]
hfiguiere: final and renamed and properly declared.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+                         G_IMPLEMENT_INTERFACE (XDP_DBUS_TYPE_USB, usb_iface_init));
+
+static gboolean
+hex_to_uint16 (const char *property,
```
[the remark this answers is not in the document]
hfiguiere: removed.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+    {
+      GUdevDevice *device = l->data;
+
+      g_assert (G_UDEV_IS_DEVICE (device));
+      g_assert (g_strcmp0 (g_udev_device_get_subsystem (device), "usb") == 0);
+
+      if (!is_gudev_device_suitable (device))
+        continue;
+
+      if (!create_unique_usb_id (self, device, NULL))
```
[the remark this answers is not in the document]
hfiguiere: renamed and addressed usage.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+  g_autoptr(GError) error = NULL;
+  GDBusConnection *connection;
+  GVariantBuilder options_builder;
+  UsbSession *usb_session;
+  Permission permission;
+  Session *session;
+  Call *call;
+  Usb *self;
+
+  self = (Usb *) object;
```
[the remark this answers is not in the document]
hfiguiere: `XDP_USB ()`

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+
+      g_assert (syspath != NULL);
+      id = g_strdup (g_hash_table_lookup (self->syspaths_to_ids, syspath));
+    }
+
+  g_assert (id != NULL);
+
+  /* Send event to all sessions that are allowed to handle it */
+  g_hash_table_iter_init (&iter, self->sessions);
+  while (g_hash_table_iter_next (&iter, (gpointer *) &usb_session, NULL))
```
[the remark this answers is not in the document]
hfiguiere: done

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+      count++;
+    }
+
+  property = g_udev_device_get_property (device, "ID_SERIAL");
+  if (property)
+    {
+      g_string_append_printf (permission_id, "%s%s", count > 0 ? "/" : "", property);
+      count++;
+    }
+
```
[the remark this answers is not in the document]
hfiguiere: actually I think we should change this to `usb:`. And I'll change the code to get worst case scenario `usb://` which should lift the confusion between model ID and serial. The thing I'm not sure it's possible to get an empty vendor ID and model ID. They are kinda required for USB.

src/usb.c
```diff
@@ -0,0 +1,1619 @@
...
+      if (parent_syspath != NULL)
+        {
+          parent_id = g_hash_table_lookup (self->syspaths_to_ids, parent_syspath);
+          if (parent_id != NULL)
+            g_variant_dict_insert (&device_variant_dict, "parent", "s", parent_id);
+        }
+    }
+
+  device_file = g_udev_device_get_device_file (device);
+  g_variant_dict_insert (&device_variant_dict, "device-file", "s", device_file);
```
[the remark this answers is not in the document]
hfiguiere: it's kinda of an identifier. It's not meant for the app to open it.

data/org.freedesktop.portal.Usb.xml
hfiguiere: It will close the file descriptor when we release the "owned device" object. I haven't checked how libusb, for example, reacts to this.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+
+        Each element of the @devices array contains the device ID of the device.
+
+        There are no supported keys in the @options vardict.
+     -->
+    <method name="ReleaseDevices">
```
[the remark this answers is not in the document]
hfiguiere: when the dbus connection is closed all the devices are removed (ie released)

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,222 @@
...
+      device types, such as the Camera portal for cameras.
+
+      This documentation describes version 1 of this interface.
+    -->
+  <interface name="org.freedesktop.portal.Usb">
+
+    <!--
+        CreateSession:
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
```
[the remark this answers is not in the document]
hfiguiere: updated. It's a session_handle

data/org.freedesktop.portal.Usb.xml
swick: Well, it closes the fd on the portal side which is good because we need to do that to not leak things, but it doesn't close the fd on the client side. In fact, you cannot close the fd on the client side from the portal side. The only thing we can do is revoke file descriptors but that currently only works for evdev and hidraw and not for usb. So as long as we don't get kernel support for that, clients will be able to continue to use the fd that they acquired even if the connection is closed. We should however say something in the interface description here which will allow us to revoke those USB devices if the connection gets closed, or if the user wants to revoke access manually, when we get USB revoke from the kernel.

src/usb.c
swick: Right, but that means we'll show USB devices to the portal clients, letting users go through the UI to approve access, just to then eventually notice that we actually can't make the device available. Would be much nicer if we didn't advertise those devices in the first place.

hfiguiere: there is not a lot we can do. We can call access on the device `is_gudev_device_suitable()` and return `FALSE` if the call fail. That will check for existance and read permission. At that leve we don't know if we want the write permissions.

hfiguiere: And we already check in `handle_acquire_devices()` as well.

Still a lot of race condition can occur it acquiring the device.

src/usb.c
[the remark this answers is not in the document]
swick: And why is it needed? The device file name depends on the udev configuration of the host and is not stable.

src/usb.c
swick: Yes, obviously there can always be cases where opening will fail because this is inherently racy, but why would we want to advertise devices which we cannot open because we have no read permission? The code right now figures out that it won't be able to open it, advertises it anyway and then fails when a client tries to acquire it.

hfiguiere: we no longer do. the change just hadn't been pushed yet.

src/xdp-app-info-host.c
```diff
@@ -73,6 +98,11 @@ xdp_app_info_host_class_init (XdpAppInfoHostClass *klass)
 static void
 xdp_app_info_host_init (XdpAppInfoHost *app_info_host)
 {
+  app_info_host->usb_queries = g_ptr_array_new_with_free_func ((GDestroyNotify) xdp_usb_query_free);
+  g_autoptr(XdpUsbQuery) query =
+    xdp_usb_query_from_string (XDP_USB_QUERY_TYPE_ENUMERABLE, "all");
+  if (query)
+    g_ptr_array_add (app_info_host->usb_queries, g_steal_pointer (&query));
```
swick, line 105:
```suggestion
  g_autoptr(XdpUsbQuery) query = NULL;

  app_info_host->usb_queries =
    g_ptr_array_new_with_free_func ((GDestroyNotify) xdp_usb_query_free);

  query = xdp_usb_query_from_string (XDP_USB_QUERY_TYPE_ENUMERABLE, "all");
  if (query)
    g_ptr_array_add (app_info_host->usb_queries, g_steal_pointer (&query));
```

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,228 @@
...
+
+        * ``id`` (``s``)
+
+          The portal-specific unique identifier of the device.
+
+        * ``parent`` (``s``)
+
+          Device ID of the parent device.
+
+        * ``readable`` (``b``)
```
swick, line 77: What use is a device which can not be read from? `AcquireDevices` also doesn't have a `readable` only a `writable`. Pretty sure this should just be removed.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,228 @@
...
+    </method>
+
+    <!--
+        FinishAcquireDevices:
+        @options: Vardict with optional further information
+        @results: Array of device ids, and the result of the access
+        @finished: Whether all device results were reported
+
+        Retrieves the file descriptors of the devices requested during
+        org.freedesktop.portal.Usb.AcquireDevices().
```
swick, line 146: And here maybe add this:

> Retrieves the file descriptors of the devices requested during org.freedesktop.portal.Usb.AcquireDevices(). The file descriptors remain usable until the underlying device is removed, they are released with `ReleaseDevices`, the dbus connections is closed, or the portal revokes the file descriptor which can happen at any point. File descriptors which are not needed anymore should be passed to `ReleaseDevices`.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -0,0 +1,228 @@
...
+      <arg type="a(sa{sv})" name="results" direction="out"/>
+      <arg type="b" name="finished" direction="out"/>
+    </method>
+
+    <!--
+        ReleaseDevices:
+        @devices: Array of device identifiers
+        @options: Vardict with optional further information
+
+        Releases (i.e. closes) previously acquired devices. The file
```
swick, line 183: This does not close the device. It only removes the record on the portal side and in the future hopefully revokes the device as well.

How about this?

> Releases (i.e. closes) previously acquired devices. The file descriptor returned by FinishAcquireDevices for each of the devices is closed.

> Releases previously acquired devices. The file descriptors of those devices might become unusable as a result of this.

data/org.freedesktop.portal.Usb.xml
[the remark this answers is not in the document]
swick: I think my suggestion on `FinishAcquireDevices` would resolve this thread.

data/org.freedesktop.portal.Usb.xml
swick: I think my suggestion on FinishAcquireDevices would resolve this thread as well.

src/usb.c
swick: The changes I proposed on AcquireDevices would resolve this thread.

src/usb.c
swick: Maybe I'm blind but I can't seem to find it? Are you sure you pushed the latest version now?

hfiguiere: https://github.com/flatpak/xdg-desktop-portal/pull/1354/files#diff-4ce7d34e66e5ca9b52f8e5666f0d4efe11e48278f34a7cdba6df390c14012a27R163

Which mean if it can't access it it won't list it. So open won't be attempted. In case of race condition, ie the device is no longer accessible between the list and this, then this will error normally.

swick: Ah, I missed that one. Perfect!

data/org.freedesktop.portal.Usb.xml
hfiguiere: Until very recently it would have listed the devices even if it couldn't access them. It is devices that can be enumerated after all.

swick:
Pushed a branch with a bunch of cleanups and changes I'd like to see https://github.com/swick/xdg-desktop-portal/commits/wip/usb-portal-fixes1/

But in general this LGTM. The biggest issue is the lack of tests IMO. Some basic python tests using umockdev would be great but I'd also like to see this merged rather sooner than later. Do you have time to do that work?

src/xdp-utils.c
hfiguiere: it's now g_return_if_fail()

swick:
I started working on python tests for this portal here: https://github.com/swick/xdg-desktop-portal/commits/wip/usb-portal-tests/

It's based on the PR which tests the location portal with the python tests because it already has a bunch of improvements.

Found a few bugs in the impl already and the branch contains fixes.

It's using umockdev to create the testbed and it succeeds in hiding the host devices but no success so far with making devices visible from the test itself.

swick:
Some prepwork for the testing is up in a PR: https://github.com/flatpak/xdg-desktop-portal/pull/1477

Also update the branch https://github.com/swick/xdg-desktop-portal/commits/wip/usb-portal-tests/ which now has passing tests with some mocked devices. It could use some more tests and it still contains fixes that this PR doesn't have.

src/usb.c
hfiguiere: this is done in the current code.

hfiguiere:
note that it the moment, without #1484, test might fail if you run them locally.

src/usb.c
hfiguiere: actually it caused a serious regression as it didn't detect removals, and a few other things.

hfiguiere:
Rebase and added the tests

hfiguiere:
The tests pass locally. Not sure yet what's happening.

swick:
I think I saw a race in the test removing a device. Maybe you can force it by adding and then removing a bunch of devices and repeating the test.

src/usb.c
```diff
@@ -1252,6 +1259,7 @@ handle_acquire_devices (XdpDbusUsb            *object,
 static gboolean
 handle_finish_acquire_devices (XdpDbusUsb            *object,
                                GDBusMethodInvocation *invocation,
+                               const char            *object_path,
```
swick, line 1246: Do we have a guarantee that this isn't the object path of another client?

hfiguiere:
> I think I saw a race in the test removing a device. Maybe you can force it by adding and then removing a bunch of devices and repeating the test.

It never receives the device event for removal. Still work locally. I can even receive it twice.

swick:
FWIW, with `podman run --rm -ti -v .:/portal:z ghcr.io/hfiguiere/xdg-desktop-portal:20241024-3 bash` I can reproduce the failure.

swick:
> It never receives the device event for removal. Still work locally. I can even receive it twice.

In the current PR the signal is connected twice so I guess that's what you're seeing.

hfiguiere:
there are a few commit for debug. Yes it's connected twice... that was to answer "does it get disconnected somehow?" and the answer is "not that problem". If can reproduce locally it will be easier. These last test commits will be stripped off.

swick:
I tried intercepting the mocked uevents but when I replace the socket (`mv "$sock" "$sock.original" && socat -t100 -v  UNIX-LISTEN:$sock,mode=777,reuseaddr,fork  UNIX-CONNECT:$sock.original`) I get nothing and dbusmock seems to crash the pytest harness.

src/usb.c
hfiguiere: I have added a check so that if it doesn't know about the object_path / token it will return an error.

hfiguiere:
The test passes as I work around skipping in CI the one that fails.

src/xdp-app-info-flatpak.c
```diff
@@ -364,6 +370,58 @@ xdp_app_info_flatpak_validate_autostart (XdpAppInfo          *app_info,
...
+      for (size_t i = 0; hidden_devices && hidden_devices[i] != NULL; i++)
+        {
+          g_autoptr(XdpUsbQuery) query =
+            xdp_usb_query_from_string (XDP_USB_QUERY_TYPE_HIDDEN, hidden_devices[i]);
+
+          if (query)
+            g_ptr_array_add (usb_queries, g_steal_pointer (&query));
+        }
+
+      g_message ("Found %d enumerable and %d hidden for app %s",
```
GeorgesStavracas, line 415: Either make this a `g_debug` or remove it entirely?

src/usb.c
```diff
@@ -0,0 +1,1571 @@
...
+  g_autoptr(GUnixFDList) fds = NULL;
+  GVariantBuilder results_builder;
+  Permission permission;
+  uint32_t accessed_devices;
+  gboolean finished;
+  GPtrArray *pending_devices = NULL;
+  Call *call;
+  XdpUsb *self;
+
+  g_print ("object path: %s\n", object_path);
```
GeorgesStavracas, line 1259: Ditto about `g_debug` or remove

src/xdp-usb-query.h
```diff
@@ -0,0 +1,89 @@
...
+void xdp_usb_query_free (XdpUsbQuery *query);
+XdpUsbQuery *xdp_usb_query_from_string (XdpUsbQueryType  query_type,
+					const char      *string);
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (XdpUsbQuery, xdp_usb_query_free);
+
+gboolean
+validate_hex_uint16 (const char *value,
+                     size_t      expected_length,
+                     uint16_t   *out_value);
```
GeorgesStavracas, line 89: Maybe move this to `xdp-utils.{c,h}`  and prefix with `xdp_` ?

GeorgesStavracas (changes requested):

src/usb.c
hfiguiere: done

src/xdp-app-info-flatpak.c
hfiguiere: done

hfiguiere: made g_debug

src/xdp-usb-query.h
hfiguiere: it's only used by usb

hfiguiere: but I prefixed it.

swick (commented):
There are a bunch of commits which fix up previous commits. I would prefer them to be squashed down.

hfiguiere:
> There are a bunch of commits which fix up previous commits. I would prefer them to be squashed down.

I will squash all the USB portal stuff into one and keep the test commits as they are.

Also I will shorten the commit message but copy that content into a separate file in `doc` that will be added to the commit. It makes more sense to have it there.

.github/workflows/container.yml
```diff
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 20241024-1
+  IMAGE_TAG: 20241024-3
```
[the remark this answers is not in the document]
hfiguiere: to what? (sorry I don't know github CI)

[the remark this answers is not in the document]
swick: `$CURRENT_DAY-1`. It's just a tag that's used to identify the image but it's better to keep it close to the date when it gets merged.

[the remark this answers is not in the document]
hfiguiere: ah ok, so like DNS serials. Will do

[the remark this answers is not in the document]
swick: Not entirely sure what went wrong but I think if you force push the CI failure should resolve itself.

swick:
LGTM

swick (commented):
Just some nit picking about the commits...

src/xdp-usb-query.c
```diff
@@ -0,0 +1,213 @@
+/*
```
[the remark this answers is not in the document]
hfiguiere: done

src/xdp-utils.c
```diff
@@ -162,7 +162,7 @@ xdp_connection_track_name_owners (GDBusConnection     *connection,
 gboolean
 xdp_filter_options (GVariant *options,
                     GVariantBuilder *filtered,
-                    XdpOptionKey *supported_options,
+                    const XdpOptionKey *supported_options,
```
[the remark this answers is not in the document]
hfiguiere: done

src/xdp-request.c
```diff
@@ -255,6 +255,12 @@ xdp_request_from_invocation (GDBusMethodInvocation *invocation)
   return g_object_get_data (G_OBJECT (invocation), "request");
 }
 
+const char *
+xdp_request_get_object_path (XdpRequest *request)
```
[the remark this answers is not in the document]
hfiguiere: done

GeorgesStavracas (approved):

data/org.freedesktop.impl.portal.Usb.xml
```diff
@@ -0,0 +1,81 @@
...
+
+ Author: Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.Usb:
+      @short_description: USB portal backend interface
+
+      This portal lets applications register global shortcuts so they can
```
CarlSchwan, line 26: The description seems wrong, this has afaik nothing to do with global shortcut

hfiguiere: oops. I missed that too. Will address this.

hfiguiere: See https://github.com/flatpak/xdg-desktop-portal/pull/1529

hfiguiere: not sure about the annotation. Will followup too.

hfiguiere: added the annotations to the same PR.

PR #1357, opened 2024-05-06
document-portal: Use portable formatting specifiers in debug strings

mardy:
This allows running the xdg-document-portal binary on 32 bit architectures.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -1428,7 +1428,7 @@ xdp_fuse_getattr (fuse_req_t req,
   double attr_valid_time = 0.0;/* Time in secs for attribute validation */
   const char *op = "GETATTR";
 
-  g_debug ("GETATTR %lx", ino);
+  g_debug ("GETATTR %" G_GINT64_MODIFIER "x", ino);
 
   if (xdp_domain_is_virtual_type (domain))
     {
@@ -1468,7 +1468,7 @@ xdp_fuse_setattr (fuse_req_t             req,
   int res;
   const char *op = "SETATTR";
 
-  g_debug ("SETATTR %lx %s", ino, to_set_string);
+  g_debug ("SETATTR %" G_GINT64_MODIFIER "x %s", ino, to_set_string);
 
   if (!xdp_document_inode_checks (op, req, inode,
                                   CHECK_CAN_WRITE | CHECK_IS_PHYSICAL))
@@ -1810,7 +1810,7 @@ xdp_fuse_lookup (fuse_req_t req,
   int open_flags = O_PATH|O_NOFOLLOW;
   const char *op = "LOOKUP";
 
-  g_debug ("LOOKUP %lx:%s", parent_ino, name);
+  g_debug ("LOOKUP %" G_GINT64_MODIFIER "x:%s", parent_ino, name);
 
   if (strcmp (name, ".") == 0 || strcmp (name, "..") == 0)
     {
@@ -1860,7 +1860,7 @@ xdp_fuse_lookup (fuse_req_t req,
       queue_invalidate_dentry (parent, name);
     }
 
-  g_debug ("LOOKUP %lx:%s => %lx", parent_ino, name, e.ino);
+  g_debug ("LOOKUP %" G_GINT64_MODIFIER "x:%s => %" G_GINT64_MODIFIER "x", parent_ino, name, e.ino);
 
   if (fuse_reply_entry (req, &e) == -ENOENT)
     abort_reply_entry (&e);
@@ -1895,7 +1895,7 @@ xdp_fuse_open (fuse_req_t req,
   XdpDocumentChecks checks;
   const char *op = "OPEN";
 
-  g_debug ("OPEN %lx %s", ino, open_flags_string);
+  g_debug ("OPEN %" G_GINT64_MODIFIER "x %s", ino, open_flags_string);
 
   checks = CHECK_IS_PHYSICAL;
   if (open_flags_has_write (open_flags))
@@ -1961,7 +1961,7 @@ xdp_fuse_create (fuse_req_t             req,
   XdpFile *file = NULL;
   const char *op = "CREATE";
 
-  g_debug ("CREATE %lx %s %s, 0%o", parent_ino, filename, open_flags_string, mode);
+  g_debug ("CREATE %" G_GINT64_MODIFIER "x %s %s, 0%o", parent_ino, filename, open_flags_string, mode);
 
   if (!xdp_document_inode_checks (op, req, parent,
                                   CHECK_CAN_WRITE |
@@ -2004,7 +2004,7 @@ xdp_fuse_read (fuse_req_t req,
   struct fuse_bufvec buf = FUSE_BUFVEC_INIT(size);
   XdpFile *file = (XdpFile *)fi->fh;
 
-  g_debug ("READ %lx size %ld off %ld", ino, size, off);
+  g_debug ("READ %" G_GINT64_MODIFIER "x size %" G_GSIZE_FORMAT " off %" G_GOFFSET_FORMAT, ino, size, (goffset)off);
 
   buf.buf[0].flags = FUSE_BUF_IS_FD | FUSE_BUF_FD_SEEK;
   buf.buf[0].fd = file->fd;
@@ -2026,7 +2026,7 @@ xdp_fuse_write (fuse_req_t             req,
   ssize_t res;
   const char *op = "WRITE";
 
-  g_debug ("WRITE %lx size %ld off %ld", ino, size, off);
+  g_debug ("WRITE %" G_GINT64_MODIFIER "x size %" G_GSIZE_FORMAT " off %" G_GOFFSET_FORMAT, ino, size, (goffset)off);
 
   res = pwrite (file->fd, buf, size, off);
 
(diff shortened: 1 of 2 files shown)


document-portal/document-portal-fuse.c
```diff
@@ -1428,7 +1428,7 @@ xdp_fuse_getattr (fuse_req_t req,
   double attr_valid_time = 0.0;/* Time in secs for attribute validation */
   const char *op = "GETATTR";
 
-  g_debug ("GETATTR %lx", ino);
+  g_debug ("GETATTR %" G_GINT64_MODIFIER "x", ino);
```
smcv, line 1431: Is `fuse_ino_t` guaranteed to be signed 64-bit on all architectures?

(If not, then this would be wrong, but `"%llx", (long long) ino` would work.)

document-portal/file-transfer.c
```diff
@@ -174,7 +174,7 @@ file_transfer_start (XdpAppInfo *app_info,
     g_free (transfer->key);
     key = g_random_int ();
     key = (key << 32) | g_random_int ();
-    transfer->key = g_strdup_printf ("%lu", key);
+    transfer->key = g_strdup_printf ("%" G_GUINT64_FORMAT, key);
```
smcv, line 177: This is really a somewhat separate issue (it's `guint64`, not `fuse_ino_t`) and is certainly correct.

document-portal/document-portal-fuse.c
mardy: Hi Simon, long time no see :-)

No, it's actually **unsigned** in all architectures: https://github.com/libfuse/libfuse/blob/master/include/fuse_lowlevel.h#L46

But there isn't a `G_GUINT64_MODIFIER` macro, and if my understanding is correct the reason is that this macro only contains the length modifier part for the sprintf operation, not the actual type, so it can be used in both cases (for 32 bit types, it expands to `ll`). I'm not sure why you think that `ino` requires a cast, since the `x` conversion specifier expects an unsigned integer, as `ino` is.

mardy: Ah, maybe your question was about the bit length rather than about the signedness. :-)

smcv: Ah, yes, I should have said: is it guaranteed to be unsigned 64-bit on all architectures? (And the important part is the width, not the signedness.)

smcv: Please mention in the commit message that this is a FUSE 3 change, rather than just saying "use portable specifiers": `%lx` was the correct, portable specifier for FUSE 2, but isn't correct any more for FUSE 3.

document-portal/file-transfer.c
smcv: Unlike the FUSE inode numbers, this was never portable, so please separate it into its own commit, perhaps something like "file-transfer: Use correct formatting specifier for guint64".

document-portal/document-portal-fuse.c
mardy: Thanks, commits updated.

document-portal/document-portal-fuse.c
```diff
@@ -2004,7 +2004,7 @@ xdp_fuse_read (fuse_req_t req,
   struct fuse_bufvec buf = FUSE_BUFVEC_INIT(size);
   XdpFile *file = (XdpFile *)fi->fh;
 
-  g_debug ("READ %lx size %ld off %ld", ino, size, off);
+  g_debug ("READ %" G_GINT64_MODIFIER "x size %" G_GSIZE_FORMAT " off %" G_GSIZE_FORMAT, ino, size, off);
```
smcv, line 2007: Sorry, I had missed that this commit is also correcting the format string for `size_t` and `off_t` in some places, not just `fuse_ino_t`.

For `size_t`, you are right that using `"%" G_GSIZE_FORMAT` or `"%zu"` is correct. I would personally suggest `"%zu"` for `size_t` in software that is not intended to be portable to non-Unix platforms, but `"%" G_GSIZE_FORMAT` is fine. (On Linux and all Unixes that I know of, `long` is the same size as `size_t` so `%ld` is equivalent to `%zd`, although strictly speaking wrong for `size_t` because of the different signedness.)

For `off_t`, unfortunately neither `"%ld"` nor `"%" G_GSIZE_FORMAT` is always correct. `off_t` is 64-bit on LP64 platforms, but can be either 32- or 64-bit on ILP32, depending on build options, toolchain, and which specific ILP32 platform (for example it's normally 32-bit on i386 but 64-bit on x32). Unfortunately there is no Standard C or POSIX modifier for an `off_t`.

I'm surprised that this compiled for you without warnings on a 32-bit platform, in fact, because I thought both Meson and FUSE enforce use of LFS (large file support), which makes `off_t` always 64-bit even on 32-bit platforms?

As far as I know, the only portable way to print an `off_t` would be with an explicit cast to a type that is at least as large as `off_t`: `("%lld", (long long) off)`, `("%jd", (intmax_t) off)`, or if you prefer the GLib type vocabulary, `("%" G_GOFFSET_FORMAT, (goffset) off)`.

document-portal/document-portal-fuse.c
```diff
@@ -2026,7 +2026,7 @@ xdp_fuse_write (fuse_req_t             req,
   ssize_t res;
   const char *op = "WRITE";
 
-  g_debug ("WRITE %lx size %ld off %ld", ino, size, off);
+  g_debug ("WRITE %" G_GINT64_MODIFIER "x size %" G_GSIZE_FORMAT " off %" G_GSIZE_FORMAT, ino, size, off);
```
smcv, line 2029: Same notes on `off_t` here.

document-portal/document-portal-fuse.c
```diff
@@ -2048,7 +2048,7 @@ xdp_fuse_write_buf (fuse_req_t             req,
   ssize_t res;
   const char *op = "WRITEBUF";
 
-  g_debug ("WRITEBUF %lx off %ld", ino, off);
+  g_debug ("WRITEBUF %" G_GINT64_MODIFIER "x off %" G_GSIZE_FORMAT, ino, off);
```
smcv, line 2051: `off_t` again here

document-portal/document-portal-fuse.c
```diff
@@ -2380,7 +2380,7 @@ xdp_fuse_readdir (fuse_req_t req,
   size_t rem;
   const char *op = "READDIR";
 
-  g_debug ("READDIR %lx %ld %ld", ino, size, off);
+  g_debug ("READDIR %" G_GINT64_MODIFIER "x %" G_GSIZE_FORMAT " %" G_GSIZE_FORMAT, ino, size, off);
```
smcv, line 2383: and here

document-portal/document-portal-fuse.c
mardy: Indeed, the changes for `size_t` and `off_t` were not in our original patch, I just added them when preparing the branch for this upstream, and I tested them on x86-46 only. I'll try your suggestion (although quite verbose, using the GLib type seems cleaner to me) on 32 bits too.

document-portal/document-portal-fuse.c
```diff
@@ -2141,7 +2141,7 @@ forget_one (fuse_ino_t ino,
 {
   g_autoptr(XdpInode) inode = xdp_inode_from_ino (ino);
 
-  g_debug ("FORGET %lx %ld", ino, nlookup);
+  g_debug ("FORGET %" G_GINT64_MODIFIER "x %" G_GSIZE_FORMAT, ino, nlookup);
```
mardy, line 2144: `nlookup` here is a long double, so `%ld` was actually correct. I'm reverting this part.

mardy: Done.

document-portal/document-portal-fuse.c
mardy: Done.

document-portal/document-portal-fuse.c
mardy: Done.

document-portal/document-portal-fuse.c
mardy: Done.

On arm32 I don't get any warnings on g_debug invocations, so this seems to be fine. I do get quite a few of these:

```
../document-portal/document-portal-fuse.c: In function 'xdp_fuse_setattr':
../document-portal/document-portal-fuse.c:1481:23: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
 1481 |       XdpFile *file = (XdpFile *)fi->fh;
      |                       ^
```
but this is harmless, since `fh` is declared as `uint64_t` and our pointers are 32 bits. I guess that an intermediate cast to `uintptr_t` would fix these warnings, if one really wanted to.

document-portal/document-portal-fuse.c
smcv: I think you mean `unsigned long` (an integer), not `long double` (a floating-point number)?

Strictly speaking it ought to be `%lu` for an `unsigned long`, but `%ld` is close enough.

smcv (approved):

document-portal/document-portal-fuse.c
smcv: This would be wrong on CHERI or other architectures with very large pointers, so strictly speaking this ought to have a `G_STATIC_ASSERT (sizeof (fi->fh) >= sizeof (void *))` or something, and then use `(XdpFile *) (uintptr_t) fi->fh` to silence the warning. But that seems more appropriate for a separate PR.

mardy: Makes sense.

PR #1358, opened 2024-05-10
portal-impl: Use fallback portals for Settings as well

liskin:
Since 1.18.2, we have a hardcoded fallback to xdg-desktop-portal-gtk as a last resort if no other portal is defined by a configuration file. This fallback mechanism, however, isn't implemented for the org.freedesktop.impl.portal.Settings interface.

The consequence is that everything seems to work just fine, but user preferences like color scheme are ignored in applications that use the portal API to retrieve them (not just flatpak apps, also libadwaita apps). That comes as a surprise to users; and since the portal stuff otherwise works fine, it's not entirely straightforward to figure out that configuring portal.conf would fix it.

I believe that if [`find_portal_implementation`](https://github.com/flatpak/xdg-desktop-portal/blob/ef1e96b51be6313f524279cd77c8261d1e5186ba/src/portal-impl.c#L598) implements fallbacks, so should [`find_all_portal_implementations`](https://github.com/flatpak/xdg-desktop-portal/blob/ef1e96b51be6313f524279cd77c8261d1e5186ba/src/portal-impl.c#L670), for consistency and to avoid surprises.

The fallback logic this commit implements closely resembles [the logic in `find_portal_implementation`](https://github.com/flatpak/xdg-desktop-portal/blob/ef1e96b51be6313f524279cd77c8261d1e5186ba/src/portal-impl.c#L621-L664): we don't look for fallbacks if there is a portal configured in portals.conf (or if portals.conf explicitly says there should be none), and we only use the x-d-p-gtk as a last resort if we haven't found any impl(s) using the legacy UseIn key.

(The above should be a self-contained description but there's some additional thoughts at https://github.com/flatpak/xdg-desktop-portal/pull/1199#issuecomment-2093902804)

Fixes: d18c563b6a34 ("portal-impl: Hard-code x-d-p-gtk as a last-resort fallback")
Related: https://github.com/flatpak/xdg-desktop-portal/issues/1102

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -653,8 +653,6 @@ PortalImplementation *
 find_portal_implementation (const char *interface)
 {
   const char **desktops;
-  GList *l;
-  int i;
 
   if (portal_interface_prefers_none (interface))
     return NULL;
@@ -677,9 +675,9 @@ find_portal_implementation (const char *interface)
   desktops = get_current_lowercase_desktops ();
 
   /* Fallback to the old UseIn key */
-  for (i = 0; desktops[i] != NULL; i++)
+  for (size_t i = 0; desktops[i] != NULL; i++)
     {
-      for (l = implementations; l != NULL; l = l->next)
+      for (GList *l = implementations; l != NULL; l = l->next)
         {
           PortalImplementation *impl = l->data;
 
@@ -703,7 +701,7 @@ find_portal_implementation (const char *interface)
    * try to fall back to x-d-p-gtk, which has historically been the portal
    * UI backend used by desktop environments with no backend of their own.
    * If it isn't installed, that is not an error: we just don't use it. */
-  for (l = implementations; l != NULL; l = l->next)
+  for (GList *l = implementations; l != NULL; l = l->next)
     {
       PortalImplementation *impl = l->data;
 
@@ -727,8 +725,6 @@ find_all_portal_implementations (const char *interface)
 {
   const char **desktops;
   GPtrArray *impls;
-  GList *l;
-  int i;
 
   impls = g_ptr_array_new ();
 
@@ -756,9 +752,9 @@ find_all_portal_implementations (const char *interface)
   desktops = get_current_lowercase_desktops ();
 
   /* Fallback to the old UseIn key */
-  for (i = 0; desktops[i] != NULL; i++)
+  for (size_t i = 0; desktops[i] != NULL; i++)
     {
-      for (l = implementations; l != NULL; l = l->next)
+      for (GList *l = implementations; l != NULL; l = l->next)
         {
           PortalImplementation *impl = l->data;
 
@@ -785,7 +781,7 @@ find_all_portal_implementations (const char *interface)
    * try to fall back to x-d-p-gtk, which has historically been the portal
    * UI backend used by desktop environments with no backend of their own.
    * If it isn't installed, that is not an error: we just don't use it. */
-  for (l = implementations; l != NULL; l = l->next)
+  for (GList *l = implementations; l != NULL; l = l->next)
     {
       PortalImplementation *impl = l->data;
 
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -725,7 +725,10 @@ find_portal_implementation (const char *interface)
 GPtrArray *
 find_all_portal_implementations (const char *interface)
 {
+  const char **desktops;
   GPtrArray *impls;
+  GList *l;
+  int i;
 
   impls = g_ptr_array_new ();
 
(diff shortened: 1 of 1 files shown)


tommie:
I like this. I just spent an hour or two figuring out why GNOME Software Center refused to go into dark mode under XFCE: https://tommie.github.io/a/2024/08/xfce-gnome-dark-mode

Turns out it was because Settings requires the `portals.conf` (Xubuntu 24.04; XDP 1.14.4) and libadwaita seems to not fall back on gsettings once it's tried XDP.

However, I'm curious why the UseIn is deprecated. Is there a replacement? Assuming one machine supports multiple desktop environments for users to choose from, it seems like a global configuration isn't expressive enough. The workaround is a user-specific configuration. Forcing users to create a config file feels like a regression.

smcv:
> However, I'm curious why the UseIn is deprecated. Is there a replacement? Assuming one machine supports multiple desktop environments for users to choose from, it seems like a global configuration isn't expressive enough.

Agreed, and that's why there's a per-desktop-environment configuration (which supersedes `UseIn`), documented in <https://flatpak.github.io/xdg-desktop-portal/docs/portals.conf.html>. XFCE provides an `xfce-portals.conf` since https://gitlab.xfce.org/xfce/xfce4-session/-/issues/181; if that hasn't made it into Xubuntu, please report that to your distro vendor.

tommie:
> xfce-portals.conf

Oh, sorry, I missed the part where it said that the prefix is matched against the desktop name. So that's the replacement.

Thanks for the link!

tommie:
Apologies for the tangential noise on this PR. Just to close this off:

Ubuntu 24.04 is unlucky to have xfce4-session [4.18.3](https://packages.ubuntu.com/noble/xfce4-session), while the new config file is in [4.18.4](https://gitlab.xfce.org/xfce/xfce4-session/-/tags/xfce4-session-4.18.4). 4.18.4 is in Debian testing.

I also found your https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1050802.

So I filed https://bugs.launchpad.net/ubuntu/+source/xfce4-session/+bug/2076999.

src/portal-impl.c
```diff
@@ -691,5 +693,56 @@ find_all_portal_implementations (const char *interface)
         }
     }
 
+  if (impls->len)
```
swick, line 696:
```suggestion
  if (impls->len > 0)
```
Just to be more clear.

src/portal-impl.c
```diff
@@ -691,5 +693,56 @@ find_all_portal_implementations (const char *interface)
...
+  desktops = get_current_lowercase_desktops ();
+
+  /* Fallback to the old UseIn key */
+  for (i = 0; desktops[i] != NULL; i++)
+    {
+     for (l = implementations; l != NULL; l = l->next)
+        {
+          PortalImplementation *impl = l->data;
+
+          if (!g_strv_contains ((const char **)impl->interfaces, interface))
```
swick, line 708:
```suggestion
          if (!portal_impl_supports_iface (impl, interface))
```

src/portal-impl.c
```diff
@@ -691,5 +693,56 @@ find_all_portal_implementations (const char *interface)
...
+              g_warning ("Choosing %s.portal for %s via the deprecated UseIn key",
+                         impl->source, interface);
+              warn_please_use_portals_conf ();
+              g_debug ("Using %s.portal for %s in %s (fallback)", impl->source, interface, desktops[i]);
+              g_ptr_array_add (impls, impl);
+            }
+        }
+    }
+
+  if (impls->len)
```
swick, line 722:
```suggestion
  if (impls->len > 0)
```

src/portal-impl.c
```diff
@@ -691,5 +693,56 @@ find_all_portal_implementations (const char *interface)
...
+   * UI backend used by desktop environments with no backend of their own.
+   * If it isn't installed, that is not an error: we just don't use it. */
+  for (l = implementations; l != NULL; l = l->next)
+    {
+      PortalImplementation *impl = l->data;
+
+      if (!g_str_equal (impl->dbus_name, "org.freedesktop.impl.portal.desktop.gtk"))
+        continue;
+
+      if (!g_strv_contains ((const char **)impl->interfaces, interface))
```
swick, line 738:
```suggestion
      if (!portal_impl_supports_iface (impl, interface))
```

swick:
The change seems fine to me but I'd like @ebassi opinion on it.

liskin:
Rebased on top on today's main and applied the suggestions.

src/portal-impl.c
```diff
@@ -741,5 +744,56 @@ find_all_portal_implementations (const char *interface)
         }
     }
 
+  if (impls->len > 0)
+    return impls;
+
+  desktops = get_current_lowercase_desktops ();
+
+  /* Fallback to the old UseIn key */
+  for (i = 0; desktops[i] != NULL; i++)
```
GeorgesStavracas, line 753: These days I've been preferring the following syntax:

```c
for (size_t i = 0; desktops[i] != NULL; i++)
```

src/portal-impl.c
```diff
@@ -741,5 +744,56 @@ find_all_portal_implementations (const char *interface)
...
+    {
+     for (l = implementations; l != NULL; l = l->next)
```
GeorgesStavracas, line 755: Ditto:

```c
for (GList *l = implementations; l != NULL; l = l->next)
```

src/portal-impl.c
```diff
@@ -741,5 +744,56 @@ find_all_portal_implementations (const char *interface)
...
+  if (impls->len > 0)
+    return impls;
+
+  /* As a last resort, if nothing was selected for this desktop by
+   * ${desktop}-portals.conf or portals.conf, and no portal volunteered
+   * itself as suitable for this desktop via the legacy UseIn mechanism,
+   * try to fall back to x-d-p-gtk, which has historically been the portal
+   * UI backend used by desktop environments with no backend of their own.
+   * If it isn't installed, that is not an error: we just don't use it. */
+  for (l = implementations; l != NULL; l = l->next)
```
GeorgesStavracas, line 782: Ditto:

```c
for (GList *l = implementations; l != NULL; l = l->next)
```

src/portal-impl.c
```diff
@@ -719,7 +719,10 @@ find_portal_implementation (const char *interface)
 GPtrArray *
 find_all_portal_implementations (const char *interface)
 {
+  const char **desktops;
   GPtrArray *impls;
+  GList *l;
+  int i;
```
GeorgesStavracas, line 725: If you apply the previous suggestions, you'll be able to remove these variables from here

GeorgesStavracas (changes requested):
Superficial comments only, but I also would like to hear what @ebassi thinks of it

src/portal-impl.c
liskin: I did apply them—in a separate commit, as I updated `find_portal_implementation` to match as well.

ebassi (approved):

swick:
LGTM and ebassi approves

GeorgesStavracas (approved):

PR #1364, opened 2024-05-27
document-portal: Implement GetHostPaths

### LINKED ISSUE #475: Real directory name not accessible from file opened via document portal

grulja:
This method allows apps to get path as exists on the host filesystem for documents exported through the document portal.

This method takes a list of document IDs as a string array and returns a dictionary mapping document IDs to the paths in the host filesystem. It is expected an app making this request to have access to given list of documents.

Based on initial work made by @JakobDev

Fixes #475

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -1416,6 +1416,101 @@ portal_list (GDBusMethodInvocation *invocation,
   return TRUE;
 }
 
+const char *
+get_host_path_internal (GDBusMethodInvocation *invocation,
+                        XdpAppInfo            *app_info,
+                        const char            *id,
+                        GError                **error)
+{
+  g_autoptr(PermissionDbEntry) entry = NULL;
+
+  XDP_AUTOLOCK (db);
+
+  entry = permission_db_lookup (db, id);
+
+  if (!entry)
+    {
+      if (error != NULL && *error == NULL)
+        {
+          g_set_error (error,
+                      XDG_DESKTOP_PORTAL_ERROR,
+                      XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                      "Invalid ID passed (%s)", id);
+        }
+
+      return NULL;
+    }
+
+  if (!xdp_app_info_is_host (app_info))
+    {
+      g_autofree const char **apps = NULL;
+      const char *app_id = NULL;
+      gboolean app_found = FALSE;
+
+      app_id = xdp_app_info_get_id (app_info);
+
+      apps = permission_db_entry_list_apps (entry);
+      for (size_t i = 0; apps[i] != NULL; i++)
+        {
+          if (g_strcmp0 (app_id, apps[i]) == 0)
+            {
+              app_found = TRUE;
+              break;
+            }
+        }
+
+      if (!app_found)
+        {
+          if (error != NULL && *error == NULL)
+            {
+              g_set_error (error,
+                           XDG_DESKTOP_PORTAL_ERROR,
+                           XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
+                           "Not enough permissions");
+            }
+
+          return NULL;
+        }
+    }
+
+  return document_entry_get_path (entry);
+}
+
+static gboolean
+portal_get_host_paths (GDBusMethodInvocation *invocation,
+                       GVariant              *parameters,
+                       XdpAppInfo            *app_info)
+{
+  g_autofree const char **id_list = NULL;
+  GVariantBuilder builder;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("(a{say})"));
+  g_variant_builder_open (&builder, G_VARIANT_TYPE ("a{say}"));
+
+  for (size_t i = 0; id_list[i] != NULL; i++)
+    {
+      g_autoptr(GError) error = NULL;
+      g_autofree const char *path = NULL;
+
+      path = get_host_path_internal (invocation, app_info, id_list[i], &error);
+      if (path == NULL)
+        {
+          g_warning ("Failed to get host path for %s: %s", id_list[i], error->message);
+          continue;
+        }
+
+      g_variant_builder_add (&builder, "{s@ay}", id_list[i], g_variant_new_bytestring (path));
+    }
+
+  g_variant_builder_close (&builder);
+
+  g_dbus_method_invocation_return_value (invocation, g_variant_builder_end (&builder));
+
+  return TRUE;
+}
+
 static void
 peer_died_cb (const char *name)
 {
@@ -1432,7 +1527,7 @@ on_bus_acquired (GDBusConnection *connection,
 
   dbus_api = xdp_dbus_documents_skeleton_new ();
 
-  xdp_dbus_documents_set_version (XDP_DBUS_DOCUMENTS (dbus_api), 4);
+  xdp_dbus_documents_set_version (XDP_DBUS_DOCUMENTS (dbus_api), 5);
 
   g_signal_connect_swapped (dbus_api, "handle-get-mount-point", G_CALLBACK (handle_get_mount_point), NULL);
   g_signal_connect_swapped (dbus_api, "handle-add", G_CALLBACK (handle_method), portal_add);
@@ -1445,6 +1540,7 @@ on_bus_acquired (GDBusConnection *connection,
   g_signal_connect_swapped (dbus_api, "handle-lookup", G_CALLBACK (handle_method), portal_lookup);
   g_signal_connect_swapped (dbus_api, "handle-info", G_CALLBACK (handle_method), portal_info);
   g_signal_connect_swapped (dbus_api, "handle-list", G_CALLBACK (handle_method), portal_list);
+  g_signal_connect_swapped (dbus_api, "handle-get-host-paths", G_CALLBACK (handle_method), portal_get_host_paths);
 
   file_transfer = file_transfer_create ();
   g_dbus_interface_skeleton_set_flags (file_transfer,
--- a/tests/test-doc-portal.c
+++ b/tests/test-doc-portal.c
@@ -688,6 +688,57 @@ test_add_named (void)
 }
 
 static void
+test_get_host_paths (void)
+{
+  g_autofree char *doc_id = NULL;
+  g_autofree char *expected_real_path = NULL;
+  g_autofree char *real_path = NULL;
+  const char *basename = "host-path";
+  GVariant *path = NULL;
+  g_autoptr(GVariant) reply = NULL;
+  g_autoptr(GVariant) result = NULL;
+  g_autoptr(GVariantIter) iter = NULL;
+  GVariantBuilder builder;
+  GError *error = NULL;
+  const gchar* key = NULL;
+
+  if (!check_fuse_or_skip_test ())
+    return;
+
+  doc_id = export_new_file (basename, "content", FALSE);
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("(as)"));
+  g_variant_builder_open (&builder, G_VARIANT_TYPE ("as"));
+  g_variant_builder_add (&builder, "s", doc_id);
+  g_variant_builder_close (&builder);
+
+  reply = g_dbus_connection_call_sync (session_bus,
+                                       "org.freedesktop.portal.Documents",
+                                       "/org/freedesktop/portal/documents",
+                                       "org.freedesktop.portal.Documents",
+                                       "GetHostPaths", g_variant_builder_end (&builder),
+                                       G_VARIANT_TYPE ("(a{say})"),
+                                       0, -1,
+                                       NULL,
+                                       &error);
+
+  g_assert_no_error (error);
+  result = g_variant_get_child_value (reply, 0);
+
+  g_assert (g_variant_is_of_type (result, G_VARIANT_TYPE ("a{say}")));
+
+  expected_real_path = g_build_filename (outdir, basename, NULL);
+  iter = g_variant_iter_new (result);
+  while (g_variant_iter_loop (iter, "{&s@ay}", &key, &path))
+    {
+      g_assert_cmpstr (key, ==, doc_id);
+      g_assert_cmpstr (g_variant_get_bytestring (path), ==, expected_real_path);
+      g_variant_unref (path);
+      path = NULL;
+    }
+}
+
+static void
 global_setup (void)
 {
   gboolean inited;
@@ -853,7 +904,7 @@ test_version (void)
   if (!check_fuse_or_skip_test ())
     return;
 
-  g_assert_cmpint (xdp_dbus_documents_get_version (documents), ==, 4);
+  g_assert_cmpint (xdp_dbus_documents_get_version (documents), ==, 5);
 }
 
 int
@@ -871,6 +922,7 @@ main (int argc, char **argv)
   g_test_add_func ("/db/recursive_doc", test_recursive_doc);
   g_test_add_func ("/db/create_docs", test_create_docs);
   g_test_add_func ("/db/add_named", test_add_named);
+  g_test_add_func ("/db/get_host_paths", test_get_host_paths);
 
   global_setup ();
 
--- a/data/org.freedesktop.portal.Documents.xml
+++ b/data/org.freedesktop.portal.Documents.xml
@@ -48,7 +48,7 @@
       bus name org.freedesktop.portal.Documents and the object path
       /org/freedesktop/portal/documents.
 
-      This documentation describes version 4 of this interface.
+      This documentation describes version 5 of this interface.
   -->
   <interface name='org.freedesktop.portal.Documents'>
     <property name="version" type="u" access="read"/>
@@ -272,6 +272,22 @@
     <method name="List">
       <arg type='s' name='app_id' direction='in'/>
       <arg type='a{say}' name='docs' direction='out'/>
+    </method>
+
+    <!--
+        GetHostPaths:
+        @doc_ids: the list of IDs of the files in the document store
+        @paths: a dictionary mapping document IDs to the paths in the host filesystem
+
+        Gets the host filesystem paths for document store entries.
+
+        This call is available inside the sandbox, if the application has the ``read`` permission for the documents.
+
+        This method was added in version 5 of this interface.
+    -->
+    <method name="GetHostPaths">
+      <arg type='as' name='doc_ids' direction='in'/>
+      <arg type='a{say}' name='paths' direction='out'/>
     </method>
   </interface>
 </node>


document-portal/document-portal.c
```diff
@@ -1416,6 +1416,91 @@ portal_list (GDBusMethodInvocation *invocation,
...
+{
+  g_autofree const char **id_list = NULL;
+  const char *path;
+  int i;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  GVariantBuilder builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{say}"));
+
+  for (i = 0; id_list[i]; i++)
```
ebassi, line 1487:
```suggestion
  for (size_t i = 0; id_list[i] != NULL; i++)
```

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,91 @@ portal_list (GDBusMethodInvocation *invocation,
...
+    {
+      path = get_host_path_internal(invocation, app_info, id_list[i]);
+      if (path == NULL)
+        {
+          g_variant_builder_clear(&builder);
+          return FALSE;
+        }
```
ebassi, line 1494:
```suggestion
      path = get_host_path_internal (invocation, app_info, id_list[i]);
      if (path == NULL)
        {
          g_variant_builder_clear (&builder);
          return FALSE;
        }
```

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,91 @@ portal_list (GDBusMethodInvocation *invocation,
...
+
+      g_variant_builder_add (&builder, "{s@ay}", id_list[i], g_variant_new_bytestring(path));
+    }
+
+  g_dbus_method_invocation_return_value (invocation, g_variant_new ("(@a{say})", g_variant_builder_end(&builder)));
```
ebassi, line 1499:
```suggestion
      g_variant_builder_add (&builder, "{s@ay}", id_list[i], g_variant_new_bytestring (path));
    }

  g_dbus_method_invocation_return_value (invocation, g_variant_new ("(@a{say})", g_variant_builder_end (&builder)));
```

ebassi:
You also need to update the documentation of the Document portal interface, since it still mentions version 4.

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+             XdpAppInfo *app_info)
+{
+  g_autofree const char **id_list = NULL;
+  const char *path;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  GVariantBuilder builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{say}"));
+
+  for (size_t i = 0; id_list[i] != NULL; i++)
```
swick, line 1486: declarations at the beginning of the function

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+    {
+      path = get_host_path_internal (invocation, app_info, id_list[i]);
+      if (path == NULL)
+        {
+          g_variant_builder_clear (&builder);
+          return FALSE;
+        }
```
swick, line 1493: Why not continue to build the rest of the dict and return that?

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+
+      g_variant_builder_add (&builder, "{s@ay}", id_list[i], g_variant_new_bytestring (path));
+    }
+
+  g_dbus_method_invocation_return_value (invocation, g_variant_new ("(@a{say})", g_variant_builder_end (&builder)));
```
swick, line 1498: should be possible to get that type from the builder directly.

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
   return TRUE;
 }
 
+const char *
+get_host_path_internal (GDBusMethodInvocation *invocation,
+             XdpAppInfo *app_info,
+             const char *id)
```
swick, line 1422: indentation

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+
+  entry = permission_db_lookup (db, id);
+
+  if (!entry)
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                             g_strdup_printf("Invalid ID passed (%s)", id));
+      return NULL;
+    }
```
swick, line 1436: maybe generate an error from the internal function and propagate it to dbus in  `portal_get_host_paths`?

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+
+  if (!xdp_app_info_is_host (app_info))
+    {
+      g_autofree const char **apps = NULL;
+      const char *app_id = NULL;
+      gboolean app_found = FALSE;
+      int i;
+
+      app_id = xdp_app_info_get_id(app_info);
```
swick, line 1445:
```suggestion
      app_id = xdp_app_info_get_id (app_info);
```

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+          return NULL;
+        }
+    }
+
+  g_autoptr (GVariant) data = permission_db_entry_get_data (entry);
+  const char *path;
+
+  g_variant_get (data, "(^&ayttu)", &path, NULL, NULL, NULL);
+
+  return path;
```
swick, line 1471: This seems wrong to me: (https://docs.gtk.org/glib/gvariant-format-strings.html#convenience-conversions)

> ^&ay 		equivalent to g_variant_get_bytestring()

and for `g_variant_get_bytestring`:

> The return value remains valid as long as value exists.

I.e, path becomes invalid at the end of the function because data will be freed.

(also declarations should be at the start of the block)

```suggestion
  data = permission_db_entry_get_data (entry);

  g_variant_get (data, "(^ayttu)", &path, NULL, NULL, NULL);

  return path;
```

but this needs to be freed in `portal_get_host_paths` then.

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+  g_autofree const char **id_list = NULL;
+  const char *path;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  GVariantBuilder builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{say}"));
+
+  for (size_t i = 0; id_list[i] != NULL; i++)
+    {
+      path = get_host_path_internal (invocation, app_info, id_list[i]);
```
swick, line 1488:
```suggestion
      g_autofree char *path = NULL;

      path = get_host_path_internal (invocation, app_info, id_list[i]);
```

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,111 @@ portal_list (GDBusMethodInvocation *invocation,
...
+                       XdpAppInfo *app_info)
+{
+  g_autofree const char **id_list = NULL;
+  GVariantBuilder builder;
+  GVariantBuilder map_builder;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("(a{say})"));
+  g_variant_builder_init (&map_builder, G_VARIANT_TYPE ("a{say}"));
```
swick, line 1498: You should be able to get away with the single builder of type `(a{say})`.

tests/test-doc-portal.c
```diff
@@ -687,6 +687,60 @@ test_add_named (void)
...
+    return;
+
+  doc_id = export_new_file (basename, "content", FALSE);
+
+  GVariantBuilder builder;
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("(as)"));
+  GVariantBuilder arrayBuilder;
+  g_variant_builder_init (&arrayBuilder, G_VARIANT_TYPE ("as"));
+  g_variant_builder_add (&arrayBuilder, "s", doc_id);
+  g_variant_builder_add_value (&builder, g_variant_builder_end (&arrayBuilder));
```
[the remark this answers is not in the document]
grulja: Thanks! This is way better option I didn't know about. I'm still learning with GVariant.

grulja:
> Thanks for bringing my work to en end. glib is not easy to use.
> 
> I have one small suggestion however: I used `y` instead of `s` as type for the path, as I was told that a path is in most cases UTF-8 but not always. I think it's better to make sure that the returned path is always UTF-8 and return `s` instead of `y`, to prevent errors, as most people will treat the bytes as an UTF-8 encoded string.

It's exactly like that. It's a `map<doc_id, path>` where `doc_id` is `s` and `path` is `ay`.

ebassi:
> I think it's better to make sure that the returned path is always UTF-8 and return `s` instead of `y`, to prevent errors, as most people will treat the bytes as an UTF-8 encoded string.

No. "Most people" will take that path and feed it to some platform-specific API, like `open()` or whatever other file-related function they have. All of those functions take a natively encoded path, which may or may not be UTF-8. Converting natively encoded paths to UTF-8 is potentially lossy, because file systems can use encodings for alphabets like Cyrillic, or CJK languages.

The only place where you want UTF-8 is displaying the path into some UI; for that you will need to convert the path to UTF-8, using functions like [`g_filename_to_utf8()`](https://docs.gtk.org/glib/func.filename_to_utf8.html).

Stop trying to enforce UTF-8 into Unix paths: that's not how Unix paths work.

grulja:
> > Thanks for bringing my work to en end. glib is not easy to use.
> > I have one small suggestion however: I used `y` instead of `s` as type for the path, as I was told that a path is in most cases UTF-8 but not always. I think it's better to make sure that the returned path is always UTF-8 and return `s` instead of `y`, to prevent errors, as most people will treat the bytes as an UTF-8 encoded string.
> 
> It's exactly like that. It's a `map<doc_id, path>` where `doc_id` is `s` and `path` is `ay`.

I should have read carefully the rest of your comment as I misunderstood so ignore what I wrote above.

I think using `ay` is consistent with the rest of the API. For example:
```
    <method name="Info">
      <arg type='s' name='doc_id' direction='in'/>
      <arg type='ay' name='path' direction='out'/>
      <arg type='a{sas}' name='apps' direction='out'/>
    </method>
```

swick:
But it shouldn't.

tests/test-doc-portal.c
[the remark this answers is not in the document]
swick: I only recently properly learned about GVariants as well and it can be really confusing :)

swick:
Paths are not UTF-8, not when you get the path from syscalls, not when you get them from clipboard, not anywhere. Apps already deal with that fact and are able to display them. Let's not make this simple thing hard for no reason.

grulja:
Hi, can we finally make a decision on this change if it's good code-wise?

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,109 @@ portal_list (GDBusMethodInvocation *invocation,
...
+  entry = permission_db_lookup (db, id);
+
+  if (!entry)
+    {
+      if (error != NULL && *error == NULL)
+        {
+          g_set_error (error,
+                      XDG_DESKTOP_PORTAL_ERROR,
+                      XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                      "%s", g_strdup_printf ("Invalid ID passed (%s)", id));
```
swick, line 1440:
```suggestion
                      "Invalid ID passed (%s)", id);
```
Shouldn't this work?

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,90 @@ portal_list (GDBusMethodInvocation *invocation,
...
+
+  g_autoptr (GVariant) data = permission_db_entry_get_data (entry);
+  const char *path;
+
+  g_variant_get (data, "(^&ayttu)", &path, NULL, NULL, NULL);
+
+  return path;
+}
+
+static gboolean
```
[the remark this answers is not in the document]
swick: Let's resolve this. Should be fine either way and can be a problem for another day.

swick:
Code LGTM but this needs a maintainer to take a look as well. @GeorgesStavracas, @ebassi, anyone else?

mcatanzaro:
A couple +1s from me:

 * +1 for "paths are not UTF-8." Other strings are guaranteed to be UTF-8 because that makes things easier for developers, but paths are consistently "filesystem encoding" everywhere else in our stack. Violating that rule here would be confusing and would make things harder for developers and not easier. Nobody wants to deal with converting to filesystem encoding before opening a file, and developers who work only with English won't even notice their mistake.
 * +1 for allowing applications to get host paths. This leaks information from the host filesystem, but it's a limited and pragmatic leak. If the user doesn't want the application to see the path to the host file, then don't open the file or copy it to a different location first.

What I'm not sure about is whether we should add this GetHostPaths, or just change the document portal to `mkdir -p` and then mount documents at the same location they exist on the host, because as soon as GetHostPaths exists, there's no longer much value to mounting documents under /run, unless I'm missing something? Is there value to putting files under /run other than to hide the host path?

mcatanzaro:
Answering my own question: it'd be really confusing if the file is in the same place in the sandbox as it is on the host system. Imagine moving the file in the sandbox from one location to another, but it stays in the same place on the host. That's just asking for bugs. The question was not smart.

smcv:
> or just change the document portal to mkdir -p and then mount documents at the same location they exist on the host

In addition to what @mcatanzaro said, this might be possible some of the time, but in general is not going to be possible all of the time - for example Flatpak often mounts locations from the host read-only, and that is irreversible while the sandbox is running (the only process that ever had the necessary capabilities to be able to change mount options has already given them up by the time the app talks to portals).

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,109 @@ portal_list (GDBusMethodInvocation *invocation,
...
+                           "Not enough permissions");
+            }
+
+          return NULL;
+        }
+    }
+
+  data = permission_db_entry_get_data (entry);
+
+  g_variant_get (data, "(^ayttu)", &path, NULL, NULL, NULL);
```
swick, line 1481: This is `document_entry_get_path`

swick:
Did some xattr experimentation for other reasons and noticed that putting the host path there is pretty easy.

https://github.com/swick/xdg-desktop-portal/commit/d55f0370b14258ef76aa4b0f183f699e350031a2

```
$ flatpak run --command=getfattr --file-forwarding  org.gnome.clocks -d -m '' -- @@ $HOME/Downloads/test.txt @@
getfattr: Removing leading '/' from absolute path names
# file: run/user/1000/doc/e5080d2/test.txt
security.selinux="system_u:object_r:fusefs_t:s0"
user.xdp-document-portal.host-path="/var/home/swick/Downloads/test.txt"
```

grulja:
> Did some xattr experimentation for other reasons and noticed that putting the host path there is pretty easy.
> 
> [swick@d55f037](https://github.com/swick/xdg-desktop-portal/commit/d55f0370b14258ef76aa4b0f183f699e350031a2)
> 
> ```
> $ flatpak run --command=getfattr --file-forwarding  org.gnome.clocks -d -m '' -- @@ $HOME/Downloads/test.txt @@
> getfattr: Removing leading '/' from absolute path names
> # file: run/user/1000/doc/e5080d2/test.txt
> security.selinux="system_u:object_r:fusefs_t:s0"
> user.xdp-document-portal.host-path="/var/home/swick/Downloads/test.txt"
> ```

We can have this as an additional way to get the host path, but I think this (or any other one) DBus API will be easier to discover and use for most of the clients. I know there is an API for this in GIO, but for example Qt doesn't seem to have nice API for this.

swick:
Yes, I'm still 100% in favor of this portal and would like to see it merged.

tests/test-doc-portal.c
```diff
@@ -687,6 +687,55 @@ test_add_named (void)
...
+  result = g_variant_get_child_value (reply, 0);
+
+  g_assert (g_variant_is_of_type (result, G_VARIANT_TYPE ("a{say}")));
+
+  expected_real_path = g_build_filename (outdir, basename, NULL);
+  iter = g_variant_iter_new (result);
+  while (g_variant_iter_loop (iter, "{&s@ay}", &key, &path)) {
+    g_assert_cmpstr (key, ==, doc_id);
+    g_assert_cmpstr (g_variant_get_bytestring (path), ==, expected_real_path);
+    return;
```
GeorgesStavracas, line 735: I think the `return` here defeats the purpose of this loop?

tests/test-doc-portal.c
```diff
@@ -687,6 +687,55 @@ test_add_named (void)
...
+  }
```
GeorgesStavracas, line 736: Please follow the project's code style:

```c
  while (g_variant_iter_loop (iter, "{&s@ay}", &key, &path))
    {
      ...
    }
```

tests/test-doc-portal.c
```diff
@@ -687,6 +687,55 @@ test_add_named (void)
...
+                                       &error);
+
+  g_assert_no_error (error);
+  result = g_variant_get_child_value (reply, 0);
+
+  g_assert (g_variant_is_of_type (result, G_VARIANT_TYPE ("a{say}")));
+
+  expected_real_path = g_build_filename (outdir, basename, NULL);
+  iter = g_variant_iter_new (result);
+  while (g_variant_iter_loop (iter, "{&s@ay}", &key, &path)) {
```
GeorgesStavracas, line 732: If this loop iters more than once, the key and path variables will leak

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,103 @@ portal_list (GDBusMethodInvocation *invocation,
...
+    {
+      g_autofree const char **apps = NULL;
+      const char *app_id = NULL;
+      gboolean app_found = FALSE;
+      int i;
+
+      app_id = xdp_app_info_get_id (app_info);
+
+      apps = permission_db_entry_list_apps (entry);
+      for (i = 0; apps[i] != NULL; i++)
```
GeorgesStavracas, line 1454: `for (size_t i = 0; apps[i] != NULL; i++)`

GeorgesStavracas (changes requested):

document-portal/document-portal.c
grulja: Ok, changing this for the third time. I had `size_t i` as suggested by @ebassi, then @swick wanted the declaration of `i` be at the beginning with the rest of declarations and now you want me to change it again :).

tests/test-doc-portal.c
grulja: I think `key` is not going to be leaked according to the docs:
```
For example, if you are only looping over simple integer and string types, g_variant_iter_next() is definitely preferred. For string types, use the ‘&’ prefix to avoid allocating any memory at all (and thereby avoiding the need to free anything as well).
```

document-portal/document-portal.c
GeorgesStavracas: Yeah sorry for this back and forth. I've been preferring this syntax because it keeps `i` scoped to the loop.

tests/test-doc-portal.c
GeorgesStavracas: So I think you might have the opposite problem, no? Declaring key and path as `g_auto*` variables will end up double freeing these values I think?

document-portal/document-portal.c
```diff
@@ -1416,6 +1416,102 @@ portal_list (GDBusMethodInvocation *invocation,
...
+  g_autofree const char **id_list = NULL;
+  GVariantBuilder builder;
+  size_t i = 0;
+
+  g_variant_get (parameters, "(^a&s)", &id_list);
+
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("(a{say})"));
+  g_variant_builder_open (&builder, G_VARIANT_TYPE ("a{say}"));
+
+  for (i = 0; id_list[i] != NULL; i++)
```
GeorgesStavracas, line 1493: `for (size_t i = 0; id_list[i] != NULL; i++)` here too

tests/test-doc-portal.c
grulja: `key` is just `const gchar *`

GeorgesStavracas (changes requested):

tests/test-doc-portal.c
swick:
```
const gchar* path = NULL;
...
while (g_variant_iter_loop (iter, "{&s^&ay}", &key, &path))
  {
```
That should prevent leaking the value as well.

document-portal/document-portal.c
swick: We're not declaring variables in loops in e.g. mutter and this project mostly doesn't either (there are a few exceptions). Do we want to start doing that? Should we get a coding style document somewhere?

GeorgesStavracas: I think it's good if we start doing that. I don't think we have x-d-p coding style documented, but that would be good to have as well.

GeorgesStavracas:
Thanks!

GeorgesStavracas:
@swick I'd like to have the xattr one too, mind proposing a PR for that?

grulja:
Thanks @GeorgesStavracas for merging this.

Also kudos to @JakobDev for initial version of this change!!

tests/test-doc-portal.c
grulja: Shouldn't it be `"{&s&@ay}"`?

swick: No, `@` gives you a GVariant:

> Used as a prefix for a GVariant type string (not a prefix for a format string, so as is a valid format string but @^as is not).

on the other hand:

> ^&ay 		equivalent to [g_variant_get_bytestring()](https://docs.gtk.org/glib/method.Variant.get_bytestring.html)

and from g_variant_get_bytestring:

> The data is owned by the instance.

grulja: I understand now. Thank you for explaining.

PR #1366, opened 2024-05-30
XdpAppInfo refactoring

swick:
This turns XdpAppInfo into a GObject, moves it into its own file, pulls engine specific things that are scattered around the codebase into XdpAppInfo, and gets rid of a bunch of API.

I expect another round of refactoring where I'm actually changing the XdpAppInfo implementation. The goal is still Containers1.


swick:
This is ready. @smcv @GeorgesStavracas

GeorgesStavracas (changes requested):
A couple of small comments, but generally looks fine.

swick:
Addressed the feedback.

GeorgesStavracas (approved):

PR #1372, opened 2024-06-07
document/fuse: Provide the document host path as xattr

swick:
Complementing https://github.com/flatpak/xdg-desktop-portal/pull/1364, this provides the host path as an xattr via the document portal fuse fs.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -64,6 +64,18 @@
 #define ENODATA ENOATTR
 #endif
 
+#define XDP_XATTR_HOST_PATH "user.document-portal.host-path"
+/* man listxattr:
+ *   The list is the set of (null-terminated) names, one after the other.
+ *
+ * Example:
+ *   user.name1\0system.name1\0user.name2\0
+ *
+ * We can statically define the string by doing this:
+ *   XDP_XATTR_MY_ATTR "\0" XDP_XATTR_ANOTHER_ATTR
+ */
+#define XDP_XATTR_ATTRIBUTE_NAME_LIST XDP_XATTR_HOST_PATH
+
 /* Inode ownership model
  *
  * The document portal exposes something as a filesystem that it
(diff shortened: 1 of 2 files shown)


document-portal/document-portal-fuse.c
```diff
@@ -64,6 +64,18 @@
 #define ENODATA ENOATTR
 #endif
 
+#define XDP_XATTR_HOST_PATH "user.xdp-document-portal.host-path"
```
GeorgesStavracas, line 67: Let's bikeshed on this a little :)

I think `user.` and `.host-path` are fine. I think the middle part should be either `document-portal`, or `portal.document` instead of `xdp-document-portal`.

For reference, the PipeWire streams we manage have `pipewire.access.portal.app_id` and `pipewire.access.portal.media_roles` on them.

Thoughts?

swick: I really don't care. Updated it to be `user.document-portal.host-path`.

swick:
This is missing documentation. Not sure where to put it.

GeorgesStavracas (approved):

swick:
Added a sentence to the document portal's `GetHostPaths` about the xattr.

document-portal/document-portal-fuse.c
```diff
@@ -2993,13 +3005,62 @@ xdp_fuse_statfs (fuse_req_t req,
...
+                              char     *buf,
+                              size_t    size)
+{
+  g_autoptr(PermissionDbEntry) entry =
+    xdp_lookup_doc (inode->domain->doc_id);
+  const char *path = NULL;
+  size_t path_size;
+
+  if (!entry)
+    return ENODATA;
```
tristan957, line 3028: Should you return `-ENODATA` here?

tristan957: Perhaps this is what the FUSE API expects, which in that case, please ignore :smile:.

swick: No, good catch, this should return -1 and set errno like in all other cases.

swick: I copied this from `xdp_fuse_getxattr` where I believe this already was a mistake. Yikes.

tristan957: No problem. Thanks for your work on desktop linux!

tristan957 (approved):

PR #1378, opened 2024-06-17
portal-impl: fix config ordering (supersedes #1240)

SoumyaRanjanPatnaik:
This PR aims to supersede #1240. 

Original PR message:
> Quoting portals.conf(5):
> 
>     Each key in the group contains a semi-colon separated list of portal backend
>     implementation, to be searched for an implementation of the requested interface,
>     in the same order as specified in the configuration file.
> 
> But this wasn't actually true. If the portals were set to z;y and z and y both implemented the same portal, y would be used.
> 
> Fixing this requires reworking how portals are selected — going through the config file and selecting the first available configured portal, rather than going through each known portal and checking whether it's allowed in the config file.
> 
> find_all_portal_implementations() is unchanged. The portal-first approach is fine for it, and it would be difficult for it to share as much code with find_portal_implementation() now that it's written to stop as soon as a configured portal is found.

Changes from the original PR include:
1. Rebased and resolved conflicts from main.
2. Cleanup redundant code for handling `none` case since `find_portal_implementation` returns early if the portal interface prefers `none`.
3. Add check to `find_portal_implementation` to ensure that the portal implementation selected from the config actually supports the interface.
4. Created a seperate function `find_default_implementation_iface`  to handle the default case, since `find_portal_implementation` only has access to `dbus_name`.  The `dbus_name` for `config->default_portal` is always set to `default`, which will never match the name of the interface we are finding the implementation for. 
5. The original PR moved the code for GTK fallback into a seperate function that was in certain conditions invoked before the default / UseIn fallbacks were tried. Moved this code back into `find_portal_implementation`.
6.  Use the same logic for selecting portals in `find_all_portal_implementations` as well to ensure consistency.
7. Cleanup unused functions.

Let me know if any changes are required or if you need to see a diff between the two PRs. I'm

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/1111

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -535,55 +535,10 @@ portal_interface_prefers_none (const char *interface)
 }
 
 static gboolean
-portal_impl_name_matches (const PortalImplementation *impl,
-                          const PortalInterface      *iface)
-{
-  /* Exact match */
-  if (g_strv_contains ((const char * const *) iface->portals, impl->source))
-    {
-      g_debug ("Found '%s' in configuration for %s", impl->source, iface->dbus_name);
-      return TRUE;
-    }
-
-  /* The "*" alias means "any" */
-  if (g_strv_contains ((const char * const *) iface->portals, "*"))
-    {
-      g_debug ("Found '*' in configuration for %s", iface->dbus_name);
-      return TRUE;
-    }
-
-  /* No portal */
-  if (portal_interface_prefers_none (iface->dbus_name))
-    {
-      g_debug ("Found 'none' in configuration for %s", iface->dbus_name);
-      return FALSE;
-    }
-
-  return FALSE;
-}
-
-static gboolean
-portal_impl_matches_config (const PortalImplementation *impl,
+portal_impl_supports_iface (const PortalImplementation *impl,
                             const char                 *interface)
 {
-  if (config == NULL)
-    return FALSE;
-
-  /* Interfaces have precedence, followed by the "default" catch all,
-   * to allow for specific interfaces to override the default
-   */
-  for (int i = 0; i < config->n_ifaces; i++)
-    {
-      const PortalInterface *iface = config->interfaces[i];
-
-      if (g_strcmp0 (iface->dbus_name, interface) == 0)
-        return portal_impl_name_matches (impl, iface);
-    }
-
-  if (config->default_portal)
-    return portal_impl_name_matches (impl, config->default_portal);
-
-  return FALSE;
+  return g_strv_contains ((const char * const *) impl->interfaces, interface);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


A6GibKm:
I tried the setup described in #1417. from the logs:

```
$ xdg-desktop-portal --replace -v
....
XDP: Preferred portals for interface 'org.freedesktop.impl.portal.Secret': oo7-portal, gnome-keyring
XDP: portal implementation supports org.freedesktop.impl.portal.Secret
XDP: portal implementation supports org.freedesktop.impl.portal.Secret
XDP: Found 'oo7-portal' in configuration for org.freedesktop.impl.portal.Secret
XDP: Using oo7-portal.portal for org.freedesktop.impl.portal.Secret (config)
XDP: providing portal org.freedesktop.portal.Secret
```

So this fixes the issue for me.

src/portal-impl.c
```diff
@@ -535,63 +535,107 @@ portal_interface_prefers_none (const char *interface)
...
+      if (g_str_equal (portal, "*"))
+        return find_any_portal_implementation (iface->dbus_name);
 
-  return FALSE;
+      PortalImplementation *impl = find_portal_implementation_by_name (portal);
+
+      if (!portal_impl_supports_iface (impl, iface->dbus_name))
+        {
+          g_info ("Requested backend %s.portal does not support %s. Skipping...",
+                      impl->source, iface->dbus_name);
```
A6GibKm, line 607: Indentation seems to be off here.

src/portal-impl.c
```diff
@@ -670,26 +715,26 @@ GPtrArray *
...
   impls = g_ptr_array_new ();
 
   if (portal_interface_prefers_none (interface))
     return impls;
 
-  for (l = implementations; l != NULL; l = l->next)
-    {
-      PortalImplementation *impl = l->data;
+  if (config)
+  {
```
A6GibKm, line 725: based on the diff only, indentation seems to be off here.

A6GibKm (commented):
This MR is missing a rebase. Thanks!

ebassi:
I apologise for taking this long, and thanks for your pull request.

It's actually hard for me to review this change, because the diff is all over the place. For future reference, please break apart the changes across multiple commits. Your steps breakdown in the PR cover description would have been a great way to split up this work.

ebassi (approved):

SoumyaRanjanPatnaik:
Thanks for the ACK @ebassi . I agree on the diffs being all over the place haha. I assumed that changes to a single file are supposed to go into a single commit based on the review on a previous PR where I was suggested to squash the commits.

I've fixed the indentation issues and re-based on main. Let me know if there's anything else that needs to go into this PR.

SoumyaRanjanPatnaik:
I tried my best to keep the code style same as the rest of the codebase. I did try to look for a style guide or a linter that enforces it but couldn't. Had to guess for the most part. Can you point me towards the correct resources for this? I'll try to make sure any future contributions will follow the code style.

SoumyaRanjanPatnaik:
Anyways thanks for merging :)

PR #1379, opened 2024-06-20
screen-cast: Add textual context hint

tytan652:
Adds an optional argument to display a textual hint on the Screen Cast selection dialog.

When a software like OBS Studio has multiple Screen Cast session, the user has no way to know from which session the dialog comes from when prompted at start.
This new argument allows to mitigate that if used by the implementation.

- [OBS Studio branch with context hint support](https://github.com/tytan652/obs-studio/tree/screen_cast_context_hint)
- [xdg-desktop-portal-gnome branch with context hint support](https://gitlab.gnome.org/tytan652/xdg-desktop-portal-gnome/-/tree/tytan652/dirty_screen_cast_context_hint), the design is a patchwork not meant to be upstream

### DIFF
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -138,6 +138,16 @@
         Start the screen cast session. This will typically result the portal presenting
         a dialog letting the user do the selection set up by SelectSources.
 
+        Supported keys in the @options vardict include:
+
+        * ``context_hint`` (``s``)
+
+          Textual hint that can be shown on the presented dialog. This allow the user to
+          know the context behind the selection dialog if the application happens use
+          multiple sessions.
+
+          The string is localized by the application.
+
         The following results get returned in the @results vardict:
 
         * ``streams`` (``a(ua{sv})``)
--- a/data/org.freedesktop.portal.ScreenCast.xml
+++ b/data/org.freedesktop.portal.ScreenCast.xml
@@ -168,6 +168,14 @@
           object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
           more information about the @handle.
 
+        * ``context_hint`` (``s``)
+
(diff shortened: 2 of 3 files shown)


tytan652:
Just found about #939 but the author didn't rebased it since a while…

tytan652:
This is not a session name and this is not about shown sources.

User have an with 3 screen cast session (OBS Screen Capture source), if the three show up a prompt at start, the user has no idea for which session the prompt is for.

The context hint shown (for OBS the name of the source) on the prompt would allow the user to know for which session is the prompt, allowing him to chose the source. Rather than dismissing all of them and check which session didn't get restored.

Mikenux:
The problem with text hints from an app to display in the portal UI is that they are unreliable. In the case of the screencast portal:
- No guarantee that the selected source will be assigned to the corresponding source in the app UI.
- The user must be able to check if said source exists in the app UI. Possible for the user, but is this a process we want for them, especially when there are multiple prompts that may cover the app UI?
- A hint can be used to prompt the user to request a specific source. This depends on how things are requested in the portal UI and how the user remembers their session.

Not knowing how users will take things, I tend to say not to use such hints. It is best to contact designers to solve such problem.

tytan652:
> It is best to contact designers to solve such problem.

Feel free to do so, I can't endlessly add stuff to work on my plate (and creating design discussions and keeping them moving is work).

Mikenux:
I can at least open a design issue on the GNOME tracker.

However, before that, I would like to know if the following alternative without a hint is reasonably feasible without too many changes and solves the problem: have a way to account for multiple tokens so that we can present only one window to restore the sources (e.g. "Select missing sources").

If this is not feasible without big changes or if it is not a solution to the problem at all, then I will be more willing to open an issue.

swick:
Not sure what to do with this PR.

@jadahl @davidedmundson opinions?

PR #1380, opened 2024-06-20
globalshorts: Do what the docs say

matthiasclasen:
The docs say that the reponse brings the session handle with the 'session_id' key. But the code was using 'session_handle'.

### DIFF
--- a/src/global-shortcuts.c
+++ b/src/global-shortcuts.c
@@ -186,7 +186,7 @@ session_created_cb (GObject *source_object,
     }
 
   g_variant_builder_add (&results_builder, "{sv}",
-                         "session_handle", g_variant_new ("s", session->id));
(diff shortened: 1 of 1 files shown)


swick:
The frontend calls this `session_token`, the backend calls it `session_id`. I think this change is wrong.

matthiasclasen:
Let me recite the documentation for you:


        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:

        * ``session_handle`` (``o``)

          The session handle. An object path for the
          org.freedesktop.portal.Session object representing the created
          session.

matthiasclasen:
Arf. I looked at the wrong interface.  There's entirely too many property bags with too many freeform keys around these apis :(

PR #1384, opened 2024-06-24
Make contribution link in README.md direct instead of requiring more clicks

gijsk:
This is just a tiny thing but it was one more hurdle - `CONTRIBUTING.md` doesn't contain anything except a link to the "real" docs, so there's not much point for the README to point there.


GeorgesStavracas (approved):
Sure, thanks

PR #1390, opened 2024-06-28
Check for empty environment variables

calendulish:
Fixes portals not being loaded when xdg-desktop-portal is invoked with empty XDG_DESKTOP_PORTAL_DIR environment variable

### DIFF
--- a/src/portal-impl.c
+++ b/src/portal-impl.c
@@ -275,7 +275,7 @@ load_installed_portals (gboolean opt_verbose)
 
   /* We need to override this in the tests */
   portal_dir = g_getenv ("XDG_DESKTOP_PORTAL_DIR");
-  if (portal_dir == NULL)
+  if (portal_dir == NULL || !*portal_dir)
     portal_dir = DATADIR "/xdg-desktop-portal/portals";
 
   g_debug ("load portals from %s", portal_dir);
@@ -436,7 +436,7 @@ load_portal_configuration (gboolean opt_verbose)
   /* We need to override this in the tests */
   portal_dir = g_getenv ("XDG_DESKTOP_PORTAL_DIR");
 
-  if (portal_dir != NULL)
+  if (portal_dir != NULL && *portal_dir)
     {
       load_config_directory (portal_dir, desktops, opt_verbose);
       /* All other config directories are ignored when this is set */


smcv:
> Fixes portals not being loaded when xdg-desktop-portal is invoked with empty XDG_DESKTOP_PORTAL_DIR environment variable

Why has that environment variable been set to an empty value?

"Set-but-empty" and "unset" exist as distinct, valid states for an environment variable. It seems equally valid to expect that setting it to an empty value should either mean "use the default" (which it seems is what you want), or "don't load portals" (which is effectively what it means now)?

calendulish:
> > Fixes portals not being loaded when xdg-desktop-portal is invoked with empty XDG_DESKTOP_PORTAL_DIR environment variable
> 
> Why has that environment variable been set to an empty value?
> 
> "Set-but-empty" and "unset" exist as distinct, valid states for an environment variable. It seems equally valid to expect that setting it to an empty value should either mean "use the default" (which it seems is what you want), or "don't load portals" (which is effectively what it means now)?

With the vast array of different desktop environments available for Linux users, there are certain situations where this variable can be incorrectly set to a empty value. This is due to the behavior of previous sessions that set the XDG_DESKTOP_PORTAL_DIR for some compatibility reasons. Subsequently, when starting a different environment, it's not correctly updated, as in the case of XFCE, which updates the "global" environment variables but doesn't handle updating the user's local variables (on per-user systemd instances and dbus). Typically, the display manager takes care of updating these instances, but in the case of starting sessions using another method, the variable may not be updated. This opens an unnecessary gap for the malfunctioning of the portals, which from the user's perspective is difficult to understand without prior knowledge of the internal workings of the both sessions and portals.

I encountered this problem after starting a gamescope session provided by Valve, ending it, and then starting a XFCE session. This happens because the use of portals within gamescope-session is undesirable, however, diagnosing the problem was difficult due to not even having the XDG_DESKTOP_PORTAL_DIR variable defined in my environment. The problem could only be solved after reading the code and realizing that the variable was defined with an empty value locally in the user's systemd instance by the gamescope session.

It's evident to me that this is improper behavior on the part of the previously started session. A quick google search with the terms "file chooser don't open", "portals not loading" shows some cases of this problem happening with users who have no idea of the reason (among other causes), as user don't expect the portal dir to be incorrectly defined without the user having done this explicitly, which means there has to be a bug report for each of these problematic sessions, a burden that could be **prevented** with this change on the xdg-desktop-portal that uses the default value for empty variables, as this simple empty value causes the overall malfunction of the portals and I don't see any situation where setting the portal dir to an empty value can be useful.

If someone wants to disable portals, I believe this should be done in more appropriate ways, that doesn't interfere in a hidden way with other unrelated user sessions. Allowing this type of behavior causes issues on day-to-day that can be completely avoided.

smcv:
> previous sessions that set the XDG_DESKTOP_PORTAL_DIR for some compatibility reasons

What component is doing this? I'm surprised that anyone is setting XDG_DESKTOP_PORTAL_DIR for reasons other than debugging/development. Each portal backend has its own file in the default directory `/usr/share/xdg-desktop-portal/portals`, and the active portal backend(s) can be selected via the portals.conf mechanism, so there shouldn't be any reason for desktop sessions to need to change this variable.

calendulish:
> > previous sessions that set the XDG_DESKTOP_PORTAL_DIR for some compatibility reasons
> 
> What component is doing this? I'm surprised that anyone is setting XDG_DESKTOP_PORTAL_DIR for reasons other than debugging/development. Each portal backend has its own file in the default directory `/usr/share/xdg-desktop-portal/portals`, and the active portal backend(s) can be selected via the portals.conf mechanism, so there shouldn't be any reason for desktop sessions to need to change this variable.

The official gamescope session implementation used by SteamOS, provided by a package available at https://steamdeck-packages.steamos.cloud/archlinux-mirror/jupiter-3.6/os/x86_64/gamescope-3.14.8-1-x86_64.pkg.tar.zst
the variable is defined by the /usr/bin/start-gamescope-session

> But that doesn't mean we should selectively ignore specific values of the variable, just because those happen to be left behind on your system. What if someone elses system leaks a nonempty value. Should we ignore that too?

We are not talking about arbitrary values or specific values, we are talking about 'no value,' empty. Back to basics, the variable is documented as a means of defining the directory of the portals. For what reason should an empty value be accepted as an appropriate definition for a directory? This goes against the purpose for which this variable was created. The empty variable is currently treated as a valid value, even trying to load the directory, which obviously is invalid (and silently ignored). Why a variable that expects a directory path is accepting an empty value as valid?

smcv:
> The official gamescope session implementation used by SteamOS [is setting XDG_DESKTOP_PORTAL_DIR to the empty string]

That's a bug: it should stop doing that. Instead, if it wants to disable all portals, it should provide a `gamescope-portals.conf` containing

```
[preferred]
default=none
```

and newer versions of the Gamescope session in SteamOS do exactly that. (I don't know the status of that change through the alpha -> beta -> stable pipeline.)

In the older versions of the Gamescope session that did set `XDG_DESKTOP_PORTAL_DIR=""`, that was a workaround for the fact that disabling unwanted portals was not possible in any other way before xdg-desktop-portal 1.17.x/1.18.x, and what it intended to achieve by setting `XDG_DESKTOP_PORTAL_DIR=""` was exactly "make portals not work".

> It's evident to me that this is improper behavior on the part of the previously started session

So please fix the previously-started session, by updating the Gamescope session to a version that does not have this bug.

GeorgesStavracas:
I think there's sufficient evidence against the behavior proposed by this PR. Thanks for the contribution and the discussion.

PR #1393, opened 2024-07-03
ci: Install libportal from git

GeorgesStavracas:
Install libportal from git, potentially checking out a similarly named branch from the author's fork.


tristan957:
Did you give any thought to managing libportal via a Meson wrap? It _might_ be easier specifying branch names there rather than in the workflow file.

GeorgesStavracas:
> Did you give any thought to managing libportal via a Meson wrap? It might be easier specifying branch names there rather than in the workflow file.

Not much, I just copied the approach that Mutter / GNOME Shell implemented. I've already spent too much time on this, but I don't have any preliminary concern or objection to using wrap files, if someone opens a PR for that.

tristan957:
I think if you find this workflow works for you, keep it. I had a similar problem at my last job that I had to solve, too.

TingPing (approved):

PR #1398, opened 2024-07-21
Fix potential segfault in permissions.c

heather7283:
Under specific circumstances, xdp could segfault due to `error` variable being NULL. This patch adds a check that prevents us from referencing `error` if it's not initialised. 

```
[174025.661871] pool-/usr/lib/x[621481]: segfault at 8 ip 000055e10b83d484 sp 00007fc411fff7c0 error 4 in xdg-desktop-portal[55e10b7d5000+86000] likely on CPU 0 (core 0, socket 0)
[174025.661899] Code: 5d c3 0f 1f 40 00 48 8b 7d d0 ff 15 6e fc 06 00 48 8b 45 d0 4c 89 f1 48 8d 15 a8 0d 02 00 be 80 00 00 00 48 8d 3d 2d 16 02 00 <4c> 8b 40 08 31 c0 ff 15 90 0a 07 00 31 db e9 6f ff ff ff 66 0f 1f
```
[xdp-bt.txt](https://github.com/user-attachments/files/16325136/xdp-bt.txt)

### DIFF
--- a/src/permissions.c
+++ b/src/permissions.c
@@ -44,8 +44,16 @@ get_permissions_sync (const char *app_id,
                                                         NULL,
                                                         &error))
     {
-      g_dbus_error_strip_remote_error (error);
-      g_debug ("No '%s' permissions found: %s", table, error->message);
+      if (error != NULL)
+        {
+          g_dbus_error_strip_remote_error (error);
+          g_debug ("No '%s' permissions found: %s", table, error->message);
+        }
+      else
+        {
+          g_debug ("No '%s' permissions found", table);
+        }
+
       return NULL;
     }
 


smcv:
I don't think this is right: I think it's more likely to be working around a different programming error.

The GLib error-handling convention is that if we have a function that looks like this:

```
gboolean do_something (..., GError **error);

gboolean result = do_something (..., &local_error);
```

then there are only two possible results in the absence of a programming error:

* result == TRUE && local_error == NULL
* result == FALSE && local_error != NULL

Normally the only time a function returns false without setting the error indicator is if incorrect arguments are passed to it. For example if `permission_store == NULL`, then that should trip a `g_return_val_if_fail()` in `g_dbus_proxy_call_sync()`, which will result in an "assertion failed" message being logged.

If that's what is happening in your scenario, then checking whether the error is NULL is the wrong solution. Instead, if `permission_store == NULL`, then we simply should not be calling `xdp_dbus_impl_permission_store_call_lookup_sync()`.

The GLib jargon is to say that it is a programming error (undefined behaviour) to pass incorrect arguments to a function, for example `xdp_dbus_impl_permission_store_call_lookup_sync (NULL, ...)`.

More information: https://docs.gtk.org/glib/error-reporting.html

smcv:
> Under specific circumstances

What specific circumstances? What else is logged before the segfault?

heather7283:
At the time of writing I was unsure what exactly causes it to segfault, but now I can reproduce it reliably. That's what triggers a segfault on my system:
1. Make sure xdg-permission-store is *not* running
2. Start xdg-desktop-portal and make sure dbus activation for xgd-permission-store fails
3. Run mpv

full xdp log with --verbose and backtrace:
[xdp-log.txt](https://github.com/user-attachments/files/16338730/xdp-log.txt)

swick:
> (/home/heather/projects/xdg-desktop-portal/build/src/xdg-desktop-portal:189018): GLib-GIO-CRITICAL **: 23:07:35.181: g_dbus_proxy_call_sync_internal: assertion 'G_IS_DBUS_PROXY (proxy)' failed

This is the root cause.

smcv:
A good starting point would be to add a `g_return_val_if_fail()` at the beginning of `get_permissions_sync()` like this:

```
char **
get_permissions_sync (const char *app_id,
                      const char *table,
                      const char *id)
{
  g_autoptr(GVariant) out_perms = NULL;
  g_autoptr(GVariant) out_data = NULL;
  g_autofree char **permissions = NULL;

  g_return_val_if_fail (G_IS_DBUS_PROXY (permission_store), NULL);

  ...
```

That would change the symptom from "assertion failure, crash" to "assertion failure, continue to run".

smcv:
> Start xdg-desktop-portal and make sure dbus activation for xgd-permission-store fails

So, the real root cause is: your xdg-permission-store (which is also part of this project) is somehow broken. xdg-permission-store is required by xdg-desktop-portal for correct operation, so please look into why that component is broken on your system (assuming you have not intentionally broken it to provide a simple reproducer for this crash), and fix it so that it can start successfully.

I would be OK with making `permissions.c` more robust against a missing or broken xdg-permission-store by adding some `g_return_[val_]if_fail()` as suggested in https://github.com/flatpak/xdg-desktop-portal/pull/1398#issuecomment-2244838674, but that is a workaround, not the solution. The only proper solution here would be to fix xdg-permission-store.

If xdg-permission-store is failing to start on your system, please report that as an issue. It's often clearest if you report issues as issues, rather than jumping directly to contributing a workaround which might be happening in the wrong layer.

swick:
There are a bunch of places where `get_permission_store` is used with the assumption that it will return a valid pointer. Adding `g_return_if_fail` is just moving the crashes further up the stack.

smcv:
> There are a bunch of places where `get_permission_store` is used with the assumption that it will return a valid pointer.

If someone wants to make those more robust against x-p-s being broken, they could contribute a PR to do that too, for example:

```diff
 static GVariant *
 get_all_permissions (void)
 {
   g_autoptr(GError) error = NULL;
   g_autoptr(GVariant) out_perms = NULL;
   g_autoptr(GVariant) out_data = NULL;
-
-  if (!xdp_dbus_impl_permission_store_call_lookup_sync (get_permission_store (),
+  XdpDbusImplPermissionStore *store = get_permission_store ();
+
+  g_return_val_if_fail (store != NULL, NULL);
+
+  if (!xdp_dbus_impl_permission_store_call_lookup_sync (store,
```

But, again, this would only be a workaround for x-p-s being broken, and the actual solution is to fix x-p-s.

heather7283:
> So, the real root cause is: your xdg-permission-store (which is also part of this project) is somehow broken.
Yes, I am aware that permission store being broken is not normal and errors are to be expected, but I think segfault is not a good way of reporting error to end user.

Should I close this PR and create an issue instead?

swick:
Technically it is logged here already:

> No permission store: Error calling StartServiceByName for org.freedesktop.impl.portal.PermissionStore: Failed to activate service 'org.freedesktop.impl.portal.PermissionStore': timed out (service_start_timeout=1ms)

> A good starting point would be to add a g_return_val_if_fail() at the beginning of get_permissions_sync() like this:

Sorry but this isn't useful. There are a bunch of cases in the code base where `get_permission_store` is used but none of them expect it to be NULL and would crash. `get_permission_sync` will return `PERMISSION_UNSET` if `get_permissions_sync` didn't crash which then might get you more permissions than you actually have (i.e. someone denied something that is granted by default).

I would very much prefer to bail out when the permission store is not available with a stronger message.

smcv:
> Should I close this PR and create an issue instead?

Yes please. The real solution to this is to fix whatever is wrong with your xdg-permission-store process.

GeorgesStavracas:
Thanks for the contribution and discussion nonetheless :)

PR #1401, opened 2024-07-29
main: Exit early when permission store or document portal are missing

swick:
And another small fix for a potential issue.

### DIFF
--- a/src/documents.c
+++ b/src/documents.c
@@ -39,17 +39,23 @@
 static XdpDbusDocuments *documents = NULL;
 static char *documents_mountpoint = NULL;
 
-void
-init_document_proxy (GDBusConnection *connection)
+gboolean
+init_document_proxy (GDBusConnection  *connection,
+                     GError          **error)
 {
   documents = xdp_dbus_documents_proxy_new_sync (connection, 0,
                                                  "org.freedesktop.portal.Documents",
                                                  "/org/freedesktop/portal/documents",
-                                                 NULL, NULL);
+                                                 NULL, error);
+  if (!documents)
+    return FALSE;
+
   xdp_dbus_documents_call_get_mount_point_sync (documents,
                                                 &documents_mountpoint,
                                                 NULL, NULL);
   xdp_set_documents_mountpoint (documents_mountpoint);
+
+  return TRUE;
 }
 
 char *
--- a/src/permissions.c
+++ b/src/permissions.c
@@ -161,18 +161,16 @@ void set_permission_sync (const char *app_id,
   set_permissions_sync (app_id, table, id, (const char * const *)perms);
 }
 
-void
-init_permission_store (GDBusConnection *connection)
+gboolean
+init_permission_store (GDBusConnection  *connection,
+                       GError          **error)
 {
-  g_autoptr(GError) error = NULL;
-
   permission_store = xdp_dbus_impl_permission_store_proxy_new_sync (connection,
                                                                     G_DBUS_PROXY_FLAGS_NONE,
                                                                     "org.freedesktop.impl.portal.PermissionStore",
                                                                     "/org/freedesktop/impl/portal/PermissionStore",
-                                                                    NULL, &error);
-  if (permission_store == NULL)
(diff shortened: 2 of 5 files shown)


src/documents.c
```diff
@@ -39,17 +39,23 @@
...
                                                  "org.freedesktop.portal.Documents",
                                                  "/org/freedesktop/portal/documents",
-                                                 NULL, NULL);
+                                                 NULL, error);
+  if (!documents)
+    return FALSE;
+
   xdp_dbus_documents_call_get_mount_point_sync (documents,
                                                 &documents_mountpoint,
                                                 NULL, NULL);
```
smcv, line 55: Not a merge blocker for this change IMO, but should this check for errors too?

src/permissions.c
```diff
@@ -161,18 +161,16 @@ void set_permission_sync (const char *app_id,
...
-
   permission_store = xdp_dbus_impl_permission_store_proxy_new_sync (connection,
                                                                     G_DBUS_PROXY_FLAGS_NONE,
                                                                     "org.freedesktop.impl.portal.PermissionStore",
                                                                     "/org/freedesktop/impl/portal/PermissionStore",
-                                                                    NULL, &error);
-  if (permission_store == NULL)
-    g_warning ("No permission store: %s", error->message);
+                                                                    NULL, error);
+  return !!permission_store;
```
smcv, line 173:
```suggestion
  return (permission_store != NULL);
```

is functionally equivalent, but would be more consistent with the coding style I generally see in GLib-based code

src/xdg-desktop-portal.c
```diff
@@ -196,26 +196,31 @@ on_bus_acquired (GDBusConnection *connection,
...
 
   /* make sure errors are registered */
   portal_errors = XDG_DESKTOP_PORTAL_ERROR;
 
   xdp_connection_track_name_owners (connection, peer_died_cb);
-  init_document_proxy (connection);
-  init_permission_store (connection);
+
+  if (!init_permission_store (connection, &error))
+    g_error ("No permission store: %s", error->message);
```
smcv, line 210: This will die with `SIGABRT` or equivalent, and dump core. Is that what we want?

If we want the portal to exit gracefully, the way to achieve that would be something like this:

```
static int global_exit_status = 0;
static GMainLoop *loop = NULL;

…

  if (!init_permission_store (connection, &error))
    {
      g_critical ("No permission store: %s", error->message);
      global_exit_status = 1;
      g_main_loop_quit (loop);
      return;
    }
  /* and the same for the documents proxy */

…

main (...)
{
  …

  loop = …;
  …
  g_main_loop_run (loop);
  …

  return global_exit_status;
}
```

src/xdg-desktop-portal.c
```diff
@@ -196,26 +196,31 @@ on_bus_acquired (GDBusConnection *connection,
...
   if (lockdown_impl != NULL)
     lockdown = xdp_dbus_impl_lockdown_proxy_new_sync (connection,
                                                       G_DBUS_PROXY_FLAGS_NONE,
                                                       lockdown_impl->dbus_name,
                                                       DESKTOP_PORTAL_OBJECT_PATH,
-                                                      NULL, &error);
-  else
+                                                      NULL, NULL);
+
+  if (!lockdown)
```
smcv, line 223:
```suggestion
  if (lockdown == NULL)
```

smcv (commented):
Thanks, I like this better than #1398.

src/xdg-desktop-portal.c
swick: updated the PR to do this

src/documents.c
swick: There is only one place that doesn't check for it to be NULL and that's in `register_document`. Not sure if that should also handle NULL or if we should exit early like in the other cases.

smcv (approved):

src/documents.c
smcv: I don't know the documents portal well enough to say. Definitely a job for a separate PR, either way.

GeorgesStavracas (approved):

PR #1402, opened 2024-07-29
README: Describe the activation environment requirements for x-d-p

smcv:
Inspired by https://github.com/flatpak/xdg-desktop-portal-gtk/issues/440 and various other issue reports that can be summarized as "I use (some minimal window manager or compositor), and it didn't propagate an environment variable to the activation environment".

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,4 +97,58 @@ When it comes to processes, passing PIDs around is not useful in a sandboxed
 world where apps are likely in their own PID namespace. And passing PIDs from
 inside the sandbox is problematic, since the app can just lie.
 
+## System Integration
+
+### D-Bus Activation Environment
+
+xdg-desktop-portal and its portal backends are activatable D-Bus services.
+This means that they inherit environment variables from the
+"activation environment" maintained by either `systemd --user`
+(on systems that use systemd) or `dbus-daemon` (on systems that do not).
+They do not inherit environment variables from the GUI environment or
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
Thanks, this looks good to me. Can I suggest, in addition to the README file, also adding documentation pages for these?

The syntax is slightly different, but it's a matter of:

 1. Add `doc/system-integration.rst` with a similar content
 2. List it in `doc/meson.build`
 3. List it under `doc/for-desktop-developers.rst`

smcv:
> Can I suggest, in addition to the README file, also adding documentation pages for these?

I'll put it on my list, but my list is not short.

I did wonder how much of this should have been in `doc/` instead, but I think we probably do want it to be in the README, because otherwise there's realistically no hope of "I built my own desktop environment" enthusiasts discovering it. I think this is close to the maximum level of detail for this stuff that we can reasonably put in the README, but it can link to a more detailed document where necessary (like I did for portals.conf).

Of course it isn't really x-d-p's job to explain what an activatable service is, how environment variables propagate around a modern "freedesktop" system, or that setting `XDG_CURRENT_DESKTOP` is a de facto standard... but it seems we're absorbing a lot of the support burden for this anyway, so we might as well try to mitigate that by educating our users.

GeorgesStavracas (approved):
> I'll put it on my list, but my list is not short.

Actually don't worry about it, you seem very busy, I can do this.

PR #1408, opened 2024-08-06
data: Document the session_handle type mismatch

### LINKED ISSUE #1359: Type of session_handle in the response of CreateSession differs from documentation
### Operating System

Fedora 40

1.18

_No response_

### Desktop Environment

KDE

_No response_

### Expected Behavior

It should return an object path as specified in the documentation (`o` type)

### Current Behavior

It returns a string (`s` type)

### Steps to Reproduce

1. Call CreateSession on RemoteDesktop, ScreenCast, GlobalShortcuts, or CreateMonitor on Inhibit
2. Listen for the Response signal
3. Look at the type of the session_handle

_No response_

whot:
This was erroneously implemented as `s` but for backwards compatibility cannot be changed as it would break existing clients. Best we can do here is fix the documentation and tell people about it.

See eebb04172d00 ("use correct type for the session object path") and its revert in e4523531c.

Closes #1359

### DIFF
--- a/data/org.freedesktop.portal.GlobalShortcuts.xml
+++ b/data/org.freedesktop.portal.GlobalShortcuts.xml
@@ -67,11 +67,15 @@
 
         The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
 
-        * ``session_handle`` (``o``)
+        * ``session_handle`` (``s``)
 
           The session handle. An object path for the
           org.freedesktop.portal.Session object representing the created
           session.
+
+          .. note::
+            The ``session_handle`` is an object path that was erroneously implemented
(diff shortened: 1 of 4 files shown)


swick:
LGTM

bilelmoussaoui:
same is also true for location portal

whot:
@bilelmoussaoui which call? I'm looking at `xdp_dbus_location_call_create_session_finish()` and it gets the handle as `o`, same for `xdp_dbus_location_call_create_session_sync()`. Same for the `call_start*` equivalents. So this looks correct?

jadahl:
lgtm, will merge in some days if noone objects

PR #1414, opened 2024-08-08
Screencast: do not destroy proxy object when destroying PipeWire remote

### LINKED ISSUE #1407: Segfault when launching screen sharing
### Operating System

KDE Neon 6.0 (Ubuntu 22.04)

1.18

### Desktop Environment

KDE Plasma 6.1.3

### Expected Behavior

When starting a screen sharing session, a window appears allowing to choose what to share (single screen, single app...).
Clicking on any option, the screen sharing should start right away.

### Current Behavior

Immediately upon clicking on any options, I have a segfault that terminates the process immediately.

### Steps to Reproduce

1. Go to https://www.webrtc-experiment.com/Pluginfree-Screen-Sharing/
2. Press Share your screen
Alternatively, trigger a screen sharing from any application.
3. Choose any option other than a specific tab inside Chrome.
Core dumped.

This seems to be related to libpipewire-module-protocol-native.so but when rolling back to an earlier release of xdg-desktop-portal (like 1.14.4 which is available in my distro) everything works fine.

The logs, attached: 
[xdg-desktop-portal.log](https://github.com/user-attachments/files/16499156/xdg-desktop-portal.log)

grulja:
This is causing a crash with older PipeWire clients. Partly undoes #1319 which introduced this issue.

Fixes #1407

### DIFF
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -222,7 +222,6 @@ pipewire_remote_destroy (PipeWireRemote *remote)
   /* This check is a workaround for older PW versions */
   if (remote->registry)
     spa_hook_remove (&remote->registry_listener);
-  g_clear_pointer (&remote->registry, pw_proxy_destroy);
   g_clear_pointer (&remote->globals, g_hash_table_destroy);
   if (remote->core)
     spa_hook_remove (&remote->core_listener);


jadahl:
Cc: @rmader this partly undoes https://github.com/flatpak/xdg-desktop-portal/pull/1319.

grulja:
Yeah, I've seen those. I'm confused and not sure what should be done here. In the examples I can see `pw_core_disconnect()` is called aftewards, but also in the PipeWire code, the `pw_core_disconnect()` [also destroys the proxy](https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/pipewire/core.c#L492). 

Maybe @wtay can give us the answer. Btw. I'm dealing with the same issue in WebRTC, where I can see the same.

grulja:
I have managed to find the cause for asserts in WebRTC and it's unrelated to this change. I'm still not sure whether this change here is correct so I'll keep it open. Since I'm not able to reproduce this issue, it would help to get a backtrace with debug symbols from PipeWire.

lmoureaux:
> Since I'm not able to reproduce this issue, it would help to get a backtrace with debug symbols from PipeWire.

Posted on #1407

PR #1415, opened 2024-08-09
Screenshot: Allow selecting the type of screenshot

### LINKED ISSUE #1065: Bring non-interactive screenshot options to parity with interactive ones
Hi. I was looking into building a screenshot tool for Wayland similar to Flameshot, and it's my understanding that I would use the desktop portal for that.

When I experimented with some prototypes, I noticed that the "silent" method call (with `interactive: false`) significantly lacks in features compared to the interactive one. On my DE (KDE), the interactive dialogue allows to select the screenshot region, whether to include the mouse cursor, set a delay, etc., none of which are available to the other version.

This is perhaps fine for applications that just want to "take a screenshot" and don't particularly care about how it's handled, but it's very inconvenient for writing custom screengrab tools which obviously do care a lot.

The screen capture tools on Linux still lag behind the equivalent solutions on Windows (such as ShareX), and it would be helpful to have the infrastructure in order to improve on that. Would it be possible to expose some of those interactive options on the D-Bus interface to bring the methods up to parity?

I found prior discussion about this in #139, and it seems like there was some initial interest in pursuing it. Would it make sense to restart the discussion?

leolost2605:
This option allows to select the type (screen, window, area) of screenshot to be taken. Useful for apps that want to do the interactive parts themselves but still allow the user these different types of screenshots.
This is only a hint, backends may not implement all of these.

The use case I'm coming from is that at elementary we would like our screenshot app to have app actions for screen, window and area that bypass any UI (given that permission was granted previously) and just take the screenshot when launched.

pantheon (elementary OS) backend implementation: elementary/portals#105

If this receives some positive feedback I will do the GNOME implementation as well and libportal too I guess.

Fixes #1065


src/screenshot.c
```diff
@@ -178,9 +179,28 @@ screenshot_done (GObject *source,
   g_task_run_in_thread (task, send_response_in_thread_func);
 }
 
+static gboolean
+validate_type (const char *key,
+               GVariant *value,
+               GVariant *options,
+               GError **error)
+{
+  uint32_t type = g_variant_get_uint32 (value);
```
hfiguiere, line 188: This returns a `guint32`. Please use that.

src/screenshot.c
```diff
@@ -178,9 +179,28 @@ screenshot_done (GObject *source,
...
+
+  if (type > 2)
+    {
+      g_set_error (error, XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   "Invalid screenshot type: %x", type);
```
hfiguiere, line 193: why `%x` ?

src/screenshot.c
leolost2605: 👍

src/screenshot.c
leolost2605: Whoops copy paste sry

src/screenshot.c
```diff
@@ -20,6 +20,7 @@
 
 #include "config.h"
 
+#include <stdint.h>
```
hfiguiere, line 23:
```suggestion
```

you can remove this too now.

leolost2605: Ah right I'm currently doing this on my phone  whoops 😬

src/screenshot.c
jadahl: We've actually to some degree been migrating away from `glib` types that have standard equivalent variants, like `gint` `guint32`, `gint64` to `int`, `uint32_t` and `int64_t`, etc. They are for all intents and purposes equivalent, but there is no reason to avoid the standard ones. They also have a higher chance of resulting in looking more sensible in editors syntax highlighting.

Mikenux:
How is this supposed to work with non-interactive mode?

I think the ability for apps to check which values ​​are supported would be useful so they can gray out actions (e.g. menu items).

leolost2605:
> How is this supposed to work with non-interactive mode?

It would take a screenshot of the requested kind right away and send it, allowing to take more sophisticated screenshots without having to go through additional UI, like the interactive dialog. All of that of course given that permission was granted by the user. 
So with screen it would behave like it currently does by default, with window it would, depending on the backend implementation, just take a screenshot of the active window and send it or allow the user to choose one and with area it would allow the user to select an area then screenshot that and send it. 

> I think the ability for apps to check which values ​​are supported would be useful so they can gray out actions (e.g. menu items).

Yeah that would certainly be useful but AFAICT it would require some more severe API breakage since the best way I can think of would probably be a bit flag dbus property. Not sure how this should be handled/was handled in other cases but I'll look into it :)

src/screenshot.c
leolost2605: I noticed in the rest of the file the glib types are used so maybe keep that for consistency but you tell me what to use :shrug:

Mikenux:
I took a look at previous issues regarding the addition of the automatic permission. It seems it has been added for having: - One less step. Before, including for interactive, and in GNOME it was: select screenshot method > take screenshot > share. Now it is, with interactive: select screenshot method > take and share screenshot.
- Allow screenshot tools like Flameshot to draw their custom interface.

That makes one reason to not have selecting the type of screenshot for the automatic permission. Also:
- For "window", this needs to be predictive so it can"t really depend on backend implementation. Currently, the interactive method does a better job by allowing to select which window to screenshot, plus allowing to user to validate this shot (take and share the screenshot).
- For "area", if it is using the system-side are selector anyway, why having it without screenshot validation as users can decide to change at any time the area?

src/screenshot.c
GeorgesStavracas: What we've been doing, as @jadahl mentioned, is that new code should use standard types (`uint32_t`) instead of GLib types, so I think this PR should use `uint32_t` etc

src/screenshot.c
```diff
@@ -178,9 +178,28 @@ screenshot_done (GObject *source,
...
+    }
+
+  return TRUE;
+}
+
 static XdpOptionKey screenshot_options[] = {
   { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
-  { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL }
+  { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "type", G_VARIANT_TYPE_UINT32, validate_type }
```
GeorgesStavracas, line 202: Nitpick:

```suggestion
  { "type", G_VARIANT_TYPE_UINT32, validate_type },
```

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -50,6 +50,17 @@
           Hint whether the dialog should offer customization before taking a screenshot.
           Defaults to no.
 
+        * ``type`` (``u``)
```
GeorgesStavracas, line 53: I think `type` is not very descriptive, because many parts of D-Bus and GLib and etc already use "type" for other things. When it comes to screenshots, I feel like `mode` fits better. What do you think?

GeorgesStavracas (changes requested):
I think the current code is alright, but incomplete: usually for features like this, it's important that we expose to apps which screenshot types the system supports. This can be done by adding a new property `AvailableModes` to the Screenshot portal.

You can look at the ScreenCast portal's `AvailableSourceTypes` for a reference code

swick:
Also missing tests

Photon89:
@leolost2605 Thanks for the PR! Is there any progress on implementing the changes requested by @GeorgesStavracas? I think, there are several screenshot tools out there eagerly waiting for this functionality to be implemented to be able to take screenshots on Wayland!

PR #1422, opened 2024-08-24
Order portal startup after graphical-session.target

Vladimir-csp:
Otherwise portal units might try to start before compositor exported `WAYLAND_DISPLAY` to activation environments.
[Reference](https://github.com/Vladimir-csp/uwsm/issues/37#issuecomment-2307818022)

### DIFF
--- a/document-portal/xdg-document-portal.service.in
+++ b/document-portal/xdg-document-portal.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=flatpak document portal service
 PartOf=graphical-session.target
+After=graphical-session.target
 
 [Service]
 BusName=org.freedesktop.portal.Documents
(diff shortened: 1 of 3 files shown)


TingPing:
I don't think this service would depend on `WAYLAND_DISPLAY`, only the frontends like -gtk would.

I guess one question is what happens when this starts earlier, does `-gtk` wait on the `After` rule or autostart immediately.

Vladimir-csp:
I do not know, I never had this problem myself. Ordering seemed like logical solution and it was reported working. I queried for more information.

smcv:
> Ordering seemed like logical solution

The solution to what problem? (Steps to reproduce / expected result / actual result)

Vladimir-csp:
Linked in the description. Judging by [reporter's log](https://github.com/user-attachments/files/16538681/uwsm-labwc-startup.log.txt) something might have triggered portals early and they just tried to start right away, long before units bound to `graphical-session.target` entered the picture. Namely, gtk portal tried to start somewhere alongside units bound to `graphical-session-pre.target`.

PR #1424, opened 2024-08-26
Use sealded memfd for icon validator

jsparber:
I split out the first few commits from https://github.com/flatpak/xdg-desktop-portal/pull/1298/ that rework how the validator works and that introduce the `XdpSealdFd`.

CC: @hfiguiere and @GeorgesStavracas

### DIFF
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -138,6 +138,33 @@ validate_desktop_file_id (const char  *app_id,
 }
 
 static gboolean
+validate_serialized_icon (GVariant  *arg_icon_v,
+                          char     **icon_format,
+                          char     **icon_size)
+{
+  GBytes *bytes;
+  g_autoptr(GIcon) icon = NULL;
+  g_autoptr(GVariant) icon_v = NULL;
+  g_autoptr(XdpSealedFd) sealed_icon = NULL;
+
+  if (!arg_icon_v)
+    return FALSE;
+
+  if (!(icon_v = g_variant_get_variant (arg_icon_v)))
+    return FALSE;
+
+  icon = g_icon_deserialize (icon_v);
+
+  if (!G_IS_BYTES_ICON (icon))
+    return FALSE;
+
+  bytes = g_bytes_icon_get_bytes (G_BYTES_ICON (icon));
+  sealed_icon = xdp_sealed_fd_new_from_bytes (bytes, NULL);
+
+  return sealed_icon && xdp_validate_icon (sealed_icon, icon_format, icon_size);
+}
+
+static gboolean
 write_icon_to_disk (GVariant    *icon_v,
                     const char  *icon_subdir,
                     const char  *icon_path,
@@ -617,9 +644,7 @@ handle_prepare_install (XdpDbusDynamicLauncher *object,
     }
 
   /* Do some validation on the icon before passing it along */
-  icon_v = g_variant_get_variant (arg_icon_v);
-  if (!icon_v || !xdp_validate_serialized_icon (icon_v, TRUE /* bytes_only */,
-                                                &icon_format, &icon_size))
+  if (!validate_serialized_icon (arg_icon_v, &icon_format, &icon_size))
     {
       g_dbus_method_invocation_return_error (invocation,
                                              XDG_DESKTOP_PORTAL_ERROR,
@@ -688,9 +713,7 @@ handle_request_install_token (XdpDbusDynamicLauncher *object,
   if (response == 0)
     {
       /* Do some validation on the icon before saving it */
-      icon_v = g_variant_get_variant (arg_icon_v);
-      if (!icon_v || !xdp_validate_serialized_icon (icon_v, TRUE /* bytes_only */,
-                                                    &icon_format, &icon_size))
+      if (!validate_serialized_icon (arg_icon_v, &icon_format, &icon_size))
         {
           g_dbus_method_invocation_return_error (invocation,
                                                  XDG_DESKTOP_PORTAL_ERROR,
--- a/src/validate-icon.c
+++ b/src/validate-icon.c
@@ -34,14 +34,14 @@
 #endif
 
 #define ICON_VALIDATOR_GROUP "Icon Validator"
+#define MAX_ICON_SIZE 512
+#define MAX_SVG_ICON_SIZE 4096
 
 static int
-validate_icon (const char *arg_width,
-               const char *arg_height,
-               const char *filename)
+validate_icon (const char *filename)
 {
   GdkPixbufFormat *format;
-  int max_width, max_height;
+  int max_size;
   int width, height;
   g_autofree char *name = NULL;
   const char *allowed_formats[] = { "png", "jpeg", "svg", NULL };
@@ -57,6 +57,12 @@ validate_icon (const char *arg_width,
       return 1;
     }
 
+  if (width != height)
+    {
+      g_printerr ("Expected a square icon but got: %dx%d\n", width, height);
+      return 1;
+    }
+
   name = gdk_pixbuf_format_get_name (format);
   if (!g_strv_contains (allowed_formats, name))
     {
@@ -64,31 +70,13 @@ validate_icon (const char *arg_width,
       return 1;
     }
 
-  if (!g_str_equal (name, "svg"))
-    {
-      max_width = g_ascii_strtoll (arg_width, NULL, 10);
-      if (max_width < 16 || max_width > 4096)
-        {
-          g_printerr ("Bad width limit: %s\n", arg_width);
-          return 1;
-        }
+  /* Sanity check for vector files */
+  max_size = g_str_equal (name, "svg") ? MAX_SVG_ICON_SIZE : MAX_ICON_SIZE;
 
-      max_height = g_ascii_strtoll (arg_height, NULL, 10);
-      if (max_height < 16 || max_height > 4096)
-        {
-          g_printerr ("Bad height limit: %s\n", arg_height);
-          return 1;
-        }
-    }
-  else
+  /* The icon is a square so we only need to check one side */
+  if (width > max_size)
     {
-      /* Sanity check for vector files */
-      max_height = max_width = 4096;
-    }
-
-  if (width > max_width || height > max_height)
-    {
-      g_printerr ("Image too large (%dx%d). Max. size %dx%d\n", width, height, max_width, max_height);
+      g_printerr ("Image too large (%dx%d). Max. size %dx%d\n", width, height, max_size, max_size);
       return 1;
     }
 
@@ -96,12 +84,6 @@ validate_icon (const char *arg_width,
   if (pixbuf == NULL)
     {
       g_printerr ("Failed to load image: %s\n", error->message);
-      return 1;
-    }
-
-  if (width != height)
-    {
-      g_printerr ("Expected a square icon but got: %dx%d\n", width, height);
       return 1;
     }
 
@@ -165,9 +147,7 @@ flatpak_get_bwrap (void)
 }
 
 static int
-rerun_in_sandbox (const char *arg_width,
-                  const char *arg_height,
-                  const char *filename)
+rerun_in_sandbox (const char *filename)
 {
   const char * const usrmerged_dirs[] = { "bin", "lib32", "lib64", "lib", "sbin" };
   int i;
@@ -234,7 +214,7 @@ rerun_in_sandbox (const char *arg_width,
   if (g_getenv ("G_MESSAGES_PREFIXED"))
     add_args (args, "--setenv", "G_MESSAGES_PREFIXED", g_getenv ("G_MESSAGES_PREFIXED"), NULL);
 
-  add_args (args, validate_icon, arg_width, arg_height, filename, NULL);
+  add_args (args, validate_icon, filename, NULL);
   g_ptr_array_add (args, NULL);
 
   execvpe (flatpak_get_bwrap (), (char **) args->pdata, NULL);
@@ -257,7 +237,7 @@ main (int argc, char *argv[])
   g_autoptr(GOptionContext) context = NULL;
   g_autoptr(GError) error = NULL;
 
-  context = g_option_context_new ("WIDTH HEIGHT PATH");
+  context = g_option_context_new ("PATH");
   g_option_context_add_main_entries (context, entries, NULL);
   if (!g_option_context_parse (context, &argc, &argv, &error))
     {
@@ -265,16 +245,16 @@ main (int argc, char *argv[])
       return 1;
     }
 
-  if (argc != 4)
+  if (argc != 2)
     {
-      g_printerr ("Usage: %s [OPTION…] WIDTH HEIGHT PATH\n", argv[0]);
+      g_printerr ("Usage: %s [OPTION…] PATH\n", argv[0]);
       return 1;
     }
 
 #ifdef HELPER
   if (opt_sandbox)
-    return rerun_in_sandbox (argv[1], argv[2], argv[3]);
+    return rerun_in_sandbox (argv[1]);
   else
 #endif
-    return validate_icon (argv[1], argv[2], argv[3]);
+    return validate_icon (argv[1]);
 }
--- a/src/validate-icon.c
+++ b/src/validate-icon.c
@@ -16,6 +16,7 @@
  *
  * Authors:
  *       Matthias Clasen <mclasen@redhat.com>
+ *       Julian Sparber <jsparber@gnome.org>
  */
 
 /* The canonical copy of this file is in:
@@ -27,7 +28,10 @@
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include <glib/gstdio.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <unistd.h>
+
+#include "xdp-utils.h"
 
 #ifdef __FreeBSD__
 #define execvpe exect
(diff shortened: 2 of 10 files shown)


swick:
@GeorgesStavracas last time we talked about this I think we could agree on making the validator take an argument which specifies what kind of icon should be validated and have all the logic about what a valid icon kind looks like in the validator itself. Do I remember that right?

src/dynamic-launcher.c
```diff
@@ -137,6 +137,31 @@ validate_desktop_file_id (const char  *app_id,
...
+
+  if (!(icon_v = g_variant_get_variant (arg_icon_v)))
+    return FALSE;
+
+  icon = g_icon_deserialize (icon_v);
+
+  if (!G_IS_BYTES_ICON (icon))
+    return FALSE;
+
+  sealed_icon = xdp_sealed_fd_new_from_bytes (g_bytes_icon_get_bytes (G_BYTES_ICON (icon)), NULL);
```
swick, line 160: bit long line, maybe split

src/xdp-sealed-fd.h
```diff
@@ -0,0 +1,37 @@
...
+                      GObject)
+
+XdpSealedFd * xdp_sealed_fd_new_take_memfd  (int           memfd,
+                                             GError      **error);
+XdpSealedFd * xdp_sealed_fd_new_from_bytes  (GBytes       *bytes,
+                                             GError      **error);
+XdpSealedFd * xdp_sealed_fd_new_from_handle (GVariant     *handle,
+                                             GUnixFDList  *fd_list,
+                                             GError      **error);
+int           xdp_sealed_fd_get_fd         (XdpSealedFd  *sealed_fd);
```
swick, line 35: Are we still trying to align all functions like  this @GeorgesStavracas ?

src/xdp-utils.c
```diff
@@ -613,42 +575,13 @@ xdp_validate_serialized_icon (GVariant  *v,
...
   args[1] = "--sandbox";
   args[2] = "--fd";
   args[3] = G_STRINGIFY (VALIDATOR_INPUT_FD);
   args[4] = NULL;
 
-  /* Ensure that we read from the beginning of the file */
-  lseek (fd, 0, SEEK_SET);
-
-  if (!(output = xdp_spawn_full (args, xdp_steal_fd (&fd), VALIDATOR_INPUT_FD, &error)))
+  if (!(output = xdp_spawn_full (args, xdp_sealed_fd_dup_fd (icon), VALIDATOR_INPUT_FD, &error)))
```
swick, line 584: Does this handle dup returning -1 correctly?

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+#include "xdp-utils.h"
+#include "xdp-sealed-fd.h"
+
+static const int required_seals = F_SEAL_GROW | F_SEAL_WRITE | F_SEAL_SHRINK;
+
+struct _XdpSealedFd {
+  GObject parent_instance;
+
+  int fd;
+  gboolean is_sealed;
```
swick, line 39: leftover

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+  return g_steal_pointer (&sealed_fd);
+}
+
+XdpSealedFd *
+xdp_sealed_fd_new_from_bytes (GBytes  *bytes,
+                              GError **error)
+{
+  gconstpointer bytes_data;
+  gsize bytes_len;
+  xdp_autofd int fd;
```
swick, line 117: have to initialize to -1

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+  sealed_fd->fd = -1;
+}
+
+XdpSealedFd *
+xdp_sealed_fd_new_take_memfd (int      memfd,
+                              GError **error)
+{
+  int saved_errno = -1;
+  g_autoptr(XdpSealedFd) sealed_fd = NULL;
+  xdp_autofd int fd = memfd;
```
swick, line 74: For some reason I'm a bit uneasy about this and would rather initialize to -1 and then assign after the declaration.

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+  gsize bytes_len;
+  xdp_autofd int fd;
+  int saved_errno = -1;
+  g_autoptr(XdpSealedFd) sealed_fd = NULL;
+  char *shm;
+  g_autoptr(GOutputStream) stream = NULL;
+
+  g_return_val_if_fail (bytes != NULL, NULL);
+
+  if ((fd = memfd_create ("sealed-fd-from-bytes", MFD_ALLOW_SEALING)) == -1)
```
swick, line 125: maybe a bettter name would be `xdp-sealed-fd` so one knows it's from xdp?

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+  if (ftruncate (fd, bytes_len) == -1)
+    {
+      int saved_errno;
+
+      saved_errno = errno;
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "ftruncate: %s", g_strerror (saved_errno));
+      return NULL;
```
swick, line 160: Here we will leak the `shm` mapping.

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,226 @@
...
+
+#include "xdp-utils.h"
+#include "xdp-sealed-fd.h"
+
+static const int required_seals = F_SEAL_GROW | F_SEAL_WRITE | F_SEAL_SHRINK;
+
+struct _XdpSealedFd {
+  GObject parent_instance;
+
+  int fd;
```
swick, line 38: It's a bit weird that the fd is the only thing left because one could just replace XdpSealedFd with an int, but I guess this at least provides nice type-safety for all things we consider sealed.

src/xdp-sealed-fd.h
GeorgesStavracas: No need to align functions names and return types, but the arguments must be aligned as always

src/xdp-sealed-fd.c
GeorgesStavracas: Maybe it would look less "risky" if this was declared as:

```c
  xdp_autofd int fd = g_steal_fd (&memfd);
```

as it makes it abundantly clear that the `fd` variable is unequivocally the owner of the file descriptor?

GeorgesStavracas:
> @GeorgesStavracas last time we talked about this I think we could agree on making the validator take an argument which specifies what kind of icon should be validated and have all the logic about what a valid icon kind looks like in the validator itself. Do I remember that right?

Yes. But I'm tending not to block this PR on it, it can be a follow-up cleanup just fine.

src/validate-icon.c
```diff
@@ -237,24 +267,42 @@ main (int argc, char *argv[])
...
+  if (opt_path != NULL && opt_fd != -1)
     {
-      g_printerr ("Usage: %s [OPTION…] PATH\n", argv[0]);
+      g_printerr ("Error: Only --path or --fd can be given\n");
+      return 1;
+    }
+
+  if (opt_path)
+    {
+      opt_fd = g_open (opt_path, O_RDONLY | O_CLOEXEC, 0);
```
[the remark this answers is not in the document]
jsparber: it's closed anyways once the validator terminates, no?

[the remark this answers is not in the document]
swick: Yes, but it's not cleaned up by your code but by the OS eventually which is a leak. So while I'd still prefer a local autofd, this isn't bad.

src/xdp-utils.c
jsparber: `xdp_spawn_full()` uses only fds > -1 so yes.

src/validate-icon.c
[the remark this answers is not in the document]
jsparber: don't we actually also leak `opt_path`?

src/validate-icon.c
```diff
@@ -36,37 +38,67 @@
...
   g_autofree char *key_file_data = NULL;
+  g_autoptr(GdkPixbufLoader) loader = NULL;
+  g_autoptr(GMappedFile) mapped = NULL;
+  int max_size, width, height;
+  g_autofree char *name = NULL;
+  GdkPixbuf *pixbuf;
 
-  format = gdk_pixbuf_get_file_info (filename, &width, &height);
-  if (format == NULL)
+  if (!(mapped = g_mapped_file_new_from_fd (input_fd, FALSE, &error)))
```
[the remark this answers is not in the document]
jsparber: According to the [docs](https://www.man7.org/linux/man-pages/man2/mmap.2.html) of mmap 

> It is unspecified whether changes made to the file after the mmap() call are visible in the mapped region.

Even though [GMappedFile](https://docs.gtk.org/glib/ctor.MappedFile.new_from_fd.html) says that the mapping will change when the file changes.

But i agree we should document it.

src/validate-icon.c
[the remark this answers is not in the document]
jsparber: I think it's pretty hard to close the fd in the right cases, since it depends on whether we run in sandbox and whether the fd given to the validator has FD_CLOSEXEC. I would leave it as is.

src/validate-icon.c
```diff
@@ -224,10 +252,14 @@ rerun_in_sandbox (const char *filename)
...
 
-static gboolean opt_sandbox;
+static gboolean  opt_sandbox;
+static gchar    *opt_path = NULL;
+static gint      opt_fd = -1;
 
 static GOptionEntry entries[] = {
   { "sandbox", 0, 0, G_OPTION_ARG_NONE, &opt_sandbox, "Run in a sandbox", NULL },
+  { "path", 0, 0, G_OPTION_ARG_FILENAME, &opt_path, "Read icon data from given file path, required to be guaranteed non modifiable", "PATH" },
+  { "fd", 0, 0, G_OPTION_ARG_INT, &opt_fd, "Read icon data from given file descriptor, required to be sealed", "FD" },
```
swick, line 269: This is a bit wrong. Both the (non-sealed) fd and path must not be from an untrusted source. It being "modifiable" isn't an issue and the file xdp passes to it is modifiable for example but xdp creates this file in a way that it's not visible to sandboxed apps (untrusted sources).

Only a sealed fd can be directly from an untrusted source.

src/validate-icon.c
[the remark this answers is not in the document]
swick: > It is unspecified whether changes made to the file after the mmap() call are visible in the mapped region.

This talks about changes to the content, not to the mapping. If you truncate a file while a mapping exist, the mapping will continue to have the same size and have parts that have no backing anymore. That's what will result in SIGBUS.

src/validate-icon.c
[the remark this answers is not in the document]
swick: Don't really see the issue... Use a local `autofd` variable for cleaning up the fd the code creates itself and always assign to opt_fd as well. Same for `opt_path` with `autofree`.

Not leaking things is really nice because tools like valgrind and lsan will work better.

src/validate-icon.c
```diff
@@ -224,10 +252,14 @@ rerun_in_sandbox (const char *filename)
 }
 #endif
 
-static gboolean opt_sandbox;
+static gboolean  opt_sandbox;
+static gchar    *opt_path = NULL;
+static gint      opt_fd = -1;
```
swick, line 264: char/int

src/xdp-utils.c
```diff
@@ -414,37 +414,34 @@ spawn_exit_cb (GObject      *obj,
   spawn_data_exit (data);
 }
 
-gboolean
-xdp_spawn (GFile       *dir,
-           char       **output,
-           GSubprocessFlags flags,
-           GError     **error,
+gchar *
```
swick, line 417: can change this to char as well given that the entire function is being changed

src/xdp-utils.h
```diff
@@ -122,17 +122,11 @@ xdp_close_fd (int *fdp)
 
 
 char *   xdp_quote_argv (const char           *argv[]);
-gboolean xdp_spawn      (GFile                *dir,
-                         char                **output,
-                         GSubprocessFlags      flags,
-                         GError              **error,
+char *   xdp_spawn      (GError              **error,
```
swick, line 125: no need to align different functions

src/xdp-utils.c
swick: Same in the next commit

src/xdp-utils.h
swick: Same in a few more places.

swick:
A few more nit picks about alignment and use of gchar/gint and documentation but when they are dealt with this is good to land IMO.

src/validate-icon.c
```diff
@@ -41,6 +41,7 @@
 #define MAX_ICON_SIZE 512
 #define MAX_SVG_ICON_SIZE 4096
 #define BUFFER_SIZE 4096
+#define MAX_FILE_SIZE 4194304 /* Max file size of 4MB */
```
swick, line 44: `(1024 * 1024 * 4)`

src/xdp-utils.h
```diff
@@ -122,17 +122,13 @@ xdp_close_fd (int *fdp)
...
-                         const gchar          *argv0,
-                         va_list               ap);
-gboolean xdp_spawnv     (GFile                *dir,
-                         char                **output,
-                         GSubprocessFlags      flags,
-                         GError              **error,
-                         const gchar * const  *argv);
+char *   xdp_spawn (GError             **error,
+                    const char          *argv0,
+                    ...) G_GNUC_NULL_TERMINATED;
```
swick, line 127:
```suggestion
char * xdp_spawn (GError             **error,
                  const char          *argv0,
                  ...) G_GNUC_NULL_TERMINATED;
```

swick: Last nitpicks, I promise.

src/validate-icon.c
```diff
@@ -27,88 +28,105 @@
...
+
 #ifdef __FreeBSD__
 #define execvpe exect
 #endif
 
 #define ICON_VALIDATOR_GROUP "Icon Validator"
+#define MAX_ICON_SIZE 512
+#define MAX_SVG_ICON_SIZE 4096
+#define BUFFER_SIZE 4096
+#define MAX_FILE_SIZE (1024 * 1024 * 4) /* Max file size of 4MB */
```
hfiguiere, line 44:
```suggestion
#define MAX_FILE_SIZE (1024 * 1024 * 4) /* Max file size of 4MiB */
```

src/validate-icon.c
```diff
@@ -27,46 +28,82 @@
...
+  int max_size, width, height;
+  g_autofree char *name = NULL;
+  GdkPixbuf *pixbuf;
 
-  format = gdk_pixbuf_get_file_info (filename, &width, &height);
-  if (format == NULL)
+  /* Ensure that we read from the beginning of the file */
+  lseek (input_fd, 0, SEEK_SET);
+
+  if (!(mapped = g_mapped_file_new_from_fd (input_fd, FALSE, &error)))
```
GeorgesStavracas, line 64:
```suggestion
  mapped = g_mapped_file_new_from_fd (input_fd, FALSE, &error);
  if (error)
```

src/validate-icon.c
```diff
@@ -27,46 +28,82 @@
...
+  loader = gdk_pixbuf_loader_new ();
+
+  if (!gdk_pixbuf_loader_write_bytes (loader, bytes, &error) ||
+      !gdk_pixbuf_loader_close (loader, &error) ||
+      !(pixbuf = gdk_pixbuf_loader_get_pixbuf (loader)))
+    {
+      g_printerr ("Failed to load image: %s\n", error->message);
+      gdk_pixbuf_loader_close (loader, NULL);
+      return 1;
+    }
```
GeorgesStavracas: Hm, I think this will eventually call `gdk_pixbuf_loader_close (loader, NULL);` too much. I think these checks have to be split after all.

```suggestion
  if (!gdk_pixbuf_loader_write_bytes (loader, bytes, &error) ||
      !gdk_pixbuf_loader_close (loader, &error))
    {
      g_printerr ("Failed to load image: %s\n", error->message);
      gdk_pixbuf_loader_close (loader, NULL);
      return 1;
    }
    
  pixbuf = gdk_pixbuf_loader_get_pixbuf (loader);
  if (!pixbuf)
    {
      g_printerr ("Failed to load image: %s\n", error->message);
      return 1;
    }
```

src/validate-icon.c
```diff
@@ -27,46 +28,82 @@
...
+
+  if (!(format = gdk_pixbuf_loader_get_format (loader)))
```
GeorgesStavracas, line 95:
```suggestion
  format = gdk_pixbuf_loader_get_format (loader);
  if (!format)
```

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,225 @@
...
+                              GError **error)
+{
+  int saved_errno = -1;
+  g_autoptr(XdpSealedFd) sealed_fd = NULL;
+  xdp_autofd int fd = g_steal_fd (&memfd);
+  int seals;
+
+  g_return_val_if_fail (fd != -1, NULL);
+
+  if ((seals = fcntl (fd, F_GET_SEALS)) == -1)
```
GeorgesStavracas, line 78:
```suggestion
  seals = fcntl (fd, F_GET_SEALS);
  if (seals == -1)
```

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,225 @@
...
+  gsize bytes_len;
+  xdp_autofd int fd = -1;
+  int saved_errno = -1;
+  g_autoptr(XdpSealedFd) sealed_fd = NULL;
+  char *shm;
+  g_autoptr(GOutputStream) stream = NULL;
+
+  g_return_val_if_fail (bytes != NULL, NULL);
+
+  if ((fd = memfd_create ("xdp-sealed-fd", MFD_ALLOW_SEALING)) == -1)
```
GeorgesStavracas, line 124:
```suggestion
  fd = memfd_create ("xdp-sealed-fd", MFD_ALLOW_SEALING);
  if (fd == -1)
```

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,225 @@
...
+
+      saved_errno = errno;
+      g_set_error (error,
+                   G_IO_ERROR,
+                   g_io_error_from_errno (saved_errno),
+                   "ftruncate: %s", g_strerror (saved_errno));
+      return NULL;
+    }
+
+  if ((shm = (char *) mmap (NULL, bytes_len, PROT_WRITE, MAP_SHARED, fd, 0)) == (void *) -1)
```
GeorgesStavracas, line 150: I think here you can also use `MAP_FAILED` instead of the raw `(void *) -1` cast:

```suggestion
  shm = mmap (NULL, bytes_len, PROT_WRITE, MAP_SHARED, fd, 0);
  if (shm == MAP_FAILED)
```

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,225 @@
...
+XdpSealedFd *
+xdp_sealed_fd_new_from_bytes (GBytes  *bytes,
+                              GError **error)
+{
+  gconstpointer bytes_data;
+  gsize bytes_len;
+  xdp_autofd int fd = -1;
+  int saved_errno = -1;
+  g_autoptr(XdpSealedFd) sealed_fd = NULL;
+  char *shm;
```
GeorgesStavracas, line 119:
```suggestion
  gpointer shm;
```

src/xdp-sealed-fd.c
```diff
@@ -0,0 +1,225 @@
...
+{
+  int fd_id;
+  xdp_autofd int fd = -1;
+
+  g_return_val_if_fail (g_variant_is_of_type (handle, G_VARIANT_TYPE_HANDLE), NULL);
+  g_return_val_if_fail (G_IS_UNIX_FD_LIST (fd_list), NULL);
+  g_return_val_if_fail (g_unix_fd_list_get_length (fd_list) > 0, NULL);
+
+  fd_id = g_variant_get_handle (handle);
+  if ((fd = g_unix_fd_list_get (fd_list, fd_id, error)) == -1)
```
GeorgesStavracas, line 205:
```suggestion
  fd = g_unix_fd_list_get (fd_list, fd_id, error);
  if (fd == -1)
```

GeorgesStavracas (changes requested):
My comments are mostly superficial at this point, my big complaint is that I think inline assignments are overused in these patches. My rule of thumb for this is, only assign variables in `if`s / `for`s / `while`s lines when the variable is only going to be used inside the block (i.e. it's an scoped variable).

src/validate-icon.c
jsparber: `_close()` checks whether the loader was already closed. https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/gdk-pixbuf/gdk-pixbuf-loader.c#L839

I still changed it since it makes it easier to read the code.

GeorgesStavracas (approved):

PR #1439, opened 2024-09-23
GVariant, GError and g_autoptr cleanups

swick:


whot (approved):
Not sure about the non-floating gvariants patch but it looks correct and the rest too.

matthiasclasen:
Pedantic grammar fix for the second commit message:

"... and do in fact re-use it ..."

matthiasclasen:
I agree with the idea of avoid floating as much as possible

PR #1440, opened 2024-09-23
Add cast and type-check functions for Session and Request

swick:
We would usually get them from the G_DECLARE_TYPE macro family but we cannot use them right now because they have no namespace and the derivable classes would need private data and our data is public.

For now, just define the functions manually where it makes sense. In the longer term, we should switch towards the G_DECLARE_TYPE macros.


whot:
does what it says on the box, LGTM

PR #1444, opened 2024-09-26
tests: install files required for `installed_tests`

Artturin:
These are required to be in `XDG_DATA_DIR` when the installed tests are run


swick:
LGTM

PR #1446, opened 2024-09-27
data: Fix license blurb and add SPDX-License-Identifier

pwithnall:
The license blurb mentioned version 2 of the Lesser GPL, which doesn’t
exist — there’s version 2.0 of the _Library_ GPL, or version 2.1 of the
_Lesser_ GPL, but no 2.0 Lesser GPL.

Given that the COPYING file correctly mentions the LGPL 2.1, and the
vast majority of software projects and contributors use version 2.1 in
preference to version 2.0, I think this is a typo which has been
copypastaed across the project.

So, change ‘version 2’ to ‘version 2.1’ in the license blurbs for the
portal API definitions. It’s particularly important that these are
right, because they may well be copied to other projects so that D-Bus
bindings can be generated from them.

In addition, add an `SPDX-License-Identifier` to the portal API
definitions so the license is machine readable.

Signed-off-by: Philip Withnall <pwithnall@gnome.org>

### DIFF
--- a/data/org.freedesktop.background.Monitor.xml
+++ b/data/org.freedesktop.background.Monitor.xml
@@ -2,10 +2,12 @@
 <!--
  Copyright (C) 2022 Endless OS Foundation, LLC
 
+ SPDX-License-Identifier: LGPL-2.1-or-later
+
  This library is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
  License as published by the Free Software Foundation; either
- version 2 of the License, or (at your option) any later version.
+ version 2.1 of the License, or (at your option) any later version.
 
  This library is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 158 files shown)


pwithnall:
These changes were all done manually, but should be identical in each file — so if you agree with the rationale for the PR, then this should be an easy review despite the number of files the PR touches.

If you don’t agree with the rationale, then the only other option for fixing the licensing (that I can think of) would be to switch all the affected files to use the Library GPL 2.0 blurb (and appropriate SPDX header).

In any case, having the `SPDX-License-Identifier` header is very useful, as it allows the licensing to be machine readable, and hence tools like `reuse lint` can be used to verify the licensing of the entire project automatically. See [here](https://gitlab.gnome.org/GNOME/glib/-/blob/main/tests/reuse.sh?ref_type=heads) for an example of how we use that in GLib’s tests. There’s an introduction to REUSE and SPDX [here](https://reuse.software/faq/).

GeorgesStavracas (approved):
Thanks, I agree with this, seems like the least disruptive way to fix this

PR #1447, opened 2024-09-27
doc: Fix Background Monitor Apps ref link

tytan652:
Fix the ref link which is linked to this warning while generating docs:
```
../doc/for-desktop-developers.rst:102: WARNING: unknown document: '/doc-org.freedesktop.background.Monitor.rst' [ref.doc]
```


GeorgesStavracas (approved):
Thank you!

PR #1450, opened 2024-10-06
Fix some memory leaks

hfiguiere:
g_build_filename() always transfer full
Add auto cleanup for Settings implementation


swick:
LGTM

GeorgesStavracas (approved):

PR #1334, opened 2024-04-11
ci: Build image from Containerfile when tag changes

swick:
This is based on work by @flexagoon (https://github.com/flatpak/xdg-desktop-portal/pull/1270) but changes how the build images are being build, has a few things renamed, fixes an issue with the artifact upload, and adds a check for POTFILES.in.

PR #1451, opened 2024-10-07
Test the location portal with a python test and a mocked GeoClue2

swick:


tests/__init__.py
```diff
@@ -332,8 +332,8 @@ class PortalMock:
     Parent class for portal tests.
     """
 
-    def __init__(self, session_bus, portal_name: str, app_id: str = "org.example.App"):
-        self.bus = session_bus
+    def __init__(self, dbus_test_case, portal_name: str, app_id: str = "org.example.App"):
+        self.test_case = dbus_test_case
```
whot, line 336: arguably, it's better to use `self.dbus_test_case` here to avoid confusion with the individual actual test case we're running.

tests/__init__.py
```diff
@@ -353,40 +353,80 @@ def dbus_con(self):
...
-        portal = portal or self.portal_name
-        self.p_mock, self.obj_portal = self.test_case.spawn_server_template(
-            template=f"tests/templates/{portal.lower()}.py",
-            parameters=params,
+    def get_bus_for_template(self, template: str):
+        # we need the bus address from the template module
+        module = dbusmock.mockobject.load_module(template)
+
+        bustype = dbusmock.BusType.SYSTEM if module.SYSTEM_BUS else dbusmock.BusType.SESSION
+        assert bustype is not None
```
whot, line 361: can't ever happen - is this to satisfy some naive linter?

tests/__init__.py
```diff
@@ -332,57 +332,101 @@ class PortalMock:
...
         )
-        flags = fcntl.fcntl(self.p_mock.stdout, fcntl.F_GETFL)
-        fcntl.fcntl(self.p_mock.stdout, fcntl.F_SETFL, flags | os.O_NONBLOCK)
-        self.mock_interface = dbus.Interface(self.obj_portal, dbusmock.MOCK_IFACE)
 
-        self.start_dbus_monitor()
+        flags = fcntl.fcntl(server.process.stdout, fcntl.F_GETFL)
+        fcntl.fcntl(server.process.stdout, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+
+        self.busses[bustype][module.BUS_NAME] = server
```
whot, line 377: Hooray, github doesn't let me do suggestions here... Anwyay, more pythonic may be:
```python
try:
    return self.busses[bustype][module.BUS_NAME]
except KeyError:
    server = ....
    return server
```

tests/__init__.py
```diff
@@ -332,57 +332,101 @@ class PortalMock:
...
         """
 
-        self.obj_portal.AddTemplate(
-            f"tests/templates/{portal.lower()}.py",
+        template = f"tests/templates/{template.lower()}.py"
+
+        module = dbusmock.mockobject.load_module(template)
+
+        bustype = dbusmock.BusType.SYSTEM if module.SYSTEM_BUS else dbusmock.BusType.SESSION
+        assert bustype is not None
```
whot, line 390: same as above, cannot happen anyway

tests/conftest.py
```diff
@@ -64,12 +94,14 @@ def app_id():
...
     """
     Fixture yielding a PortalMock object with the impl started, if applicable.
     """
-    pmock = PortalMock(session_bus, portal_name, app_id)
-    if portal_has_impl:
-        pmock.start_impl_portal(params)
+    pmock = PortalMock(dbus_test_case, portal_name, app_id)
+    for template, params in required_templates.items():
+        if template in required_params:
+            params = required_params[template]
```
whot, line 104:
```suggestion
        params = required_params.get(template, params)
```

tests/test_location.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+    def get_client_mock(self, portal_mock):
+        geoclue_manager_proxy = portal_mock.dbus_con_sys.get_object(
+            "org.freedesktop.GeoClue2",
+            "/org/freedesktop/GeoClue2/Manager",
+        )
+        geoclue_manager = dbus.Interface(
+            geoclue_manager_proxy,
+            f"org.freedesktop.GeoClue2.Manager"
```
whot, line 41:
```suggestion
            "org.freedesktop.GeoClue2.Manager"
```

tests/test_clipboard.py
```diff
@@ -43,36 +47,30 @@ def start_session(self, pmock, params={}):
...
+    def test_request_clipboard_and_start_session(self, portal_mock):
+        _, clipboard_enabled = self.start_session(portal_mock)
 
         assert clipboard_enabled
 
-    def test_clipboard_checks_clipboard_enabled(self, dbus_test_case):
-        pmock = PortalMock(dbus_test_case, "Clipboard")
-        session, clipboard_enabled = self.start_session(pmock)
-        clipboard_interface = pmock.get_dbus_interface()
+    @pytest.mark.parametrize("required_params", ({"RemoteDesktop": {"force-clipboard-enabled": False}},))
```
whot, line 55: bikeshed: `required_params` sounds like these are, well, required. But these are "just" parameters passed into the test for the test to pass. How about `template_params`?

swick: Always happy to take suggestions for names :)

tests/__init__.py
swick: Copied from python-dbusmock but you're right. Removed it.

tests/__init__.py
swick: Reworked this commit a bit. Would be good if you can take another look.

swick:
@whot CI fails with `AttributeError: module 'dbusmock' has no attribute 'BusType'` and I'm a bit confused why CI has this problem and locally it works

swick:
@GeorgesStavracas is it okay to add a more recent python-dbusmock for the tests?

GeorgesStavracas:
> @GeorgesStavracas is it okay to add a more recent python-dbusmock for the tests?

I'm not sure, it probably is fine. @whot knows better about the Python testing infrastructure and might have a more informed answer though.

swick:
If we get the CI MR (https://github.com/flatpak/xdg-desktop-portal/pull/1334) in then we'd be using ubuntu:latest which ships with a recent python-dbusmock.

swick:
CI is green now.

swick:
@whot could you take another look please? I have more python test changes in the pipeline.

tests/__init__.py
```diff
@@ -353,40 +353,73 @@ def dbus_con(self):
...
+        try:
+            return self.busses[bustype][module.BUS_NAME]
+        except KeyError:
+            server = dbusmock.SpawnedMock.spawn_for_name(
+                module.BUS_NAME,
+                "/dbusmock",
+                dbusmock.OBJECT_MANAGER_IFACE,
+                bustype,
+                stdout=subprocess.PIPE,
+            )
```
whot, line 366: theoretically you could get that key error if you specify and invalid bus type though then everything else will fall over too I guess. a simple
```py
assert bustype in dbusmock.BusType
```
should do the trick here but I also accept a "meh".

whot (approved):
LGTM now, thanks

Issue #1350, opened 2024-04-26
GlobalShortcuts.BindShortcut unclear return value


dcz-self:
### Operating System

n/a

### XDG Desktop Portal version

1.18

### Desktop Environment

Deepin

### Expected Behavior

The docs cover all the relevant use cases.

## Disabled shortcuts

[BindShortcuts description](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html#org-freedesktop-portal-globalshortcuts-bindshortcuts) says:

> Each element of the shortcuts array returned by the org.freedesktop.portal.Request::Response signal is a tuple composed of a shortcut id, and a vardict with the following keys:

about the list, and pretty much nothing else. What's missing in particular is whether all shortcuts are expected to come back.

> trigger_description (s)
> 
> User-readable text describing how to trigger the shortcut for the client to render.

is clearly meant for the application to communicate how to trigger the shortcut. What if the user decided to make the shortcut impossible to trigger? An application assuming that whatever comes back is set may end up printing "Press `unset` to trigger this", which is silly.
On the other hand, if skipped shortcuts are not present in the return value, the application may crash when trying to print a message("Press {} to trigger this", shortcuts["this"]).

### Steps to Reproduce

1. Read BindShortcuts description
2. Try to understand the return value across different use cases

swick:
I think the intention here is to send back the shortcuts that were bound which might be a subset, including an empty subset of the shortcuts which were passed in the `shortcuts` array.

KDE seems to have implemented this in a way that it returns as many entries as shortcuts were passed in. The frontend does not have this restriction.

It does make sense to me to clarify that a subset of shortcuts can be returned.

PR #1462, opened 2024-10-09
globalshortcuts: Specify the shortcuts result of BindShortcuts

swick:
The method result's shortcuts list is the list of shortcuts which were actually bound. It's valid to send any subset of shortcuts that were passed in by the caller.

See: https://github.com/flatpak/xdg-desktop-portal/issues/1350

@dcz-self @davidedmundson


matthiasclasen:
Looks good, +1 from me.

PR #1464, opened 2024-10-09
globalshortcuts: Remove session_id option from CreateSession impl

swick:
It is not used in the frontend and not in the backends. This is most likely a mistake from copying things from another portal.


matthiasclasen:
Looks fine to me, +1

PR #1471, opened 2024-10-15
ci: Use images from actor instead of owner

swick:
This allows PRs to get through CI when they have changes in the required image.


GeorgesStavracas (approved):
Sweet and small, thanks

PR #1472, opened 2024-10-15
ci: Use images from actor for pull requests

swick:
This allows PRs to get through CI when they have changes in the required image.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 20241008-1
+  IMAGE_TAG: 20241015-2
 
 on:
   workflow_call:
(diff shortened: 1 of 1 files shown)


.github/workflows/container.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
           password: ${{ github.token }}
 
-      - name: Set lower case owner name
-        env:
-          OWNER: '${{ github.repository_owner }}'
+      - name: Find the registry owner to use
         run: |
-          echo "OWNER_LOWERCASE=${OWNER,,}" >>${GITHUB_ENV}
+          OWNER="${{ github.repository_owner }}"
+          [[ "${{github.event_name}}" == "pull_request" ]] && OWNER="${{ github.actor }}"
```
GeorgesStavracas, line 33: I think it'd prefer the boring, obvious if-then-else here:

```suggestion
          if [[ "${{github.event_name}}" == "pull_request" ]]; then
              OWNER="${{ github.repository_owner }}"
          else
              OWNER="${{ github.actor }}"
          fi
```

GeorgesStavracas (changes requested):

.github/workflows/container.yml
swick: It's the other way around, but that's maybe a good argument for doing it the more normal way.

GeorgesStavracas (approved):
Let's see how it goes this time

PR #1473, opened 2024-10-16
Implement a signal handler to help with asan

hfiguiere:
This is the same as in the gnome portal.

Kill the portal with SIGHUP, it will run asan if it was built with asan.


hfiguiere:
The GNOME portal patch https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/186

GeorgesStavracas (approved):

PR #1477, opened 2024-10-17
pytest: Isolate directories, start permission store and add umockdev support

swick:
Based on https://github.com/flatpak/xdg-desktop-portal/pull/1451

What the title says. This is prep work for being able to test the USB portal with the python harness.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -337,9 +337,9 @@ class PortalMock:
     ):
         self.dbus_test_case = dbus_test_case
         self.portal_name = portal_name
-        self.xdp = None
-        self.portal_interfaces: Dict[str, dbus.Interface] = {}
+        self.portal_frontend = None
         self.dbus_monitor = None
+        self.portal_interfaces: Dict[str, dbus.Interface] = {}
         self.app_id = app_id
         self.busses = {dbusmock.BusType.SYSTEM: {}, dbusmock.BusType.SESSION: {}}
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -94,13 +94,25 @@ def app_id():
...
 
 @pytest.fixture
-def portal_mock(dbus_test_case, portal_name, required_templates, template_params, app_id) -> PortalMock:
+def portal_mock(
+    dbus_test_case,
+    portal_name,
+    required_templates,
+    template_params,
+    app_id,
+) -> PortalMock:
```
whot, line 153: on that note, might be best to add a `ruff` job via precommit.ci, that way we don't have to manually do this.

tests/conftest.py
```diff
@@ -94,13 +94,25 @@ def app_id():
...
     """
     Fixture yielding a PortalMock object with the impl started, if applicable.
     """
-    pmock = PortalMock(dbus_test_case, portal_name, app_id)
+    pmock = PortalMock(
+        dbus_test_case,
+        portal_name,
+        app_id=app_id,
```
whot, line 160: Any reason you're passing the app_id by name but not the others?

tests/conftest.py
```diff
@@ -6,10 +6,51 @@
...
+        'XDG_CACHE_HOME',
+        'XDG_CONFIG_HOME',
+        'XDG_DATA_HOME',
+        'XDG_RUNTIME_DIR',
+    ]
+    for env_dir in env_dirs:
+        directory = os.path.join(test_root.name, env_dir.lower())
+        os.mkdir(directory, mode=0o700)
+        os.environ[env_dir] = directory
+        print('Setup', env_dir, 'as', directory)
```
whot, line 47:
```suggestion
        directory = test_root / env_dir.lower()
        directory.mkdir(mode=0o700, parents=True)
        os.environ[env_dir] = directory.absolute()
        print('Setup', env_dir, 'as', directory)
```
because pathlib is nicer and iirc `tempfile` gives a Path object anyway.4


Does the `print` actually do something? iirc printing during conftest was a bit tricky (and arguably should be logger anyway)

tests/__init__.py
```diff
@@ -420,68 +428,118 @@ def start_xdp(self):
         Start the xdg-desktop-portal process
         """
 
+        portal_dir = (
+            Path(os.getenv("G_TEST_BUILDDIR") or "tests")
```
whot, line 432:
```suggestion
            Path(os.getenv("G_TEST_BUILDDIR", "tests"))
```
though admittedly it does the same thing...

tests/__init__.py
```diff
@@ -420,68 +428,118 @@ def start_xdp(self):
...
+
+        # This roughly resembles test-portals.c and glib's test behavior
+        # but preferences in-tree testing by running pytest in meson's
+        # project_build_root
+        libexecdir = os.getenv("LIBEXECDIR")
+        if libexecdir:
+            permission_store_path = Path(libexecdir) / "xdg-permission-store"
+        else:
+            permission_store_path = (
+                Path(os.getenv("G_TEST_BUILDDIR") or "tests")
```
whot, line 501:
```suggestion
                Path(os.getenv("G_TEST_BUILDDIR", "tests"))
```
same as above but feel free to ignore, the `or` is more expressive

whot (approved):
a few style nitpicks on style but this LGTM, thanks

tests/conftest.py
swick: Added ruff via pre-commit in https://github.com/flatpak/xdg-desktop-portal/pull/1486 and this PR now depends on that.

tests/conftest.py
swick: No, passing everything by position now.

tests/conftest.py
swick: Removed the printing as well. It adds no value.

TingPing:
Needs rebase. The leak commits could also be its own PR.

TingPing (approved):

PR #1478, opened 2024-10-18
Avoid forwarding all notification file desciptors to backend

jadahl:
Main commit:

```
    notification: Use separate fd list for input and output
    
    Doing this means we don't unnecessarily forward file descriptors that
    weren't intended to be passed along.
    
    For the notification portal, this avoids forwarding file descriptors
    from the app to the backend that were dropped due to being placed in an
    unsupported property.
```

I haven't tested this yet, I just ran the tests.


swick:
LGTM but CI is currently a bit weird because it requires the image in the repository of the user of the pull request. If you force push again it should get green.

GeorgesStavracas (approved):

PR #1479, opened 2024-10-18
meson: Make libportal a meson subproject

swick:


subprojects/libportal.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory=libportal
+url=https://github.com/jsparber/libportal.git
```
whot, line 3: Is this correct? if so it should be noted in the commit message at least that we're not using upstream.

whot (approved):
Does what it says on the box :) See my comment regarding the libportal repo though please

subprojects/libportal.wrap
swick: Yes, we currently require this version because we merged the notifications portal v2 changes which depend on a libportal branch/PR to add support for the notifications portal v2 API.

This is an unfortunate circular dependency we should drop in the longer term by moving the tests to the pytest.

hfiguiere: the unfortunate part is that the **build** has been broken since and the proper solution instead is

1. disable said test
or
2. implement the missing function directly in the test.

swick: I don't understand. With this PR, the local and CI builds both use libportal as subproject which implements all the functions required by the tests.

hfiguiere: using a fork like that is NOT the proper fix. If you are blocked by some PR being merged elsewhere then the QUICK solution is the one I gave above, that will UNBLOCK other contributors. I still don't understand how CI didn't break (tbf I didn't look at the problem) but then that mean the CI workflow has a bug there too.

hfiguiere: See https://github.com/flatpak/xdg-desktop-portal/pull/1484

swick: Sorry, but I don't see what's wrong with specifying which libportal commit is required to run the tests. Disabling some tests on the other hand just makes our testing worse while also not making it clear which libportal version is required and potentially having different versions in CI and on development setups.

swick: Changed it to build libportal from upstream main.

swick:
Requires https://github.com/flatpak/xdg-desktop-portal/pull/1484 to fix CI now.

meson.build
```diff
@@ -129,7 +128,23 @@ if not bwrap.found()
     ''')
 endif
 
-have_libportal = libportal_dep.found()
+if get_option('libportal-subproject')
```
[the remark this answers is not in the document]
tristan957: From discussion on Matrix, you can set `force_fallback_for` as a default project option, which can then be overridden by packagers.

tristan957 (approved):

meson.build
```diff
@@ -4,13 +4,18 @@ project(
...
 
 ###### various directories we'll use later
 # foodir are built-in ones, foo_dir are our options
 
 prefix = get_option('prefix')
 datadir = prefix / get_option('datadir')
+libdir = prefix / get_option('libdir')
 libexecdir = prefix / get_option('libexecdir')
 sysconfdir = prefix / get_option('sysconfdir')
 localedir = prefix / get_option('localedir')
```
tristan957, line 21: FWIW, all this prefix / stuff is unnecessary. Meson appends the prefix automatically at time of use if the variable is not an absolute path.

swick: That doesn't seem to be the case at least in some cases. If I put `pkgconfig.generate(..., variables: {'datadir':get_option('datadir')})` for example gives me `datadir=share`.

tristan957: Good point. In that case, Meson cannot know how the value will be used. I meant more for `install_dir: get_option('localedir')`.

subprojects/libportal.wrap
whot: I'm guessing we're talking about https://github.com/flatpak/libportal/pull/147/?

Best to check in meson if `xdp_portal_get_supported_options` exists and ifdef those out

meson.build
swick: Sure, but given that we might want to use them in various places I think it's better to leave it as is.

TingPing:
Looks fine to me outside of the unrelated notification changes.

swick:
Rebased

meson.build
```diff
@@ -4,13 +4,18 @@ project(
   version: '1.19.1',
   meson_version: '>= 0.60',
   license: 'LGPL-2.0-or-later',
-  default_options: ['warning_level=2'])
+  default_options: [
+    'warning_level=2',
+    'force_fallback_for=libportal',
```
TingPing, line 9: Do we really want to force this? Seems fine to support both ways.

smcv: With Debian hat on, I'd prefer to use the packaged libportal for the tests (I already do, and I'd like to continue to do that). If the released libportal doesn't have something that the x-d-p's test suite needs, the solution should be to release libportal.

swick: This isn't really forced as it changes the default. One can compile with `-Dforce_fallback_for=[]` to use the system provided libportal.

meson.build
eli-schwartz: This is also fine because:

1. you do NOT want to use 
    ```meson
    datadir = prefix / get_option('datadir')
    pkgconfig.generate(..., variables: ['datadir='+datadir])
    ```
    since that will cause your .pc file to say "prefix=/usr; datadir=/usr/share" and you actually want `datadir_for_pcfile = '${prefix}' / get_option('datadir')`

2. pkgconfig.generate will, since 0.62.0, solve this issue automatically by allowing you to use `pkgconfig.generate(..., variables: ['foobar=${datadir}/thingy'])` and automatic references for datadir will be generated for you.

meson.build
eli-schwartz: @swick what you are suggesting is unnatural and non-idiomatic usage of meson. Nobody expects to have to do this.

Why do you even want to "change the default"? Surely changing a default should come with a rationale for why the default is bad.

swick: libportal is a dependency for the tests and the tests only. The tests are mainly a developer tool and getting the right version to developers as soon as possible is important. There are definitely packages which pass `-Dforce_fallback_for=[]` to meson already.

@eli-schwartz if you continue with this level of hostility I will block you

thesamesam: We don't pass `force_fallback_for` for any of the packages using Meson in Gentoo. This is the first time it's come up.

Can you explain what the actual need here is? Why not just set the minimum version appropriately?

swick: Got rid of `force_fallback_for=[]` from the default_options but still using it in the CI. It now also has a version requirement.

PR #1480, opened 2024-10-18
ci: Prefer repository owner image if it exists

swick:
Only fall back to the actor's image if the repository owner ones does not exist.


swick:
You can see the action started by this PR now used the image `ghcr.io/flatpak/xdg-desktop-portal:20241016-2` because it already exists.

GeorgesStavracas (approved):

PR #1484, opened 2024-10-21
Disable broken test

hfiguiere:
This test require unreleased code in a dependency. This fixes the build

### DIFF
--- a/tests/notification.c
+++ b/tests/notification.c
@@ -10,12 +10,16 @@
   "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" \
   "<svg xmlns=\"http://www.w3.org/2000/svg\" height=\"16px\" width=\"16px\"/>"
 
+
+/* reenable when the proper PR in libportal is merged */
+#if 0
 static const guchar SOUND_DATA[] = {
   0x52, 0x49, 0x46, 0x46, 0x24, 0x00, 0x00, 0x00, 0x57, 0x41, 0x56, 0x45,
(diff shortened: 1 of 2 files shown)


hfiguiere:
see https://github.com/flatpak/xdg-desktop-portal/pull/1479#discussion_r1808664882

swick:
I don't see why this would be necessary when we can just pull in the required libportal commit.

hfiguiere:
because the commit doesn't exist in the libportal repository yet. If you really want to do that, get it merged.

Using a fork is NOT a solution. It's another breakage in the making.

swick:
It works, so it is a solution.

hfiguiere:
It is fragile - what happen when the fork disappear? And it blocks any upgrade to libportal - aka it blocks other contributions. So it makes it worse.

swick:
Alright, let's go with this one.

swick:
As for why CI didn't notice: libportal is missing from the image. https://github.com/flatpak/xdg-desktop-portal/pull/1479 fixes that.

swick:
Had to ifdef out a few more things to get tests passing on current libportal main: https://github.com/flatpak/xdg-desktop-portal/pull/1479/commits/ebb8d907a2e39ef78aff17bd0e793db058c0c0ab

hfiguiere:
Updated the patch. Test passes.

swick:
LGTM

hfiguiere:
btw just ICYMI, I don't have permissions to merge, so someone needs to do this.

GeorgesStavracas (approved):

PR #1485, opened 2024-10-22
ci: Install doc dependencies via pip

bbhtt:
Fixes the current messed up CSS/theme in https://flatpak.github.io/xdg-desktop-portal/docs/index.html, deployed at https://bbhtt.github.io/xdg-desktop-portal/docs/index.html

### DIFF
--- a/.github/workflows/Containerfile
+++ b/.github/workflows/Containerfile
@@ -49,8 +49,6 @@ RUN apt install -y --no-install-recommends \
     python3-dbus
 
 # Install doc dependencies
-RUN apt install -y --no-install-recommends \
-    python3-sphinx-copybutton \
-    python3-sphinxext-opengraph \
-    furo \
-    python3-sphinx
+RUN apt install -y --no-install-recommends python3-pip
+RUN pip install --user --break-system-packages furo">=2024.04.27" \
+    sphinx-copybutton sphinxext-opengraph matplotlib
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 20241016-2
+  IMAGE_TAG: 20241022-0
 
 on:
   workflow_call:
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -27,6 +27,8 @@ jobs:
 
       - name: Build docs
         run: |
+          export PYTHONPATH="/root/.local/lib/python$(python3 -c 'import sys; print("{}.{}".format(*sys.version_info))')/site-packages:$PYTHONPATH"
+          export PATH="/root/.local/bin:$PATH"
           meson setup builddir -Ddocumentation=enabled
           ninja -C builddir doc/html
 


.github/workflows/container.yml
```diff
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 20241016-2
+  IMAGE_TAG: 20241022-2
```
swick, line 2:
```suggestion
  IMAGE_TAG: 20241022-0
```
New day, new 0 :)

.github/workflows/Containerfile
```diff
@@ -49,8 +49,6 @@ RUN apt install -y --no-install-recommends \
     python3-dbus
 
 # Install doc dependencies
-RUN apt install -y --no-install-recommends \
-    python3-sphinx-copybutton \
-    python3-sphinxext-opengraph \
-    furo \
-    python3-sphinx
+RUN apt install -y --no-install-recommends python3-pip
+RUN pip install --user --break-system-packages furo">=2024.04.27" \
```
swick, line 53: Is there a reason to install this into user instead of system? If it's installed to system we wouldn't need to adjust the PATHs in the workflow.

swick:
Thanks! Didn't even notice something was going wrong. Can we figure out why the website build step didn't fail to avoid similar mistakes in the future?

.github/workflows/Containerfile
bbhtt: I generally wanted to avoid conflicts/overlaps. If one of the Python modules gets installed via `apt` first through some dependency, `pip` will fail to install it in system.

bbhtt:
It's not on xdp's end, sphinx made some breaking changes in 7.2.0->7.2.6 and Furo needed some adjustments. It's on the distribution's end to ship compatible/working packages.

I guess you can set up RTD previews for PRs like we have in flatpak-docs.

.github/workflows/Containerfile
swick: Mh, isn't that what `--ignore-installed` is for?

bbhtt:
This was the change that broke the copy button extension https://github.com/sphinx-doc/sphinx/commit/8e730ae303ae686705ea12f44ef11da926a87cf5

swick:
Alright, thanks for digging into it. A bug in some other component isn't something we can catch in CI.

The changes LGTM in general.

.github/workflows/Containerfile
bbhtt: `--ignore-installed` is a bit slippery-slope, because it overwrites the system version with the pip's version. 

If `apt` installed `foobar 2.2.0`, the something might depend on that version and might break with pip which always fetches the latest right now under the current config.

bbhtt: which is also why I set these env vars here so that it doesn't pollute anything else. Without the env var, the packages won't be found.

swick: Alright, let's go with the current version then.

GeorgesStavracas (approved):

Issue #916, opened 2022-11-24
[Bug]: App Name is not available when using Portal Notifications


Crylia:

GeorgesStavracas:
Please be more specific. Which specific apps have you used that trigger this behavior, and what did you expect to see?

Crylia:
One example would be spotify, I would expect a notification to also contain the applications name, but it doesn't.

https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html The freedesktop spec defines a "app_name" which is not set when using any application from within flatpak

swick:
Is the app flatpak, snap or host? Which portal backend do you use?

Crylia:
It would be a flatpak, a good example would be Spotify (app/com.spotify.Client/x86_64/stable)

swick:
This is very likely an issue with a specific backend + compositor. Which ones do you use? I suspect xdg-desktop-portal-gtk and not gnoms-shell?

https://github.com/flatpak/xdg-desktop-portal-gtk/blob/366d3349041c2e06658c7c5a8b5bfd5f345d8157/src/fdonotification.c#L348

Crylia:
How do I check/change the backend? I just use awesomewm with picom

swick:
`ps ax| grep xdg-desktop-portal` might be sufficient, otherwise check which component supports the notification interface in `/usr/share/xdg-desktop-portal/portals/` or use `dbus-monitor` to figure out what gets called.

Crylia:
```
3239 ?        SNsl   0:00 /usr/libexec/xdg-desktop-portal
   3256 ?        SNsl   0:00 /usr/libexec/xdg-desktop-portal-gnome
   3265 ?        SNsl   0:00 /usr/libexec/xdg-desktop-portal-gtk
   3270 ?        SNsl   0:00 /usr/lib/x86_64-linux-gnu/libexec/xdg-desktop-portal-kde
  46431 pts/0    SN+    0:00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn --exclude-dir=.idea --exclude-dir=.tox xdg-desktop-portal
  ```
  is the output from the first command, so ig I got three running?

swick:
[...]

PR #1488, opened 2024-10-25
notification: Add the desktop-file-id key for portal backends

swick:
To fix https://github.com/flatpak/xdg-desktop-portal/issues/916, we need to get the application name from a `org.freedesktop.impl.portal.Notification.AddNotification` call which is currently not possible because we only get an app id which does not always map to a dekstop file (https://github.com/flatpak/xdg-desktop-portal-gtk/blob/366d3349041c2e06658c7c5a8b5bfd5f345d8157/src/fdonotification.c#L348).

This PR adds the `desktop-file-id` key to `AddNotification` impl calls to make it possible for backends to show an app name.


src/xdp-app-info.c
```diff
@@ -155,6 +168,18 @@ xdp_app_info_get_instance (XdpAppInfo *app_info)
   return priv->instance;
 }
 
+const char *
+xdp_app_info_get_desktop_file (XdpAppInfo *app_info)
```
jsparber, line 172: Would it be better to call it `xdp_app_info_get_desktop_file_id ()` to be consistent? Same in other locations where you use `desktop_file`

src/xdp-app-info.c
```diff
@@ -108,18 +110,29 @@ xdp_app_info_initialize (XdpAppInfo *app_info,
...
 {
   XdpAppInfoPrivate *priv = xdp_app_info_get_instance_private (app_info);
+  g_autoptr(GAppInfo) gappinfo = NULL;
+
+  if (desktop_file != NULL)
+    {
+      g_autoptr(GDesktopAppInfo) desktop_appinfo = NULL;
+
+      desktop_appinfo = g_desktop_app_info_new (desktop_file);
+      if (desktop_appinfo)
```
jsparber, line 126: do we actually need this dance of assigning the value first to `desktop_appinfo`?

src/notification.c
```diff
@@ -1077,6 +1078,10 @@ handle_add_in_thread_func (GTask        *task,
       return;
     }
 
+  desktop_file = xdp_app_info_get_desktop_file (call_data->app_info);
+  if (desktop_file)
```
jsparber, line 1082: probably we should check here whether `desktop_file == ""`.  Or maybe a follow up MR could make sure that the `desktop_file` is never set to "".

data/org.freedesktop.impl.portal.Notification.xml
```diff
@@ -47,6 +47,12 @@
 
         The format of the @notification is the same as for
         :ref:`org.freedesktop.portal.Notification.AddNotification`.
+        It additionally can contain the following keys:
+
+        * ``desktop-file-id`` (``s``)
+
+          The desktop file of the app adding the notification. It can be
```
jsparber, line 54: Best to mention that this id contains the `.desktop` prefix.

jsparber (commented):
Looks good to me.  This improves the situation but doesn't resolve the issue when a flatpak exports multiple desktop-files, but this does move in the correct direction so I'm fine with this changes.

We probably can do some clean up e.g. remove `xdp_get_app_id_from_desktop_id()` and replace fallback of "" for host apps with` NULL` .The clean ups should probably happen in a different MR.

src/xdp-app-info.c
swick: The casts don't like getting NULL and there is no guarantee that we'll be able to find the desktop file that we "detected" so I don't see how.

src/xdp-app-info.c
swick: Yup, good call

jadahl:
It was decided against this last time the same thing was proposed, as it is many places assumed the app ID resolves to a desktop file. See https://github.com/flatpak/xdg-desktop-portal/pull/904.

src/notification.c
swick: Changed it so we never get an empty string or ".desktop" as `desktop_file_id` but in general, it needs to handle not finding the file anyway so this doesn't really change anything.

data/org.freedesktop.impl.portal.Notification.xml
swick: Yup, makes sense.

jadahl:
> This is in the backend i.e. the impl only.

That is what was proposed last time too.

swick:
Sorry, but assuming that an app id maps to a desktop file is just plain wrong. Not sending the desktop file is wrong. Nothing will change for impls which currently get away with assuming the app id maps to a desktop file.

jadahl:
Resolving an app id that is derived in xdg-desktop-portal to a desktop file is assumed to work in the multiple portal backends, I don't know why you are claiming it's "plain wrong", it's how it's designed to work. It's not only a notification portal thing.

swick:
There are literally two choices here:
1. Accept that an app id and desktop files can be different and start fixing up things in the stack
2. Force app ids and desktop files to be identical

What we currently have is that desktop files and app ids can be different but the stack assumes that they are not. If we're not going for 1. then I want to know what the plan for 2. is.

jadahl:
For flatpak, the only time it doesn't work fully as expected is apps with multiple "apps" inside, where the sub-app are distinguished by a suffix, while sharing the same app id (that still resolves to a desktop file) as the prefix. But that is a different story to what is done here.

swick:
Portals are used not only by flatpak.

src/xdp-app-info.c
jsparber: oh, i didn't know that.

swick:
The implementation now more or less has adopted the fact that the app id, desktop id and permission id are different concepts. @jadahl do you still oppose the idea here?

PR #1491, opened 2024-10-28
Fix memory leaks in globalshortcuts and email portals

swick:
```
    globalshortcuts: Fix memory leaks

    email: Fix memory leaks
```

### DIFF
--- a/src/global-shortcuts.c
+++ b/src/global-shortcuts.c
@@ -233,7 +233,7 @@ handle_create_session (XdpDbusGlobalShortcuts *object,
   Request *request = request_from_invocation (invocation);
   g_autoptr(GError) error = NULL;
   g_autoptr(XdpDbusImplRequest) impl_request = NULL;
-  GVariantBuilder options_builder;
+  g_auto(GVariantBuilder) options_builder;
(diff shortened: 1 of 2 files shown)


src/global-shortcuts.c
```diff
@@ -233,7 +233,7 @@ handle_create_session (XdpDbusGlobalShortcuts *object,
   Request *request = request_from_invocation (invocation);
   g_autoptr(GError) error = NULL;
   g_autoptr(XdpDbusImplRequest) impl_request = NULL;
-  GVariantBuilder options_builder;
+  g_auto(GVariantBuilder) options_builder;
```
GeorgesStavracas, line 236: I don't see how this is a leak? In the original code, `GVariantBuilder` is stack allocated, and `g_variant_builder_end` cleans it up as expected

jadahl: Indeed, this part was unnecessary. The actual leak fix is not to steal the options pointer in the function call .

jadahl: (same applies below, some unrelated changes, should have caught that)

PR #1493, opened 2024-10-29
Backport g_autofd and g_clear_fd, cleanup GLib backports

GeorgesStavracas:
See commits. Nothing groundbreaking and shouldn't have any practical changes in behaviour.


jsparber:
lgtm

swick:
LGTM as well

src/glib-backports.h
```diff
@@ -25,6 +25,52 @@
 
 #include <gio/gio.h>
 
+#if GLIB_CHECK_VERSION(2, 76, 0)
+
+#include <glib/gstdio.h>
```
jsparber, line 30: Doesn't this need to be included always? the else branch uses `g_close`

GeorgesStavracas: Sure

document-portal/document-portal-fuse.c
```diff
@@ -2492,7 +2492,7 @@ xdp_fuse_readdir (fuse_req_t             req,
       if (off < d->dirbuf_size)
         {
           gsize reply_size = MIN (d->dirbuf_size - off, size);
-          g_autofree char *buf = g_memdup (d->dirbuf + off, reply_size);
+          g_autofree char *buf = g_memdup2 (d->dirbuf + off, reply_size);
```
hfiguiere, line 2495: This is glib >= 2.68 but we only require >= 2.66

jsparber: The first commits bumps glib to 2.72

hfiguiere: nevermind I missed the bump in version in meson.build.

hfiguiere:
if we require glib, 2.72 which long term support has it? People still run 4 year old distros...

GeorgesStavracas:
> if we require glib, 2.72 which long term support has it? People still run 4 year old distros...

We looked at Ubuntu 22.04 which ships with GLib 2.72

hfiguiere:
That's only two years old

hfiguiere:
20.04 has 2.64 which is lower than what was before...

GeorgesStavracas:
I don't think we ever had any policy to how far back the project would support old distros. I assumed we'd support current Ubuntu LTS + previous Ubuntu LTS at most. I don't think it's practical to support anything older than the previous Ubuntu LTS version.

smcv:
> People still run 4 year old distros...

I think it's fine to say that if they do, then they can run the xdg-desktop-portal version that is supported by their distro vendor (after all, the whole point of LTS distributions is that your distro vendor will support them for some reasonable lifetime); and if that version is too old for the apps and desktop environments they want to run, then they should re-evaluate their choice of distro branch.

I would like the latest x-d-p to continue to be buildable on the latest Ubuntu LTS (released once per 2 years, in April of even-numbered years) and the latest Debian stable (released approximately once per 2 years, normally in the middle of odd-numbered years), but I don't think it's necessary to go to heroic efforts to be backportable beyond that.

As a data point, Debian 12 has GLib 2.74.x, which is slightly newer than Ubuntu 22.04. If the freeze for Debian 13 happens at around the date I expect, then it will likely have GLib 2.82.x.

We don't routinely backport a new x-d-p in <https://launchpad.net/~flatpak/+archive/ubuntu/stable/+packages>, mainly because x-d-p is only useful if it is accompanied by a desktop environment backend like x-d-p-gnome (or x-d-p-kde, or whatever is appropriate for the user's environment), and those tend to be too tightly coupled to the rest of the desktop environment to be useful to backport. If x-d-p offers new features, but they don't actually work because the LTS distribution's x-d-p-gnome is the same age as the rest of the LTS distribution and therefore does not implement the backend for those features, then we gain nothing by backporting it.

PR #1503, opened 2024-11-01
Use g_auto and G_VARIANT_BUILDER_INIT for all GVariantBuilders

swick:
* Always initialize g_auto(GVariantBuilder)
* Use g_auto and G_VARIANT_BUILDER_INIT for all GVariantBuilders
    
    One has to be very careful about the control flow and early exits
    without the g_auto which can easily result in memory leaks. I noticed at
    least 5 cases where the change does fix a leak, even though it might be
    a bit hard to trigger.
    
    This also makes sure that all g_auto usages are initialized correctly.

### DIFF
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -230,14 +230,14 @@ create_session_done (GObject *source_object,
   Session *session;
   guint response = 2;
   gboolean should_close_session;
-  GVariantBuilder results_builder;
+  g_auto(GVariantBuilder) results_builder =
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
   g_autoptr(GError) error = NULL;
 
   REQUEST_AUTOLOCK (request);
 
   session = g_object_get_qdata (G_OBJECT (request), quark_request_session);
 
-  g_variant_builder_init (&results_builder, G_VARIANT_TYPE_VARDICT);
   SESSION_AUTOLOCK_UNREF (g_object_ref (session));
   g_object_set_qdata (G_OBJECT (request), quark_request_session, NULL);
 
@@ -281,10 +281,6 @@ out:
                                       g_variant_builder_end (&results_builder));
       request_unexport (request);
     }
-  else
-    {
-      g_variant_builder_clear (&results_builder);
-    }
 
   if (should_close_session)
     session_close (session, FALSE);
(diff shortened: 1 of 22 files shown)


src/remote-desktop.c
```diff
@@ -383,9 +379,9 @@ select_devices_done (GObject *source_object,
     {
       if (!results)
         {
-          GVariantBuilder results_builder;
+        g_auto(GVariantBuilder) results_builder =
+          G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
```
whot, line 383: looks like wrong indentation

whot (approved):
LGTM

src/glib-backports.h
```diff
@@ -67,3 +67,10 @@ _g_clear_fd_ignore_error (int *fd_ptr)
 
 #define g_autofd __attribute__((cleanup(_g_clear_fd_ignore_error)))
 #endif
+
+#ifndef G_VARIANT_BUILDER_INIT_UNSET
+#define G_VARIANT_BUILDER_INIT_UNSET() \
```
jsparber, line 72: for what is this needed?

jsparber (commented):
lgtm but not sure what the backport does, would be good to explain it in the commit message.

src/restore-token.c
```diff
@@ -108,7 +108,8 @@ xdp_session_persistence_set_persistent_permissions (Session *session,
                                                     GVariant *restore_data)
 {
   g_autoptr(GError) error = NULL;
-  GVariantBuilder permissions_builder;
+  g_auto(GVariantBuilder) permissions_builder =
+    G_VARIANT_BUILDER_INIT_UNSET ();
```
jsparber, line 112:
```suggestion
    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE ("a{sas}"));
```

jsparber: Don't see the point of not using ` G_VARIANT_BUILDER_INIT` in this case.

swick:
Dropped BUILDER_INIT_UNSET and its usage.

jsparber:
> Dropped BUILDER_INIT_UNSET and its usage.

the commit is still there.

swick:
Woops, my bad.

GeorgesStavracas (approved):

PR #1506, opened 2024-11-05
Fixes for GUnixFDList related issues

swick:


src/clipboard.c
```diff
@@ -168,54 +168,74 @@ handle_set_selection (XdpDbusClipboard *object,
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
-static void
```
jadahl, line 171: This commit does a lot more than fixing errors. I suggest undoing the unrelated changes, and just add what I'm guessing is the only relevant change:

```
  fd_id_in = g_variant_get_handle (fd_handle);
  if (fd_id_in < 0 || fd_id_in >= g_unix_fd_list_get_length (fd_list))
```

I'm not sure the others are that useful, I would always expect the first thing in a async I/O function callback to be the actual finish callback, instead of putting that away in a helper.

swick: Reworked this and split it into multiple commits each explaining the reason for the change.

swick:
@jadahl ping

jsparber:
lgtm

src/clipboard.c
jadahl: That's better, thanks.

PR #1507, opened 2024-11-05
Move plumbing to the Xdp prefix

swick:
This moves the plumbing code to the Xdp prefix which is required for the GObject boilerplate macros. Newer plumbing already is in that namespace so this is just porting over the remaining things.

This doesn't move files around (other than adding the `xdp-` prefix) and also doesn't include any semantic changes.


GeorgesStavracas (approved):

GeorgesStavracas:
LGTM, will merge after the rebase

swick:
Rebased, and ran `git rebase main --exec='meson compile -C build'`.

swick:
CI failed. Will investigate.

PR #1519, opened 2024-11-27
Improve host XdpAppInfo handling and force empty app id for Camera portal

swick:
* The first commit adds a missing environment variable to the tests README
* The next commit tightens the requirements for host app id's detected from the cgroup to also require a matching GAppInfo
* The third commit cleans up the camera portal function for querying permissions
* The last commit forces an empty app id for the Camera portal (essentially the same as https://github.com/flatpak/xdg-desktop-portal/pull/1512)

### DIFF
--- a/src/xdp-app-info-host.c
+++ b/src/xdp-app-info-host.c
@@ -159,11 +159,16 @@ _xdp_app_info_host_parse_app_id_from_unit_name (const char *unit)
 }
 #endif /* HAVE_LIBSYSTEMD */
 
-static char *
-get_appid_from_pid (pid_t pid)
+static gboolean
+get_app_from_pid (pid_t      pid,
+                  char     **app_id_out,
+                  GAppInfo **app_info_out)
 {
 #ifdef HAVE_LIBSYSTEMD
   g_autofree char *unit = NULL;
+  g_autofree char *desktop_id = NULL;
+  g_autofree char *app_id = NULL;
+  g_autoptr(GDesktopAppInfo) desktop_app_info = NULL;
   int res;
 
   res = sd_pid_get_user_unit (pid, &unit);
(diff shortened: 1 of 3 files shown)


swick:
Now that https://github.com/flatpak/xdg-desktop-portal/pull/1512 has been merged this PR only contains what I hope to be uncontroversial changes.

jadahl:
> This branch has conflicts that must be resolved

swick:
Rebased

src/xdp-app-info-host.c
```diff
@@ -197,6 +197,12 @@ xdp_app_info_host_new (int pid,
   desktop_id = g_strconcat (appid, ".desktop", NULL);
   gappinfo = G_APP_INFO (g_desktop_app_info_new (desktop_id));
 
+  if (!gappinfo)
+    {
+      g_clear_pointer (&desktop_id, g_free);
+      desktop_id = g_strdup ("");
```
jadahl, line 203: `desktop_id` isn't used after this, so this doesn't actually do anything.

Also this came up on Matrix; it might be useful to pass app ids to the backend in some cases where one doesn't necessarily need a .desktop file, more specifically when the desktop environment launches a service that should use portals in a specially crafted cgroup, that the backend will special case.

The real use case is running Xwayland under a cgroup giving it an app ID, and special casing its use of liboeffis (libei via remote desktop portal).

jadahl (changes requested):

src/xdp-app-info-host.c
swick: > desktop_id isn't used after this, so this doesn't actually do anything.

Yikes, that was supposed to be `appid`. Reworked this and moved the logic to `get_appid_from_pid`.

> The real use case is running Xwayland under a cgroup giving it an app ID, and special casing its use of liboeffis (libei via remote desktop portal).

I'd rather we pass the pidfd to backends which gives them the ability to look up cgroups and other kinds of special handling. Abusing the thing that's supposed to identify apps isn't a great idea imo.

jadahl: Passing pidfds could perhaps be useful, but the thing that is already there is the app ID. Perhaps split that part out, so the rest doesn't need to linger, then we can consider what to do with requiring a desktop file to have host app ID's be a thing separately.

swick: The app id has never meant the cgroup of a process and just because that kind-of practically is the case right now in some cases doesn't mean that we should keep that behavior or encourage people to build on top of that!

I'm really annoyed that any kind of push towards improving and tightening our app detection is met with resistance.

PR #1521, opened 2024-12-02
Allow host apps to register themselves

jadahl:
For host apps, we currently require applications to run inside a cgroup with a name that allows xdg-desktop-portal to derive an application ID from it. This does not work reliably, and while the situation can improve, it's unlikely to ever be fool proof, as it in theory requires every potential piece of software that launches other applications to know how to launch it in a compatible way.

Cgroups is also a Linux specific API, which is problematic from a *nix wide portability perspective.

To address this, introduce a host-only portal API that allows applications to register their D-Bus connections, associating them with an application ID. This moves back the burden of associating an application with an application ID back to the application itself, instead of the application launcher.

Applications calling this overrides any applicion ID derived from the name fo the cgroup the application runs inside.

------

I'm not saying we should do this, but perhaps it is something we should consider doing. Relying only on cgroups is a noble idea, but it hasn't worked well so far, and while it's possible that at least a majority of cases can be handled by peeking at it and deriving the application ID, it's also likely that there will always be cases where it won't work, without any way for applications to get themselves to work.

Some downsides include:

 * Giving up on the idea to always have *host* apps run in a properly named cgroup, at least for any portal purposes. It's a good end goal in itself, but hard to say how reliable it can ever be.
 * Yet again more than one way to get the same piece of information, as in we're creating another mess for ourself.

Some upsides:

 * Applications can now fix themselves. They can't realistically do that by relying on cgroups.
 * Not Linux specific. While sandbox engines probably are, it's a good end goal to have portal API's be as widely adopted as possible.

### DIFF
--- a/src/xdg-desktop-portal.c
+++ b/src/xdg-desktop-portal.c
@@ -140,7 +140,7 @@ authorize_callback (GDBusInterfaceSkeleton *interface,
   g_autoptr(XdpAppInfo) app_info = NULL;
   g_autoptr(GError) error = NULL;
 
-  app_info = xdp_invocation_lookup_app_info_sync (invocation, NULL, &error);
+  app_info = xdp_invocation_ensure_app_info_sync (invocation, NULL, &error);
   if (app_info == NULL)
     {
       g_dbus_method_invocation_return_error (invocation,
--- a/src/xdg-desktop-portal.c
+++ b/src/xdg-desktop-portal.c
@@ -60,6 +60,7 @@
 #include "print.h"
 #include "proxy-resolver.h"
 #include "realtime.h"
+#include "registry.h"
 #include "remote-desktop.h"
 #include "xdp-request.h"
 #include "screen-cast.h"
@@ -174,6 +175,33 @@ export_portal_implementation (GDBusConnection *connection,
                                        G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
   g_signal_connect (skeleton, "g-authorize-method",
                     G_CALLBACK (authorize_callback), NULL);
+
+  if (!g_dbus_interface_skeleton_export (skeleton,
+                                         connection,
+                                         DESKTOP_PORTAL_OBJECT_PATH,
+                                         &error))
+    {
+      g_warning ("Error: %s", error->message);
+      return;
+    }
+
+  g_debug ("providing portal %s", g_dbus_interface_skeleton_get_info (skeleton)->name);
+}
+
+static void
+export_host_portal_implementation (GDBusConnection        *connection,
+                                   GDBusInterfaceSkeleton *skeleton)
+{
+  g_autoptr(GError) error = NULL;
+
+  if (skeleton == NULL)
+    {
+      g_warning ("No skeleton to export");
+      return;
+    }
+
+  g_dbus_interface_skeleton_set_flags (skeleton,
+                                       G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD);
 
   if (!g_dbus_interface_skeleton_export (skeleton,
                                          connection,
@@ -372,6 +400,8 @@ on_bus_acquired (GDBusConnection *connection,
     export_portal_implementation (connection,
                                   xdp_usb_create (connection, implementation->dbus_name));
 #endif
+
+  export_host_portal_implementation (connection, registry_create (connection));
 }
 
 static void
--- a/src/xdp-app-info.c
+++ b/src/xdp-app-info.c
@@ -742,6 +742,21 @@ on_peer_died (const char *name)
 }
 
 static XdpAppInfo *
+maybe_create_test_app_info (void)
+{
+  const char *test_override_app_id;
+  const char *test_override_usb_queries;
+
+  test_override_app_id = g_getenv ("XDG_DESKTOP_PORTAL_TEST_APP_ID");
+  if (!test_override_app_id)
+    return NULL;
+
+  test_override_usb_queries = g_getenv ("XDG_DESKTOP_PORTAL_TEST_USB_QUERIES");
+  return xdp_app_info_test_new (test_override_app_id,
+                                test_override_usb_queries);
+}
+
+static XdpAppInfo *
 xdp_connection_create_app_info_sync (GDBusConnection  *connection,
                                      const char       *sender,
                                      GCancellable     *cancellable,
@@ -750,20 +765,12 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
   g_autoptr(XdpAppInfo) app_info = NULL;
   g_autofd int pidfd = -1;
   uint32_t pid;
-  const char *test_override_app_id;
-  const char *test_override_usb_queries;
   g_autoptr(GError) local_error = NULL;
 
   if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
     return NULL;
 
-  test_override_app_id = g_getenv ("XDG_DESKTOP_PORTAL_TEST_APP_ID");
-  test_override_usb_queries = g_getenv ("XDG_DESKTOP_PORTAL_TEST_USB_QUERIES");
-  if (test_override_app_id)
-    {
-      app_info = xdp_app_info_test_new (test_override_app_id,
-                                        test_override_usb_queries);
-    }
+  app_info = maybe_create_test_app_info ();
 
   if (app_info == NULL)
     app_info = xdp_app_info_flatpak_new (pid, pidfd, &local_error);
--- a/src/xdp-app-info.c
+++ b/src/xdp-app-info.c
@@ -712,6 +712,19 @@ cache_lookup_app_info_by_sender (const char *sender)
   return app_info;
 }
 
+static gboolean
+cache_has_app_info_by_sender (const char *sender)
+{
+  gboolean has_app_info = FALSE;
+
+  G_LOCK (app_infos);
+  if (app_info_by_unique_name)
+    has_app_info = !!g_hash_table_lookup (app_info_by_unique_name, sender);
+  G_UNLOCK (app_infos);
+
+  return has_app_info;
+}
+
 static void
 ensure_app_info_by_unique_name (void)
 {
@@ -753,6 +766,21 @@ maybe_create_test_app_info (void)
 
   test_override_usb_queries = g_getenv ("XDG_DESKTOP_PORTAL_TEST_USB_QUERIES");
   return xdp_app_info_test_new (test_override_app_id,
+                                test_override_usb_queries);
+}
+
+static XdpAppInfo *
+maybe_create_registered_test_app_info (const char *registered_app_id)
+{
+  const char *test_override_app_id;
+  const char *test_override_usb_queries;
+
+  test_override_app_id = g_getenv ("XDG_DESKTOP_PORTAL_TEST_APP_ID");
+  if (!test_override_app_id)
+    return NULL;
+
+  test_override_usb_queries = g_getenv ("XDG_DESKTOP_PORTAL_TEST_USB_QUERIES");
+  return xdp_app_info_test_new (registered_app_id,
                                 test_override_usb_queries);
 }
 
@@ -806,6 +834,58 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
   return g_steal_pointer (&app_info);
 }
 
+static XdpAppInfo *
+xdp_connection_create_host_app_info_sync (GDBusConnection  *connection,
+                                          const char       *sender,
+                                          const char       *app_id,
+                                          GCancellable     *cancellable,
+                                          GError          **error)
+{
+  g_autoptr(XdpAppInfo) app_info = NULL;
+  g_autofd int pidfd = -1;
+  uint32_t pid;
+
+  if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
+    return NULL;
+
+  app_info = maybe_create_registered_test_app_info (app_id);
+
+  if (!app_info)
+    {
+      gboolean is_sandboxed = FALSE;
+
+      if (!xdp_is_flatpak (pid, &is_sandboxed, error))
+        return NULL;
+
+      if (is_sandboxed)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Can't manually register a flatpak application");
+          return NULL;
+        }
+
+      if (!xdp_is_snap (pid, &is_sandboxed, error))
+        return NULL;
+
+      if (is_sandboxed)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Can't manually register a snap application");
+          return NULL;
+        }
+
+      app_info = xdp_app_info_host_new_registered (pidfd, app_id, error);
+      if (!app_info)
+        return NULL;
+    }
+
+  cache_insert_app_info (sender, app_info);
+
+  xdp_connection_track_name_owners (connection, on_peer_died);
+
+  return g_steal_pointer (&app_info);
+}
+
 XdpAppInfo *
 xdp_invocation_ensure_app_info_sync (GDBusMethodInvocation  *invocation,
                                      GCancellable           *cancellable,
@@ -823,4 +903,27 @@ xdp_invocation_ensure_app_info_sync (GDBusMethodInvocation  *invocation,
                                               sender,
                                               cancellable,
                                               error);
+}
+
+XdpAppInfo *
+xdp_invocation_register_host_app_info_sync (GDBusMethodInvocation  *invocation,
+                                            const char             *app_id,
+                                            GCancellable           *cancellable,
+                                            GError                **error)
+{
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  const char *sender = g_dbus_method_invocation_get_sender (invocation);
+
+  if (cache_has_app_info_by_sender (sender))
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Connection already associated with an application ID");
+      return NULL;
+    }
+
+  return xdp_connection_create_host_app_info_sync (connection,
+                                                   sender,
+                                                   app_id,
+                                                   cancellable,
+                                                   error);
 }
--- a/src/xdp-app-info.c
+++ b/src/xdp-app-info.c
@@ -794,14 +794,21 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
   g_autofd int pidfd = -1;
   uint32_t pid;
   g_autoptr(GError) local_error = NULL;
+  const char *app_info_kind = NULL;
 
   if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
     return NULL;
 
   app_info = maybe_create_test_app_info ();
+  if (app_info)
+    app_info_kind = "test";
 
   if (app_info == NULL)
-    app_info = xdp_app_info_flatpak_new (pid, pidfd, &local_error);
+    {
+      app_info = xdp_app_info_flatpak_new (pid, pidfd, &local_error);
+      if (app_info)
+        app_info_kind = "flatpak";
+    }
 
   if (!app_info && !g_error_matches (local_error, XDP_APP_INFO_ERROR,
                                      XDP_APP_INFO_ERROR_WRONG_APP_KIND))
@@ -812,7 +819,11 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
   g_clear_error (&local_error);
 
   if (app_info == NULL)
-    app_info = xdp_app_info_snap_new (pid, pidfd, &local_error);
+    {
+      app_info = xdp_app_info_snap_new (pid, pidfd, &local_error);
+      if (app_info)
+        app_info_kind = "snap";
+    }
 
   if (!app_info && !g_error_matches (local_error, XDP_APP_INFO_ERROR,
                                      XDP_APP_INFO_ERROR_WRONG_APP_KIND))
@@ -823,9 +834,14 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
   g_clear_error (&local_error);
 
   if (app_info == NULL)
-    app_info = xdp_app_info_host_new (pid, pidfd);
+    {
+      app_info = xdp_app_info_host_new (pid, pidfd);
+      app_info_kind = "derived host";
+    }
 
   g_return_val_if_fail (app_info != NULL, NULL);
+
+  g_debug ("Adding %s app '%s'", app_info_kind, xdp_app_info_get_id (app_info));
 
   cache_insert_app_info (sender, app_info);
 
@@ -878,6 +894,8 @@ xdp_connection_create_host_app_info_sync (GDBusConnection  *connection,
       if (!app_info)
         return NULL;
     }
+
+  g_debug ("Adding registered host app '%s'", xdp_app_info_get_id (app_info));
 
   cache_insert_app_info (sender, app_info);
 
--- a/src/xdp-app-info.c
+++ b/src/xdp-app-info.c
@@ -742,7 +742,7 @@ on_peer_died (const char *name)
 }
 
 static XdpAppInfo *
-xdp_connection_lookup_app_info_sync (GDBusConnection  *connection,
+xdp_connection_create_app_info_sync (GDBusConnection  *connection,
                                      const char       *sender,
                                      GCancellable     *cancellable,
                                      GError          **error)
@@ -753,10 +753,6 @@ xdp_connection_lookup_app_info_sync (GDBusConnection  *connection,
   const char *test_override_app_id;
   const char *test_override_usb_queries;
   g_autoptr(GError) local_error = NULL;
-
-  app_info = cache_lookup_app_info_by_sender (sender);
-  if (app_info)
-    return g_steal_pointer (&app_info);
 
   if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
     return NULL;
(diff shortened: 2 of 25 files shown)


ebassi:
> Applications can now fix themselves. They can't realistically do that by relying on cgroups.

As much as I admire your optimism, I have a pretty hard time believing apps will "fix themselves" by calling a D-Bus API (or using libportal) for the unsandboxed case. You're still asking application developers to deal with this case. Might as well ask them to call:

```c
setenv ("XDP_PROP_application_id", "com.example.App");
```

in their `main()`, like Pulseaudio requires, and then have every single portal-aware library check for that environment variable.

swick:
One could equally argue that apps only have to call `StartTransientUnit("app-none-{appid}", pid())` in its main but that also hasn't happened so far...

jadahl:
> As much as I admire your optimism, I have a pretty hard time believing apps will "fix themselves" by calling a D-Bus API (or using libportal) for the unsandboxed case. You're still asking application developers to deal with this case. Might as well ask them to call:

For apps that wants to support running non-sandboxed that uses libportal or ashpd could perhaps use API like `xdp_portal_set_fallback_app_id()`. No need to go via env vars for that.

For apps that interact directly with the portal using D-Bus it seems like a trivial thing to make just another method call before anything else.

> One could equally argue that apps only have to call StartTransientUnit("app-none-{appid}", pid()) in its main but that also hasn't happened so far...

Using `StartTransientUnit()` is a bit problematic since it might already be in a proper unit, not to mention it's systemd specific, which might hinder adoption of portal APIs as "universal". I also imagine app developers are more hesitant to start to become self-service managers like that, compared to just passing a bit of metadata.

swick:
> Using StartTransientUnit() is a bit problematic since it might already be in a proper unit

One could check that and skip `StartTransientUnit()`.

Anyway, I do agree that calling one more portal helper function or calling one more dbus method when one wants to interact with portals shouldn't be much of an issue.

hsitter:
@Sodivad and I were musing on this a bit from the KDE side of things. It generally seems like a good idea, so +1 from us

A getter might be handy for the unforeseen scenario where the app (e.g. inside a library) wants to know its own app id.

It'd be good to document whether the registration is per-pid or per-connection (e.g. say I open multiple dbus connections in my app - do I want to make sure that all of them are registered with the portal or is one sufficient so the pid is associated?)

Maybe also document what happens when Register is called more than once (possibly with conflicting information?).

jadahl:
> A getter might be handy for the unforeseen scenario where the app (e.g. inside a library) wants to know its own app id.

Seems like a harmless thing to add. Apps must know it's only a host-app thing though, since this API namespace won't be available in sandboxes.

> It'd be good to document whether the registration is per-pid or per-connection (e.g. say I open multiple dbus connections in my app - do I want to make sure that all of them are registered with the portal or is one sufficient so the pid is associated?)

That is a good question. Making it per pid makes sense unless apps expect to be able to have different hats from the same process. Seems unlikely, and simply impossible would they run in a sandbox.

Sodivad:
> That is a good question. Making it per pid makes sense unless apps expect to be able to have different hats from the same process. Seems unlikely, and simply impossible would they run in a sandbox.

The famous one would be Libre Office where afaiu everything is a single process even windows can appear as differnt apps writer, calc, impress..

jadahl:
> The famous one would be Libre Office where afaiu everything is a single process even windows can appear as differnt apps writer, calc, impress..

Then perhaps for consistency, it should still be per pid here still, so that it doesn't work all too differently running LibreOffice inside or outside of a sandbox, to avoid both developer and user confusion.

hsitter:
Sounds reasonable :+1:

checkraisefold:
This would solve the massive headache of launching apps from terminal windows giving incorrect app IDs and absolutely requiring either a wrapper to execute an app or launching the .desktop through a graphical interface

When working on global shortcuts portal adoption, it's not been great running an app from a terminal window, or as a child process of something, and then it has a blank appid so the portal backend either doesn't register your shortcuts or it generates a random appid based on the session token. In KDE specifically, you will then have a random token as the app name in the system settings, which is pretty bad for UX. It turns out most apps don't run their children in [spec conformant](https://systemd.io/DESKTOP_ENVIRONMENTS/) systemd units

I have a feeling though that the automatic app id registration should be kept as a fallback in case an app doesn't register an app id. App ID being guaranteed to match a desktop entry by default, or being empty, (or flatpak/snap) is useful for fetching display names from the .desktop and other info automagically, which is now done in xdp-kde (see https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/341). But an app not being able to register its own app id is a much bigger headache than not being able to fetch a display name.

jadahl:
Pushed an implementation, but now noticed we had discussed pid vs peer:

> Then perhaps for consistency, it should still be per pid here still, so that it doesn't work all too differently running LibreOffice inside or outside of a sandbox, to avoid both developer and user confusion.

It's not yet doing that, and doing so will already address issues I ran into when updating libportal's qt examples, where the app ID for connections from Qt resolve differently to the libportal connection.

src/xdp-app-info.c
```diff
@@ -806,6 +834,58 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
...
+  uint32_t pid;
+
+  if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
+    return NULL;
+
+  app_info = maybe_create_registered_test_app_info (app_id);
+
+  if (!app_info)
+    {
+      gboolean is_sandboxed = FALSE;
```
swick, line 870: Maybe instead of having this duplication we could call `xdp_connection_create_app_info_sync` and `xdp_app_info_is_host`.

src/xdg-desktop-portal.c
```diff
@@ -187,6 +188,33 @@ export_portal_implementation (GDBusConnection *connection,
   g_debug ("providing portal %s", g_dbus_interface_skeleton_get_info (skeleton)->name);
 }
 
+static void
+export_host_portal_implementation (GDBusConnection        *connection,
```
swick, line 192: It's a bit unfortunate that this function doesn't by itself make sure that only host apps can call interfaces exported by it. The burden is currently completely on the implementation.

src/xdp-app-info.c
```diff
@@ -742,35 +755,60 @@ on_peer_died (const char *name)
...
+  app_info = maybe_create_test_app_info ();
+  if (app_info)
+    app_info_kind = "test";
 
   if (app_info == NULL)
-    app_info = xdp_app_info_flatpak_new (pid, pidfd, &local_error);
+    {
+      app_info = xdp_app_info_flatpak_new (pid, pidfd, &local_error);
+      if (app_info)
+        app_info_kind = "flatpak";
```
swick, line 810:
```suggestion
      app_info_kind = "flatpak";
```

tests/__init__.py
```diff
@@ -91,18 +91,23 @@ def get_mock_iface(bus: dbus.Bus, bus_name: Optional[str] = None):
     return dbus.Interface(obj, dbusmock.MOCK_IFACE)
 
 
-def portal_interface_name(portal_name) -> str:
+def portal_interface_name(portal_name, domain: Optional[str] = None) -> str:
```
swick, line 94: `portal_name` should probably also have a type annotation

src/xdp-app-info.c
jadahl: Creating `app_info` might have failed, and don't want to set it if it did, even if it'd be overwritten eventually.

src/xdp-app-info.c
jadahl: I did that in an earlier iteration, but changed to this approach to not have this code path go through potential creation of non-host apps. I think this ended up nicer.

src/xdp-app-info.c
swick: The other ones do exactly that though, so this isn't consistent.

src/xdp-app-info.c
swick: Why is that undesirable? Creating XdpAppInfo's is side-effect free.

jadahl: In my opinion, trying to fully fledged objects, when there were no intention to use them, since only the condition for them to be created was of interest, wasn't nice. The relevant condition also became somewhat hidden. Thus I ditched complicating `create_app_info_sync()` and made the paths separate.

src/xdp-app-info.c
jadahl: What other ones do you mean?

src/xdp-app-info.c
swick:
```diff
@@ -867,25 +866,19 @@ xdp_connection_create_host_app_info_sync (GDBusConnection  *connection,
 
   if (!app_info)
     {
-      gboolean is_sandboxed = FALSE;
-
-      if (!xdp_is_flatpak (pid, &is_sandboxed, error))
-        return NULL;
-
-      if (is_sandboxed)
-        {
-          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                       "Can't manually register a flatpak application");
-          return NULL;
-        }
+      g_autoptr(XdpAppInfo) detected_app_info = NULL;
 
-      if (!xdp_is_snap (pid, &is_sandboxed, error))
+      detected_app_info = xdp_connection_create_app_info_sync (connection,
+                                                               sender,
+                                                               cancellable,
+                                                               error);
+      if (!detected_app_info)
         return NULL;
 
-      if (is_sandboxed)
+      if (!xdp_app_info_is_host (detected_app_info))
         {
           g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                       "Can't manually register a snap application");
+                       "Can't manually register non-host applications");
           return NULL;
         }
 
```
What's the problem with this? It exactly tells the reader the condition under which it is not allowed while also abstracting away any engine specifics.

src/xdp-app-info.c
swick: `app_info_kind = "snap";` and `app_info_kind = "derived host";` below

jadahl: Ah, indeed. The intention was to make those who conditionally work conditionally set it, but seems I did the wrong thing for the snap case.

tests/test_registry.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import tests as xdp
+
+import dbus
+import pytest
+import os
```
swick, line 7: Unused. A two more details which ruff check and ruff format will fix.

src/xdp-app-info.c
swick: (applied the rest of the review changes https://github.com/swick/xdg-desktop-portal/tree/refs/heads/wip/host-portal-registry)

swick: Sorry, but you're creating more code to make this piece here more error prone than it needs to be.

jadahl: Just a different opinions here I guess, you say "create dummy objects" which I found awkward because one use API in a way it isn't intended to, and I say "don't create dummy objects" which you find awkward because things become a bit more open coded.

src/xdp-app-info.c
jadahl: On the other hand, not conditioning them ends up with intermediate incorrect state, which I found misleading.

src/xdp-app-info.c
swick: It's not a dummy object, it's the XdpAppInfo that results from auto-detection and you ask it the question you actually want an answer to which is "is this connection from a host" instead of "is this connection a flatpak or snap".

A6GibKm:
Would it make sense to do this in GApplication (in addition to libportal/ashpd) if the application-id property is set?

jadahl:
> Would it make sense to do this in GApplication (in addition to libportal/ashpd) if the application-id property is set?

Perhaps, I wonder if it should be possible to calling Register() multiple times with the same application ID to make this easier, or whether apps should know whether Q/GApplication does it for them. Pushed a branch that allows that here: https://github.com/jadahl/xdg-desktop-portal/commits/wip/redundant-register

tests/test_registry.py
```diff
@@ -0,0 +1,105 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
A6GibKm, line 1: It would be good to to have a test to demonstrate the expected behavior when calling `Register` twice.

tests/test_registry.py
```diff
@@ -103,3 +103,17 @@ def test_multiple_connections(self, portals, dbus_con):
...
+        mock_intf = xdp.get_mock_iface(dbus_con)
+
+        expected_app_id = "org.example.CorrectAppId"
+
+        registry_intf.Register(expected_app_id, {})
+        session = self.create_dummy_session(dbus_con)
+        app_id = mock_intf.GetSessionAppId(session.handle)
+        assert app_id == expected_app_id
+
+        with pytest.raises(dbus.exceptions.DBusException):
```
swick, line 118: All of  those should probably test against a more specific error. a timeout for example shouldn't make this succeed.

tests/test_registry.py
jadahl: Added one. It tests the opposite of what https://github.com/jadahl/xdg-desktop-portal/commits/wip/redundant-register expects.

tests/test_registry.py
jadahl: Indeed, they now do that.

swick:
Not a fan is tracking things per PID. Connections can move around between PIDs via exec and fd-passing. Per-connection is great because it's so simple and nothing can go wrong.

ebassi:
> Not a fan is tracking things per PID. Connections can move around between PIDs via exec and fd-passing. Per-connection is great because it's so simple and nothing can go wrong.

Is this an actual case, or is it just a last minute idea?

"Not a fan" is not a substantive objection: there are plenty of things we all are "not a fan" of; a real world example where this approach fails would, instead, be a much more relevant thing to bring up.

swick:
> "Not a fan" is not a substantive objection: there are plenty of things we all are "not a fan" of; a real world example where this approach fails would, instead, be a much more relevant thing to bring up.

Alright, let me rephrase that. If things are tracker per-PID then you now have a global resource (the app id) that needs to be synchronized between possibly multiple components (e.g. different libraries). For example, what happens after exec when the PID changes? What happens when gtk sets the app id and then ashpd tries to set the app id? What happens when the connection is shared between two processes and each has multiple components trying to set the app id independent from the other?

This is much harder to reason about, and because of that I'd like to keep the scope to a minimum, release that, see how the ecosystem is coping with it, and then maybe reconsider if we need to take on more complexity.

nicolasfella:
> and it would be nice to get some feedback from Gtk/Qt folks and potentially even patches already :)

We have a use case for this in KDE Connect, so I went and implemented it there: https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/779

Patching Qt to always send the app's desktop file name to this would be easy, but I'm less sure it's needed or a good idea

Sodivad:
> > and it would be nice to get some feedback from Gtk/Qt folks and potentially even patches already :)
> 
> We have a use case for this in KDE Connect, so I went and implemented it there: https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/779
> 
> Patching Qt to always send the app's desktop file name to this would be easy, but I'm less sure it's needed or a good idea

I think it should because it will make at least one call to the portal when unsandboxed, also QMultimedia for screencast...

Sodivad:
Regardless of the conclusion if multiple calls are allowed or not, should the behavior be documented or left as implementation detail

jadahl:
I think in practice, the decision to make is:

Should it be possible to have `[Gtk/QtGui)Application` to set the app ID and then use portal APIs (via D-Bus or via libportal) without explicitly registering again? Then it needs to be per pid. It's a bit awkward because it's unclear who does the registration, and has the "what is a pid" question.

If each D-Bus connection should need to register themselves (e.g. both have `[QGui/Gtk]Application` register, and doing something like `xdp_portal_regiser()` for the portal instance, then it should be per connection. This is a bit awkward because it makes apps have a bit more split brain - the GDBus connection and zbus, or QtGuiApplication and libportal, can have different app IDs even though it's the same app.

>  If things are tracker per-PID then you now have a global resource (the app id) that needs to be synchronized between possibly multiple components (e.g. different libraries).

Is that a problem? That is already more or less the case for Flatpak - only that the app ID is external.

> For example, what happens after exec when the PID changes?

Can you expand what actual scenario this is about?

> What happens when gtk sets the app id and then ashpd tries to set the app id?

The implementation fails the latter call, but I think it'd make sense to let it succeed if it was identical, so applications can register, and not start to fail if the toolkit started to integrate. This was not in the implementation, but I did add such semantics in a branch: https://github.com/jadahl/xdg-desktop-portal/commits/wip/redundant-register

In a way, getting this out of the way with specified semantics is to some degree more sensible to me than the alternative, which is the same app, with different connections, being potentially and unknowingly treated as different apps. Would the tracking be per pid it'd at least be discoverable with a the register call resulting in an error. 

> What happens when the connection is shared between two processes and each has multiple components trying to set the app id independent from the other?

I think what would make sense here is for the first one to have effect, and any conflicting later call to result in an error. That avoids race conditions, and is arguably less confusing, since there is only one answer to "what app ID is this app using?".

swick:
> Is that a problem? That is already more or less the case for Flatpak - only that the app ID is external.

For Flatpak the app ID is constant because everything that it can exec will end up with the same app ID. With PID tracking, any sort of exec will result in loosing the app ID.

>  Can you expand what actual scenario this is about?

Let's say that you have a gtk app, you use portal APIs, then exec a helper process for some reason (let's say to drop some privileges) and you use the portal APIs from there as well.

If the gtk app registered the app ID with the portal, then the usage in the first process might use another app ID as in the second process. The worst part is that it's not obvious why.

Sodivad:
> Either way, I'd be fine with making it per connection, but it effectively makes it impossible to rely on `GtkApplication`, `QtGuiApplication` or up front portal call to do anything automatically for the app process. If people think that is no big deal, then at last going per connection is a start.

For Qt it's possible and easy to use the same connection that the app uses. And is the common pattern when interacting with dbus when writing a Qt Application/library.  Of course that doesnt work as straightforward if you use some random library that makes it own connection

smcv:
> Losing the app ID on fork+exec must be the semantics otherwise one app launching another, which is partly why we need this to begin with, still wouldn't work.

I think @swick's concern is about the situation where one app is a wrapper/launcher for another, and calls `execve()` to *replace itself* with the wrapped app, *without* forking first. Like this pseudocode:

```
#!/usr/bin/hypothetical

show gui {
    label "Are you sure you want to run Chrome? You do realise it's proprietary?"
    button "Yes" -> {
        sys.execve("/opt/bin/chrome", ["chrome"] + argv[1:], environ)
    }
    button "Maybe not" -> {
        exit(0);
    }
}
```

If that happens, the process ID of the wrapper is reused as the process ID of the wrapped program - it is in some sense "the same process", but the stack, heap, data and executable code of the wrapper has been deleted from memory and replaced by the wrapped program.

jadahl:
> For Qt it's possible and easy to use the same connection that the app uses.

Even when the app uses libportal, thus GDBus?

jadahl:
> Like this pseudocode:

True, e.g. a gtk app that does exec(/somethingelse) without forking, would inherit the original gtk app's app id here, that would be problematic. Perhaps very unlikely, but a corner case it'd be hard to crawl out of.

jadahl:
The per pid tracking was removed from this branch now. That means every D-Bus connection needs to be registered, hopefully with the same app ID within the same process. There is no checking that this actually is the case.

Unless Qt uses GDBus for D-Bus IPC, this means Qt applications must both make sure QGuiApplication can register its connection, as well as manually registering using libportal API if it uses that.

While GtkApplication can help registering on behalf of libportal since they share the same D-Bus connection, ashpd users, or e.g. python GTK applications using python-dbus, must also register separately, hopefully with the same app ID, within the same application.

tests/__init__.py
```diff
@@ -195,14 +200,16 @@ def get_xdp_dbus_object(bus: dbus.Bus) -> dbus.proxies.ProxyObject:
     return obj
 
 
-def check_version(bus: dbus.Bus, portal_name: str, expected_version: int):
+def check_version(
```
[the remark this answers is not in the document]
swick: This is also a bit out of place in this commit.

tests/test_registry.py
```diff
@@ -0,0 +1,87 @@
...
+        request = xdp.Request(dbus_con, remotedesktop_intf)
+
+        session_counter_attr_name = "session_counter"
+        if hasattr(self, session_counter_attr_name):
+            session_counter = getattr(self, session_counter_attr_name)
+        else:
+            session_counter = 0
+        setattr(self, session_counter_attr_name, session_counter + 1)
+
+        print(f"session_handle_token: {session_counter}")
```
swick, line 34: use a logger instead (or remove it): `logger = xdp.init_logger(__name__)`

tests/test_registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def create_dummy_session(self, dbus_con):
+        remotedesktop_intf = xdp.get_portal_iface(dbus_con, "RemoteDesktop")
+        request = xdp.Request(dbus_con, remotedesktop_intf)
+
+        session_counter_attr_name = "session_counter"
+        if hasattr(self, session_counter_attr_name):
+            session_counter = getattr(self, session_counter_attr_name)
+        else:
+            session_counter = 0
```
swick, line 31: `session_counter = getattr(self, session_counter_attr_name, 0)`

tests/test_registry.py
```diff
@@ -0,0 +1,87 @@
...
+            options=options,
+        )
+        assert response.response == 0
+
+        return xdp.Session.from_response(dbus_con, response)
+
+    def test_registerless(self, portals, dbus_con):
+        mock_intf = xdp.get_mock_iface(dbus_con)
+
+        expected_app_id = "org.example.WrongAppId"
```
swick, line 49: This is the fixture app id, so you should use it.

swick:
I think the big question left is if we want to allow registering with the same app id multiple times without raising an error. I don't really have a preference now that things are per-connection again. Both are sensible IMO.

The code also looks fine except some nit picks.

jadahl:
I have no opinion on error vs success on registering with the same ID as well.

GeorgesStavracas:
GLib implementation: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4465

swick:
> I have no opinion on error vs success on registering with the same ID as well.

Alright, let's go with the current implementation then. We could relax this later on.

mcatanzaro:
What happens if xdg-desktop-portal isn't running yet when the application attempts to register itself, or if xdg-desktop-portarl crashes, or if it gets manually restarted by running `xdg-desktop-portal -r`? I think it is going to forget about apps that have previously registered themselves, right?  Will we wind up with apps sometimes falling back to permissions for the empty string app instead of using the permissions for its app ID?

pwithnall:
This looks a lot like it duplicates the session management protocol: `GtkApplication` already calls `RegisterClient` with the application’s app ID [here](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkapplication-dbus.c?ref_type=heads#L419).

It would be great if we could avoid another sync call in the startup path for every application, given that information is already available to the session manager.

ebassi:
> This looks a lot like it duplicates the session management protocol: `GtkApplication` already calls `RegisterClient` with the application’s app ID [here](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkapplication-dbus.c?ref_type=heads#L419).

There is no session management shared interface: it's only available on GNOME and Xfce. We might end up with a [Wayland protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/18) for session management, but right now it's still an open issue if the session object ought to be tied to an application id.

Portals are supposed to work in multiple environments, and not all of them have a session manager.

jadahl:
> I think it is going to forget about apps that have previously registered themselves, right?

They indeed need to re-register, ideally by listening to the portal name coming back. Something libportal ideally should handle itself.

> It would be great if we could avoid another sync call in the startup path for every application, given that information (app ID, app’s D-Bus connection name) is already available to the session manager.

It doesn't need to be sync, does it? The app can continue the same no matter if the call succeeded or not.

mcatanzaro:
> They indeed need to re-register, ideally by listening to the portal name coming back. Something libportal ideally should handle itself.

OK, that should work, but it should be documented more clearly. Plenty of applications talk to portals directly without using libportal, and developers won't know to implement this manually if we don't draw some attention to it.

pwithnall:
> It would be great if we could avoid another sync call in the startup path for every application, given that information (app ID, app’s D-Bus connection name) is already available to the session manager.

Georges asked me to comment further on this, and I think I’ve settled my thoughts. So:
 - I still believe this should be implemented by gnome-session on GNOME/Xfce, i.e. via `xdg-desktop-portal-gnome` interfacing with `gnome-session`.
 - On other platforms, something like this host portal registry will be needed.
 - In particular, gnome-session continues to be needed for non-flatpak apps, and it’s what we’re (still, slowly) hanging session save/restore support off, so it’s not going to go away any time soon.
 - It already has a mapping between app and D-Bus connection.
 - It seems really unfortunate that this feature has got so far into development without gnome-session being explored. That leads us to a sunk cost situation where nobody wins. I’m not a xdg-desktop-portal developer so I don’t really have any sway here.
 - However, I will not merge changes into GLib which make a second session registration call in `GApplication`.
 - Running two parallel session management protocols sounds like a great way to end up in corner case situations (what if one of them times out, or one succeeds but the other one fails, why should we pay the cost of two D-Bus calls on _every_ application startup)
 - So if an implementation for this protocol is to land in GLib, it will either need to replace the gnome-session protocol (I don’t see how that’s possible), or only be used if the gnome-session protocol is not used

pwithnall:
> > I still believe this should be implemented by gnome-session on GNOME/Xfce, i.e. via xdg-desktop-portal-gnome interfacing with gnome-session.
> > On other platforms, something like this host portal registry will be needed.
> 
> So we need this after all.

Yes, I said that.

> > It seems really unfortunate that this feature has got so far into development without gnome-session being explored. That leads us to a sunk cost situation where nobody wins. I’m not a xdg-desktop-portal developer so I don’t really have any sway here
> 
> xdg-desktop-portal is desktop agnostic why should it explore an interface that only gnome implements. And as you said other platforms would need this any way.

`xdg-desktop-portal-gnome` is not desktop agnostic. aiui the portal APIs have always been designed to allow for different desktop-specific implementations.

> > However, I will not merge changes into GLib which make a second session registration call in GApplication.
> > Running two parallel session management protocols sounds like a great way to end up in corner case situations (what if one of them times out, or one succeeds but the other one fails, why should we pay the cost of two D-Bus calls on every application startup)
> 
> This is not a session managment protocol.

aiui it’s about associating an app’s D-Bus **session** bus connection with the app ID and cgroup. That counts as session management, even if it doesn’t do all the same things as other session management protocols.

> If you worry about sync dbus calls on startup, it doesn't need to by sync and can be async just fine (that's what my Qt impl does)

Perhaps I shouldn’t have emphasised sync. Sync calls on startup are terrible, but async calls still raise all the questions of what to do if they fail, time out, etc. IPC is not simple, doing less of it is always better.

A6GibKm:
> Nothing. This whole registry is band aid.

I think the bigger issue would be potential races. In snapshots we call the camera portal asap and this is yet another thing that might be not be done before we need the app id.

swick:
Then you can call it yourself and let the gtk/glib one potentially fail.

swick:
It bears repeating: this is band-aid. It's a tool for *apps* to work around the system not putting them into an appropriate cgroup.

I honestly don't care if we put the band-aid into gtk/glib or not. What we really should put there is some code that makes sure we spawn apps into the right cgroup.

mcatanzaro:
My concern with the registry is it's being presented as a temporary band-aid solution, which doesn't make sense to me at all since creating a new D-Bus API seems like a very permanent solution.

But I have no objection to just keeping this registry API around forever. It works and it solves the problem, right? Why not?

PR #1525, opened 2024-12-04
Install pytest based tests

swick:
I personally don't really understand the value of installed tests, so this is more of an question if it makes sense to do it.  The plan is to remove the C based tests soon which would mean we'd have no more installed tests.

Only the last commit is relevant.

/cc @smcv

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -278,7 +278,7 @@ enable_pytest = get_option('pytest') \
   .allowed()
 
 if enable_pytest
-  pytest_args = [meson.current_source_dir(), '--verbose', '--log-level=DEBUG']
+  pytest_args = ['--verbose', '--log-level=DEBUG']
 
   pytest_env = environment()
   pytest_env.set('XDG_DESKTOP_PORTAL_PATH', xdg_desktop_portal.full_path())
@@ -318,17 +318,89 @@ if enable_pytest
     'test_usb.py',
     'test_wallpaper.py',
   ]
+
+  template_files = [
+    'templates/access.py',
+    'templates/account.py',
+    'templates/appchooser.py',
+    'templates/background.py',
+    'templates/clipboard.py',
+    'templates/email.py',
+    'templates/filechooser.py',
+    'templates/geoclue2.py',
+    'templates/globalshortcuts.py',
+    'templates/inhibit.py',
+    'templates/__init__.py',
+    'templates/inputcapture.py',
+    'templates/lockdown.py',
+    'templates/notification.py',
+    'templates/print.py',
+    'templates/remotedesktop.py',
+    'templates/screenshot.py',
+    'templates/settings.py',
+    'templates/usb.py',
+    'templates/wallpaper.py',
+  ]
+
   foreach pytest_file : pytest_files
     testname = pytest_file.replace('.py', '')
     test(
       'pytest/@0@'.format(testname),
       pytest,
-      args: pytest_args + ['-k', testname],
+      args: [meson.current_source_dir()] + pytest_args + ['-k', testname],
       env: pytest_env,
       suite: ['pytest'],
       timeout: 120,
     )
   endforeach
+
+  if enable_installed_tests
+    install_data(
+      pytest_files,
+      '__init__.py',
+      'conftest.py',
+      'asan.suppression',
+      install_dir: installed_tests_dir / 'tests',
+    )
+    install_data(
+      template_files,
+      install_dir: installed_tests_dir / 'tests' / 'templates',
+    )
+
+    installed_env = {
+      'XDG_DESKTOP_PORTAL_PATH': libexecdir / 'xdg-desktop-portal',
+      'XDG_PERMISSION_STORE_PATH': libexecdir / 'xdg-permission-store',
+      'XDG_DOCUMENT_PORTAL_PATH': libexecdir / 'xdg-document-portal',
+      'XDP_VALIDATE_AUTO': '1',
+    }
+    env = ''
+    foreach key, value : installed_env
+      env += f'@key@=@value@ '
+    endforeach
+
+    foreach pytest_file : pytest_files
+      testname = pytest_file.replace('.py', '')
+
+      exec = [pytest.full_path(), installed_tests_dir / 'tests']
+      exec += pytest_args
+      exec += ['-p', 'no:cacheprovider']
+      exec += ['-k', testname]
+      exec = ' '.join(exec)
+
+      data = configuration_data()
+      data.set('exec', exec)
+      data.set('env', env)
+      data.set('libdir', libdir)
+
+      configure_file(
+        input: 'template.test.in',
+        output: '@0@.test'.format(testname),
+        configuration: data,
+        install: true,
+        install_dir: installed_tests_data_dir,
+      )
+    endforeach
+  endif
 endif
 
 if enable_installed_tests
@@ -339,6 +411,9 @@ if enable_installed_tests
     install_dir: installed_tests_dir / 'services',
   )
 
+  installed_tests_env = 'XDG_DATA_DIRS=@0@/share:@1@:/usr/local/share:/usr/share'
+  installed_tests_env = installed_tests_env.format(installed_tests_data_dir, datadir)
+
   testfiles = [
     'testdb',
     'test-doc-portal',
@@ -347,13 +422,11 @@ if enable_installed_tests
   ]
   foreach tf : testfiles
       data = configuration_data()
-      data.set('datadir', datadir)
-      data.set('installed_testdir', installed_tests_dir)
-      data.set('installed_tests_data_dir', installed_tests_data_dir)
+      data.set('exec', installed_tests_dir / tf)
+      data.set('env', installed_tests_env)
       data.set('libdir', libdir)
-      data.set('exec', tf)
       configure_file(
-        input: 'template.test.in',
+        input: 'template-tap.test.in',
         output: '@0@.test'.format(tf),
         configuration: data,
         install: true,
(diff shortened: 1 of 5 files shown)


swick:
The part that's quite awkward is that pytest-tap isn't in the ubuntu repos which means I have to install it using pip into the system for the installed tests to work.

The alternative is to just not use TAP output. Again, I don't know enough about those tests to make a reasonable decision here.

smcv:
> I personally don't really understand the value of installed tests, so this is more of an question if it makes sense to do it.

Most of the value of as-installed tests is that distributions can re-run the test against a new version of a dependency (like GLib or pytest), *without* recompiling the dependent software (in this case xdg-desktop-portal), to simulate what will happen on end-user systems in `apt upgrade` or equivalent.

This avoids spending time (a lot of time in some cases) on recompiling the dependent software. More importantly, it avoids hiding problems - if a library like GLib breaks its ABI, recompiling the dependent software under test against the new version will usually work around the ABI break by building new binaries that target the new ABI, whereas re-running previously-compiled tests (which expect the old ABI) will detect the ABI break so that it can be investigated and fixed.

Debian and Ubuntu routinely do this, via the `autopkgtest` framework.

A secondary reason to use as-installed tests is that they are often running in a less "fake" environment than build-time tests. For example, during as-installed tests, we can assume that x-d-p is a fully working D-Bus-activatable service, so if we let the message bus start it via service activation, that will detect any brokenness that might exist in its `.service` file. Similarly, we implicitly check for its ability to load the "real" `bwrap` from the `PATH` and its ability to find the "real" `xdg-desktop-portal-validate-*` executables in `/usr/libexec`; and unlike build-time tests being run by a distro autobuilder, the as-installed tests will usually not be running in a chroot or a heavily-constrained container, so we won't usually need workarounds like #1498.

Debian and Ubuntu currently run as-installed tests in either privileged lxc, some sort of lxd (privileged, I think), or a qemu VM, depending on CPU architecture, with Podman as a possible future replacement for lxc, and it is possible to flag tests as `isolation-machine` (meaning "needs machine-level testbed isolation"; we do this for xdg-desktop-portal) so that they will automatically be skipped on the architectures that are not using qemu. This makes the test environment a much more realistic reflection of what will happen on end-user systems than the heavily-constrained environment of an autobuilder.

I think Fedora also has some amount of "as-installed" testing, although I don't know the finer details. I believe Fedora uses `mock` (a restricted chroot/container vaguely similar to Debian's `schroot`) for autobuilders and build-time testing, but then uses qemu or some sort of more permissive container technology for "as-installed" testing.

swick:
Thanks for the thorough explanation.

The first reason would indicate that we should install the pytest based tests as well.

The second reason cannot be realized with the pytest based tests because they tightly control startup of all components themselves. The framework is flexible enough that maybe one could add a special mode for installed tests. The current C portal tests also tightly control the startup so that isn't a regression. Some of the other C based tests however rely on a less "fake" environment.

Given that I want to drop all the C tests, do you have concerns with that?

smcv:
> The part that's quite awkward is that pytest-tap isn't in the ubuntu repos
...
> The alternative is to just not use TAP output.

TAP output is a nice-to-have, but I don't consider it to be important.

We have https://github.com/python-tap/tappy in Debian/Ubuntu (as `python3-tap`), but I don't think we have its pytest plugin.

smcv:
> The second reason cannot be realized with the pytest based tests because they tightly control startup of all components themselves. The framework is flexible enough that maybe one could add a special mode for installed tests. The current C portal tests also tightly control the startup so that isn't a regression.

I think that's reasonable: I can see that the tests for x-d-p probably do need to exercise quite tight control over it so that it will select mock portals (which are testable) instead of the real implementations (which are not).

swick:
@smcv Would be great if you could review this one. We now have all the pytest based integration tests.

tests/meson.build
```diff
@@ -324,17 +324,87 @@ if enable_pytest
...
+      install_dir: installed_tests_dir / 'tests' / 'templates',
+    )
+
+    installed_env = {
+      'XDG_DESKTOP_PORTAL_PATH': libexecdir / 'xdg-desktop-portal',
+      'XDG_PERMISSION_STORE_PATH': libexecdir / 'xdg-permission-store',
+      'XDG_DOCUMENT_PORTAL_PATH': libexecdir / 'xdg-document-portal',
+      'XDP_VALIDATE_ICON': libexecdir / 'xdg-desktop-portal-validate-icon',
+      'XDP_VALIDATE_SOUND': libexecdir / 'xdg-desktop-portal-validate-sound',
+    }
```
smcv, line 382: Do we need to force all of these? Can't we just leave some (ideally all) of them at their defaults, which verifies that they will work as intended on end-user systems with no special environment variable overrides?

smcv (commented):
I'd like to test this with Debian's autopkgtest before saying yes or no to it, but it looks plausible. Thanks!

tests/meson.build
smcv: I expect that we will need to force anything that is an input to the pytest scripts, but we shouldn't need to force anything that is only run indirectly (via the production executables) and gets a suitable default hard-coded into the executables.

So we probably do need to force `*_PATH`, but not `XDP_VALIDATE_ICON` or `XDP_VALIDATE_SOUND`.

swick:
Hey @smcv. I don't want to rush you but Georges mentioned that he wants to release x-d-p soon and I'd like to get this one in if possible.

smcv:
> I don't want to rush you but Georges mentioned that he wants to release x-d-p soon and I'd like to get this one in if possible.

I'm trying. As a first step (to make the tests pass *without* this change): #1577.

smcv:
https://github.com/smcv/xdg-desktop-portal/tree/installed-pytest mostly works, but I'm seeing one failure - in `test_document_fuse.py::TestDocumentFuse::test_multi_thread`, `g-io-error-quark: Error calling StartServiceByName for org.freedesktop.portal.Documents: The connection is closed (18)` - and a *lot* of warnings.

If we can tell pytest to use a different cache directory (environment variable?), I think that would help to avoid warnings like this:

`214s   /usr/lib/python3/dist-packages/_pytest/stepwise.py:51: PytestCacheWarning: could not create cache path /usr/libexec/installed-tests/xdg-desktop-portal/tests/.pytest_cache/v/cache/stepwise: [Errno 13] Permission denied: '/usr/libexec/installed-tests/xdg-desktop-portal/tests/pytest-cache-files-btnx48xn'`

swick:
I've addressed everything except the TestDocumentFuse failure which I can now reproduce locally and without installing. This might be tricky to resolve, so I'm inclined to maybe skip the test for now. WDYT?

swick:
Funnily, `./tests/run-test.sh -n 0 tests/test_document_fuse.py::TestDocumentFuse::test_multi_thread` works fine, but running `./tests/run-test.sh -n 0 tests/test_document_fuse.py::TestDocumentFuse` makes `test_multi_thread` fail, so this somehow is connected to the previous test messing up some state.

swick:
@smcv I've added a commit which disables the racy test for now. I can remove it if we'd rather have to racy test in there. Just tell me your preference. @GeorgesStavracas wants to release later today and says that you'd have to approve this PR for it to land in the release.

smcv:
Re-testing this now

smcv:
> I've added a commit which disables the racy test for now

I think that's appropriate, given that it fails for you too.

If it works reliably when the two test-cases are not combined into one run, then a nice followup would be to generate two separate `.test` files, each of which runs one of the two test-cases (single- and multi-threaded), as two separate processes - but that's not a blocker for this IMO. (Or obviously if someone can figure out the actual problem and fix it, e.g. by cleaning up state more thoroughly between the two test-cases, that would be even better.)

tests/meson.build
```diff
@@ -318,17 +318,89 @@ if enable_pytest
...
+    foreach key, value : installed_env
+      env += f'@key@=@value@ '
+    endforeach
+
+    foreach pytest_file : pytest_files
+      testname = pytest_file.replace('.py', '')
+
+      exec = [pytest.full_path(), installed_tests_dir / 'tests']
+      exec += pytest_args
+      exec += ['-o', 'cache_dir=/tmp/xdp-pytest-cache']
```
smcv, line 386: Let's not introduce /tmp symlink attacks via hard-coded filenames in /tmp, please. ([protected_symlinks](https://docs.kernel.org/admin-guide/sysctl/fs.html#protected-symlinks) mitigates this into only being a denial of service, but it's still best avoided.)

Can we use `-o cache_dir=.` or `-o cache_dir=/proc/self/cwd`, relying on the fact that ginsttest-runner runs each test with its current working directory set to a temporary directory that it automatically cleans up afterwards?

Or perhaps `-p no:cacheprovider`?

smcv (changes requested):
Mostly LGTM. I merged the two prerequisites #1578 and #1579, so it might benefit from rebasing.

tests/meson.build
smcv: > relying on the fact that ginsttest-runner runs each test with its current working directory set to a temporary directory

Reference: https://wiki.gnome.org/Initiatives(2f)GnomeGoals(2f)InstalledTests.html

smcv: Testing https://github.com/smcv/xdg-desktop-portal/tree/wip/pytest-installed now

smcv: Yes, that works. Please consider applying the last commit of that branch, either squashed into [build/tests: Install pytest based tests](https://github.com/flatpak/xdg-desktop-portal/pull/1525/commits/e190f5ce64848326cb2a5f439a3719fa94accff0) or separately.

smcv: wtf. It works, and says `cachedir: .`, but I still get (non-fatal!) warnings about attempting to write into `/usr/libexec/installed-tests`. Sorry, I don't understand pytest well enough to diagnose this... but the tests pass so that's an excellent start.

smcv: Trying now

smcv: Yeah, this works nicely

smcv (approved):

swick:
Thanks a lot Simon!

PR #1527, opened 2024-12-05
test: Depend on a recent enough version of libportal

hfiguiere:
Without this the build fail if you have libportal 0.8.1 installed

This is needed for the notification work.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -120,7 +120,7 @@ geoclue_dep = dependency(
 )
 libportal_dep = dependency(
   'libportal',
-  version: '>= 0.8.1',
+  version: '>= 0.9.0',
   required: get_option('libportal'),
   fallback: ['libportal', 'libportal_dep'],
   default_options: [


swick:
Would be nice to have the "why?" answered in the commit message and not just in a github comment.

smcv (changes requested):
As @swick said,
> Would be nice to have the "why?" answered in the commit message and not just in a github comment.

meson.build
```diff
@@ -120,7 +120,7 @@ geoclue_dep = dependency(
 )
 libportal_dep = dependency(
   'libportal',
-  version: '>= 0.8.1',
+  version: '>= 0.9.0',
```
smcv, line 123: libportal 0.9.0 has not been released yet. If x-d-p needs a newer libportal than 0.8.1, then we need a libportal release first.

hfiguiere: the thing is:

1. it's only for the test
2. it used to be that it wanted a special branch off the repository and I pushed back. This is currently in master in the official repo.
3. this is meant to trigger the meson subproject. I have 0.8.1 installed and the build fail because of that (it build the tests in the default target).

So we can keep the build broken until a release, or we can just accept it and move forward. Or disable the tests again.

swick: Yeah, the projects have a cyclic dependency so depending on an unreleased version is fine, until we want to have a release.

GeorgesStavracas:
libportal uses the post-release version bump scheme, so technically the main branch is already 0.9.0 (but still untagged) and projects can start requiring this version for features that will only exist in the 0.9.0 release.

hfiguiere:
I started building in a fresh toolbox that has not libportal-devel. The subproject fetches 0.8.1 and thus the build fail.

hfiguiere:
So I'm not sure in which circumstances this is supposed to work OOB, but with this PR it does in that case as well.

swick:
ping @smcv

smcv (commented):
hopefully leaving this comment will clear my previous "changes requested"?

hfiguiere:
Merging this would help https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/159 to pass.

GeorgesStavracas (approved):

GeorgesStavracas:
libportal 0.9.0 has been released so this is unblocked now

PR #1530, opened 2024-12-06
Settings portal: Date and Time formats

orowith2os:
This is my proposal for new settings under the Settings portal. I got the idea from Android, and it should help with a tiny bit more consistency in apps and desktops.

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -58,6 +58,25 @@
 
       Unknown values should be treated as 0 (no preference).
 
+    * ``org.freedesktop.date-time`` ``time-format`` (``u``)
+
+      Indicates the system's preferred time format, 12h or 24h.
+      Supported values are:
+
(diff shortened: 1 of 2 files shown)


orowith2os:
I'll be completely honest, everything about time and date here on Linux is weird, so I might not fully understand all this, but it's an idea. I was also curious about seeing if a time zone preference is necessary, for apps to do calculations with?

swick:
-> discussions

jadahl:
> -> discussions

Can't turn pull requests into discussions.

With that said, 12/24h etc is already handled via `LC_TIME`. For example, run `flatpak run --command=sh some.flatpak.App` and then `locale` inside.

orowith2os:
Ah. Sorry, then.

swick:
Yeah, might be nice to have a portal for this but I guess you first have to convince toolkits.

orowith2os:
Maybe if a portal like this gets the go-ahead, we can just add *all* the `LC_*` variables? Thoughts?

PR #1531, opened 2024-12-09
Readd Qt annotations for a{sv} being QVariantMap 

Sodivad:


hfiguiere:
LGTM, but we wonder of the opportunity to run dbusxml2cpp in CI to ensure we never forget.

Sodivad:
I've added now a step that runs qdbusxml2cpp on all xml files which fail on the USB portal atm

But with the tuple annoation from https://github.com/flatpak/xdg-desktop-portal/pull/1529 it seems to pass

swick:
Thank you! That seems to do the job.

So, let's merge https://github.com/flatpak/xdg-desktop-portal/pull/1529 first, then rebase this PR on top of it.

swick (approved):

swick:
Or you pull in all the annotation changes from https://github.com/flatpak/xdg-desktop-portal/pull/1529. That would make things a bit easier.

jadahl:
> But with the tuple annoation from https://github.com/flatpak/xdg-desktop-portal/pull/1529 it seems to pass

The author suggested doing it the other order.

> Or you pull in all the annotation changes from https://github.com/flatpak/xdg-desktop-portal/pull/1529. That would make things a bit easier.

This works too. @Sodivad, what do you prefer?

Sodivad:
> > But with the tuple annoation from #1529 it seems to pass
> 
> The author suggested doing it the other order.
> 
> > Or you pull in all the annotation changes from #1529. That would make things a bit easier.
> 
> This works too. @Sodivad, what do you prefer?

I can do that, although I am not sure about adding annotations just to make the check that is added here happy but requires the user to still basically do the (de)marshalling themselves.

GeorgesStavracas (approved):

PR #1533, opened 2024-12-11
meson: Allow test cases to be optional

sewnie:


meson_options.txt
```diff
@@ -38,6 +38,10 @@ option('datarootdir',
         type: 'string',
         value: '',
         description: 'Define the datarootdir for the pkgconf file')
+option('tests',
+        type: 'boolean',
+        value: true,
+        description: 'Build unit tests')
```
whot, line 44: i'd argue for making this a feature and setting the default value to `enabled`, feels like the more consistent approach with all the other features.

swick:
Enabling pytest and disabling tests results in pytest being disabled as well...

apprehensions:
Is pytest is part of the test suite? if so, then this is to be expected. If not, then that means that the test suite is simply just pytest, and the two options should be merged.

swick:
It should throw an error because the configuration is nonsensical instead of getting ignored silently.

PR #1534, opened 2024-12-11
org.freedesktop.portal.Notification: clarify icon validation requirem…

whereswaldon:
…ents

### DIFF
--- a/data/org.freedesktop.portal.Notification.xml
+++ b/data/org.freedesktop.portal.Notification.xml
@@ -96,7 +96,8 @@
 
         * ``icon`` (``v``)
 
-          A serialized icon to add to the notification. The format for
+          A serialized icon to add to the notification. The icon must pass
+          `icon validation <icons.html>`_ in order to be used. The format for
           serialized icon is a tuple (sv) with the following supported keys:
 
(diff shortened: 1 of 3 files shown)


matthiasclasen:
+1 from me

swick:
Could be exhaustive:
```
    .name = "notification",
    .max_icon_size = 512,
    .max_svg_icon_size = 4096,
    .max_file_size = 1024 * 1024 * 4 /* 4MB */,
```

so file size limit for svg 4k, raster 4mb.

GeorgesStavracas:
I appreciate the documentation improvement, but I think this should be a separate doc page containing some sort of table, or exhaustive list of criteria. Then this page can be referenced from various D-Bus files.

whereswaldon:
I've done my best to move this to a new page with a table of requirements. Feel free to tweak the content to reflect whatever the preferred verbiage is.

GeorgesStavracas (approved):

swick:
What's still kind of problematic is the fact that the file types that are supported depends on the gdk pixbuf backends which are available to the portal. You can do all the "right" things regarding file size, dimensions etc and still get your icon rejected.

PR #1536, opened 2024-12-17
Use all requested setting backends again

### LINKED ISSUE #1535: GTK 3 apps on Wayland have broken anti-aliasing and theming, under KDE Plasma with GTK_USE_PORTAL=1
### Operating System

Neon 6.2

Other

1.19.0

### Desktop Environment

KDE

_No response_

### Expected Behavior

with `GTK_USE_PORTAL=1` (needed in order for GTK apps to use the desktop environment file dialogs) on Wayland, GTK3 apps should look and behave like they do on X11 and also on Wayland without `GTK_USE_PORTAL` (except for the portal file dialogs):

This is how `awf-gtk3` looks for me on Plasma Wayland, without `GTK_USE_PORTAL=1` :
Screenshot_20241212_104034

We can also get the same behavior by setting `GTK_USE_PORTAL=1` and forcing XWayland rendering.

### Current Behavior

This is how `awf-gtk3` looks for me on Plasma Wayland, with `GTK_USE_PORTAL=1` 
Screenshot_20241212_103930

This behavior is new for 1.19.0 - it works fine if I downgrade to 1.18.3. Looking at the D-Bus calls, it looks like XDG Desktop Portal does not call the `gtk` implementation service (and depending on its startup triggers, it may no even start if it is set to start lazily with systemd D-Bus trigger).

### Steps to Reproduce

1. Install xdg-desktop-portal 1.19.0
2. Set the environment variable `GTK_USE_PORTAL=1` 
3. Start a Plasma Wayland session
4. Start any GTK 3 application.

This behavior was originally reported on the closed issue #1123 - that was closed as dup of #1112. I'm reporting this again as (a) probably nobody pays attention to closed issues; (b) at least this behavior seems unrelated to #1112 (and I believe also the originally reported behavior there).

My `/usr/share/xdg-desktop-portal/kde-portals.conf` - which I know XDG Desktop Portal reads, because I `strace`d it - looks like this:

```
[preferred]
default=kde
org.freedesktop.impl.portal.Settings=kde;gtk;
org.freedesktop.impl.portal.Secret=kwallet
```

Sodivad:
This broke with 4db11f7 which made find_all_portal_implementations use the same logic as find_portal_implementation and only make it use the first configured portal that was found. Notably if there's no portal config we would stil load multiple backends. Fixes #1535

### DIFF
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -622,6 +622,40 @@ find_portal_implementation_iface (const PortalInterface *iface)
   return NULL;
 }
 
+static GPtrArray *
+find_all_portal_implementations_iface (const PortalInterface *iface)
+{
+  GPtrArray *impls;
+  g_autofree char *portals = NULL;
+
+  impls = g_ptr_array_new ();
+
+  if (iface == NULL)
+    return impls;
+
+  portals = g_strjoinv (";", iface->portals);
+  g_debug ("Found '%s' in configuration for %s", portals, iface->dbus_name);
+
+
+  for (const GList *l = implementations; l != NULL; l = l->next)
+    {
+      XdpPortalImplementation *impl = l->data;
+
+      for (size_t i = 0; iface->portals && iface->portals[i]; i++)
+        {
+          if (g_str_equal (impl->source, iface->portals[i]) || g_str_equal (iface->portals[i], "*"))
+            {
+              if (portal_impl_supports_iface (impl, iface->dbus_name))
+                g_ptr_array_add(impls, impl);
+              else
+                g_info ("Requested backend %s.portal does not support %s. Skipping...", impl->source, iface->dbus_name);
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


swick:
The pytest framework should make it reasonably easy to write some tests so we don't accidentally break this again.

swick:
`find_default_implementation_iface` also only returns a single impl when it should return multiple. We also looked at `settings_handle_read_all` and it looks like it prioritizes the wrong impls.

Sodivad:
So if 
```
default=foo;bar
```
settings should be read from both. Makes sense

> settings_handle_read_all and it looks like it prioritizes the wrong impls.

To make it clear for myself: In the case as above if both provide the same setting, the one from `foo` should be preferred?

swick:
Found a few more things to do, so I opened https://github.com/flatpak/xdg-desktop-portal/pull/1538. Will try to add tests later...

PR #705, opened 2022-02-01
webextensions: add a portal for managing WebExtensions native messaging servers

jhenstridge:
This is an implementation of a slightly modified version of the interface proposed in #655  for starting WebExtensions native messaging servers on behalf of a confined web browser.

The main change was to remove the use of Request objects, as I'm not waiting for a backend impl server to do the work. It might be worth adding that back though to give more flexibility in the future though.

There are some missing features at present:

1. a functional test for the new portal
2. add some restrictions on use of the portal. In the issue, the suggestions were either (a) a fixed permission from the package manifest, or (b) present access control dialogs, and record result in permission store. If we go with (a), that will need to be implemented for both flatpaks and snaps.

Fixes #655.

PR #1537, opened 2024-12-17
webextensions: add a portal for managing WebExtensions native messaging servers

xhorak:
This MR continues the work from the abandoned MR #705.

The following updates and improvements have been made:
- rebase
- annotation for qt
- fixed test 

@jhenstridge  thank you for your work on this! Let me know if you'd like to collaborate further. 

This is intended to provide a way for a confined web browser to start native code helpers for their extensions. At present it can start the servers installed on the host system. But in future this could be extended to cover sandboxed native messaging servers too.


Rob--W:
Has the other PR really been abandoned, or is the author merely focused on other tasks or even away from work? The latter would not be surprising considering that we are approaching the end of the year.

This patch squashes all changes from #705 into one, along with new changes. I suppose that this was the easiest way to rebase, but it causes the lost of context because some individual commits offered context that are missing from the squashed commit.

Out of all, I think that it would make most sense to include the context of https://github.com/flatpak/xdg-desktop-portal/pull/705/commits/59d7b4bc1b447fb56ac2d25d9e3f1300e93c208a addresses https://github.com/flatpak/xdg-desktop-portal/issues/769 . E.g. by including the issue reference in the commit message ("Fixes xxx") and a code comment pointing to the discussion.
I have not reviewed the updated patch yet, but note that in the previous PR there was still an active discussion on this topic at https://github.com/flatpak/xdg-desktop-portal/pull/705#discussion_r1879469857

swick:
I'd really like if the test was moved over to python.

xhorak:
I've been trying to contact authors since late October but did not received any reply. I would happily keep the original PR but I want to move this forward. Especially now when the Firefox part landed.

The squash was suggested, so I did  it. I'll check the https://github.com/flatpak/xdg-desktop-portal/commit/59d7b4bc1b447fb56ac2d25d9e3f1300e93c208a
I plan to do some testing because I'm getting various app_ids as I've mentioned in https://github.com/flatpak/xdg-desktop-portal/pull/705#discussion_r1880689192

xhorak:
In latest update I've fixed the `xdp_app_info_get_gappinfo` call and stopped to use `g_desktop_app_info_new`.

data/org.freedesktop.portal.WebExtensions.xml
```diff
@@ -0,0 +1,160 @@
...
+        @session_handle: Object path for the #org.freedesktop.portal.Session object
+        @name: name of the native messaging host
+        @extension_or_origin: extension ID or origin URI identifying the extension
+        @json_manifest: the JSON manifest for the native messaging host
+
+        Return the JSON manifest of the native messaging host that
+        Start would invoke.
+    -->
+    <method name="GetManifest">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In3" value="QVariantMap"/>
```
hfiguiere, line 94: I don't see the argument matching this annotation

data/org.freedesktop.portal.WebExtensions.xml
```diff
@@ -0,0 +1,160 @@
...
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="name" direction="in"/>
+      <arg type="s" name="extension_or_origin" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
```
hfiguiere, line 98: I don't see the argument matching this annotation

grulja:
>I plan to do some testing because I'm getting various app_ids as I've mentioned in [#705 (comment)](https://github.com/flatpak/xdg-desktop-portal/pull/705#discussion_r1880689192)

That's the usual issue we have with host apps. The application id comes from cgroups, which is set depending on the way you start the app, e.g. you start it from a terminal app it might get app id of the terminal app. Or you start it with Alt + F2 in GNOME and the app id will be just `firefox`. Flatpak apps are reliable in this case.

See https://github.com/flatpak/xdg-desktop-portal/pull/1522 what we did for the camera portal, where this was causing issues.

swick:
Regarding the appid mismatch that unfortunately still happens: I'd much rather see us adding https://github.com/flatpak/xdg-desktop-portal/pull/1521 than special casing every portal that firefox actually depends on.

xhorak:
> I'd really like if the test was moved over to python.

Looking at the python vs C tests, there seems to be missing permission store service in Python. I'm not sufficiently familiar with the code to implement it for python tests or is it done somewhere already?

swick:
The permission store is supposed to get started (https://github.com/flatpak/xdg-desktop-portal/blob/1c902cc77e53b422d49988bab58b6cca0ed9b112/tests/conftest.py#L458C5-L458C25) when the test case uses either the `xdg_permission_store` or the more general `portals` fixtures.

xhorak:
The proposal (by @swick) suggests replacing xdg-desktop-portal with a separate service, `org.freedesktop.native_messaging`, for native messaging API support. This would make the feature exclusive to browsers using the `--talk-name=org.freedesktop.native_messaging` permission. On the other hand it removes the user access dialog provided by `xdg-desktop-portal`. 

The service API remains unchanged, but as noted by @swick, this approach does not address potential sandbox escapes but it will be exclusive to those who want to use the `org.freedesktop.native_messaging`.

@Rob--W please share your thoughts from the Firefox point of view and all the work that has been done on the Firefox side.

xhorak:
> Could you elaborate on the changes that you're considering? What are the current values and the proposed values?
@swick I'm not sure how the separate service will support creating and maintaining sessions currently implemented by xdp. Could you help?

swick:
Maybe I should try to explain why I would like to keep the native messaging host functionality out of portals and expand a bit on the alternative that I'm suggesting.

Portal APIs are meant to be generically useful platform APIs with high stability which work with the restrictions of a sandbox, and in particular don't allow sandbox escapes. In my opinion the native messaging host portal doesn't meet those criteria:

* It is only exposed to very specific apps instead of being generically useful
* It still can be used to escape the sandbox in some circumstances
* An API that doesn't have the potential sandbox escape might look very different and we wouldn't be able to guarantee stability

The potential sandbox escape comes from the fact that native messaging hosts (NMH) didn't have to deal with a sandboxed browser before as they were both running in the trusted computing base (TCB). If an attacker takes control over firefox, it can use the portal to use all APIs exposed via any NMH. If any of those provides an API that can be used to escape the sandbox, so can the attacker. This could be something as simple as providing an arbitrary path to write a file into.

The portal does ask the user the question "Allow %s to start WebExtension backend" but in general a user which has NHMs installed wants to use them and doesn't know if the browser is malicious, so this doesn't prevent the attack.

Besides not preventing the sandbox escape, the question serves no other purpose because installing the browser extension and having the NMH installed is already enough intent to let them communicate (under the assumption that there is no sandbox escape).

So I'm suggesting to move the code out of the x-d-p project and expose it on a bus name that is not visible to sandboxed apps such as `org.freedesktop.NativeMessagingProvider`. The API has to change slightly because we no longer have a concept of a session. We could merge everything into the `GetManifest` and `Start` methods: mode from the session becomes an argument to both and `GetPipes` can become part of `Start`.

It doesn't fix the inherent sandbox escape problem, but at least it gets us some time to figure out how to get there properly.

Rob--W:
@jhenstridge resumed work on the original PR (#705). Have you all synchronized with each other? Which PR is intended to be the one to be merged?

To me (as a relatively outsider in this project who merely contributes reviews as a subject matter expert in the topic of extensions and NativeMessaging), it is unclear what the desired intermediate and long-term direction is:
- #705 was the original PR, and has now been revived.
- before 705 was revived, this PR was created.
- in comments above, a different design was sketched, that does not only diverge in terms of code, but also external API.

Could someone clarify?

Mikenux:
@Rob--W: See https://github.com/flatpak/xdg-desktop-portal/discussions/1191

Maybe this:
- A containerized browser can only exchange data with another containerized app. Need to determine how much the app needs to be containerized.
- The request to the portal should be about requesting to exchange data with a specific app.
- The discovery of the app to be contacted is the responsibility of the portal or the sandbox system. Same for launching the app (we should have the possibility to present a dialog box to confirm the launch of the app) .
- I wonder if something general can be proposed to allow the general exchange of data between app.

xhorak:
> @jhenstridge resumed work on the original PR (#705). Have you all synchronized with each other? Which PR is intended to be the one to be merged?
> 
> To me (as a relatively outsider in this project who merely contributes reviews as a subject matter expert in the topic of extensions and NativeMessaging), it is unclear what the desired intermediate and long-term direction is:
> 
>     * [webextensions: add a portal for managing WebExtensions native messaging servers #705](https://github.com/flatpak/xdg-desktop-portal/pull/705) was the original PR, and has now been revived.
> 
>     * before 705 was revived, this PR was created.
> 
>     * in comments above, a different design was sketched, that does not only diverge in terms of code, but also external API.
> 
> 
> Could someone clarify?
I've tried to contact @jhenstridge in past few months without any luck, so I've decide to move things forward with this PR. 
We're trying to figure out how to proceed further.

PhrozenByte:
I'd like to chip in a thought: Do we really need the portal to be able to start other applications?

I know that the original idea is about supporting the known and well established Native Messaging API with Flatpak and Snap. However, the API's capability of running and communicating with arbitrary binaries somewhat conflicts with the concept of containerization. The debate revolves around the question of how the Native Messaging API with its powerful capabilities could be limited to fit into Flatpak's security model and has ultimately led to a more or less finished solution not being integrated for several years now.

So I'd like to ask whether we even need it: Is there a reason (besides requiring changes to applications using Native Messaging, I rather mean conceptionally) why we couldn't flip things over, i.e. requiring other applications to provide the necessary communication endpoints and the browser connecting to these endpoints, rather than the browser starting other applications?

The exact solution doesn't really matter right now: It could be a DBus interface (requiring `--talk-name`), or it could even be a well-known runtime directory applications create socket files in for sandboxed browsers to use (requiring `--filesystem`). It would require the other applications to start in advance and possibly run in background to wait for a browser connecting with their endpoint to do their thing though. Are there any extensions that wouldn't work that way? I mean, applications running in background would fit into Flatpak's security model quite well: We even have great desktop integration and it is under full user control.

swick:
The problem isn't who execs the native messaging hosts. The problem is that the native messaging hosts might expose APIs on that socket which can be used to escape the sandbox.

PhrozenByte:
> The problem isn't who execs the native messaging hosts. The problem is that the native messaging hosts might expose APIs on that socket which can be used to escape the sandbox.

@swick As you correctly [stated before](https://github.com/flatpak/xdg-desktop-portal/pull/1537#issuecomment-2579960342), this is by design and more or less the point of Native Messaging. I personally don't see much issue with it, because this isn't something Flatpak needs to or even can sort out, but something the browser vendors must sort out with the developers of applications using the Native Messaging API. This issue isn't even special to the Native Messaging API: As you correctly pointed out in the same post, the same could happen with any custom bus.

I therefore suggest focusing on things that increase the attack surface in comparison to the status quo, like the Native Messaging API being able to more or less execute arbitrary binaries outside the sandbox. IMHO this is a critical issue and requires not just "some" user interaction, but must allow the user to fully understand what's going on. @Mikenux's [suggestions](https://github.com/flatpak/xdg-desktop-portal/pull/1537#issuecomment-2599813416) address this pretty well IMHO.

However, if the Native Messaging API couldn't start applications, I personally don't see much of an issue: Browsers then can only communicate with applications that already run and advertised that they want to talk with a browser using the Native Messaging API (i.e. e.g. by using a custom bus, or creating a socket file in a well-known runtime directory). The issue is then similar to any other custom bus, and users are made aware of running applications by other means already.

Long story short: I'm asking whether we even need a portal. And whether I've missed other issues.

Mikenux:
@PhrozenByte: To clarify, what I've written is for long-term direction, not for intermediate direction. However, I guess people want to know what to make an intermediate solution on which the "right" solution can be built.

swick:
> As you correctly https://github.com/flatpak/xdg-desktop-portal/pull/1537#issuecomment-2579960342, this is by design and more or less the point of Native Messaging

This is not by design, it's just how it turned out to be because native messaging hosts didn't have to care about providing sandbox escapes for the browser before.

Focusing on another aspect is pointless because this is a fundamental flaw that must get sorted out before it can be exposed.

> This issue isn't even special to the Native Messaging API: As you correctly pointed out in the same post, the same could happen with any custom bus.

That's why we only expose APIs (including busses) where the end-points understand the threat-model of having a potentially malicious actor on the other side which will try to escape the sandbox or otherwise mess up the system. The the Native Messaging APIs are not one of those.

> like the Native Messaging API being able to more or less execute arbitrary binaries outside the sandbox

It cannot. There is no such issue. This is distracting from the main problem.

PhrozenByte:
> This is not by design, it's just how it turned out to be because native messaging hosts didn't have to care about not providing sandbox escapes for the browser before.
> 
> Focusing on another aspect is pointless because this is a fundamental flaw that must get sorted out before it can be exposed.

I don't agree with that assessment. However, I'd only repeat things that were said in #705 multiple times before. I agree that this whole discussion is a distraction though. Let's get this PR or #705 merged first and discuss alternative approaches later.

swick:
Sorry, but the correct response to "this PR contains a security vulnerability" is not to go ahead and merge it.

swick:
Rebased because of the testing changes. Works locally but somehow CI fails with the `GetPipes` call with `org.freedesktop.DBus.Error.AccessDenied: Invalid session`.  https://github.com/swick/xdg-desktop-portal/commits/wip/webextensions

tests/test_webextensions.py
```diff
@@ -0,0 +1,126 @@
...
+
+        response = self.start(name, extension_or_origin, dbus_con)
+        assert response, "Response cannot be None"
+        assert response.response == 0, "Returned invalid response code, expected 0"
+
+        pipes = self.get_pipes(dbus_con)
+        assert pipes, "No pipes returned"
+        assert pipes[0].take() > 0, "stdin file descriptor should be valid"
+        assert pipes[1].take() > 0, "stdout file descriptor should be valid"
+        assert pipes[2].take() > 0, "stderr file descriptor should be valid"
```
jhenstridge, line 78: I remember I had used `cat` as the test server in order to test that communication with the server was possible. Would it be worth including that in this test?

Something like `os.write(stdin_fileno, b"Hello")` followed by `assert os.read(stdout_fileno, 100) == b"Hello"`.

Also, the C version of this test tried to manually close the session. Would it make sense to do that here, or is it handled by other parts of the test harness?

jhenstridge (approved):
Sorry for not getting back to you guys on this. This PR looks like a good improvement over my original, and is probably the right way forward to get it merged.

The Python based testing code also looks a lot simpler than what I had written. I wonder if it would be worth testing that you can actually communicate with the messaging server. I had meant to add that to my version but hadn't managed to get it to work reliably. Maybe it will be easier with the Python based tests.

jbicha:
We applied swick's version of this pull request in Ubuntu 25.04 (plucky-proposed) but the webextension tests are [unreliable](https://autopkgtest.ubuntu.com/packages/x/xdg-desktop-portal/plucky/ppc64el) in our infrastructure.

swick:
The fundamental security concerns are still valid and seem to require more fundamental changes to how native messaging works. This is still an important feature and we'd rather have it work now, even with those security concerns. However, portals are meant to provide interfaces which are *safe* and *secure* for sandboxed applications which makes this a bad fit. We therefore decided to move the functionality out of xdg-desktop-portal, into a small stand-alone dbus service:

https://github.com/swick/xdg-native-messaging-proxy

The README in there has a more thorough explanation of the rationale. We also have an implementation for Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1955255.

I believe this is the right way forward because it can fill in the gap now, until we figure out how to change native messaging to become secure for sandboxed browsers.

Would other stakeholders please share if they have any concerns with this direction? If there are no objections, can we proceed with closing the existing PRs?

ssokolow:
> However, if the Native Messaging API couldn't start applications, I personally don't see much of an issue: Browsers then can only communicate with applications that already run and advertised that they want to talk with a browser using the Native Messaging API (i.e. e.g. by using a custom bus, or creating a socket file in a well-known runtime directory). The issue is then similar to any other custom bus, and users are made aware of running applications by other means already.

Extensions which are OK with that dynamic, like [WebScrapbook](https://addons.mozilla.org/en-US/firefox/addon/webscrapbook/) already achieve what they want without waiting for a portal by asking you to run a local HTTP daemon like [PyWebScrapbook](https://pypi.org/project/webscrapbook/).

As a user, I don't consider it acceptable to clutter up my `top`/`ps` editor with dozens of daemons that exist just to wait for me to use the one of the "View source with gVim"/"View selection with gVim"/"Edit text with gVim" entries injected by and extension like [withExEditor](https://addons.mozilla.org/en-US/firefox/addon/withexeditor/) and I don't think it's fair to expect every other user who cares to learn how to configure systemd socket activation for user sessions.

It's bad enough that, when Flatpak gets buggy, it can flood out my `mount` listings with `revokefs-fuse` entries. That sort of behaviour was literally one of the two deal-breakers for me with snaps. (The other being how much it slows down application startup.)

A built-in, cross-platform equivalent to systemd socket activation is an integral part of Native Messaging's value proposition.

jhenstridge:
Looking over the `xdg-native-messaging-proxy` code, my main concerns are:

1. there is no access control. Any decisions about access will need to be made at package install time and cover all native messaging hosts.
2. The simplified API looks like it would make it difficult to retrofit access control onto: clients would need to make sure they don't time out method calls. It might be possible for clients to cause the server problems by creating too many pending method calls.
3. It looks like the proxy uses `g_unix_fd_list_append`, which duplicates the file descriptor into the list. This is going to leak file descriptors in the proxy, and prevent the browser from signalling a native messaging host to exit by closing its stdin.
4. It doesn't look like the proxy will clean up spawned processes if the browser drops off the bus. And due to the file descriptor leak, the hosts won't notice the browser has gone either.

We could add access to this proxy in snapd's browser-support interface, but it is not obviously an improvement over the xdg-desktop-portal version of the feature.

PhrozenByte:
This whole story is a bit... unusual.

I think I can speak for everyone here that the work is very much appreciated. However, I don't see how this brings us closer to get Native Messaging finally working. The new "proxy approach" (?) surely looks promising and indeed has the potential to be advantageous over the portal (which was never denied in principle after its proposal btw), but the project more looks like a Proof of Concept right now. It lacks some testing, a code review and concerns have been raised, both here on GitHub and on Bugzilla.

The portal approach is under discussion for more than three years now. There was a lengthy discussion about it and even though most agree that it's no perfect solution and that it indeed has issues, there was clear consensus that the solution is acceptable. This new proxy approach really just addresses a single issue with the existing portal approach. This basically means switching from one solution with issues to another solution with issues. There's one major difference though: The portal approach more or less matches that of Snap and is used in production for years now, it follows consensus, both sides are fully developed, mostly tested (yes, some minor work needs to be done), reviewed extensively, and the MRs are basically just waiting to be merged.

I thus strongly advocate for being pragmatic: The MRs of the portal approach are merge-able as-is. Let's merge them now. But don't stop then, keep working on a proper solution addressing *all* issues, not just a single one. This new improved solution can then be merged as well and the portal can be deprecated. This also allows for a proper transition period for LTS releases and keeps Snap and Flatpak aligned in this regard.

rugk:
> The important part to nice here is that a user cannot know if a particular host program will allow a sandbox escape or not.

Why?
IMHO this premise is not true. As a user of KeePassXC I totally trust their proxy to not be a sandbox escape, it is well designed etc. Some users may not know their software but many will and can evaluate whether they are trustworthy etc. IMHO.

ssokolow:
@rugk If I remember the rest of the discussion correctly, the argument being made is that access control lends a false sense of security. (eg. Maybe someone pulls a Jia Tan on your native messaging server. You can't apply "If that happens to my password manager, I have much bigger problems" to other native messaging servers.)

rugk:
Sure, but would not the thing or calls also be sandboxed in a flatpak? (so we keep compartimentalisation)
I mean, running on the host can be a real escape but like this?

Then maybe limit the whole thing to flatpaks/other sandboxed services again? The permission print can alos again show what permission the to-be-called app has?

Just as an idea.

ssokolow:
> Sure, but would not the thing or calls also be sandboxed in a flatpak? (so we keep compartimentalisation)

Not necessarily. One of the recurring themes for existing portal APIs has been trying to remain interoperable, regardless of which applications are or aren't installed as Flatpaks.

> Then maybe limit the whole thing to flatpaks/other sandboxed services again? The permission print can alos again show what permission the to-be-called app has?

That would likely invite an immediate bug report/feature request asking for support to be extended to non-Flatpak'd native messaging servers.

(eg. for the kind of Firefox extensions where the developer doesn't use Flatpak, may not get any closer to Linux than running `GOOS=linux go build` on Windows, and sees it as a Flatpak bug if a native messaging server that works with one of Mozilla's Firefox tarballs doesn't work with the Flatpak release.)

Mikenux:
@rugk 

Doing only flatpak to flatpak (or snap to snap) and depending on permissions is already an idea for the long term solution :wink: 

That doesn’t exclude support for host apps. Just that it would be host to host only.

tazihad:
Hi! Just a gentle reminder to consider merging this PR. It's been open for quite some time, and it would be greatly beneficial for Fedora Atomic and Bazzite users, especially if it already supports the essential browser native messaging features.

Even if it's not fully complete, we could improve or update the patches gradually over time. Thank you for your time and consideration!

Mikenux:
This portal cannot be merged as is for security reasons. To be merged, it is either to ask the user for permission with a big security warning, which might be too worrying for the naive user, or to give automatically the permission to use extensions when the portal is used. However, for the latter option, stores must be able to know in advance that an app will use this portal (afaik, they only know about static permissions). Otherwise, that's swick's solution (see https://github.com/flatpak/xdg-desktop-portal/pull/1537#issuecomment-2747856356).

gabeklavans:
> ask the user for permission with a big security warning, which might be too worrying for the naive user,

pardon me if you're just simplifying the reality for the sake of brevity, but if this is truly an option, I believe it is perfectly fine to merge as a first pass. Currently, Native Messaging doesn't work _at all_, so ostracizing users who click away from a scary looking prompt would at the very least allow users who understand the prompt to use this functionality. That seems perfectly reasonable as a first-pass implementation, and can be tweaked later on to make the process more automatic.

If it is not that simple or if the consensus does not agree with my take, I understand, but I felt it should be said as I presume many other users will feel the same way.

Mikenux:
Yes, this was to be brief. This might be an option, but:
- That would mean signifying to users that it isn’t recommended to give access. This would indeed allow knowledgeable users to allow (or disallow) the use of extensions, but then that’s the other kind of users that will be "ostracized " . Knowledgeable users are also expected to know how to manage permissions (cli, flatseal for flatpak).
- As far as I know, nobody has the UI. If someone has, they can show it for review.
- Depending on how things goes for the long term solution, the UI might change and a mechanism to transition the permission would be probably to be developed.

In sum, this requires additional work which can even slow down things, especially UI-side at first. So, the best, for now, is to use a static permission and let knowledgeable users modify it if they want a more sandboxed web browser.

The only problem that might happen is people aren’t happy with swick's solution (on aspects other than that’s a static permission).

Mikenux:
> * As far as I know, nobody has the UI. If someone has, they can show it for review.

Found it on https://www.omgubuntu.co.uk/2022/07/ubuntu-devs-fix-another-frustrating-firefox-snap-flaw.

Not very good in my opinion, but it is for snaps (where I don't know what model they use).

If a user interface is wanted, I can try, but I won't do anything without approval.

Utini2000:
This issue really needs more attention and a quick solution.

Flatpaks are already widely used by Linux users and will become even more used with atomic OS becoming more popular (e.g. Fedora will default to Atomic OS like Kinoite or Silverblue by 2028).

We already spent 3 years with this issue...

swick:
This portal is not needed anymore because we now have [xdg-native-messaging-proxy](https://github.com/flatpak/xdg-native-messaging-proxy) and support for it is being implemented by firefox.

GeorgesStavracas:
Please do not write unrelated comments in pull requests. Also, please do not spam issues and PRs with requests like these.

sudoshindo:
I just need a bit of clarification on the xdg-native-messaging-proxy repo.

Is that proxy still in an unusable state and not ready to be implemented by applications that need it?

Or does it make use of the work made in this pull request and is basically ready to use?

GeorgesStavracas:
xdg-native-messaging-proxy contains the code in this pull request, but in a separate project with a separate executable. So it should be ready to use yes.

PR #1538, opened 2024-12-18
Use all requested setting backends again [cont]

swick:
Based on https://github.com/flatpak/xdg-desktop-portal/pull/1536

### DIFF
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -629,6 +629,56 @@ find_portal_implementation_iface (const PortalInterface *iface)
   return NULL;
 }
 
+static void
+_add_all_portal_implementations_iface (const PortalInterface *iface,
+                                       const char            *interface,
+                                       GPtrArray             *impls)
+{
+  g_autofree char *portals = NULL;
+
+  portals = g_strjoinv (";", iface->portals);
+  g_debug ("Found '%s' in configuration for %s", portals, iface->dbus_name);
+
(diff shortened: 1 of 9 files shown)


src/xdp-portal-impl.c
```diff
@@ -622,6 +629,51 @@ find_portal_implementation_iface (const PortalInterface *iface)
...
+
+  for (size_t i = 0; iface->portals && iface->portals[i]; i++)
+    {
+      const char *portal = iface->portals[i];
+
+      for (const GList *l = implementations; l != NULL; l = l->next)
+        {
+          XdpPortalImplementation *impl = l->data;
+
+          if (!g_str_equal (impl->source, portal) && !g_str_equal (portal, "*"))
```
Sodivad, line 650: If someone has 

`org.freedesktop.impl.portal.Settings=foo;bar;*`

then  what we end up with adding `foo` and `bar` twice

Sodivad: > True, I should add tests for * as well. The reason why I changed the loops is that iterating over the impls in the outer loop destroys the config order

Makes sense.

swick: Took the easy route and just check if the impl is already in the impls list. This is now loops in a loops in a loop but should be fine as the number of items is low.

Also added tests which use `*` and actually test that the backend is only called once with a config of `test1;*`.

GeorgesStavracas (approved):

PR #1539, opened 2024-12-20
settings: Add cursor-theme and cursor-theme-size

jadahl:
These aim to provide cursor theming settings that isn't to a desktop environment or toolkit.

### DIFF
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -62,6 +62,14 @@
 
       Unknown values should be treated as ``0`` (no preference).
 
+    * ``org.freedesktop.appearance``  ``cursor-theme`` (``s``)
+
+      Indicates the system's preferred cursor theme.
+
+    * ``org.freedesktop.appearance``  ``cursor-theme-size`` (``u``)
+
+      Indicates the system's preferred cursor theme size.
+
     This documentation describes version 2 of this interface.
   -->
   <interface name="org.freedesktop.portal.Settings">


data/org.freedesktop.portal.Settings.xml
```diff
@@ -62,6 +62,14 @@
 
       Unknown values should be treated as ``0`` (no preference).
 
+    * ``org.freedesktop.appearance``  ``cursor-theme`` (``s``)
+
+      Indicates the system's preferred cursor theme.
+
+    * ``org.freedesktop.appearance``  ``cursor-theme-size`` (``u``)
+
+      Indicates the system's preferred cursor theme size.
```
ideasman42, line 71: It would be good to mention if this includes output scaling (how it's expected to handle Hi-DPI / fractional-scaling) for e.g.

jadahl: Indeed, I would assume it be in "logical pixels" like size, meaning equivalent to the surface size one would set with `wl_pointer.set_cursor`, which is more or less what the existing cursor theme size settings do, and what `gsettings get org.gnome.desktop.interface cursor-size` would return.

ideasman42: This could be mentioned in the description then (or a reference elsewhere - to more detailed docs).
On my own desktop I've noticed the cursor size jumping erratically between windows. So having a convention for calculating the final cursor size would be good (including rounding behavior with fractional scaling).
Although that doesn't have to be handled in this PR, failure to define this anywhere likely means cursor size will "jump" from a user perspective.

JakobDev:
Why is this needed? With the Cursor Shape Wayland Protocol the Cursor is drawn by the Compositor, so Apps don't need to know about the Cursor theme.

swick:
The only cursors that must be supported on wayland are surface-based cursors, so they will stick around.

ideasman42:
> Why is this needed? With the Cursor Shape Wayland Protocol the Cursor is drawn by the Compositor, so Apps don't need to know about the Cursor theme.

Applications may need cursors that aren't available as presets in the cursor-shape protocol, in this case there is no unified way to know how large to draw the cursor:
See the second paragraph in this email:
https://www.mail-archive.com/wayland-devel@lists.freedesktop.org/msg42651.html

data/org.freedesktop.portal.Settings.xml
jadahl: I added some description how to treat the theme size, let me know if you think it's enough. It is somewhat vague, to not fully define all of Wayland scaling semantics, but tries to be clear enough that the size is not "true pixel size".

swick:
> There are hypothetical scenarios that this could fix so I don't object to it if others want it, but I'm not aware of any real case.

LTS distros where flatpak and x-d-p is updated but the rest of the stack is not.

Anyway, fair enough.

PR #1550, opened 2024-12-24
tests: disable only sound-related notification tests if without wavparse

ziyao233:
Only sound-related tests require gstreamer wavparse plugin for validation, so it's okay to disable test_notification_sound but keep other notification tests.

This also fixes link failures when wavparse isn't available: we previously excluded notification.c from the source list in the missing case, but still used test_notification_* symbols in test-portals.c.

Fixes: 1c6dd18 ("tests: Check for required WAV decoder for notification tests")

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -144,6 +144,9 @@ if gst_inspect.found()
 else
   have_wav_parse = false
 endif
+if have_wav_parse
+  config_h.set('HAVE_WAV_PARSE', 1)
+endif
 
 bwrap = find_program('bwrap', required: get_option('sandboxed-image-validation').allowed() or get_option('sandboxed-sound-validation').allowed())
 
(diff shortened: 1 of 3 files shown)


tests/test-portals.c
```diff
@@ -589,7 +589,9 @@ main (int argc, char **argv)
   g_test_add_func ("/portal/notification/bad-priority", test_notification_bad_priority);
   g_test_add_func ("/portal/notification/bad-button", test_notification_bad_button);
   g_test_add_func ("/portal/notification/icon", test_notification_icon);
+#ifdef HAVE_WAV_PARSE
   g_test_add_func ("/portal/notification/sound", test_notification_sound);
+#endif
```
swick, line 594: Can we instead `g_test_skip` this test?

ziyao233: it's okay. changed in the new HEAD

swick (approved):

swick:
Thanks, LGTM now.

PR #1552, opened 2024-12-29
settings: Do not export without implementations

kleifgch:
If no settings portal implementations are found, the `settings_create()` function currently returns NULL and the `No skeleton to export` warning is emitted. This change guards against that, which is more consistent with the other portals.

### DIFF
--- a/src/xdg-desktop-portal.c
+++ b/src/xdg-desktop-portal.c
@@ -253,7 +253,8 @@ on_bus_acquired (GDBusConnection *connection,
   export_portal_implementation (connection, realtime_create (connection));
 
   impls = find_all_portal_implementations ("org.freedesktop.impl.portal.Settings");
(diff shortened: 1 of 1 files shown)


swick:
Might make sense to always export the settings portal even if there are no backends.

src/xdg-desktop-portal.c
```diff
@@ -253,7 +253,8 @@ on_bus_acquired (GDBusConnection *connection,
   export_portal_implementation (connection, realtime_create (connection));
```
swick, line 253: Commit message: `main: Do not export without implementations`

swick:
Having the settings portal exposed and returning nothing at all doesn't get us anything, so I'm now in favor of this change.

This LGTM with the nit resolved.

swick (approved):

Issue #1528, opened 2024-12-06
Restore device removal test for USB


hfiguiere:
### Operating System

any

### XDG Desktop Portal version

Git

### Desktop Environment

GNOME

### Expected Behavior

All test run

### Current Behavior

Some test have been disabled on CI because they fail but work locally, see commit 06a5e502c3359cf4751ed5250eceee20cbdaae27

### Steps to Reproduce

1. Just run the test

hfiguiere:
Closed with #1556

PR #1556, opened 2025-01-07
tests/usb: Skip the removal tests when the UMockdev version is too old

swick:
We disabled the test in CI because the tests were failing but we didn't understand why in

06a5e502 ("test: Disable USB removal test in a container")

It turns out that UMockdev simply didn't send the remove event before version 0.18.4. Let's change the skip condition to actually check for the right thing instead of just disabling it in CI. This means that the test will run in CI once ubuntu:latest does have the right version.

### DIFF
--- a/tests/test_usb.py
+++ b/tests/test_usb.py
@@ -7,6 +7,8 @@ import tests as xdp
 import pytest
 import os
 import gi
+import subprocess
+import re
 
 gi.require_version("UMockdev", "1.0")
 from gi.repository import GLib, UMockdev  # noqa E402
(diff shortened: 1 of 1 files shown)


hfiguiere:
This would close #1528

tests/test_usb.py
```diff
@@ -22,6 +24,17 @@ def umockdev():
     return UMockdev.Testbed.new()
 
 
+def umockdev_has_working_remove():
```
hfiguiere, line 27: shouldn't we also try to enforce this in our CI images?

swick: We could but don't have to. At least with this PR we will automatically get the test in the future. As long as it's a single test I'm inclined to leave the CI as is.

hfiguiere: fair enough

GeorgesStavracas (approved):

PR #1557, opened 2025-01-07
build: Rename meson_options.txt to meson.options

swick:


hfiguiere:
> For versions of meson before 1.1, this file was called meson_options.txt

So we need to require 1.1. Currently 0.60 is required.

swick:
Yup, I read the docs and thought 1.1 is a really old version but it turns out to be rather recent so I'll leave it as draft.

eli-schwartz:
The old name isn't deprecated. I don't understand the rush...

PR #1558, opened 2025-01-08
tests/conftest.py: Error out if xdp exited before exporting its bus

swick:
Otherwise we'll be stuck in the loop forever and it's not clear what's going on.


GeorgesStavracas:
@whot

whot (approved):

PR #1559, opened 2025-01-08
Wait for debugger

swick:
This is especially useful when the binaries are run as part of a test and getting the debugger to the right place takes a bit of effort.

Without the `raise (SIGCONT);` I get into a loop of SIGSTOPs in gdb and I don't understand why.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -1719,6 +1719,21 @@ main (int    argc,
   g_autoptr(GOptionContext) context = NULL;
   GDBusMethodInvocation *invocation;
 
+  if (g_getenv ("XDG_DOCUMENT_PORTAL_WAIT_FOR_DEBUGGER") != NULL)
+    {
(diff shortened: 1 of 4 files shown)


swick:
@matthiasclasen would appreciate if you could review this one

swick:
You can try this with by setting the env var, run a test and then follow the instructions printed on screen to connect a debugger.

The code itself has no impact if the env vars are not set, so even if the added code would not work, this has no risk.

jadahl:
Lets go with this, it's a debug feature, hidden behind an env var, that is particularly useful given how tests are written. Also, to some degree it makes sense, the SIGCONT is confusing, but it seems to be what is needed to not make this annoying to use.

jadahl (approved):

swick:
rebased

GeorgesStavracas:
Thanks for reviewing is @jadahl

Issue #603, opened 2021-07-18
Search in $XDG_DATA_DIRS for portals


concatime:
Hi,
just like how [D-Bus finds services](https://dbus.freedesktop.org/doc/dbus-daemon.1.html), could you search in `XDG_DATA_DIRS` for portals? 

I recommend this logic:
- If `XDG_DESKTOP_PORTAL_DIR` is set, look into it (personally, `XDG_DESKTOP_PORTAL_DIRS` would have been better).
- Otherwise,
  - if `XDG_DATA_DIRS` is set, look into them with suffix `/xdg-desktop-portal/portals`.
  - Othersise, look into `/usr/local/share` and `/usr/share` [as per the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables), with suffix `/xdg-desktop-portal/portals`.
  - Finally look in `${datadir}/xdg-desktop-portal/portals` just like D-Bus.

And quite honestly, with this logic, `XDG_DESKTOP_PORTAL_DIR` can be dropped all together, but that would be a breaking change.

WDYT?

podiki:
I like this idea, can keep `XDG_DESKTOP_PORTAL_DIR` around, but using the standard XDG directories makes sense. Related on this front, see the issue I just filed #616 (doesn't handle multiple directories).

andychenbruce:
Another thing is that if you set `XDG_DESKTOP_PORTAL_DIRS` then also [prevents it from reading any configuration files](https://github.com/flatpak/xdg-desktop-portal/blob/a18f69ffdf43edddb9207d63a212b827382c4404/src/xdp-portal-impl.c#L451-L452).

PR #1561, opened 2025-01-09
Also load portals from XDG_DATA_DIRS

andychenbruce:
Addresses issue #603. The only place it searches for portals is the `DATADIR` or `XDG_DESKTOP_PORTAL_DIR`, but if `XDG_DESKTOP_PORTAL_DIR` is set, then it [doesn't read configuration files](https://github.com/flatpak/xdg-desktop-portal/blob/a18f69ffdf43edddb9207d63a212b827382c4404/src/xdp-portal-impl.c#L451-L452).

This is a problem because any system that doesn't follow the FHS has to set `XDG_DESKTOP_PORTAL_DIR` for it to find the portals, but then can't load any configuration files which leaves everything broken. A simple solution is to search all the directories in `XDG_DATA_DIRS`.


src/xdp-portal-impl.c
```diff
@@ -319,6 +314,33 @@ load_installed_portals (gboolean opt_verbose)
...
+  bool dir_set_from_env = (portal_dir == NULL);
+  if (portal_dir == NULL)
+    portal_dir = DATADIR "/xdg-desktop-portal/portals";
+
+  load_installed_portals_dir(opt_verbose, portal_dir);
+
+  if(dir_set_from_env)
+    return
+  
+  const char * const *dirs = g_get_system_data_dirs ();
```
smcv, line 335: If we're going to load from `XDG_DATA_DIRS` then we should also try `XDG_DATA_HOME`, for completeness and to be consistent with the basedir spec.

src/xdp-portal-impl.c
```diff
@@ -319,6 +314,33 @@ load_installed_portals (gboolean opt_verbose)
...
+
+  /* We need to override this in the tests */
+  portal_dir = g_getenv ("XDG_DESKTOP_PORTAL_DIR");
+  bool dir_set_from_env = (portal_dir == NULL);
+  if (portal_dir == NULL)
+    portal_dir = DATADIR "/xdg-desktop-portal/portals";
+
+  load_installed_portals_dir(opt_verbose, portal_dir);
+
+  if(dir_set_from_env)
```
smcv, line 332:
```suggestion
  if (dir_set_from_env)
```

src/xdp-portal-impl.c
```diff
@@ -269,16 +269,11 @@ sort_impl_by_use_in_and_name (gconstpointer a,
 }
 
 void
-load_installed_portals (gboolean opt_verbose)
+load_installed_portals_dir (gboolean    opt_verbose,
+                        const char *portal_dir)
```
smcv, line 274: Something has gone wrong with the indentation here, are you using tabs? (If yes: please don't)

src/xdp-portal-impl.c
```diff
@@ -319,6 +314,32 @@ load_installed_portals (gboolean opt_verbose)
...
+
+  if(dir_set_from_env)
+    return;
+  
+  const char * const *dirs = g_get_system_data_dirs ();
+  
+  for (iter = dirs; iter != NULL && *iter != NULL; iter++)
+    {
+      g_autofree char *dir = g_build_filename (*iter, "xdg-desktop-portal", "portals", NULL);
+      load_installed_portals_dir(opt_verbose, dir);
```
smcv, line 340:
```suggestion
      load_installed_portals_dir (opt_verbose, dir);
```

swick:
I really would want to have tests for the configs. With the pytest setup we currently have, we can only test config in $HOME but we could modify things for this test to launch x-d-p in a new mount namespace with  `/usr/local/share/xdg-desktop-portal` etc under our control.

Is that something you could look into?

smcv (changes requested):
Reserving judgement on whether this is a desired feature or not (other maintainers might have an opinion on that, I think it's a reasonable thing to want):

If x-d-p is going to look in more than one directory, then I think it should remember all the filenames that it has already seen, and ignore any file of the same name in the lower-precedence directories. (Probably easiest done with a global hash table.)

smcv:
> With the pytest setup we currently have, we can only test config in $HOME but we could modify things for this test to launch x-d-p in a new mount namespace with `/usr/local/share/xdg-desktop-portal` etc under our control.

If we do this, please only do it as additional test coverage, and not as the only way some other thing gets tested. Linux distributions' autobuilders usually can't create new mount namespaces so this would have to be skipped.

andychenbruce:
For reference [this](https://github.com/NixOS/nixpkgs/blob/20755fa05115c84be00b04690630cb38f0a203ad/pkgs/development/libraries/xdg-desktop-portal/default.nix#L64-L67) is how nix deals with the not loading configuration files problem.

Apteryks:
Somewhat related; this is what we use in GNU Guix: https://github.com/flatpak/xdg-desktop-portal/pull/1716

hfiguiere:
2 wrongs don't make a right.

swick:
Yes, please drop the patch that abuses `XDG_DESKTOP_PORTAL_DIR`. It's for the test suite only.

Issue #1549, opened 2024-12-24
Requesting a `handle_token` with `org.freedesktop.portal.RemoteDesktop.CreateSession` containing a special character causes the portal to crash


Kale-Ko:
### Operating System

Arch Linux

### XDG Desktop Portal version

1.18

### Desktop Environment

KDE

### Expected Behavior

An error response is returned or the message is discarded.

### Current Behavior

The portal crashes.

### Steps to Reproduce

1. Run `gdbus call --session --dest org.freedesktop.portal.Desktop --object-path /org/freedesktop/portal/desktop --method org.freedesktop.portal.RemoteDesktop.CreateSession '{"handle_token": <"token_with_special_&">, "session_handle_token": <"valid_token">}'` in a terminal

### Anything else we should know?

Core Dump: [core.dump.tar.gz](https://github.com/user-attachments/files/18237400/core.dump.tar.gz)

If the arguments are reversed the portal seems to just discard the message as expected.

PR #1562, opened 2025-01-10
session: Validate the token for correctness

whot:
The token is used as part of an object path so it has to meet those requirements. We can't escape it since the caller presumably expects to use the token as-is so where it fails the validity simply error out.

Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1549

### DIFF
--- a/src/xdp-session.c
+++ b/src/xdp-session.c
@@ -312,6 +312,23 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
 }
 
 static gboolean
+is_valid_token (const char *token)
+{
+  g_autofree char *path = NULL;
+  int i;
+
+  for (i = 0; token[i]; i++)
+    {
+      if (token[i] == '/')
+        return FALSE;
+    }
+
+  path = g_strdup_printf ("/foo/%s", token);
+
+  return g_variant_is_object_path (path);
+}
+
+static gboolean
 xdp_session_initable_init (GInitable     *initable,
                            GCancellable  *cancellable,
                            GError       **error)
@@ -332,6 +349,15 @@ xdp_session_initable_init (GInitable     *initable,
   if (!session->token)
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Missing token");
+      return FALSE;
+    }
+
+  if (!is_valid_token (session->token))
+    {
+      g_set_error (error,
+                   XDG_DESKTOP_PORTAL_ERROR,
+                   XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   "Invalid token '%s'", session->token);
       return FALSE;
     }
 
--- a/tests/test_remotedesktop.py
+++ b/tests/test_remotedesktop.py
@@ -46,6 +46,20 @@ class TestRemoteDesktop:
         session.close()
         xdp.wait_for(lambda: session.closed)
 
+    @pytest.mark.parametrize("token", ("Invalid-Token&", "", "/foo"))
+    def test_remote_desktop_create_session_invalid(self, portals, dbus_con, token):
+        remotedesktop_intf = xdp.get_portal_iface(dbus_con, "RemoteDesktop")
+
+        request = xdp.Request(dbus_con, remotedesktop_intf)
+        options = {"session_handle_token": token}
+
+        with pytest.raises(dbus.exceptions.DBusException) as excinfo:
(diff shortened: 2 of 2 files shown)


tests/test_remotedesktop.py
```diff
@@ -46,6 +46,19 @@ def test_remote_desktop_create_close_session(self, portals, dbus_con):
...
         xdp.wait_for(lambda: session.closed)
 
+    def test_remote_desktop_create_session_invalid(self, portals, dbus_con):
+        remotedesktop_intf = xdp.get_portal_iface(dbus_con, "RemoteDesktop")
+
+        request = xdp.Request(dbus_con, remotedesktop_intf)
+        options = {"session_handle_token": "Invalid-token&"}
+
+        with pytest.raises(dbus.exceptions.DBusException) as excinfo:
+            request.call("CreateSession", options=options)
```
swick, line 56: ohh, that's nice

src/xdp-session.c
```diff
@@ -311,6 +311,20 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
...
 }
 
+static gboolean
+is_valid_token(const char *token)
+{
+  int i;
+
+  for (i = 0; token[i]; i++)
+    {
+      if (!g_ascii_isalnum(token[i]) && token[i] != '_')
```
swick, line 321: Is `_` a valid token? Doesn't gdbus already provide something like this?

smcv: If this is being used in object paths, https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names is the canonical specification.

gdbus probably doesn't have a function for "is this a valid object-path *component*?", though, just "is this a valid complete object path?" (which is `g_variant_is_object_path()`).

smcv: > Is `_` a valid token?

If our only functional requirement is "can we append it to `/blah/blah/` and get a valid object path?" then yes it is.

src/xdp-session.c
```diff
@@ -311,6 +311,20 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
...
+        return FALSE;
+    }
+
+  return TRUE;
```
smcv, line 325: This implementation would return true for the empty string, but `"/blah/blah/" + ""` is not a syntactically valid object path (an object path must not end with `/` unless it is exactly `/`). So we must check for that.

(We are probably appending it to some string in the `/org/freedesktop/...` namespace rather than literally `/foo/` or `/blah/blah/`, but it's the "shape" that matters.)

smcv (changes requested):

src/xdp-session.c
smcv, line 321:
```suggestion
      if (!g_ascii_isalnum (token[i]) && token[i] != '_')
```

src/xdp-session.c
```diff
@@ -311,6 +311,20 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
   return TRUE;
 }
 
+static gboolean
+is_valid_token(const char *token)
```
smcv, line 315:
```suggestion
is_valid_token (const char *token)
```

src/xdp-session.c
```diff
@@ -335,6 +349,15 @@ xdp_session_initable_init (GInitable     *initable,
       return FALSE;
     }
 
+  if (!is_valid_token(session->token))
```
smcv, line 352:
```suggestion
  if (!is_valid_token (session->token))
```

tests/test_remotedesktop.py
```diff
@@ -46,6 +46,19 @@ def test_remote_desktop_create_close_session(self, portals, dbus_con):
         session.close()
         xdp.wait_for(lambda: session.closed)
 
+    def test_remote_desktop_create_session_invalid(self, portals, dbus_con):
+        remotedesktop_intf = xdp.get_portal_iface(dbus_con, "RemoteDesktop")
+
+        request = xdp.Request(dbus_con, remotedesktop_intf)
+        options = {"session_handle_token": "Invalid-token&"}
```
smcv, line 53: Please can we specifically test `"/foo"` and `""` here, too? Those are special cases would be easy to get wrong.

```
for badness in ("Invalid-token&", "/foo", ""):
    options = { "session_handle_token": badness }

    with pytest.raises (etc.)
    ...
```

swick (approved):

swick:
LGTM now

PR #1568, opened 2025-01-12
conf: Enable html permalinks for headers on hover

razzeee:
Based on https://github.com/flatpak/flatpak-docs/commit/29cd6f081cd921d25550eb7a5c4e99f792a5c321


GeorgesStavracas (approved):
Thanks!

PR #1571, opened 2025-01-15
Try harder to find the host path for the documents

xhorak:
When the access to the directory is provided by the portal (like browser download directory) the stored files in the directory are not found by xdp_get_real_path_for_doc_path and /run/user/... path is opened instead of host path by the OpenFile or OpenDirectory portal.

### DIFF
--- a/src/open-uri.c
+++ b/src/open-uri.c
@@ -682,19 +682,14 @@ handle_open_in_thread_func (GTask *task,
       g_autofree char *path = NULL;
       gboolean fd_is_writable;
       g_autoptr(GError) local_error = NULL;
-      g_autofree char *host_path = NULL;
 
       path = xdp_app_info_get_path_for_fd (request->app_info, fd, 0, NULL, &fd_is_writable, &local_error);
 
       if (path != NULL)
         {
-          host_path = xdp_get_real_path_for_doc_path (path, request->app_info);
-          if (host_path)
-            {
-              g_debug ("OpenFile: translating path value '%s' to host path '%s'", path, host_path);
-              g_clear_pointer (&path, g_free);
-              path = g_steal_pointer (&host_path);
-            }
+          char *resolved_path = xdp_resolve_document_portal_path (path);
+          g_clear_pointer (&path, g_free);
+          path = g_steal_pointer (&resolved_path);
         }
 
       if (path == NULL ||
@@ -724,14 +719,17 @@ handle_open_in_thread_func (GTask *task,
 
       if (open_dir)
         {
-          g_autofree char *real_path = xdp_get_real_path_for_doc_path (path, request->app_info);
           /* Try opening the directory via the file manager interface, then
              fall back to a plain URI open */
           g_autoptr(GError) local_error = NULL;
           g_autoptr(GVariant) result = NULL;
           g_autoptr(GVariantBuilder) uris_builder = NULL;
-          g_autofree char* item_uri = g_filename_to_uri (real_path, NULL, NULL);
+          g_autofree char *item_uri = NULL;
           g_autoptr(GDBusConnection) bus = NULL;
+          g_autofree char *real_path = NULL;
+
+          real_path = xdp_resolve_document_portal_path (path);
+          item_uri = g_filename_to_uri (real_path, NULL, NULL);
 
           bus = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &local_error);
 
--- a/src/file-chooser.c
+++ b/src/file-chooser.c
@@ -155,53 +155,6 @@ out:
   g_task_return_boolean (task, TRUE);
 }
 
-/* Calling Lookup on a nonexisting path does not work, so we
- * pull the doc id out of the path manually.
- */
-static gboolean
-looks_like_document_portal_path (const char *path,
-                                 char **guessed_docid)
-{
-  const char *prefix = "/run/user/";
-  char *docid;
-  char *p, *q;
-
-  if (!g_str_has_prefix (path, prefix))
-    return FALSE;
-
-  p = strstr (path, "/doc/");
-  if (!p)
-    return FALSE;
-
-  p += strlen ("/doc/");
-  q = strchr (p, '/');
-  if (q)
-    docid = g_strndup (p, q - p);
-  else
-    docid = g_strdup (p);
-
-  if (docid[0] == '\0')
-    {
-      g_free (docid);
-      return FALSE;
-    }
-
-  *guessed_docid = docid;
-  return TRUE;
-}
-
-static char *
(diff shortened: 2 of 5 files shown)


src/open-uri.c
```diff
@@ -585,6 +585,38 @@ app_exists (const char *app_id)
   return (info != NULL);
 }
 
+static char *
+get_host_path_for_document_path (const char* document_path,
```
grulja, line 589: Would it make sense for this to be part of `xdp_get_real_path_for_doc_path()` itself?

src/open-uri.c
```diff
@@ -585,6 +585,38 @@ app_exists (const char *app_id)
...
+                                 XdpAppInfo* app_info)
+{
+  char *host_path = xdp_get_real_path_for_doc_path (document_path,
+                                                    app_info);
+  GList *path_parts = NULL;
+  /* Try harder when still getting /run/user path, the document portal
+     could provide directory instead of file or subdirectory. */
+  while (g_str_has_prefix (host_path, "/run/user"))
```
grulja, line 597: Can you maybe move https://github.com/flatpak/xdg-desktop-portal/blob/7928c833597726f7128bc707ee4f0e24868c516a/src/file-chooser.c#L162 to something shared and use instead?

src/open-uri.c
xhorak: I'm not completely sure about it, the `xdp_get_real_path_for_doc_path` is used in the `handle_save_file` to match the `current_file` to the file which is already in the document storage. If these matches, the host path of the file is returned in `current_file`, when they don't, the path relevant to doc id is returned (which could be folder where the file has been saved). 

This might be a bug actually, imagine you have had used the `SaveFiles` for giving access to `/home/user/project_folder`, portal had created the document  storage record and app could later create structure like:
```
/home/user/project_folder
└── new_folder
    └── another_subfolder
        └── file.txt
```
Then the xdp_get_real_path_for_doc_path fails to return the host path of the `file.txt` and the `xdp_get_real_path_for_doc_id` returns host path of `project_folder`. The `current_file` returned to the application will be `project_folder`.

src/open-uri.c
xhorak: I'd rather not, having the doc id won't help me much since I had to recreate the whole path. Using stack like approach makes more sense to me than splitting the path to the pre-id- post-id part and then joining it together to get a full path.

grulja: My fault, my understanding of this function was it just does more advanced check whether the path is document_portal_path, basically equivalent to your `g_str_has_prefix (host_path, "/run/user")`.

grulja: Actually, wouldn't it be a better approach to use the above mentioned method?

You get the doc ID as a fallback so you can lookup the host path on the first try, without trying to get it for every subpath, which might be an overhead, because it's basically a sync DBus call. The tricky part would be to identify what is the remaining part of the path that you would need to append to the returned host path. A call to get the document path for a document ID would be helpful here, but there is none currently, you can only get the host path. 

I don't know if the document path is always in form of `/run/user/$UID/doc/$DOC_ID/filename`, but the documentation to the document store says so, in this case it would be easy, because you would just append what remains after the `/run/user/$UID/doc/$DOC_ID/filename` path.

tests/notification.c
```diff
@@ -545,7 +545,7 @@ test_notification_supported_properties (void)
 
   portal = xdp_portal_new ();
 
-  response = xdp_portal_get_supported_notification_options (portal, &error);
+  //response = xdp_portal_get_supported_notification_options (portal, &error);
```
grulja, line 548: Is this commented out by a mistake? Looks like the reason for the failing test.

swick:
Not a full review but changes like this really need tests.

swick:
There are now python based tests for the document portal (https://github.com/flatpak/xdg-desktop-portal/pull/1594). Would be great if a test for the case which this PR is trying to fix was added.

src/open-uri.c
```diff
@@ -585,6 +585,23 @@ app_exists (const char *app_id)
   return (info != NULL);
 }
 
+static char*
+build_host_path_from_doc_id(const char* doc_id, const char* doc_path)
+{
+  gchar *result;
+  g_autofree char *host_path = xdp_get_real_path_for_doc_id (doc_id);
+  g_autofree char *doc_path_regex = g_strconcat("/run/user/[0-9]+/doc/", doc_id, "/[^/]+(.*)", NULL);
```
swick, line 593: I think this could be avoided if `xdp_looks_like_document_portal_path` would add another out param which returns the path relative to the document.

swick:
Pushed some changes to https://github.com/swick/xdg-desktop-portal/pull/new/wip/document-path-resolve. Didn't look into why the tests are not working reliably, yet.

swick:
Just a note, you can download the container image and `podman run` to reproduce the issue locally. Probably much easier than having to rely on the CI logs.

swick:
This version has my changes to the implementation in xdp and your changes to the tests, with the race fixed: https://github.com/swick/xdg-desktop-portal/tree/wip/document-path-resolve

swick:
The diff between my branch and this PR
```diff
diff --git ./tests/meson.build ../tests/meson.build
index 88e7729e..1cc6a5ed 100644
--- ./tests/meson.build
+++ ../tests/meson.build
@@ -149,6 +149,7 @@ if enable_installed_tests
     '__init__.py',
     'conftest.py',
     'xdp_utils.py',
+    'xdp_doc_utils.py',
     'asan.suppression',
     install_dir: installed_tests_dir / 'tests',
   )
```
causes the CI failure.

There is also a merge commit in the git history which we avoid.

Besides that, the PR now LGTM.

swick (approved):

xhorak:
Ah, thank you for pointing that out and for the review.

mcatanzaro:
> Rebased and added the cleanup for the tests. From my POV this looks good to merge. @GeorgesStavracas @jadahl

Don't forget this!

swick:
(only rebased to make sure everything still works)

src/open-uri.c
```diff
@@ -724,14 +719,17 @@ handle_open_in_thread_func (GTask *task,
...
       if (open_dir)
         {
-          g_autofree char *real_path = xdp_get_real_path_for_doc_path (path, request->app_info);
           /* Try opening the directory via the file manager interface, then
              fall back to a plain URI open */
           g_autoptr(GError) local_error = NULL;
           g_autoptr(GVariant) result = NULL;
           g_autoptr(GVariantBuilder) uris_builder = NULL;
-          g_autofree char* item_uri = g_filename_to_uri (real_path, NULL, NULL);
+          g_autofree char* item_uri = NULL;
```
jadahl: `char* item_uri` -> `char *item_uri`

Issue #1575, opened 2025-01-16
pytest/test_document_fuse exits with unsuccessful status if all tests are skipped


smcv:
### Operating System

Debian unstable

### XDG Desktop Portal version

Git

### XDG Desktop Portal version (Other)

1.19.1-24-g7928c83

### Desktop Environment

Other

### Desktop Environment (Other)

Minimal autobuilder environment

### Expected Behavior

`/dev/fuse` is not available in a minimal autobuilder environment, so `test_document_fuse.py` should be skipped (which in Meson's default test protocol means exit status 77).

### Current Behavior

pytest exits with status 5 as documented in https://docs.pytest.org/en/stable/reference/exit-codes.html, but this is not an exit status that Meson considers to be a skip.

### Steps to Reproduce

`meson test` in a minimal chroot

smcv:
An easy way to reproduce this:

```diff
diff --git a/tests/test_document_fuse.py b/tests/test_document_fuse.py
index 29f9f26..ef88ce4 100644
--- a/tests/test_document_fuse.py
+++ b/tests/test_document_fuse.py
@@ -1290,6 +1290,8 @@ def run_bash(cmd):
     return proc.returncode == 0
 
 
+pytest.skip('no', allow_module_level=True)
+
 if not run_bash("fusermount3 --version"):
     pytest.skip("no fusermount3", allow_module_level=True)
 
```

smcv:
I believe this was fixed when #1578 was merged.

PR #1577, opened 2025-01-16
tests: Translate pytest exit status to Meson/ginsttest-runner convention

smcv:
Resolves: https://github.com/flatpak/xdg-desktop-portal/issues/1575

---

This is a pretty crude way to remap the exit status, but it's better than nothing.

#1525 is probably going to need something a lot like this too.


smcv:
#1578 is better. (I did try doing similarly, but I don't know pytest well enough to make it work!)

PR #1578, opened 2025-01-17
tests: Translate pytest exit status to Meson/ginsttest-runner convention [pytest approach]

swick:
Alternative to https://github.com/flatpak/xdg-desktop-portal/pull/1577 which implements it directly in the pytest conftest which should work transparently for build and installed tests.

/cc @smcv


smcv (commented):
Yes, this is better. I'll try it with autopkgtest.

smcv (approved):

PR #1581, opened 2025-01-17
githooks: Replace pre-commit tool with simple shell script

swick:
This gets is mypy in addition to ruff check and ruff format. I spent way too many hours trying to make things work with pre-commit and I don't think it's even possible. CI runs the check script directly, the git-hooks integration only when a commit changes something in tests.

    githooks: Replace pre-commit tool with simple shell script
    
    It's basically impossible to run mypy with pre-commit because it will
    ignore the repo config and check all files changed the commit. The
    dbusmock templates are special and mypy fails on them so we use the mypy
    config to skip them.
    
    This introduces a new script which checks the python tests and installs
    all the required tools into a python virtual environment. It also
    changes to pre-commit hook to call the script if any of the files in
    tests changed and uses git stash push/pop to get the test a clean state
    of the repo.


.githooks/check-tests-python.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PYTHON=${PYTHON:-python3}
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+VENV="${HERE}/.venv"
+
+[ -d $VENV ] || $PYTHON -m venv $VENV
+source $VENV/bin/activate
```
whot, line 10:
```suggestion
[ -d "$VENV" ] || $PYTHON -m venv "$VENV"
source "$VENV/bin/activate"
```

whot (commented):
Isn't this throwing out the baby with the bathwater? pre-commit is relatively standard and if mypy is the issue here would it make more sense to just treat mypy special (e.g. even invoking it as part of meson test) and leave everything else that works to pre-commit?

Or like [this example](https://jaredkhan.com/blog/mypy-pre-commit) to wrap the mypy run into a pre-commit local script so we control mypy only. But I guess you found that one so - maybe running mypy as part of meson test is the easiest?

Oh, and adding this line to the top of really stubborn files that we cannot fix seems to help (see [docs](https://mypy.readthedocs.io/en/stable/common_issues.html#ignoring-a-whole-file):

```python
# mypy: disable-error-code="annotation-unchecked"
```

What were the specific unfixable issues you ran into here?

swick:
Alright, a fresh mind seems to have helped. Managed to make mypy work with pre-commit by using a bunch of `mypy` annotations.

whot (approved):

tests/test_document_fuse.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
 # This file is formatted with Python Black
+# mypy: ignore-errors
```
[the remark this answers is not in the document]
swick: Figured out the issue and it was a real bug. Added two commits at the beginning to address it.

tests/test_document_fuse.py
```diff
@@ -28,7 +28,7 @@ def filename_to_ay(filename):
 dir_prefix = "dir"
 ensure_no_remaining = True
 
-running_count = {}
+running_count: dict[str, int] = {}
```
whot, line 31: slightly unrelated but this is a use-case for a [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)
```py
running_count: dict[str, int] = defaultdict(int)

def get_a_count(counter: str):
    global running_count
    running_count[counter] += 1
    return running_count[counter]
```

whot (approved):
noice!

PR #1590, opened 2025-01-20
settings: Add animation-scale, status-shapes, and always-show-scrollbars keys

orowith2os:
This allows applications to follow more accessibility preferences across different desktops, without relying on keys under the org.gnome namespace. See the descriptions of the keys themselves for more details.

Client implementations:

- [ ] GTK
- [ ] QT

Portal implementations: 
- [x] GTK: https://github.com/flatpak/xdg-desktop-portal-gtk/pull/510
- [ ] GNOME
- [ ] KDE
- [ ] Cosmic
- [ ] Pantheon

Ack on Budgie, from @serebit

### DIFF
--- a/data/org.freedesktop.impl.portal.Settings.xml
+++ b/data/org.freedesktop.impl.portal.Settings.xml
@@ -58,6 +58,27 @@
 
       Unknown values should be treated as 0 (no preference).
 
+    * ``org.freedesktop.appearance`` ``animation-scale`` (``d``)
+
+      Indicates whether the system prefers an interface that removes, 
+      speeds up, slows down, or replaces non-essential animations.
+      Supported values are:
+
+      - ``0.0``: All non-essential animations disabled
+      - ``1.0``: All non-essential animations enabled
+
+      Unknown values should be treated as 1.0 (all non-essential animations enabled).
+
+    * ``org.freedesktop.appearance`` ``status-shapes`` (``b``)
+
+      Indicates whether the system prefers an interface that uses shapes
+      to indicate status in addition to or instead of color. 
+
+    * ``org.freedesktop.appearance`` ``always-show-scrollbars`` (``b``)
+
+      Indicates whether the system prefers an interface where scrollbars
+      are always visible. 
+
     Implementations can provide other keys; they are entirely
     implementation details that are undocumented. If you are a
     toolkit and want to use this please open an issue.
--- a/data/org.freedesktop.portal.Settings.xml
+++ b/data/org.freedesktop.portal.Settings.xml
@@ -62,7 +62,28 @@
 
       Unknown values should be treated as ``0`` (no preference).
 
-    This documentation describes version 2 of this interface.
+    * ``org.freedesktop.appearance`` ``animation-scale`` (``d``)
+
+      Indicates whether the system prefers an interface that removes, 
+      speeds up, slows down, or replaces non-essential animations.
+      Supported values are:
+
+      - ``0.0``: All non-essential animations disabled
+      - ``1.0``: All non-essential animations enabled
+
+      Unknown values should be treated as 1.0 (all non-essential animations enabled).
+
+    * ``org.freedesktop.appearance`` ``status-shapes`` (``b``)
+
+      Indicates whether the system prefers an interface that uses shapes
+      to indicate status in addition to or instead of color. 
+
+    * ``org.freedesktop.appearance`` ``always-show-scrollbars`` (``b``)
+
+      Indicates whether the system prefers an interface where scrollbars
+      are always visible. 
+
+    This documentation describes version 3 of this interface.
   -->
   <interface name="org.freedesktop.portal.Settings">
 


orowith2os:
Noting that all of these keys are already exposed within GNOME Settings - the only change here this is realistically making, is making them cross-platform. There should be no issue implementing this, as I expect most clients supporting this via the old interfaces (GNOME namespace, etc) should wouldn't have an issue moving to something more global. 

GNOME can implement this in their portal implementation quite easily, as it's just reading their gsettings and putting them in the Settings portal. KDE and other desktops may require a bit more work.

Maybe just one implementation in xdp-gtk is enough to get this in, plus patches for GTK?

orowith2os:
I'm done writing the xdp-gtk implementation, if somebody on the xdp-gnome and xdp-kde portals gives me the go-ahead, I'll work on the xdp-gnome one.

data/org.freedesktop.impl.portal.Settings.xml
```diff
@@ -58,6 +58,21 @@
 
       Unknown values should be treated as 0 (no preference).
 
+    * ``org.freedesktop.appearance`` ``reduce-animation`` (``b``)
```
orowith2os, line 61: I wonder how this would play with desktops that want to set custom animation speeds instead of just disabling them entirely... Maybe changing this to a number, and applications can disable animations if it's 0, and if they don't support changing animation speed, run as normal otherwise?

chrisawi: I think less is more here. What would a numeric value mean, generically? Percentage of default animation duration?

You linked to the [web spec](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) eariler, and it says about Firefox's implementation:

> For Firefox, the reduce request is honoured if:
> [...]
>   *  In Plasma/KDE: System Settings > Workspace Behavior -> General Behavior > "Animation speed" is set all the way to right to "Instant".

So x-d-p-kde could reasonably implement a boolean setting the same way.

TingPing:
I'm a fan of `reduce-animation`. Not sure about the others.

orowith2os:
@TingPing all of these are available within GNOME Settings, so imo, if they're not suitable as a key on the Settings portal, they shouldn't be in Settings either.

nicolasfella:
In KDE we use an animation speed factor of 0 to denote "No animations".

As far as I can tell Qt doesn't have any API to expose a "No animations" preference, but it probably should.

reduce-animations and status-shapes seem like _almost_ the same thing to me? From a Qt perspective it also matters how other platforms (Windows, macOS, Android, ...) model such things to find a common enough abstraction

orowith2os:
> reduce-animations and status-shapes seem like almost the same thing to me?

This is with status-shapes - without it, this could potentially affect grayscale users (with the recent Wellbeing changes in GNOME, or colorblind individuals). They are absolutely not the same thing.

image

nicolasfella:
> > reduce-animations and status-shapes seem like almost the same thing to me?
> 
> This is with status-shapes - without it, this could potentially affect grayscale users (with the recent Wellbeing changes in GNOME, or colorblind individuals). They are absolutely not the same thing.

Oh nevermind me, I didn't read correctly :)

nicolasfella:
> As far as I can tell Qt doesn't have any API to expose a "No animations" preference, but it probably should.

I opened https://bugreports.qt.io/browse/QTBUG-133042 for this

nicolasfella:
> Indicates whether the system prefers an interface where scrollbars are always visible. 

The closest thing I can find to this in Qt is QStyle::SH_ScrollBar_Transient. It's described as 

> Determines if the style supports transient scroll bars. Transient scroll bars appear when the content is scrolled and disappear when they are no longer needed.

It's phrased the other way around than here, it's more about allowing/using transient scrollbars rather than preventing them. It seems to be always false except on macOS where this is a system preference. 

My question would be, if this value is false or not set, should application take this as _encouragement_ to use overlay scrollbars?

As for KDE: As a design choice we pretty much never use overlay scrollbars, so our portal would most likely always return false here

nicolasfella:
As for status-shapes: I can't think of a concrete place where I would make use of this in Qt/KDE. We could easily hook up the portal to a new entry in the accessibility settings module though.

I wonder though if it should be called something like no-color or avoid-color. That would make its purpose more clear to me

orowith2os:
> My question would be, if this value is false or not set, should application take this as encouragement to use overlay scrollbars?

With scrollbars, it's not meant to enable or disable them entirely, but rather configure whether they should always be available, or only when they're useful (such as if the content is currently being scrolled).

> As for status-shapes: I can't think of a concrete place where I would make use of this in Qt/KDE. We could easily hook up the portal to a new entry in the accessibility settings module though.

Perhaps toggles, where they'd do like GNOME does and have checks or X'es depending on the state. But I don't think KDE should have to handle this, really. If it doesn't see a use for it, would it be fine just hard coding `false`?

> I wonder though if it should be called something like no-color or avoid-color. That would make its purpose more clear to me

I don't think colors and shapes are mutually exclusive. Anything involving telling clients about colors would be contained in color deficiency hints, not really something like this. One can have colors and shapes to represent a UI at the same time.

swick:
I'd rather have a setting which talks about color vision deficiencies than "status-shapes".

data/org.freedesktop.impl.portal.Settings.xml
orowith2os: I will also note that being able to set custom animation speeds system-wide may also assist in debugging and developing animations. I think we should have the ability for this to be a number, since it would already be usable as a boolean anyways (0 = no animation, anything nonzero = use animation).

orowith2os:
I'm going to switch reduce-animation over to something more future-proof unless someone raises concerns with me about it.

orowith2os:
If anybody wants to push this with reduce-animation as a bool instead of animation-scale as a double, the last commit containing that is https://github.com/flatpak/xdg-desktop-portal/commit/525d4061f6ea5071b0b6a87648639a2385649beb. The xdp-gtk branch is here: https://github.com/orowith2os/xdg-desktop-portal-gtk/tree/a11y-settings-bool

PR #1594, opened 2025-01-21
tests/documents: Add document portal tests

swick:
This one fell through the cracks for some reason.

### DIFF
--- a/tests/test_documents.py
+++ b/tests/test_documents.py
@@ -1,0 +1,381 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# This file is formatted with Python Black
+
+import tests as xdp
+
+import pytest
+import dbus
+from pathlib import Path
+import os
+from gi.repository import GLib
+
+
+def path_from_null_term_bytes(bytes):
+    path_bytes = bytes.split(b"\x00")[0]
+    return Path(os.fsdecode(path_bytes))
+
+
+def get_mountpoint(documents_intf):
+    mountpoint = documents_intf.GetMountPoint(byte_arrays=True)
+    mountpoint = path_from_null_term_bytes(mountpoint)
+    assert mountpoint.exists()
+    return mountpoint
+
+
+def write_bytes_atomic(file_path, bytes):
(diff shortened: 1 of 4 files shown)


swick:
CI is failing because there is a leak in the document portal that the tests uncovered.

tests/test_documents.py
```diff
@@ -0,0 +1,381 @@
...
+import pytest
+import dbus
+from pathlib import Path
+import os
+from gi.repository import GLib
+
+
+def path_from_null_term_bytes(bytes):
+    path_bytes = bytes.split(b"\x00")[0]
+    return Path(os.fsdecode(path_bytes))
```
smcv, line 16: Shouldn't this assert that the path contains exactly one `\0`, and it is the last byte?

tests/test_documents.py
```diff
@@ -0,0 +1,381 @@
...
+
+    return doc_id
+
+
+def export_file_named(documents_intf, folder_path, name, unique=False):
+    assert folder_path.exists()
+
+    # why does this need to be NULL terminated? otherwise:
+    # dbus.exceptions.DBusException: org.freedesktop.portal.Error.InvalidArgument
+    name_nt = name.encode("utf8") + b"\x00"
```
smcv, line 55: This is necessary because when you give dbus-python a `bytes` object, it assumes you want to send exactly those bytes. It does not know about the "bytestring" convention that originated in GVariant (representing "text in an unspecified encoding" as the bytes + `\0`), which is newer than the dbus-python API design.

This should probably be `os.fsencode(path) + b"\x00"` to handle being run from a directory whose name is in a legacy non-UTF-8 encoding.

tests/test_documents.py
swick: I'm unsure. In C you'd pass around the pointer and everything up to the first `\0` is what things would work with, and the python code here mimics that behavior.

swick:
This now contains https://github.com/flatpak/xdg-desktop-portal/pull/1601 which should make CI happy.

GeorgesStavracas (approved):

PR #1595, opened 2025-01-21
app-info/host: Only use the cgroup app id if we have a matching GAppInfo

swick:
This should prevent cases where the connection is from something that's not actually an app but follows the systemd cgroup standard.

This is safe because
* it's a host app anyway
* we always use an empty app id when x-d-p is built without systemd
* it's the default when the systemd cgroup standard is not followed


swick:
I still think this is a good idea because it should lead to fewer bad surprises.

AdrianVovk (approved):
LGTM!

Issue #1583, opened 2025-01-18
Dynamic Launcher portal fails to install when ~/.local/share/applications/ does not exist yet


IlyaBizyaev:
### Operating System

Fedora KDE 41

### XDG Desktop Portal version

1.18

### Desktop Environment

KDE

### Expected Behavior

`~/.local/share/applications/` is created if it does not exist yet (e.g., on a newly installed system)

### Current Behavior

An error is returned by the Install method when it fails to create a symlink in a non-existent directory

(tested with https://flathub.org/apps/com.belmoussaoui.ashpd.demo)

### Steps to Reproduce

1. Ensure that there is no `~/.local/share/applications/`
2. Install and launch ASHPD Demo
3. From the "Dynamic Launcher" page, select /usr/share/icons/Adwaita/16x16/devices/audio-headphones.png as the icon and click "Install"

### Anything else we should know?

https://github.com/flatpak/xdg-desktop-portal/blob/736002f95a6bfd22b8076154fe20acba0471a03d/src/dynamic-launcher.c#L278

swick:
Could you try https://github.com/flatpak/xdg-desktop-portal/pull/1596?

IlyaBizyaev:
Yep, this fixes it :)

PR #1596, opened 2025-01-21
dynamic-launcher: Ensure $XDG_DATA_HOME/applications exists

swick:
It is possible that the directory has not been created by anything else before which would make dynamic launcher creation fail.

Resolves: #1583


swick:
Side note: this portal is missing tests. Someone should write them.

swick:
Had a bit of time, added a basic test, found a memory leak, a double free and a misleading error message.

GeorgesStavracas (approved):

PR #1599, opened 2025-01-23
screenshot: Forward to modality to access dialog

Sodivad:


swick:
I sound a bit like a broken record right now, but we should add tests! This one requires a non-interactive screenshot request which shall result in the access backend getting the right data.

tests/test_screenshot.py
```diff
@@ -26,18 +26,27 @@ def required_templates():
         },
     }
 
+def screenshot_and_access():
+    return {
+        "screenshot": {
+            "results": SCREENSHOT_DATA,
+        },
+        "access": {
```
swick, line 34: I think it's fine to add this to `required_templates` instead of parametrizing

swick:
https://flatpak.github.io/xdg-desktop-portal/docs/building-and-running.html#testing helps you with enabling git hooks for the python checks/formatting

tests/test_screenshot.py
Sodivad: For some reason that I cannot figure out if I add `access` to `required_templates` all other test cases fail.
The screenshot template is loaded with the correct param but it returns 0 when we want to close or cancel instead.

tests/templates/__init__.py
```diff
@@ -107,6 +107,9 @@ def export(self, close_callback: Optional[Callable] = None):
         self._close_callback = close_callback
         return self
 
+    def unexport(self):
```
swick, line 110: seems like a good idea, but deserves its own commit with an explanation

swick:
Thanks for adding tests!

tests/test_screenshot.py
swick: You found a big issue with the way the test harness starts the templates. It has to load templates into the same process because they are on the same bus, and they are on the same object because that's just how the portal system works, and that means that we call AddTemplate on the same object which means `def load` in all the different templates actually operates on the same python object.

I'll try to find a better solution for this, but in the meantime, move access over screenshot, and it should work.

tests/templates/access.py
```diff
@@ -53,9 +53,10 @@ def closed_callback():
             logger.debug(f"AccessDialog Close() response {response}")
             cb_success(response.response, response.results)
 
-        def reply_callback():
+        def reply_callback(request):
             response = Response(self.response, {})
             logger.debug(f"AccessDialog with response {response}")
+            request.unexport()
```
swick, line 59: The ImplRequest already unexports the object when the request is being closed, so I'm not entirely sure why this is needed?

Sodivad: But it is not closed. The screenshot portal will use the same request handle for the access and screenshot portal call.  (Actually I just notice without hooking up the access impl request to the app request).

Anyway  if we do not unexport here, the screenshot portal will not be able to add a mock for the request.
` xdg-desktop-portal-WARNING **: 13:41:59.942: A backend call failed: object /org/freedesktop/portal/desktop/request/1_7/request3 already exists`

I checked xdg-desktop-portal-gtk and it works the same. It unexports the request when sending the reponse.

swick: Right, I think we have a more fundamental problem here and don't close requests in a lot of cases where we should (i.e. whenever we finished handling the request).

swick:
Alright, LGTM.

tests/templates/access.py
swick: To avoid that problem in the future: https://github.com/flatpak/xdg-desktop-portal/pull/1607

tests/test_screenshot.py
swick: Trying to mitigate the strongest offenders here: https://github.com/flatpak/xdg-desktop-portal/pull/1608

PR #1600, opened 2025-01-23
RFC: Only grant permission once for unidentified apps

aleasto:
The status-quo is problematic because:
* It is not comunicated well enough that granting the permission would also grant it for any other unidentified program. That includes all unsandboxed programs started from a terminal, without a systemd scope.
* It is not true, or not true for all Desktops, that "This permission can be changed at any time from the privacy settings." because the "privacy settings" app would need to have a section dedicated to unidentified apps, which gnome-control-center for example does not have.

Both problems could be addressed differently, but I believe this is a better solution overall:

I propose we don't use the PermissionStore for unidentified apps, at all. It is generally unsensible to grant such a wide permission, and I don't think we should allow it.
Instead, such unidentified apps will always prompt whenever they want to take a privileged action.

### DIFF
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -206,6 +206,7 @@ handle_screenshot_in_thread_func (GTask *task,
   gboolean interactive;
   const char *parent_window;
   const char *app_id;
+  gboolean store_permission;
 
   REQUEST_AUTOLOCK (request);
 
@@ -213,10 +214,15 @@ handle_screenshot_in_thread_func (GTask *task,
   parent_window = ((const char *)g_object_get_data (G_OBJECT (request), "parent-window"));
   options = ((GVariant *)g_object_get_data (G_OBJECT (request), "options"));
 
+  /* Only memorize the permission for applications that can be uniquely
+   * identified by their app ID.
+   */
+  store_permission = g_strcmp0 (app_id, "") != 0;
+
   if (xdp_dbus_impl_screenshot_get_version (impl) < 2)
     goto query_impl;
 
-  permission = xdp_get_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID);
+  permission = store_permission ? xdp_get_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID) : XDP_PERMISSION_UNSET;
 
   if (!g_variant_lookup (options, "interactive", "b", &interactive))
     interactive = FALSE;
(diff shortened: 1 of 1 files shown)


aleasto:
I've only made changes to the screenshot portal, but the idea would be to replicate the same changes to other interfaces which involve the PermissionStore

src/screenshot.c
```diff
@@ -206,17 +206,23 @@ handle_screenshot_in_thread_func (GTask *task,
...
+  /* Only memorize the permission for applications that can be uniquely
+   * identified by their app ID.
+   */
+  store_permission = g_strcmp0 (app_id, "") != 0;
+
   if (xdp_dbus_impl_screenshot_get_version (impl) < 2)
     goto query_impl;
 
-  permission = xdp_get_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID);
+  permission = store_permission ? xdp_get_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID) : XDP_PERMISSION_UNSET;
```
aleasto, line 225: This is a upgrade path, so that if we previously had stored a permission for `app_id=""` we will stop honouring it from now on.

aleasto:
As a self-critique to the change: it would get annoying if one or more unidentified apps keep requesting bogus permissions that the user does not intend on granting, because we don't memorize the preference.

It would obviously be the application's fault, and I would say that one should uninstall such application. On the other hand, if we can't identify the app we can't also provide the user with the info about which application to uninstall.

I would appreciate to hear your thoughts and some actionable steps to mitigate this, if you feel that is is necessary.

swick:
I like the idea that requests from unidentified apps (`app_id == ""`) shall always prompt the user. I do however think that we should only make changes in that direction after https://github.com/flatpak/xdg-desktop-portal/pull/1521 has landed because that gives apps a way to avoid running into this case.

aleasto:
> Furthermore, what is this proposal to inform about uninstalling an app?

That is obviously not what I intended. My point is that if an unidentified application keeps asking for some permission but we don't show the name of the app (because we don't know it), then the user wouldn't know how to take action.

Mikenux:
No need to tell the user to take an action if no permission prompt is shown, at least for Allow/Deny permissions...

Sodivad:
I have a draft change locally that would add an "ask always" checkbox to the dialog via the `choices` feature. Maybe this might be a direction we could pursue?

PR #1601, opened 2025-01-23
document-portal: Track open files

swick:
The fuse lowlevel ops release vfunc says:

  For every open call there will be exactly one release call (unless the
  filesystem is force-unmounted).

Which means that if the filesystem is force-unmounted, there won't be a release call which can free a XdpFile allocated in open or create.

Introduce a list which keeps track of open files and cleans them up in the end so that we don't leak the memory. Technically this leak doesn't hurt because we will terminate anyway, but otherwise, asan will be very unhappy and we get random test failures.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -144,6 +144,9 @@ static pthread_t fuse_pthread = 0;
 static uid_t my_uid;
 static gid_t my_gid;
 
+static GList *open_files = NULL;
+G_LOCK_DEFINE (open_files);
+
 /* from libfuse */
 #define FUSE_UNKNOWN_INO 0xffffffff
 
@@ -242,6 +245,7 @@ struct _XdpInode {
 
 typedef struct {
   int fd;
+  GList *link;
 } XdpFile;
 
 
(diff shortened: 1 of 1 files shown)


document-portal/document-portal-fuse.c
```diff
@@ -1901,12 +1905,23 @@ xdp_file_new (int fd)
...
+  file->link = open_files;
+
   return file;
 }
 
 static void
 xdp_file_free (XdpFile *file)
 {
+  GList *link = file->link;
+  file->link = NULL;
```
GeorgesStavracas, line 1920: Perhaps:

```suggestion
  GList *link = g_steal_pointer (&file->link);
```

document-portal/document-portal-fuse.c
```diff
@@ -3544,6 +3559,7 @@ xdp_fuse_init (GError **error)
   my_gid = getgid ();
 
   all_inodes = g_hash_table_new_full (g_int64_hash, g_int64_equal, NULL, NULL);
+  open_files = NULL;
```
GeorgesStavracas, line 3562: The assignment doesn't seem necessary, perhaps a `g_assert (open_files == NULL);` to be extra safe

hfiguiere:
I wonder if that idea can help with #1234 - my proposed PR for it is just wrong.

GeorgesStavracas:
> I wonder if that idea can help with #1234 - my proposed PR for it is just wrong.

Off hand it doesn't seem like it - AFAIU this PR only helps with leaked objects when the process is shutting down - but worth testing, I guess

GeorgesStavracas (approved):

PR #1604, opened 2025-01-26
build: Change the pytest invocation to run a specific file

swick:
Instead of filtering with -k. This is more precise and allows adding additional filters with -k more easily.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -348,7 +348,7 @@ if enable_pytest
     test(
       'pytest/@0@'.format(testname),
       pytest,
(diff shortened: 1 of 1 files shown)


Sodivad:
would it allow selecting a specific method with -k?

swick:
Yes, `meson test -C build pytest/test_trash --test-args='-k test_trash_file_fails'` would run a single test for example.

Sodivad:
Works beautifully. LGTM.

GeorgesStavracas:
Thanks for testing it

PR #1606, opened 2025-01-27
Pytest improvements

swick:
First two commits are very mechanical, the third one lets one override the dbus timeout using an environment variable.


whot (approved):
noice :)

PR #1607, opened 2025-01-27
tests/templates: Rework ImplRequest

swick:
The portal backend mocks would all implement the same pattern by copy
    and pasting a bunch of code which is not that easy to understand and
    also subtly broken because they would often not unexport the request
    object.
    
    This changes the ImplRequest to own the control flow and do all the
    right things in a single place and provides callbacks for more advanced
    use cases.

### DIFF
--- a/tests/templates/__init__.py
+++ b/tests/templates/__init__.py
@@ -8,7 +8,7 @@ import dbusmock
 import logging
 
 
-def init_logger(name: str):
+def init_logger(name: str) -> logging.Logger:
     """
     Common logging setup for the impl.portal templates. Use as:
 
--- a/tests/templates/__init__.py
+++ b/tests/templates/__init__.py
@@ -25,7 +25,7 @@ def init_logger(name: str):
     return logger
 
 
-logger = init_logger("request")
+logger = init_logger("utils")
 
 
 class Response(NamedTuple):
--- a/tests/templates/__init__.py
+++ b/tests/templates/__init__.py
@@ -3,6 +3,7 @@
 # This file is formatted with Python Black
 
 from typing import Callable, Dict, Optional, NamedTuple
+from gi.repository import GLib
 import dbus
 import dbusmock
 import logging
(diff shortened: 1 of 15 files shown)


tests/templates/__init__.py
```diff
@@ -35,62 +36,103 @@ class Response(NamedTuple):
 
 class ImplRequest:
     """
-    Implementation of a org.freedesktop.impl.portal.Request object. Typically
-    this object needs to be merely exported:
+    Implementation of a `org.freedesktop.impl.portal.Request`` object exposed
```
GeorgesStavracas, line 39:
```suggestion
    Implementation of a ``org.freedesktop.impl.portal.Request`` object exposed
```

tests/templates/__init__.py
```diff
@@ -35,62 +36,103 @@ class Response(NamedTuple):
 
 class ImplRequest:
     """
-    Implementation of a org.freedesktop.impl.portal.Request object. Typically
-    this object needs to be merely exported:
+    Implementation of a ``org.freedesktop.impl.portal.Request`` object exposed
```
whot, line 39:
```suggestion
    Implementation of an ``org.freedesktop.impl.portal.Request`` object exposed
```

might as well change this while here :)

tests/templates/__init__.py
```diff
@@ -35,62 +36,103 @@ class Response(NamedTuple):
...
         bus = mock.connection
         proxy = bus.get_object(busname, handle)
-        mock_interface = dbus.Interface(proxy, dbusmock.MOCK_IFACE)
+        self.mock_interface = dbus.Interface(proxy, dbusmock.MOCK_IFACE)
+        self.handle = handle
+        self.logger = logger
+        self.cb_success = cb_success
+        self.cb_error = cb_error
+
+    def respond(self, response, delay=0, done_cb=None):
```
whot, line 67:
```suggestion
    def respond(self, response: Callable | Response, delay=0, done_cb=None):
```
would be a nice annotation here (I think that's the type)

tests/templates/wallpaper.py
```diff
@@ -34,29 +33,24 @@ def load(mock, parameters={}):
...
-            logger.debug(f"scheduling delay of {self.delay}")
-            GLib.timeout_add(self.delay, reply_callback)
-    except Exception as e:
-        logger.critical(e)
-        cb_error(e)
+    logger.debug(
+        f"SetWallpaperURI({handle}, {app_id}, {parent_window}, {uri}, {options})"
+    )
+
+    # This is unregular: the backend doesn't return the results vardict
```
whot, line 40:
```suggestion
    # This is irregular: the backend doesn't return the results vardict
```

tests/templates/__init__.py
```diff
@@ -35,62 +36,103 @@ class Response(NamedTuple):
...
+                "RequestClosed",
+                "s",
+                (self.handle,),
+            )
+
+            if close_callback:
+                try:
+                    close_callback()
+                except Exception as e:
+                    logger.critical(f"Request {self.handle}: failed running close: {e}")
```
whot, line 114:
```suggestion
                    logger.critical(f"Request {self.handle}: failed running close callback: {e}")
```

whot (approved):

tests/templates/__init__.py
swick: Yeah, I forgot to add typing annotations because mypy and the templates don't work well together so I excluded them.

PR #1608, opened 2025-01-27
tests/templates: Store config in dataclass

swick:
The templates are loaded onto the MAIN_OBJ which is the same for all
    portal impls. That means if a test case sets up multiple templates, they
    all operate on the same object.
    
    If we set, for example, `mock.delay` in both templates, the latter will
    overwrite the former.
    
    This commit loads the parameters into a dataclass that's available as
    mock.${classname}_params.
    
    Technically we have similar issues whenever we access self in the dbus
    method implementations, but for the most part we use different variables
    in different templates.

depends on https://github.com/flatpak/xdg-desktop-portal/pull/1607

### DIFF
--- a/tests/templates/__init__.py
+++ b/tests/templates/__init__.py
@@ -25,7 +25,7 @@ def init_logger(name: str):
     return logger
 
 
-logger = init_logger("request")
+logger = init_logger("utils")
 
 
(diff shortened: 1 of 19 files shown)


whot (approved):
:+1:

PR #1609, opened 2025-01-28
Include host app registry in the docs

GeorgesStavracas:
See commits

### DIFF
--- a/doc/api-reference.rst
+++ b/doc/api-reference.rst
@@ -4,8 +4,20 @@ API Reference
 Portal interfaces are available to sandboxed applications with the default
 filtered session bus access of Flatpak.
 
-Unless otherwise specified, they appear under the bus name ``org.freedesktop.portal.Desktop``
-and the object path ``/org/freedesktop/portal/desktop`` on the session bus.
+Desktop portals appear under the bus name ``org.freedesktop.portal.Desktop``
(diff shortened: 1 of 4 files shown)


doc/api-reference.rst
```diff
@@ -4,8 +4,16 @@ API Reference
...
 
-Unless otherwise specified, they appear under the bus name ``org.freedesktop.portal.Desktop``
-and the object path ``/org/freedesktop/portal/desktop`` on the session bus.
+Desktop portals appear under the bus name ``org.freedesktop.portal.Desktop``
+and the object path ``/org/freedesktop/portal/desktop`` on the session bus,
+unless specified otherwise.
+
+Apps running on the host system have access to a special interface for
+registering the app with XDG Desktop Portal. Registering a host app with
+XDG Desktop Portal improves the user experience when the app uses portals.
```
swick, line 13: Maybe make it a bit more concrete?
```suggestion
registering the app with XDG Desktop Portal. Registering a host app with XDG Desktop Portal overwrites the automatic detection based on the XDG cgroup pathname standardization for applications (https://systemd.io/DESKTOP_ENVIRONMENTS/#xdg-standardization-for-applications). This might improve the user experience when the host app was launched in a way that doesn't follow the standard.
```

PR #1614, opened 2025-01-31
Transition host apps to registered state if necessary

GeorgesStavracas:


src/xdp-app-info.c
```diff
@@ -923,6 +923,64 @@ xdp_invocation_ensure_app_info_sync (GDBusMethodInvocation  *invocation,
...
+        return NULL;
+
+      registered_app_info = xdp_app_info_host_new_registered (pidfd, app_id, error);
+      if (!registered_app_info)
+        return NULL;
+    }
+
+  g_debug ("Replacing host app info with a registered one for '%s'", xdp_app_info_get_id (app_info));
+
+  cache_insert_app_info (sender, app_info);
```
grulja, line 979: Two things:

It should be `registered_app_info` and not `app_info` and also a new function like `cache_replace_app_info()` should be added, because from my understanding `g_hash_table_insert()` doesn't replace existing item so one has to use `g_hash_table_replace()`.

grulja:
I have been testing this now with Firefox. The issue I hit with the original code is that when running Firefox from a command line, which is obvious since I've been testing a new feature. It fails to register, because it already has an entry it guessed from the PID, but it ends up with an empty app ID. For that reason I tried to apply this change which fixes the issue for me, but I had to fix the two issues I mentioned.

src/xdp-app-info-host.c
```diff
@@ -200,6 +202,7 @@ xdp_app_info_host_new_full (const char *app_id,
                            /* supports_opath */ TRUE,
                            /* has_network */ TRUE,
                            /* requires_pid_mapping */ FALSE);
+  app_info_host->registered = registered;
```
swick, line 205: The XdpAppInfoTest also can be registered which is needed for the tests and they now have divergent behavior.

src/xdp-app-info.c
GeorgesStavracas: `g_hash_table_insert` replaces the value, and unrefs the previous value

src/xdp-app-info-host.c
GeorgesStavracas: I've reworked XdpAppInfo internals quite intensely in this MR, and I think that's about as far as I can go without turning this PR into a monstrosity. Should make it easier to do any follow-up if necessary.

src/xdp-app-info.c
```diff
@@ -1104,6 +1104,60 @@ xdp_invocation_ensure_app_info_sync (GDBusMethodInvocation  *invocation,
                                               cancellable,
                                               error);
 }
```
jadahl, line 1106: Missing empty line.

src/xdp-app-info.c
```diff
@@ -1104,6 +1104,60 @@ xdp_invocation_ensure_app_info_sync (GDBusMethodInvocation  *invocation,
...
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  const char *sender = g_dbus_method_invocation_get_sender (invocation);
+  g_autoptr(XdpAppInfo) registered_app_info = NULL;
+  XdpAppInfo *app_info;
+
+  g_assert (app_info_by_unique_name != NULL);
+
+  G_LOCK (app_infos);
+  app_info = g_hash_table_lookup (app_info_by_unique_name, sender);
+  G_UNLOCK (app_infos);
```
jadahl, line 1122: Might want to hold the lock until the end of this function, and have a "already locked" version of `cache_insert_app_info()`.

tests/test_registry.py
```diff
@@ -78,17 +78,14 @@ def test_late_register(self, portals, dbus_con, app_id):
         registry_intf = xdp.get_portal_iface(dbus_con, "Registry", domain="host")
         mock_intf = xdp.get_mock_iface(dbus_con)
 
-        expected_app_id = app_id
-        unexpected_app_id = "org.example.CorrectAppId"
-
         session = self.create_dummy_session(dbus_con)
 
+        expected_app_id = "org.example.WrongAppId"
```
jadahl, line 83:
```suggestion
        expected_app_id = app_id
```

jadahl (commented):
The first few commits can land independently I guess, but the last few that moves towards allowing transitioning is what needs consideration.

The portal service going away is an annoying race to deal with, and with multiple entities sharing the same connection (GApplication, app and libportal) I guess it'll be hard to handle registration when the name disappears and reappears, so with that in mind, this "best effort" might be the best we can do.

What is a different but related concern is that it might paper over incorrect implementations where an app does portal call, then registers, when starting up. A bit tricky to distinguish a broken app, from an app having trouble racing to win recovering from the portal service restarting.

GeorgesStavracas:
> The first few commits can land independently I guess

Opened https://github.com/flatpak/xdg-desktop-portal/pull/1619 with them.

> What is a different but related concern is that it might paper over incorrect implementations where an app does portal call, then registers, when starting up. A bit tricky to distinguish a broken app, from an app having trouble racing to win recovering from the portal service restarting.

Yeah, this is the main reason I'm not sure about this PR. I thought about adding a warning to scare app developers slightly, but that will probably be insufficient.

PR #1619, opened 2025-02-07
XdpAppInfo cleanups

GeorgesStavracas:
See commits. This is the first part of https://github.com/flatpak/xdg-desktop-portal/pull/1614

### DIFF
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -732,8 +732,6 @@ xdp_app_info_flatpak_new (int      pid,
   const char *group;
   g_autofree char *id = NULL;
   g_autofree char *instance = NULL;
-  g_autofree char *desktop_id = NULL;
-  g_autoptr(GAppInfo) gappinfo = NULL;
   g_auto(GStrv) shared = NULL;
   gboolean has_network;
   g_autofd int bwrap_pidfd = -1;
@@ -788,9 +786,6 @@ xdp_app_info_flatpak_new (int      pid,
   if (instance == NULL)
     return NULL;
 
-  desktop_id = g_strconcat (id, ".desktop", NULL);
-  gappinfo = G_APP_INFO (g_desktop_app_info_new (desktop_id));
-
   shared = g_key_file_get_string_list (metadata,
                                        FLATPAK_METADATA_GROUP_CONTEXT,
                                        FLATPAK_METADATA_KEY_SHARED,
@@ -826,7 +821,6 @@ xdp_app_info_flatpak_new (int      pid,
                                      error,
                                      "engine", FLATPAK_ENGINE_ID,
                                      "flags", flags,
-                                     "g-app-info", gappinfo,
                                      "id", id,
                                      "instance", instance,
                                      "pidfd", bwrap_pidfd,
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -824,10 +824,11 @@ xdp_app_info_flatpak_new (int      pid,
                                      NULL,
                                      error,
                                      "engine", FLATPAK_ENGINE_ID,
+                                     "id", id,
                                      NULL);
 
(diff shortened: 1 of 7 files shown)


src/xdp-app-info.c
```diff
@@ -67,7 +67,6 @@ typedef struct _XdpAppInfoPrivate
   GAppInfo *gappinfo;
   gboolean supports_opath;
   gboolean has_network;
-  gboolean requires_pid_mapping;
```
swick, line 70: Maybe I screwed something up when refactoring XdpAppInfo? Might be worth double checking what this was used for...

src/xdp-app-info-flatpak.c
```diff
@@ -823,11 +823,13 @@ xdp_app_info_flatpak_new (int      pid,
   app_info_flatpak = g_initable_new (XDP_TYPE_APP_INFO_FLATPAK,
```
swick, line 819: commit msg: `Turn "engine" into a construct-only property`

src/xdp-app-info.c
GeorgesStavracas: It was never used:

 * First mention: 80c88a457b8de6f188f8568ff4d82841b02924af
 * Struct field added: 642ebb31e731f6e9782ea7a8ad6f7a776ab658b2

src/xdp-app-info.c
```diff
@@ -175,6 +180,11 @@ xdp_app_info_set_property (GObject      *object,
       priv->instance = g_value_dup_string (value);
       break;
 
+    case PROP_PIDFD:
+      g_assert (priv->pidfd == -1);
+      priv->pidfd = dup (g_value_get_int (value));
```
swick, line 227: `typedef enum _XdpAppInfoFlags`

tbh, I'm unsure why this even is a pattern in e.g. mutter but at least it makes it easy to grep for the declaration...

src/xdp-app-info-private.h
```diff
@@ -23,6 +23,12 @@
 
 #include "xdp-app-info.h"
 
+typedef enum
```
swick, line 26: `typedef enum _XdpAppInfoFlags`

tbh, I'm unsure why this even is a pattern in e.g. mutter but at least it makes it easy to grep for the declaration...

swick:
I'm a bit on the fence about the approach. Wouldn't it be better to move the code we currently have in `xdp_app_info_foo_new` to `xdp_app_info_foo_initable_init` and only have the pid and pidfd as props?

GeorgesStavracas:
> Wouldn't it be better to move the code we currently have in `xdp_app_info_foo_new` to `xdp_app_info_foo_initable_init` and only have the pid and pidfd as props?

That was my original approach, but it got pretty spectacularly messy and I just settled for these construct-only properties.

swick:
> That was my original approach, but it got pretty spectacularly messy and I just settled for these construct-only properties.

Mh, I couldn't resist and tried it as well. The result is pretty nice, but I need to untangle the changes into smaller commits: https://github.com/swick/xdg-desktop-portal/commits/wip/app-info-initable/

src/xdp-app-info.c
jadahl: It's nice because one can use ctags to go directly to the actual struct, instead of the typedef, by adding a `_` in front of the name.

src/xdp-app-info-private.h
jadahl: It's nice because one can use ctags to go directly to the actual struct, instead of the typedef, by adding a _ in front of the name.

jadahl:
I think this approach (no private setters) follows existing GObject convention closer (semi-private construct properties). It lgtm, but adding the `_Typedef` would be a nice touch.

swick:
Not going to block this or anything, but I just don't see how GObject properties are better for this at all. They have huge overhead (string comparisons, GValue packing and unpacking, ...) compared to direct setters. There also doesn't seem to be any benefits to having the properties. So the argument "it's GObject convention" isn't very convincing when it's not backed up with actual benefits the convention brings.

Issue #1611, opened 2025-01-30
xdg-desktop-portal fails to compile on Debian Testing


n3rdopolis:
### Operating System

Debian Testing

### XDG Desktop Portal version

Git

### Desktop Environment

Other

### Desktop Environment (Other)

Built Time Error

### Expected Behavior

Hi

For a while xdg-desktop-portal has not been building on Debian Testing, I am also building my own pipewire, It's the the latest from main, so I am not sure what I am missing

```
FAILED: src/xdg-desktop-portal.p/pipewire.c.o 
cc -Isrc/xdg-desktop-portal.p -Isrc -I../src -Idocument-portal -I../document-portal -I. -I.. -I/opt/include/glib-2.0 -I/opt/lib/x86_64-linux-gnu/glib-2.0/include -I/opt/include/sysprof-6 -I/usr/include/libmount -I/usr/include/blkid -I/opt/include/gio-unix-2.0 -I/opt/include/json-glib-1.0 -I/opt/include/gudev-1.0 -I/opt/include/libgeoclue-2.0 -I/opt/include/pipewire-0.3 -I/opt/include/spa-0.2 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Os -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers -DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_72 -Os -Wno-error -D_REENTRANT -pthread -MD -MQ src/xdg-desktop-portal.p/pipewire.c.o -MF src/xdg-desktop-portal.p/pipewire.c.o.d -o src/xdg-desktop-portal.p/pipewire.c.o -c ../src/pipewire.c
../src/pipewire.c: In function ‘pipewire_remote_new_sync’:
../src/pipewire.c:334:35: error: passing argument 1 of ‘pw_registry_add_listener’ from incompatible pointer type [-Wincompatible-pointer-types]
  334 |   pw_registry_add_listener (remote->registry,
      |                             ~~~~~~^~~~~~~~~~
      |                                   |
      |                                   struct pw_proxy *
In file included from /opt/include/pipewire-0.3/pipewire/context.h:48,
                 from /opt/include/pipewire-0.3/pipewire/protocol.h:26,
                 from /opt/include/pipewire-0.3/pipewire/proxy.h:90,
[...]

PR #1624, opened 2025-02-12
Fix build with pipewire 1.3.82

antlarr-suse:
While submitting pipewire 1.3.82 to openSUSE Tumbleweed I noticed this cast is needed to build xdg-desktop-portal. This was already submitted as an issue in #1611 in Debian testing, in which the reporter doesn't mention the used pipewire version but says "I am also building my own pipewire" so I guess it's one of the latest RCs (1.3.81 or 1.3.82).

Fixes: #1611

### DIFF
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -331,7 +331,7 @@ pipewire_remote_new_sync (struct pw_properties *pipewire_properties,
   remote->registry = (struct pw_proxy*) pw_core_get_registry (remote->core,
                                                               PW_VERSION_REGISTRY,
                                                               0);
-  pw_registry_add_listener (remote->registry,
+  pw_registry_add_listener ((struct pw_registry*)remote->registry,
                             &remote->registry_listener,
                             &registry_events,
                             remote);


swick:
I think the better way here would be to store the registry as the type it actually is (`struct pw_registry`) in `PipeWireRemote` and then cast to `struct pw_proxy` on demand.

antlarr-suse:
> I think the better way here however would be to store the registry as the type it actually is (`struct pw_registry`) in `PipeWireRemote` and then cast to `struct pw_proxy` on demand.

Yes, I also thought of that, but I wasn't sure if you'd prefer that and this way it was just a one-liner.

> Do you want to update the patch accordingly or do you want me to take over?

Sure, I'll update it.

smcv (approved):

antlarr-suse:
Hi, I just noticed that I force pushed the same commit instead of the requested changes that I had already prepared:
```
diff --git a/src/pipewire.c b/src/pipewire.c
index 39b05c7..7b8d104 100644
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -224,7 +224,8 @@ pipewire_remote_destroy (PipeWireRemote *remote)
   /* This check is a workaround for older PW versions */
   if (remote->registry)
     spa_hook_remove (&remote->registry_listener);
-  g_clear_pointer (&remote->registry, pw_proxy_destroy);
+  pw_proxy_destroy((struct pw_proxy *)remote->registry);
+  remote->registry = NULL;
   g_clear_pointer (&remote->globals, g_hash_table_destroy);
   if (remote->core)
     spa_hook_remove (&remote->core_listener);
@@ -328,10 +329,10 @@ pipewire_remote_new_sync (struct pw_properties *pipewire_properties,
                         &core_events,
                         remote);
 
-  remote->registry = (struct pw_proxy*) pw_core_get_registry (remote->core,
-                                                              PW_VERSION_REGISTRY,
-                                                              0);
-  pw_registry_add_listener ((struct pw_registry*)remote->registry,
+  remote->registry = pw_core_get_registry (remote->core,
+                                           PW_VERSION_REGISTRY,
+                                           0);
+  pw_registry_add_listener (remote->registry,
                             &remote->registry_listener,
                             &registry_events,
                             remote);
diff --git a/src/pipewire.h b/src/pipewire.h
index f861449..354f8a1 100644
--- a/src/pipewire.h
+++ b/src/pipewire.h
@@ -53,7 +53,7 @@ struct _PipeWireRemote
 
   int sync_seq;
 
-  struct pw_proxy *registry;
+  struct pw_registry *registry;
   struct spa_hook registry_listener;
 
   GHashTable *globals;
```
@smcv , @swick, Should I open another PR with those changes or is the merged one ok?

PR #1625, opened 2025-02-12
Store the registry as a struct pw_registry instead of a pw_proxy

antlarr-suse:
This is a better fix for #1611 and an improvement over #1624 as requested by @swick.

### DIFF
--- a/src/pipewire.c
+++ b/src/pipewire.c
@@ -224,7 +224,7 @@ pipewire_remote_destroy (PipeWireRemote *remote)
   /* This check is a workaround for older PW versions */
   if (remote->registry)
     spa_hook_remove (&remote->registry_listener);
-  g_clear_pointer (&remote->registry, pw_proxy_destroy);
+  g_clear_pointer ((struct pw_proxy **)&remote->registry, pw_proxy_destroy);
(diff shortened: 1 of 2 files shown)


src/pipewire.c
```diff
@@ -224,7 +229,7 @@ pipewire_remote_destroy (PipeWireRemote *remote)
   /* This check is a workaround for older PW versions */
   if (remote->registry)
     spa_hook_remove (&remote->registry_listener);
-  g_clear_pointer (&remote->registry, pw_proxy_destroy);
+  g_clear_pointer (&remote->registry, pipewire_registry_destroy);
```
swick, line 232:
```suggestion
  g_clear_pointer ((struct pw_proxy **)&remote->registry, pw_proxy_destroy);
```

antlarr-suse: ouch, I would have said I tried that and it didn't build, before adding the helper function. Thanks for suggesting that change. I made that change and I guess you can squash all commits before merging, right? or do you want me to merge them into one and force push?

swick: For the future: we also try to prefix the commit message with the area the code touches. In this case `pipewire: ...`.

swick:
Thanks, this LGTM.

PR #1631, opened 2025-02-13
Unneutralize host registry and document the plans for the future

jadahl:
From the last commit:

    Since the plan is for apps to eventually ensure they are placed in the
    right cgroup and rely on the portal to derive the app ID from the cgroup
    name, mention this in the api reference.

Time will tell when we'll be able to make this transition, but at least document the "stop gap"-ness of the registry solution.

Cc: @pwithnall

### DIFF
--- a/doc/api-reference.rst
+++ b/doc/api-reference.rst
@@ -17,6 +17,14 @@ This might improve the user experience when the host app was launched in a way
 that doesn't follow the standard. See
 :doc:`org.freedesktop.host.portal.Registry <doc-org.freedesktop.host.portal.Registry>`
 
+Disclaimer: The host app registry is expected to eventually be deprecated and
+may be removed. Applications should gracefully handle interface or method no
+longer being available to be forward compatible. App launchers, or apps
+themselves, should place the app in a cgroup named according to specific naming
+conventions. When the host app registry becomes deprecated, the details of the
+replacement will be documented in :doc:`org.freedesktop.host.portal.Registry
+<doc-org.freedesktop.host.portal.Registry>`.
+
 All apps have access to the portals below:
 
 .. toctree::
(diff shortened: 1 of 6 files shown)


grulja:
I would still vote for this to be in the stable 1.20.0 release if the documented "stop gap" is enough for glib maintainers to give this a go.

doc/api-reference.rst
```diff
@@ -4,8 +4,26 @@ API Reference
...
+<https://systemd.io/DESKTOP_ENVIRONMENTS/#xdg-standardization-for-applications>`_.
+This might improve the user experience when the host app was launched in a way
+that doesn't follow the standard. See
+:doc:`org.freedesktop.host.portal.Registry <doc-org.freedesktop.host.portal.Registry>`
+
+Disclaimer: The host app calls to the registry will eventually be replacable
+with the app placing itself in a cgroup named according to specific naming
+conventions. It will be documented in
+:doc:`org.freedesktop.host.portal.Registry <doc-org.freedesktop.host.portal.Registry>`
+when this will be possible.
```
swick, line 24:
```suggestion
Disclaimer: The host app registry will eventually be deprecated and removed. App launchers, or apps themselves, should place the app in a cgroup named according to specific naming conventions. When the host app registry becomes deprecated, the details of the replacement will be documented in :doc:`org.freedesktop.host.portal.Registry <doc-org.freedesktop.host.portal.Registry>`.
```

swick: Even though we already know the conventions that shall be used right now. We should only deprecated the registry once we have managed to get a majority of applications to use the new convention (one way or another).

jadahl: Made the deprecation note a bit softer, i.e. "deprecated and may be removed" to be more about intention and not a 100% prediction of the future - the important part is that apps eventually should be able to remove the call and rely on cgroups.

swick:
LGTM

GeorgesStavracas (approved):

PR #1633, opened 2025-02-14
run-test.sh: Quote more defensively

smcv:
This resolves several warnings from shellcheck about patterns that will not do what was intended, because shell script is a language constructed entirely from sharp edges.


swick:
Thanks, LGTM. I wonder if it's worth it to add spellcheck to CI.

Issue #1545, opened 2024-12-22
Notifications: should escape <a href> value


smcv:
### Operating System

n/a, source code inspection

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.19.1

### Desktop Environment

Other

### Desktop Environment (Other)

n/a, source code inspection

### Expected Behavior

If a client sends for example `{"markup-body": <'<a href="&lt;&quot;&gt;">'>}`, the href attribute should be re-escaped when emitting the filtered/sanitized markup-body.

### Current Behavior

If I'm reading the code correctly, this

```
+          if (strcmp (attribute_names[i], "href") == 0)
+            {
+              g_string_append_printf (composed, "<a href=\"%s\">", attribute_values[i]);
+              break;
+            }
```

will emit the attribute unescaped.

### Steps to Reproduce

n/a, source code inspection

PR #1635, opened 2025-02-14
notification: Escape the href attribute of links when re-serializing

smcv:
* notification: Escape the href attribute of links when re-serializing

    Otherwise they'll be emitted unescaped, changing the meaning of the
    notification body.
    
    Fixes: ebff872d "notification: Add new property markup-body to support markup"
    Resolves: https://github.com/flatpak/xdg-desktop-portal/issues/1545

* tests: Assert that href attributes are escaped correctly
    
    Strictly speaking the exact form of escaping used here is an
    implementation detail of GLib (it would be equally valid to re-escape
    the encoded characters with numeric character references like
    `&#34;&#62;&#60;html&#62;` or similar), but it seems unlikely that
    GLib will change this in practice.

* notification: Escape the body text of the markup-body when re-serializing
    
    Similar to #1545, otherwise they'll be emitted unescaped, changing the
    meaning of the notification body.
    
    Fixes: ebff872d "notification: Add new property markup-body to support markup"

* tests: Assert that markup-body text is escaped correctly

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -253,7 +253,15 @@ markup_parser_text (GMarkupParseContext  *context,
 {
   GString *composed = user_data;
 
-  g_string_append_len (composed, text, text_len);
+  while (text_len > 0)
+    {
+      gsize len = MIN (text_len, G_MAXSSIZE);
+      g_autofree char *escaped = g_markup_escape_text (text, len);
+
+      g_string_append (composed, escaped);
+      text_len -= len;
+      text += len;
+    }
 }
 
 static void
--- a/src/notification.c
+++ b/src/notification.c
@@ -282,7 +282,9 @@ markup_parser_start_element (GMarkupParseContext  *context,
         {
           if (strcmp (attribute_names[i], "href") == 0)
             {
-              g_string_append_printf (composed, "<a href=\"%s\">", attribute_values[i]);
+              g_autofree char *escaped = g_markup_escape_text (attribute_values[i], -1);
+
+              g_string_append_printf (composed, "<a href=\"%s\">", escaped);
               break;
             }
         }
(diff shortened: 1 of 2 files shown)


smcv:
cc @jsparber

smcv:
Actually the body text needs escaping, too.

src/notification.c
```diff
@@ -253,7 +253,15 @@ markup_parser_text (GMarkupParseContext  *context,
...
-  g_string_append_len (composed, text, text_len);
+  while (text_len > 0)
+    {
+      gsize len = MIN (text_len, G_MAXSSIZE);
+      g_autofree char *escaped = g_markup_escape_text (text, len);
+
+      g_string_append (composed, escaped);
+      text_len -= len;
+      text += len;
+    }
```
smcv, line 264: This loop is because strictly speaking there is no API guarantee that `text_len` won't exceed `G_MAXSSIZE`, but `g_markup_escape_text()` can "only" escape text in chunks of length no more than `G_MAXSSIZE` bytes.

(On 32-bit platforms, if you have somehow managed to send a notification with more than 2 GiB of text... well done, I suppose? And on 64-bit platforms this clearly can't actually happen, but writing the code in a way that defends against it seemed easier than justifying why it can't.)

swick:
Changes LGTM. ruff wants to reformat the python file. Probably wants to put the strings on the same line. Could either just call ruff format or add `# fmt: skip` to keep it the way it is.

smcv:
> Changes LGTM. ruff wants to reformat the python file. Probably wants to put the strings on the same line.

FYI, what it wanted was for all strings that don't contain embedded double quotes to be double-quoted (which is not the coding style I'm used to, and not mandated by PEP 8, but whatever).

GeorgesStavracas (approved):

PR #1638, opened 2025-02-15
Tests: Use pytest instead of pytest-3.

fabiscafe:
Back in a time of transition from Python 2 to 3, `pytest-3` was an indicator to use the Python 3 edition. However, this is no longer the case, and if `pytest-3` still exists, it's only for compatibility reasons. Let's not use it.


GeorgesStavracas:
CC @swick

whot:
best approach here would be to try pytest-3 and if that fails, use pytest (see meson.build)

GeorgesStavracas:
Superseded by #1641

PR #1641, opened 2025-02-17
tests/run-test.sh: Fall back from pytest-3 to pytest

swick:
Solving the problem that https://github.com/flatpak/xdg-desktop-portal/pull/1638 is trying to solve but with a fallback from pytest-3 to pytest.

### DIFF
--- a/tests/run-test.sh
+++ b/tests/run-test.sh
@@ -26,6 +26,8 @@ function fail()
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
+PYTEST=$(command -v "pytest-3" || command -v "pytest") || fail "pytest is missing"
+
 BUILDDIR=${BUILDDIR:-$(find "${SCRIPT_DIR}/.." -maxdepth 2 -name 'build.ninja' -printf "%h\n" -quit)}
 
 [ ! -f "${BUILDDIR}/build.ninja" ] && fail "Path '${BUILDDIR}' does not appear to be a build dir"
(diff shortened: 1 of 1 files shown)


tests/run-test.sh
```diff
@@ -19,7 +19,7 @@ set -euo pipefail
 
 function fail()
 {
-  sed -n '/^#$/,/^$/p' "${BASH_SOURCE[0]}"
+  sed -n "/^#$/,/^$/p" "${BASH_SOURCE[0]}"
```
smcv, line 22: Please don't apply this part; if this still works as intended, then it's relying on subtle details of how `$` expands or doesn't expand as a variable reference inside double quotes. It will be easier to understand with single quotes.

Unlike Python, in shell script there is a functional difference between single and double quotes which matters a lot.

tests/run-test.sh
```diff
@@ -28,7 +28,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 PYTEST=$(command -v "pytest-3" || command -v "pytest") || fail "pytest is missing"
 
-BUILDDIR=${BUILDDIR:-$(find "${SCRIPT_DIR}/.." -maxdepth 2 -name 'build.ninja' -printf "%h\n" -quit)}
+BUILDDIR=${BUILDDIR:-$(find "${SCRIPT_DIR}/.." -maxdepth 2 -name "build.ninja" -printf "%h\n" -quit)}
```
smcv, line 31: This part is OK.

smcv (changes requested):

tests/run-test.sh
swick: Woops, I didn't notice the `$` signs... Fixed.

smcv (approved):

PR #1645, opened 2025-02-20
Avoid spurious warnings when dbus.service stops

andrew-sayers:
Services in this project generate a warning when they lose ownership of their D-Bus name, which happens routinely when dbus.service exits (e.g. when the user logs out).

Stop services before dbus.service exits, to avoid this warning.


swick:
LGTM

PR #1652, opened 2025-02-25
Make sure the Host Registry XdpAppInfo updates become visible

swick:
The invocations are currently delegated to threads which means that the next invocation gets run before the first one finished. This is fine for all the other portals, but because the Registry modifies the identity of the caller for all following calls, we must ensure that it finishes before the next invocation. This means in particular no g-authorize-method and no FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD.


swick:
Should probably add a test for it as well.

Issue #1612, opened 2025-01-31
Host App registry: Reregister is problematic as a toolkit


Sodivad:
### Operating System

Neon

### XDG Desktop Portal version

Git

### Desktop Environment

KDE

### Expected Behavior

In https://github.com/flatpak/xdg-desktop-portal/pull/1521#issuecomment-2619118745 writes

> They indeed need to re-register, ideally by listening to the portal name coming back. Something libportal ideally should handle itself.

But  in certain cases this can cause the registering to be to late. In the current implementation registration only succeeds if it's the first call made to the portal.  However consider this case:

- toolkit registers app
[time passes, calls are made]
- xdg-desktop-portal stops for some reason (crashes/is restarted)
- app makes a dbus call to the portal
- xdg-desktop-portal is started again
- name appears on the bus and the toolkit issues a register request

The register attempt will arrive after the call that caused the portal to be started and fail.

### Current Behavior

My suggestion would be drop the requirement that no other call to the portal can happen before registering.  This would also solve a problem that I found while testing. Apps might do things that causes calls to the portal before instantiating their central application object by accident  as it happens in transitive libraries that they are using.

### Steps to Reproduce

na

swick:
> My suggestion would be drop the requirement that no other call to the portal can happen before registering. 

This would mean we will have inconsistent state. I'm not entirely sure about the consequences but the app id being invariant over a connection is what we have assumed so far and made things relatively simple.

LyHug:
Seeing what looks like the same host-app-registry/toolkit edge case from a Quickshell/Qt stack on Arch Linux.

Environment:
- `qt6-base 6.11.0-2`
- `xdg-desktop-portal 1.20.4-1`
- `xdg-desktop-portal-gtk 1.15.3-1`
- Wayland session on `niri`
- `QT_QPA_PLATFORMTHEME=qt6ct`
- `XDG_CURRENT_DESKTOP=niri`

The app itself does not explicitly register a portal app id in shell code; the warning is emitted from Qt (`qt.qpa.services`) during startup:

```text
WARN qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")
```

This looks like another example where toolkit-managed registration is hitting the host-app registry constraints, instead of something app-specific in the shell itself.

Sodivad:
[...]

PR #1654, opened 2025-02-26
registry: Add hint about reregistration on restarts

jadahl:
The portal service may be restarted for various reasons, and ideally applications should handle that gracefully by re-registering.


swick:
LGTM

Sodivad:
What about: https://github.com/flatpak/xdg-desktop-portal/issues/1612 ?

jadahl:
> What about: https://github.com/flatpak/xdg-desktop-portal/issues/1612 ?

I would say that is orthogonal to what is added here, but indeed need to make a decision about it.

Issue #1653, opened 2025-02-26
Realtime error: Could not get pidns: Could not fstatat ns/pid: Not a directory


andrew-sayers:
### Operating System

Debian

### XDG Desktop Portal version

Git

### Desktop Environment

Cinnamon / MATE / Xfce

### Expected Behavior

Playing a YouTube video should not cause any error messages.

### Current Behavior

The following message appears in my systemd log:

```
Realtime error: Could not get pidns: Could not fstatat ns/pid: Not a directory
```

It is often, but not always, accompanied by messages from pipewire.  These usually include at least one warning that mentions "xrun" and sometimes include an error that says "snd_pcm_mmap_commit error: Broken pipe".  They are not included here because this problem can happen without them.

### Steps to Reproduce

1. run `sudo journalctl -f` in a terminal
2. open Firefox
3. play any YouTube video
4. observe the output of `sudo journalctl -f`

(this probably isn't specific to YouTube or even Firefox, I just happened to trigger the issue with them)

### Anything else we should know?

The message is directly generated by [xdp_pidfd_get_namespace() in xdp-utils.c](https://github.com/flatpak/xdg-desktop-portal/blob/f7837cc4a93c32e783ab102c790114a49444960b/src/xdp-utils.c#L924), called from [map_pid() in realtime.c](https://github.com/flatpak/xdg-desktop-portal/blob/f7837cc4a93c32e783ab102c790114a49444960b/src/realtime.c#L68).

I'm not familiar with pidfds and namespaces, but the function seems to be passed an anonymous pidfd, but expects an fd for a directory like `/proc/$pid/task/$pid`.  Might this be getting called incorrectly?

Note: PR #1655 improved the error message so I could debug the issue, but doesn't fix anything.

kotauskas:
> It is often, but not always, accompanied by messages from pipewire. These usually include at least one warning that mentions "xrun"

FYI: this is because PipeWire's `libpipewire-module-rt` module relies on [RTKit](https://github.com/heftig/rtkit) or the [Realtime portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Realtime.html) to set real-time priority for audio playback threads. On my Arch system, setting `rtportal.enabled = false` doesn't seem to help, which suggests that the Realtime portal may be intercepting RTKit calls for one reason or another. With no real-time priority, audio threads are prone to being preempted or not scheduled at inopportune times, leading to periodic xruns.

[...]

PR #1655, opened 2025-02-26
realtime: Include PID/TID in some error messages

andrew-sayers:
The PID is useful when debugging issues related to these messages, so we can check whether the PID has some strange property that caused the problem.

In my case, this made it easier to debug #1653 - the PID showed it was caused by Firefox, so I knew to look there.

### DIFF
--- a/src/realtime.c
+++ b/src/realtime.c
@@ -62,24 +62,30 @@ static gboolean
 map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
 {
   ino_t pidns_id;
+  /*
+   * the unmapped pid is more useful for debugging, and doesn't leak
+   * information into the sandbox - see discussion in
+   * https://github.com/flatpak/xdg-desktop-portal/pull/1655
+   */
+  const pid_t unmapped_pid = *pid;
 
   if (!xdp_app_info_get_pidns (app_info, &pidns_id, error))
     {
-      g_prefix_error (error, "Could not get pidns: ");
+      g_prefix_error (error, "Could not get pidns for pid %d: ", unmapped_pid);
       g_warning ("Realtime error: %s", (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_pids (pidns_id, pid, 1, error))
     {
-      g_prefix_error (error, "Could not map pid: ");
+      g_prefix_error (error, "Could not map pid %d: ", unmapped_pid);
       g_warning ("Realtime error: %s", (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_tids (pidns_id, *pid, tid, 1, error))
     {
-      g_prefix_error (error, "Could not map tid: ");
+      g_prefix_error (error, "Could not map tid %d of pid %d: ", *tid, unmapped_pid);
       g_warning ("Realtime error: %s", (*error)->message);
       return FALSE;
     }


src/realtime.c
```diff
@@ -65,21 +65,21 @@ map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
...
-      g_prefix_error (error, "Could not map pid: ");
+      g_prefix_error (error, "Could not map pid %d: ", *pid);
       g_warning ("Realtime error: %s", (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_tids (pidns_id, *pid, tid, 1, error))
     {
-      g_prefix_error (error, "Could not map tid: ");
+      g_prefix_error (error, "Could not map tid %d of pid %d: ", *tid, *pid);
```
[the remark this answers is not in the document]
andrew-sayers: Good point - updated to "mapped pid".  The original pid would need to be stored separately if we wanted to print it here - seems like overkill unless a use case pops up for this particular message.

src/realtime.c
```diff
@@ -65,21 +65,21 @@ map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
...
-      g_prefix_error (error, "Could not map pid: ");
+      g_prefix_error (error, "Could not map pid %d: ", *pid);
       g_warning ("Realtime error: %s", (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_tids (pidns_id, *pid, tid, 1, error))
     {
-      g_prefix_error (error, "Could not map tid: ");
+      g_prefix_error (error, "Could not map tid %d of mapped pid %d: ", *tid, *pid);
```
swick, line 82: Does this propagate to the app and are we okay with leaking root pidns details to apps?

andrew-sayers: Hmm, this goes in the journal, and I guess an unscrupulous app could scrape the data out of there.

I'm not sure what the security implications are, so I've just updated the patch to print the unmapped pid.  Again, one to discuss if a use case ever arises.

swick: The journal would be unproblematic because it cannot be accessed by sandboxed apps. The problem is that the error here gets passed to dbus, and the calling client will see the message.

IOW, this should probably modify the g_warning (which will show up only in the xdp journal) and not the GError message.

swick: Right, missed the update, but we should still keep those things in mind.

swick:
LGTM

src/realtime.c
andrew-sayers: This is getting a bit bikesheddy, but...

The current version doesn't leak anything, but could theoretically encourage a reckless future developer to change the code thinking it was just an oversight on our parts.  I've pushed a comment linking back to this discussion.

The g_warning version would look like this - slightly less readable, and could theoretically be less useful to a developer reading dbus messages for some reason.  I mildly prefer the existing version, but this is fine too.

```diff
--- src/realtime.c
+++ src/realtime.c
@@ -62,25 +62,31 @@ static gboolean
 map_pid (XdpAppInfo *app_info, pid_t *pid, pid_t *tid, GError **error)
 {
   ino_t pidns_id;
+  /*
+   * the unmapped pid is more useful for debugging, and doesn't leak
+   * information into the sandbox - see discussion in
+   * https://github.com/flatpak/xdg-desktop-portal/pull/1655
+   */
+  const pid_t unmapped_pid = *pid;
 
   if (!xdp_app_info_get_pidns (app_info, &pidns_id, error))
     {
       g_prefix_error (error, "Could not get pidns: ");
-      g_warning ("Realtime error: %s", (*error)->message);
+      g_warning ("Realtime error mapping pid %d: %s", unmapped_pid, (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_pids (pidns_id, pid, 1, error))
     {
       g_prefix_error (error, "Could not map pid: ");
-      g_warning ("Realtime error: %s", (*error)->message);
+      g_warning ("Realtime error mapping pid %d: %s", unmapped_pid, (*error)->message);
       return FALSE;
     }
 
   if (pidns_id != 0 && !xdp_map_tids (pidns_id, *pid, tid, 1, error))
     {
       g_prefix_error (error, "Could not map tid: ");
-      g_warning ("Realtime error: %s", (*error)->message);
+      g_warning ("Realtime error mapping tid %d of pid %d: %s", *tid, unmapped_pid, (*error)->message);
       return FALSE;
     }
 
```

andrew-sayers:
> LGTM

For the record, I was adding the comment at the top of the patch at the same time @swick was typing this, based on feedback in the now-resolved discussion.  The PR is otherwise identical.

GeorgesStavracas (approved):

PR #1669, opened 2025-03-16
RFC: Add desktop portal diagnostic interface

tytan652:
I thought about how to improve [Door Knocker](https://flathub.org/apps/xyz.tytanium.DoorKnocker) like:
- being able to fetch if the access portal is present which if missing cause some portal to be disabled
- avoid relying on `--talk-name=org.freedesktop.Flatpak` but add a `--talk-name` to a specific interface

At the same time I tried to make sure that this interface is not meant only for my application but also could be used by desktop environment to have a good insight of the state of the portal service.

This interface is not meant to be accessible directly by Flatpak application.


data/meson.build
```diff
@@ -74,6 +74,10 @@ background_monitor_sources = files(
   'org.freedesktop.background.Monitor.xml',
 )
 
+diagnostic_portal_sources = files(
+  'org.freedesktop.diagnostic.portal.Desktop.xml',
```
swick, line 78: We now have the `org.freedesktop.host.portal` namespace, might make sense to re-use that.

data/org.freedesktop.diagnostic.portal.Desktop.xml
```diff
@@ -0,0 +1,125 @@
...
+        Name of the service.
+
+      * ``version`` (``s``)
+
+        Version of the service.
+
+      * ``stable`` (``b``)
+
+        Indicate if the version of the service is unstable.
+        Assume unstable if ``name``, ``version`` or ``stable`` is not present
```
swick, line 52: We have unstable portals?

data/org.freedesktop.diagnostic.portal.Desktop.xml
```diff
@@ -0,0 +1,125 @@
...
+
+        Name of the implementation. If this properties is not present, no
+        implementation were found.
+
+      * ``version`` (``u``)
+
+        Version of the implementation if the interface is versioned.
+        Assume 1 if not present
+    -->
+    <property name="lockdown-impl" type="a{sv}" access="read">
```
swick, line 75: Any reason why there are a bunch of properties for specific impls instead of exposing a single property with all the impls?

data/org.freedesktop.diagnostic.portal.Desktop.xml
```diff
@@ -0,0 +1,125 @@
...
+    <property name="access-impl" type="a{sv}" access="read">
+      <annotation name="org.qtproject.QtDBus.QtTypeName" value="QVariantMap"/>
+    </property>
+
+    <!--
+      portals:
+
+      The list of portal with details about their state:
+
+      * ``name``` (``s``)
```
swick, line 96: From looking at the signature we have a tuple where the first element is probably the name, and the second element is probably the vardict documented below. So \`\`name\`\` implies this is a vardict, but it's not, it's just a string.

Could just make this an array of vardicts with name as one possible property.

data/org.freedesktop.diagnostic.portal.Desktop.xml
```diff
@@ -0,0 +1,125 @@
...
+
+        Short name of the portal, usually the last element of their interface
+        name.
+
+        Each portal can contain the following properties:
+
+          * ``implementation`` (``a{sv}``)
```
swick, line 103: Could just provide the name of the impl and let the caller correlate that with the impl data from another property.

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: It's xdp own version, not everyone is aware of "odd minor means unstable".

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: Access and lockdown state affect various portals (and don't have non-impl counterpart).

One being missing will cause some portals to be missing.

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: It's made that way to avoid having to update the interface for every new portal, unless it's an impl like Access or Lockdown.

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: IIRC, I made it this way so you can just check for the key to know if the portal has an impl and not loop to find it (even if internally it may loop anyway).

data/org.freedesktop.diagnostic.portal.Desktop.xml
swick: I don't see why this needs to be encoded in the API though; especially because versioning schemes can change.

data/org.freedesktop.diagnostic.portal.Desktop.xml
swick: I'm aware of how that works. This doesn't answer the question. Why only expose some of the impls and not all?

data/org.freedesktop.diagnostic.portal.Desktop.xml
swick: That doesn't seem very useful to me. Vardicts are also just arrays. Duplicating all the data here just seems unnecessary.

data/org.freedesktop.diagnostic.portal.Desktop.xml
swick: You don't have to add a new property for every impl. Just one property that enumerates all the impls.

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: Because those are excluded from the `portals` property, since it only list impls meant to specific portal.

swick: Right, but that means you have impls in two different properties; imo it would make more sense to just list all of them in `impls` and reference them from `portals`.

In general, I think you're designing this too closely for what you want to use in your app. If we want a diagnostics interface, it should expose as much data in the most generic way possible. Let the app handle making sense of it.

data/meson.build
tytan652: While trying to work on changes, I realize that this namespace is unfiltered where the diagnostic interface should not be usable by sandboxed apps unless the static perm is set.

swick: Yeah, `org.freedesktop.diagnostic.portal` or `org.freedesktop.host.portal.diagnostics` are only interface names. You'd have to own another name on the bus if it should be filtered. The alternative is to explicitly disallow access to non-host XdpAppInfos.

data/org.freedesktop.diagnostic.portal.Desktop.xml
tytan652: This is why `stable` is a boolean set by xdp itself it avoid having consumer having to decipher the version.

I feel like being able to warn the user (through the DE or an app) that an unstable version of portal is in place is an important piece of information.

`version` is there to get the version without going through sscan `xdg-desktop-portal --version`, and `name` is the name of the binary even if alternative implementation of xdp are unlikely to be.

swick: > I feel like being able to warn the user (through the DE or an app) that an unstable version of portal is in place is an important piece of information.

That seems useful but it doesn't require the information to be exposed directly by the diagnostics interface and you can do so on the client side by simply looking at the version number.

tytan652:
I think this would be easier to locked it behind entitlement rather than owning another D-Bus name.

PR #1676, opened 2025-03-19
meson: remove wavparse dependency

Hello71:
the related test was removed in 6b424abaa5 ("tests: Remove C based integration tests")


swick:
Wavparse is required for the sound validator to not reject sound files. It is still required for the tests. The `config.h` could be removed. The python notification tests could also check for wavparse and skip it if they were not found, instead of doing it globally in meson.

PR #1677, opened 2025-03-19
validate-{icon,sound}: mount tmpfs,proc,dev first

Hello71:
otherwise running from /tmp (e.g. for tests) will fail


swick (approved):
Thank you. This LGTM.

GeorgesStavracas (approved):

Issue #1674, opened 2025-03-19
Cannot create/ install dynamic launchers


eyekay-gh:
### Operating System

Debian sid

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.20.0

### Desktop Environment

GNOME

### Expected Behavior

The dynamic launcher should be installed successfully.

### Current Behavior

I get an error that the desktop entry given to install() is not valid.

`g-io-error-quark: GDBus.Error:org.freedesktop.portal.Error.InvalidArgument: Desktop entry given to Install() not valid (36)`

### Steps to Reproduce

1. Install [Web Apps](https://flathub.org/apps/net.codelogistics.webapps) and create a new web app
2. Click 'Install' in the dialog
3. The web app isn't installed, the call to `portal.dynamic_launcher_install()` fails with this error in the terminal:
`g-io-error-quark: GDBus.Error:org.freedesktop.portal.Error.InvalidArgument: Desktop entry given to Install() not valid (36)`

### Anything else we should know?

This is the code I am running (python):
`desktop_file = '[Desktop Entry]\nType=Application\nExec=webapps {}\nTerminal=false\nCategories=Network;'.format(app_id)`

...

`portal.dynamic_launcher_install(token, "net.codelogistics.webapps." + app_id + ".desktop", desktop_file)`

where `app_id` is in the format "webapp-{UUID}".

This was working fine till a few months ago in both GNOME and KDE. I don't know whether this is a bug in xdg-desktop-portal or libportal, or in my app.

swick:
I assume the app which tries to use the dynamic launcher portal is a flatpak? We recently tightened the app id requirements to follow what flatpak considers a valid app id: https://github.com/flatpak/xdg-desktop-portal/blob/119aad03548f85ed249b7f21cd551f77bc74e4dd/src/xdp-app-info-flatpak.c#L173C1-L173C41 In other words, if your app is a flatpak, it should not be able to create sub-app-id's which are not valid flatpak app ids.

swick:
You might run into the 255 bytes limit.

eyekay-gh:
Thanks, but that is not the case, the total length of the desktop file's name is under 255 chars.

I cannot install web apps using GNOME Web (installed via Flatpak) either. (The text of the notification gets cut off in the notification panel, so I cannot attach the full error message.)

mcatanzaro:
[...]

PR #1682, opened 2025-03-25
dynamic-launcher: Check Desktop file after AppInfo can rewrite Exec

swick:
The g_desktop_app_info_new_from_keyfile function requires the the Exec line in the keyfile to be an executable (in the context of the current environment, so the portal). Flatpak apps do not have the executable on the host, but use `flatpak run`. The portal rewrites those Exec lines, but it did so after calling g_desktop_app_info_new_from_keyfile, so it would fail if whatever the Exec line specified was not a valid executable on the host.

Resolves: #1674

### DIFF
--- a/src/dynamic-launcher.c
+++ b/src/dynamic-launcher.c
@@ -294,7 +294,6 @@ get_desktop_entry (const char  *desktop_file_id,
                    char        *icon_path,
                    GError     **error)
 {
-  g_autoptr(GDesktopAppInfo) desktop_app_info = NULL;
   g_autoptr(GKeyFile) key_file = g_key_file_new ();
   g_auto(GStrv) groups = NULL;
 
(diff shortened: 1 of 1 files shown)


src/dynamic-launcher.c
```diff
@@ -401,6 +392,16 @@ handle_install (XdpDbusDynamicLauncher *object,
       return G_DBUS_METHOD_INVOCATION_HANDLED;
     }
 
+
```
jadahl, line 395: Stray newline.

jadahl (approved):
Makes sense.

GeorgesStavracas (approved):

mcatanzaro:
I'd appreciate a 1.20.1 release for this, since it's not possible to install web apps without this fix.

PR #1683, opened 2025-03-26
add clipboard support to input capture portal

sithlord48:
hook the clipboard to input capture portal 

 - [X] Finalized XML bits
 - [X] Ported clipboard parts from remotedesktop portal
 - [ ] Proper `handle_create_draft_session`
 - [ ] Proper `handle_init_session` 
 - [ ] have old `handle_capture_session`  call  `handle_create_draft_session` then `handle_init_session`

Taken over by #1803

### DIFF
--- a/data/org.freedesktop.impl.portal.InputCapture.xml
+++ b/data/org.freedesktop.impl.portal.InputCapture.xml
@@ -72,6 +72,94 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <!--
+        CreateDraftSession:
+        @app_id: App id of the application
+        @options: Vardict with optional further information
+        @results: Vardict with the results of the call
+
+        Creates a draft capture input session. A successfully created session can at
+        any time be closed using :ref:`org.freedesktop.portal.Session.Close`, or may
+        at any time be closed by the portal implementation, which will be
+        signalled via :ref:`org.freedesktop.portal.Session::Closed`.
+
+        Supported keys in the @options vardict include:
+
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        The following results get returned in the @results vardict:
+
+        * ``session_id`` (``s``)
+
+          The session id. A string representing the created capture input session.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
+        @handle: Object path for the :ref:`org.freedesktop.impl.portal.Request` object representing this call
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Initializes a capture input session. This may result in the portal backend querying the user whether
+        they want to allow the input of the session to be captured or not, and what capabilities to support.
+
+        Supported keys in the @options vardict include:
+
+        * ``handle_token`` (``s``)
+
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
+          more information about the @handle.
+
+        * ``capabilities`` (``u``)
+
+          Bitmask of requested capabilities, see the SupportedCapabilities property.
+          This value is required and must not be zero.
+
+        The following results get returned in the @results vardict:
+
+        * ``capabilities`` (``u``)
+
+          The capabilities available to this session. This is always a
+          subset of the requested capabilities.
+          See the SupportedCapabilities property for details. Note that
+          while a capability may be available to a session, there is no
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="u" name="response" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
     <!--
         GetZones:
         @handle: Object path for the :ref:`org.freedesktop.impl.portal.Request` object representing this call
--- a/data/org.freedesktop.portal.InputCapture.xml
+++ b/data/org.freedesktop.portal.InputCapture.xml
@@ -67,6 +67,9 @@
         at any time be closed by the portal implementation, which will be
         signalled via :ref:`org.freedesktop.portal.Session::Closed`.
 
+        Clients implementing version 2 of this interface should prefer CreateDraftSession followed by
+        InitSession over the version 1 CreateSession method.
+
         Supported keys in the @options vardict include:
 
         * ``handle_token`` (``s``)
@@ -110,6 +113,90 @@
     <method name="CreateSession">
       <arg type="s" name="parent_window" direction="in"/>
       <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        CreateDraftSession:
+        @options: Vardict with optional further information
+        @results: Vardict with the results of the call
+
+        Creates a draft capture input session. A successfully created session can at
+        any time be closed using :ref:`org.freedesktop.portal.Session.Close`, or may
+        at any time be closed by the portal implementation, which will be
+        signalled via :ref:`org.freedesktop.portal.Session::Closed`.
+
+        Supported keys in the @options vardict include:
+
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        The following results get returned in the @results vardict:
+
+        * ``session_handle`` (``o``)
+
+          The session handle. An object path for the
+          org.freedesktop.portal.Session object representing the created
+          session.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:org.freedesktop.portal.Request object representing this call
+
+        This method may only be called once and only on a session previously created with CreateDraftSession.
+        Initializes a capture input session previously created with CreateDraftSession. This may cause
+        the portal backend to query the user whether they want to allow the input of the session to be
+        captured or not, and what capabilities to support.
+
+        Supported keys in the @options vardict include:
+
+        * ``handle_token`` (``s``)
+
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
+          more information about the @handle.
+
+        * ``capabilities`` (``u``)
+
+          Bitmask of requested capabilities, see the SupportedCapabilities property.
+          This value is required and must not be zero.
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
+
+        * ``capabilities`` (``u``)
+
+          The capabilities available to this session. This is always a
+          subset of the requested capabilities.
+          See the SupportedCapabilities property for details. Note that
+          while a capability may be available to a session, there is no
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
       <arg type="a{sv}" name="options" direction="in"/>
       <arg type="o" name="handle" direction="out"/>
     </method>
--- a/src/input-capture.c
+++ b/src/input-capture.c
@@ -31,7 +31,8 @@
 #include "xdp-impl-dbus.h"
 #include "xdp-utils.h"
 
-#define VERSION_1  1 /* Makes grep easier */
+#define VERSION_1 1 /* Makes grep easier */
+#define VERSION_2 2 /* Makes grep easier */
 
 typedef struct _InputCapture InputCapture;
 typedef struct _InputCaptureClass InputCaptureClass;
@@ -73,6 +74,10 @@ typedef struct _InputCaptureSession
   XdpSession parent;
 
   InputCaptureSessionState state;
+
+  gboolean clipboard_requested;
+  gboolean clipboard_enabled;
+
 } InputCaptureSession;
 
 typedef struct _InputCaptureSessionClass
@@ -84,16 +89,39 @@ GType input_capture_session_get_type (void);
 
 G_DEFINE_TYPE (InputCaptureSession, input_capture_session, xdp_session_get_type ())
 
-G_GNUC_UNUSED static inline InputCaptureSession *
-INPUT_CAPTURE_SESSION (gpointer ptr)
+gboolean
+input_capture_session_can_request_clipboard (InputCaptureSession *session)
 {
-  return G_TYPE_CHECK_INSTANCE_CAST (ptr, input_capture_session_get_type (), InputCaptureSession);
+  if (session->clipboard_requested)
+    return FALSE;
+
+  if (xdp_dbus_impl_input_capture_get_version (impl) < 2)
+    return FALSE;
+
+  switch (session->state)
+    {
+    case INPUT_CAPTURE_SESSION_STATE_INIT:
+      return TRUE;
+    case INPUT_CAPTURE_SESSION_STATE_ENABLED:
+    case INPUT_CAPTURE_SESSION_STATE_ACTIVE:
+    case INPUT_CAPTURE_SESSION_STATE_DISABLED:
+    case INPUT_CAPTURE_SESSION_STATE_CLOSED:
+      return FALSE;
+    }
+
+  g_assert_not_reached ();
 }
 
-G_GNUC_UNUSED static inline gboolean
-IS_INPUT_CAPTURE_SESSION (gpointer ptr)
+gboolean
+input_capture_session_is_clipboard_enabled (InputCaptureSession *session)
 {
-  return G_TYPE_CHECK_INSTANCE_TYPE (ptr, input_capture_session_get_type ());
+  return session->clipboard_enabled;
+}
+
+void
+input_capture_session_clipboard_requested (InputCaptureSession *session)
+{
+  session->clipboard_requested = TRUE;
 }
 
 static InputCaptureSession *
@@ -226,12 +254,93 @@ static XdpOptionKey input_capture_create_session_options[] = {
   { "capabilities", G_VARIANT_TYPE_UINT32, validate_capabilities },
 };
 
+
+static gboolean handle_create_draft_session (XdpDbusInputCapture   *object,
+                                         GDBusMethodInvocation *invocation,
+                                         const char            *arg_app_id,
+                                         GVariant              *arg_options)
+{
+
+  return FALSE;
+  //Not Sure what happens here
+}
+
+static gboolean
+handle_init_session (XdpDbusInputCapture     *object,
+                       GDBusMethodInvocation *invocation,
+                       const char            *arg_session_handle,
+                       const char            *arg_parent_window,
+                       GVariant              *arg_options)
+{
+  XdpRequest *request = xdp_request_from_invocation (invocation);
+  g_autoptr(XdpDbusImplRequest) impl_request = NULL;
+  g_autoptr(GError) error = NULL;
+  g_auto(GVariantBuilder) options_builder =
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
+  g_autoptr(GVariant) options = NULL;
+  XdpSession *session;
+
+  REQUEST_AUTOLOCK (request);
+
+  impl_request =
+    xdp_dbus_impl_request_proxy_new_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (impl)),
+                                          G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
+                                          g_dbus_proxy_get_name (G_DBUS_PROXY (impl)),
+                                          request->id,
+                                          NULL, &error);
+  if (!impl_request)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  xdp_request_set_impl_request (request, impl_request);
+  xdp_request_export (request, g_dbus_method_invocation_get_connection (invocation));
+
+  session = XDP_SESSION (input_capture_session_new (arg_options, request, &error));
+  if (!session)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  if (!xdp_filter_options (arg_options, &options_builder,
+                           input_capture_create_session_options,
+                           G_N_ELEMENTS (input_capture_create_session_options),
+                           &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+  options = g_variant_ref_sink (g_variant_builder_end (&options_builder));
+
+  g_object_set_qdata_full (G_OBJECT (request),
+                           quark_request_session,
+                           g_object_ref (session),
+                           g_object_unref);
+
+  xdp_dbus_impl_input_capture_call_create_session (impl,
+                                                   request->id,
+                                                   session->id,
+                                                   xdp_app_info_get_id (request->app_info),
+                                                   arg_parent_window,
+                                                   options,
+                                                   NULL,
+                                                   create_session_done,
+                                                   g_object_ref (request));
+
+  xdp_dbus_input_capture_complete_create_session (object, invocation, request->id);
+
+  return G_DBUS_METHOD_INVOCATION_HANDLED;
+}
+
 static gboolean
 handle_create_session (XdpDbusInputCapture   *object,
                        GDBusMethodInvocation *invocation,
                        const char            *arg_parent_window,
                        GVariant              *arg_options)
 {
+  //TODO Reduce this to calling handle_create_draft_session and handle_init_session
   XdpRequest *request = xdp_request_from_invocation (invocation);
   g_autoptr(XdpDbusImplRequest) impl_request = NULL;
   g_autoptr(GError) error = NULL;
@@ -1150,7 +1259,7 @@ input_capture_init (InputCapture *input_capture)
 {
   unsigned int supported_capabilities;
 
-  xdp_dbus_input_capture_set_version (XDP_DBUS_INPUT_CAPTURE (input_capture), VERSION_1);
+  xdp_dbus_input_capture_set_version (XDP_DBUS_INPUT_CAPTURE (input_capture), VERSION_2);
 
   supported_capabilities =
     xdp_dbus_impl_input_capture_get_supported_capabilities (impl);
--- a/src/input-capture.c
+++ b/src/input-capture.c
@@ -52,7 +52,7 @@ static InputCapture *input_capture;
 
 static GQuark quark_request_session;
 
-GType input_capture_get_type (void);
+GType input_capture_get_type (void)  G_GNUC_CONST;
 static void input_capture_iface_init (XdpDbusInputCaptureIface *iface);
 
 G_DEFINE_TYPE_WITH_CODE (InputCapture, input_capture, XDP_DBUS_TYPE_INPUT_CAPTURE_SKELETON,
--- a/src/input-capture.h
+++ b/src/input-capture.h
@@ -23,5 +23,27 @@
 #include <gio/gio.h>
 #include <stdint.h>
 
+typedef struct _InputCaptureSession InputCaptureSession;
+
+GType input_capture_session_get_type (void);
+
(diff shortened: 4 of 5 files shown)


jadahl:
This was discussed a bit on Matrix, and we need to make a decision how to integrate with the clipboard portal here.

The input capture portal has the `CreateSession()` method as the main entry which queries the user etc (it is the only method with `parent_window`). However the existing way to integrate clipboard access with another portal (the remote desktop portal) is to `RequestClipboard()` on an exsiting session before it's "started", where starting is assumed to be the action that takes a `parent_window` and potentially queries the user.

It is important, however, that `RequestClipboard()` happens before the user is queried, so nagging only potentially happens once, as is the case with the clipboard portal. Thus, I see two options of how to deal with this:

1. Split up `org.freedesktop.portal.InputCapture.CreateSession()` into two parts: `CreateSession2()` (with a better name) and `InitSession()`. The `CreateSession2()` method would practically do nothing more than creating a dormant session that doesn't do anything, and `Init()` would carry the `parent_window` and carry all the parameters that `CreateSession()` has. `CreateSession()` would be a short hand for `CreateSession2(); InitSession();`. This way, clipboard integration can be done in a consistent manner, i.e. `CreateSession2(); RequestClipboard(); InitSession();`.
2. Make clipboard requesting part of `CreateSession()`, e.g. via a `request_clipboard` option in the vararg. This has the benefit of not having to split up `CreateSession()`, but also makes clipboard integration inconsistent and messy, e.g. tricker to carry the (so far unused `options` vararg of `RequestClipboard` and any semantics.

Cc: @whot, do you have any preference? Personally I'm somewhat more fond of option 1, assuming one can come up with a better name.

whot:
what may work: Instead of splitting you could have a `draft` boolean option in `CreateSession` that tells it to create a draft session that is later initialized with `Init()`? This way we don't need to split them up, I think. But otherwise I agree, option 1 seems like the more correct one.

sithlord48:
is there somewhere i can see an exmple of a draft option ? feel free to ping me on matrix or respond here.

jadahl:
> what may work: Instead of splitting you could have a draft boolean option in CreateSession that tells it to create a draft session that is later initialized with Init()? This way we don't need to split them up, I think. But otherwise I agree, option 1 seems like the more correct one.

A problem I see with that is that the CreateSession is a "request" that takes a parent_window, and it's a bit unorthodox to make an assumption that such a parent window is valid after the request was finished. As in, the potentially temporary resource "parent_window" would not be used until `Init()`.

whot:
An empty string for a parent window is valid, right? So the combination of empty parent window and draft set is sufficient (empty parent window can be a requiremen if draft is set?)

> is there somewhere i can see an exmple of a draft option ? feel free to ping me on matrix or respond here.

To copy/paste from how the documentation would look like:

```
 Supported keys in the options vardict include:

   draft (b)

   If true the created session is dormant until a subsequent call to Init(). This allows a caller 
   to invoke other functions on the session before prompting the user, e.g. RequestClipboard().
```
IOW it's just a boolean shoved into the `options` `a{sv}`.

sithlord48:
@whot  is that update correct in the xml  ?

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -86,6 +86,11 @@
           Bitmask of requested capabilities, see the SupportedCapabilities property.
           This value is required and must not be zero.
 
+        * draft (b)
```
whot, line 89:
```suggestion
        * ``draft`` (b)
```
I think, just check with the others please.

sithlord48:
If i could get some further guidance on this i added the draft but i am not sure what other changes in the code / xml files are needed.

jadahl:
> IOW it's just a boolean shoved into the options a{sv}.

What I found awkward is that it goes the whole "Request/Response" dance with nonsensical arguments when it could just be dumb method call, followed by the RequestResponse dance. It's also not unthinkable we might want to distinguish between options used when creating the session and initializing, which would be a bit messy if they were mixed between a Create and Init method. The alternative, deprecating CreateSession I'd imagine is something like:

```xml
    <!--
        CreateDraftSession:
        @session_handle: Object path for the created :ref:`org.freedesktop.portal.Session` object

        Creates a draft capture input session. A successfully created session can at
        any time be closed using :ref:`org.freedesktop.portal.Session.Close`, or may
        at any time be closed by the portal implementation, which will be
        signalled via :ref:`org.freedesktop.portal.Session::Closed`.

        Supported keys in the @options vardict include:

        * ``session_handle_token`` (``s``)

          A string that will be used as the last element of the session handle. Must be a valid
          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
          more information about the session handle.

        This method was added in version 2 of this interface.
    -->
    <method name="CreateDraftSession">
      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
      <arg type="o" name="session_handle" direction="out"/>
    </method>
    <!--
        InitSession:
        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
        @session_handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call

        Initializes a capture input session. This may result the portal backend querying the user whether
        they want to allow the input of the session to be captured or not, and what capabilities to support.

        Supported keys in the @options vardict include:

        * ``handle_token`` (``s``)

          A string that will be used as the last element of the @handle. Must be a valid
          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
          more information about the @handle.

        * ``capabilities`` (``u``)

          Bitmask of requested capabilities, see the SupportedCapabilities property.
          This value is required and must not be zero.

        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:

        * ``capabilities`` (``u``)

          The capabilities available to this session. This is always a
          subset of the requested capabilities.
          See the SupportedCapabilities property for details. Note that
          while a capability may be available to a session, there is no
          guarantee a device with that capability is currently available
          or if one does become available that it will trigger input capture.

          It is best to view this set as a negative confirmation - a
          capability that was requested but is missing is an indication that
          this application may not capture events of that capability.

        This method was added in version 2 of this interface.
    -->
    <method name="InitSession">
      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
    </method>
    <!--
        CreateSession:
        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call

        A short hand for calling :ref:`org.freedesktop.portal.InputCapture.CreateDraftSession()`
        followed by :ref:`org.freedesktop.portal.InputCapture.InitSession()`. See
        the corresponding methods for allowed options, and
        :ref:`org.freedesktop.portal.InputCapture.InitSession()` for the contents of the
        :ref:`org.freedesktop.portal.Request::Response` signal.

        Since version 2, this method is deprecated.
    -->
    <method name="CreateSession">
      <annotation name="org.freedesktop.DBus.Deprecated" value="true"/>
      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
    </method>
```

whot:
> The alternative, deprecating CreateSession I'd imagine is something like

without my brain fully engaged yet: this LGTM :)


> I have no idea what exactly i need todo here for that.

copy the new methods CreateDraftSession and Init to the XML and then implement the hooks/calls to it if a client registers version 2 of the protocol. Basically grep for anything CreateSession does and copy it for the draft session.

jadahl:
>  @whot just to clarify i need to make new methods in the input_capture.h named init and create_draft_session

More or less, I'd try to restructure things so that what `create_session` currently does (except the D-bus method call / Request plumbing) is split up in two helpers, one that creates the session, one that initializes. Then `create_session0` call both, `create_draft_session()` only calls the first and `init()` calls the second.

sithlord48:
Quick question for the xml changes do i add them to both just the impl or the non impl ?

whot:
> Quick question for the xml changes do i add them to both just the impl or the non impl ?

Both non-impl and impl

sithlord48:
updated the xml files to what i think is correct.

KoljaFrahm:
So is something still missing? I hope this can be merged soon!

sithlord48:
> So is something still missing? I hope this can be merged soon!

Yeah the code is missing then portal implementations have to also use v2 of the spec. I'm just focused on getting the xml first since those names are used in the code and that is the part im least sure of.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -56,44 +56,55 @@
...
    -->
   <interface name="org.freedesktop.portal.InputCapture">
-    <!--
-        CreateSession:
-        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+     <!--
+        CreateDraftSession:
         @options: Vardict with optional further information
-        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+        @session_handle: Object path for the created :ref:`org.freedesktop.portal.Session` object
```
whot, line 62: This may be intentional or it may be a misunderstanding how the requests work, I'm not sure :) Are you intentionally making this *not* go through a Request here? 

Makes sense since there's no reason this can fail or be delayed so requests aren't necessary but let's make sure it's spelled out in the commit message at least.

@jadahl - do you prefer this via Request for consistency with the others?

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -56,44 +56,55 @@
...
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
+    </method>
+    <!--
+        InitSession:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
```
whot, line 88:
```suggestion
        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
```
This one needs to go through a request.

The TLDR regarding requests: dbus times out faster than the user clicks so the workaround for the portals is that calls immediately return the object path for a temporary object (a `Request`) and that request then triggers a signal when the user is done. That request signal is the `Response` and contains the data we actually one (the `the following results get returned...` bit). The doc has a bit more details but the short summary is: anything that requires user interaction needs to go through a Request.

This doesn't apply to the impls since we can control the timeouts there so they don't need the request dance.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -106,14 +117,46 @@
...
+
+        * ``session_id`` (``s``)
+
+          The session id. A string representing the created capture input session.
+
+        * ``capabilities`` (``u``)
+
+          The capabilities available to this session. This is always a
+          subset of the requested capabilities.
+    -->
```
whot, line 159: Removing the `<method name="CreateSession">` is an API break, we need to keep that one around as-is.

src/input-capture.c
```diff
@@ -52,7 +52,7 @@ static InputCapture *input_capture;
 
 static GQuark quark_request_session;
 
-GType input_capture_get_type (void);
+GType input_capture_get_type (void) G_GNUC_CONST;
```
whot, line 55: Unrelated, should be a separate commit please, same with the clipboard hunks here which are a separate thing too.

src/input-capture.c
```diff
@@ -31,7 +31,7 @@
 #include "xdp-impl-dbus.h"
 #include "xdp-utils.h"
 
-#define VERSION_1  1 /* Makes grep easier */
+#define VERSION  2 /* Makes grep easier */
```
whot, line 34:
```suggestion
#define VERSION_1 1 /* Makes grep easier */
#define VERSION_2 2
```
And then use `VERSION_2` where we want that newer version.

whot (changes requested):

src/input-capture.c
sithlord48: Just to make sure i have this correct  these are the two commits you want ? 
1. commit that adds V2 methods 
2. commit that adds the clipboard chunks

data/org.freedesktop.portal.InputCapture.xml
sithlord48: I literally have no idea what i am doing. I was told on the matrix that support was as easy as copying over the methods used in remote desktop. But after i was told here in review to add these new methods etc.. So im just trying todo what was asked to get this working so we can use clipboard with deskflow, input-leap etc... 

If you or anyone who knows what they are doing wants to take this over im ok with that also . Otherwise ill keep chugging along until it gets there (just maybe alot more questions along the way)

data/org.freedesktop.portal.InputCapture.xml
sithlord48: I replaced the method in both XML files.

src/input-capture.c
jadahl: I think the point was to make the change about `G_GNUC_CONST` its own commit, since it's unrelated to everything else.

data/org.freedesktop.portal.InputCapture.xml
jadahl: I don't think we need to go through Request when creating a draft session.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -31,6 +31,119 @@
...
+        * ``session_id`` (``s``)
+
+          The session id. A string representing the created capture input session.
+
+        * ``capabilities`` (``u``)
+
+          The capabilities available to this session. This is always a
+          subset of the requested capabilities.
+    -->
+    <method name="CreateSession">
```
jadahl, line 64: There is already `CreateSession` in this interface, so no need to add this.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -113,7 +113,107 @@
...
+    <method name="InitSession">
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        CreateSession:
```
jadahl, line 188: There is already documentation for `CreateSession` at line 59.

data/org.freedesktop.impl.portal.InputCapture.xml
sithlord48: oh i see thanks updating soon.

data/org.freedesktop.portal.InputCapture.xml
sithlord48: noticed and removed the duplicate

data/org.freedesktop.portal.InputCapture.xml
sithlord48: Updated

sithlord48:
@jadahl  @whot is the XML correct now ?

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +114,78 @@
...
+      <arg type="o" name="session_handle" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+
+        Initializes a capture input session. This may result the portal backend querying the user whether
```
whot:
```suggestion
        Initializes a capture input session previously created with CreateDraftSession. This may result the portal backend querying the user whether
```

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +114,78 @@
...
+        they want to allow the input of the session to be captured or not, and what capabilities to support.
```
whot: For the more final versions we should also add things like:
```
This method may only be called once and only on a session previously 
created with CreateDraftSession.
```
and possibly

```
Clients implementing version 2 of this interface should prefer CreateDraftSession 
followed by InitSession over the version 1 CreateSession method.
```

whot (commented):
The XML looks good now, ready for implementing :)

data/org.freedesktop.portal.InputCapture.xml
sithlord48: I added the second note at the bottom of the createSession area line 70

sithlord48:
> The XML looks good now, ready for implementing :)

I changed the handle_create_session to be handle_init_session and added a for now empty
`handle_create_draft_session` as im not sure really what the return or input in there i also modified the old `handle_create_session` to call handle draft then init session

Thanks for your help and input I think we are getting close now.

data/org.freedesktop.portal.InputCapture.xml
sithlord48: Changed the description to 
```
 This method may only be called once and only on a session previously created with CreateDraftSession.
 Initializes a capture input session previously created with CreateDraftSession. This may cause
 the portal backend to query the user whether they want to allow the input of the session to be
 captured or not, and what capabilities to support.
```

sithlord48:
Oh i broke a test but not sure why or what it was testing either :(

sithlord48:
Every thing is updated, Tests no pass locally , gonna spin the review buttons

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,81 @@
...
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
```
swick: This describes the first in argument (0), so it should be 

```suggestion
      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
```

A `QVariantMap` maps to type `a{sv}` so that's right.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,80 @@
...
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
```
swick:
```suggestion
      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
```

sithlord48:
> Only test the version clipboard integration is added.

ok

> Perhaps do more or less what the remote desktop clipboard tests does?

You know i should have realized this

src/input-capture.c
```diff
@@ -226,8 +254,18 @@ static XdpOptionKey input_capture_create_session_options[] = {
   { "capabilities", G_VARIANT_TYPE_UINT32, validate_capabilities },
 };
 
+
+static gboolean handle_create_draft_session (XdpDbusInputCapture   *object,
```
sithlord48, line 258: According to the new Spec we no longer are taking a Parent_Window  in either `CreateDraftSession` or `InitSession` should one of them have a parent window ?

src/input-capture.c
```diff
@@ -226,8 +254,18 @@ static XdpOptionKey input_capture_create_session_options[] = {
...
+                                         GDBusMethodInvocation *invocation,
+                                         const char            *arg_parent_window,
+                                         GVariant              *arg_options)
+{
+  return FALSE;
```
sithlord48, line 264: Can anyone tell me what part of the old handle_session happens here now ?

src/input-capture.c
```diff
@@ -294,6 +332,16 @@ handle_create_session (XdpDbusInputCapture   *object,
...
 }
 
+static gboolean
+handle_create_session (XdpDbusInputCapture   *object,
+                       GDBusMethodInvocation *invocation,
+                       const char            *arg_parent_window,
+                       GVariant              *arg_options)
+{
+  handle_create_draft_session(object, invocation, arg_parent_window, arg_options);
+  return handle_init_session(object,invocation, arg_parent_window, arg_options);
```
jadahl: I think this function should remain unchanged. Only conceptually is it now `create_draft` + `init`, but since we need to support portal impls not updating in sync, we must still call the same `CreateSession` method on the impl object.

src/input-capture.c
```diff
@@ -226,8 +254,18 @@ static XdpOptionKey input_capture_create_session_options[] = {
...
+                                         const char            *arg_parent_window,
+                                         GVariant              *arg_options)
+{
+  return FALSE;
+  //Not Sure what happens here
+}
+
 static gboolean
-handle_create_session (XdpDbusInputCapture   *object,
+handle_init_session (XdpDbusInputCapture   *object,
```
jadahl: The function below should call `xdp_dbus_impl_input_capture_call_create_session()` and filter the options using a new `input_capture_init_session_options` array.

src/input-capture.c
```diff
@@ -226,8 +254,18 @@ static XdpOptionKey input_capture_create_session_options[] = {
   { "capabilities", G_VARIANT_TYPE_UINT32, validate_capabilities },
 };
 
+
+static gboolean handle_create_draft_session (XdpDbusInputCapture   *object,
+                                         GDBusMethodInvocation *invocation,
+                                         const char            *arg_parent_window,
```
jadahl: This one doesn't have a parent window, so should drop this argument.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,80 @@
...
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
```
jadahl, line 155: Needs a
```
@session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
```

src/input-capture.c
jadahl: Will also need a
```
const char *arg_session_handle,
```

src/input-capture.c
sithlord48: Removed

data/org.freedesktop.portal.InputCapture.xml
sithlord48: edit: (clarity)
Will this replace this in the options
```
        * ``handle_token`` (``s``)

          A string that will be used as the last element of the @handle. Must be a valid
          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
          more information about the @handle.
```

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,81 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <!--
+        CreateDraftSession:
```
jadahl, line 77: Should consider adding `app_id` here too. I see in the implementation in xdp-gnome that it would only be actually used for the request, but theoretically we might want to attach an app id to the session too, for whatever reason, and that means it needs to be passed here too.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,81 @@
...
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
```
jadahl, line 152: Also needs a
```
<arg type="o" name="handle" direction="in"/>
```
for the request part.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,81 @@
...
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
```
jadahl: It'd be safer to pass `<arg type="a{sv}" name="results" direction="out"/>` as the result value, just like `CreateSession`, with the following docs:

```
        The following results get returned in the @results vardict:
        * ``session_id`` (``s``)
The session id. A string representing the created capture input session.
```

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,81 @@
...
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
+          more information about the @handle.
+
+        * ``capabilities`` (``u``)
+
+          Bitmask of requested capabilities, see the SupportedCapabilities property.
+          This value is required and must not be zero.
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
```
jadahl: This should be
```
        The following results get returned in the @results vardict:
```
Signal emission by the `o.f.portal.InputCapture` part.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,80 @@
...
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
```
jadahl: Could consider making this a
```
```
with the `session_handle` being specified with 
```
        The following results get returned in the @results vardict:

        * ``session_handle`` (``o``)

          The session handle. An object path for the
          org.freedesktop.portal.Session object representing the created
          session.
```

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,80 @@
...
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
```
jadahl: This should be
```
```
since this is the handle of the `Request`.

data/org.freedesktop.portal.InputCapture.xml
jadahl: No, it refers to the `Session` created in `CreateDraftSession()`. See e.g. `GetZones()`.

sithlord48: Think i got this

sithlord48:
I think i got those all @jadahl can you check the xml is correct if so ill work to update the code

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,90 @@
...
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateDraftSession">
+      <arg type="o" name="session_handle" direction="in"/>
```
jadahl: Drop this, session handle is only valid after this call (since it crease the session).

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,90 @@
...
+               The following results get returned in the @results vardict:
+
+        * ``session_id`` (``s``)
+
+          The session id. A string representing the created capture input session.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
+      <arg type="o" name="handle" direction="in"/>
```
jadahl, line 153: Add
```
<arg type="o" name="session_handle" direction="in"/>
```
as the second argument.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,90 @@
...
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
```
jadahl: Make sure documented args matches the actual ones, including the order.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,90 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <!--
+        CreateDraftSession:
+        @options: Vardict with optional further information
+        @session_handle: Object path for the created :ref:`org.freedesktop.portal.Session` object
```
jadahl: Drop this, and at the end add 
```
```

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,90 @@
...
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+               The following results get returned in the @results vardict:
+
+        * ``session_id`` (``s``)
+
+          The session id. A string representing the created capture input session.
```
jadahl: The `session_id` part should be moved to `CreateDraftSession`.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,88 @@
...
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
```
jadahl, line 196: The first argument to this should be
```
      <arg type="o" name="session_handle" direction="in"/>
```

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,88 @@
...
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
```
jadahl: With the session handled argument added, this `In1` will be `In2`.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,88 @@
...
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
+    </method>
+
+    <!--
+        InitSession:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
```
jadahl: Make sure the argument documentation matches the actual arguments below.

jadahl (commented):
Getting there!

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +72,94 @@
...
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="InitSession">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="o" name="handle" direction="in"/>
```
jadahl, line 153: the order is different here from the other session methods (e.g. GetZones), here it's handle, session_handle, but below it's session_handle, handle. It seems the other methods below it's "handle, session_handle".

jadahl (commented):
Just a tiny thing left, (argument order), but other than that, the XML looks good now!

src/input-capture.c
sithlord48: Where are these defined ?

sithlord48:
Updated w/ the args @jadahl  They should be correct in code now (at least the args) . 
 I am still unsure how the code gets split into the two new methods

PR #1687, opened 2025-04-01
open-uri: Refactor handle Open to use a single thread

swick:


GeorgesStavracas:
I have no strong opinions about it, but I don't understand the reason for doing this change either.

AdrianVovk:
Why do we require a thread here at all?

PR #1690, opened 2025-04-03
Introduce Speech portal.

eeejay:
I think there is some confusion about what the speech provider API is and how it relates to libspiel. There are two parts:

1. A speech provider D-Bus interface that any session service can implement. If the session service's name ends with `.Speech.Provider`, apps can discover it, query its voices and synthesize text into audio. The interface defines two properties and one method: Name, Voices and Synthesize respectively. The spec for this interface is housed in its [own repo](https://github.com/project-spiel/libspeechprovider) that also describes a special binary stream format.
2. A client library called `libspiel` that makes the use of a distributed set of providers easy and straightforward for app developers. The API of libspiel is a lot more "evolved" and does not resemble the D-Bus API at all. Some of the features of the library include:
    1. A `Speaker` class that manages the play state of a queue of `Uterance` instances and emits signals for different speech and media events.
    2. An optional GStreamer source to use instead of the default output sink.
    3. Automatic selection of voices to use for each utterance depending on the language of the utterance and the user's preferences.
    4. A live list of available voices across all available providers
    5. A live list of providers that is kept updated by listening to the session bus for `ActivatableServicesChanged` and `NameOwnerChanged` signals.
    6. Other cool things I can't think of at the moment.

The portal API proposed here is aligned with the D-Bus interface above. It does not "proxy" or conform at all to `libspiel`! Because a sandboxed app cannot use or query the session bus the portal API also needs a way for apps to discover speech providers. So the only addition to the portal API over the host's speech provider interface is the `GetProviders` method. Again, this is not a "libspiel" concept, this functionality is a substitute for the session bus access that a host app would have.

## So why does the portal implementation depend on libspiel?
One of the hardest and most complex parts of libspiel is its ability to track available speech providers. It uses the two dbus signals mentioned above and works to keep a live list updated working through different kind of cases (installing uninstalling, activating, killing, and combinations of all). This is a brittle job and prone to bugs. I worked hard to have this be as robust as possible in libspiel and have a pretty comprehensive test suite for this.

Instead of copying all that logic into xdp I chose to rely on the implementation in libspiel and not duplicate the bugs and complexity into another project. Aside from the provider detection code, no other parts of libspiel are used in this portal. The synthesize method is called directly via a dbus proxy and the voices are taken from the proxy as well.

## Why expose providers at all? Why not just show a flat list of voices?
1. The current client implementation already exposes providers to apps. There was a request for that, and we added it to libspiel.
2. I don't think it is the portal's job to further abstract the host APIs and add more logic and complexity for voice management.
3. Exposing the providers allows sandboxed and non-sandboxed clients to share the same model of providers and voices. Why not just use the portal API in non-sandboxed environments? You could. But then you make the portal a hard requirement in the speech stack and I don't think that is wise.
4. My plan is to write a drop-in implementation in libspiel for sandboxed apps, so having a 1:1 relationship for voices and providers makes that easy. Otherwise we will need to deobfuscate and demarshall each voice and its associated provider adding more complexity.
5. We have been working through the speech API and figuring out real world cases for a while, I don't want to re-litigate those decisions and reinvent this all for portals.

## What data is streamed over the file descriptor?
Each voice provides a string that defines its audio format. In its simplest form it just writes raw audio to the file descriptor. The format string is parsed by gstreamer and looks like this:
`audio/x-raw,format=S16LE,channels=1,rate=22050`

Synthesized text can also provide embedded boundary events for when words or sentences or certain text offsets are spoken. For this we have another kind of mime type. A format string for that would look like this:
`audio/x-spiel,format=S16LE,channels=1,rate=22050`

This format is basically a stream of either audio data or event data chunks. The mimetype has the unfortunate name of `audio/x-spiel`, but is not related to libspiel, and it is implemented in a [standalone library](https://github.com/project-spiel/libspeechprovider) so that speech providers don't depend on libspiel.


Mikenux:
As for libspiel, I believe that if it is adopted by all DEs, it's acceptable. However, if there are different positions on which library to use, this might be problematic, as it would require supporting multiple libraries (especially for parts used without a corresponding method or option/property in the portal). But this will depend on the maintainers.

There may be a problem. If I understand correctly, you seem to be saying that apps can be providers, not just the system. If so, this is, in my opinion, incorrect because it implies that data (in text form) from one app can be shared with another app. App data can only be shared with the system, not with other apps, and only the system can be a provider. (sorry, I had missed this part)

eeejay:
## Shouldn't this be called the libspiel portal?

The speech providers that the client apps are accessing are not based on libspiel but on a simple [standard-ish d-bus protocol](https://project-spiel.org/dbus_iface.html). There is an option for providers to stream a special kind of audio format described above, but doing so is not implemented in libspiel, but a [separate library](https://project-spiel.org/libspeechprovider/) that is designed to be as minimal and portable as possible.

## Yeah, but isn't this still a "spiel" thing, not a generic speech service?

Spiel is the name I chose for the project of a standardized desktop speech platform. It would be the same as referring to the Location portal as the Geoclue2 portal.

## But aren't there other ways for applications to speak with other frameworks?

Yes. There is speech dispatcher, for example. Speech dispatcher is not a general use speech framework but designed for screen readers and other audio-only interfaces. "Spiel" has been adopted by Orca, WebKitGTK, and hopefully Firefox soon. A quick search on flathub reveals that the amount of talking applications is still pretty limited, and there is still a lot of potential for more adoption.

bilelmoussaoui:
Making a portal optional is imho a good way for not having any one implement it :)

data/org.freedesktop.portal.SpeechSynthesis.xml
```diff
@@ -0,0 +1,169 @@
...
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
+
+        * ``voices`` (``a(ssstas)``)
+
+          An array of voices. Each voice in the array is structure with the following members:
+            * A human readable name
+            * A unique identifier
+            * Synthesis output format
+            * A voice features bit field
```
bilelmoussaoui, line 114: any chance the bit field could be expanded with it possible values?

data/org.freedesktop.portal.SpeechSynthesis.xml
```diff
@@ -0,0 +1,169 @@
...
+        Synthesize:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+        @pipe_fd: File descriptor of pipe to write to.
+        @text: The text to be spoken.
+        @voice_id: The voice identifier for the voice that should be spoken.
+        @pitch: The voice pitch in which the text should be spoken.
+        @rate: The rate in which the text should be spoken.
+        @is_ssml: True if the text should be interpretted as an SSML snippet.
```
bilelmoussaoui, line 135: what is an ssml snippet?

bilelmoussaoui (commented):
didn't review it fully, just looked through the dbus iface

eeejay:
> Making a portal optional is imho a good way for not having any one implement it :)

What is this in response to?

data/org.freedesktop.portal.SpeechSynthesis.xml
eeejay: Yeah, I can add that. Will do that on monday.

data/org.freedesktop.portal.SpeechSynthesis.xml
eeejay: https://www.w3.org/TR/speech-synthesis/

smcv:
> Making a portal optional is imho a good way for not having any one implement it

All portals are optional: we can't force desktop environment developers to implement something that they don't want to implement.

bilelmoussaoui:
> > Making a portal optional is imho a good way for not having any one implement it
> 
> All portals are optional: we can't force desktop environment developers to implement something that they don't want to implement.

Sorry, should have been clearer. What I meant is having it optional at build time on xdp side.

data/org.freedesktop.portal.SpeechSynthesis.xml
bilelmoussaoui: would be nice to document it there or at least link to this.

eeejay:
> > > Making a portal optional is imho a good way for not having any one implement it
> > 
> > 
> > All portals are optional: we can't force desktop environment developers to implement something that they don't want to implement.
> 
> Sorry, should have been clearer. What I meant is having it optional at build time on xdp side.

libspiel is a relatively new library and I wanted to give the option to build xdp without it. Should I change that and make it a hard dependency?

eeejay:
I took inspiration from geoclue that has a similar build time option.

swick:
My 2 cents: I want a speech synthesis portal in xdp. I don't like that libspiel is supposed to be used in xdp when all it has to do is figure out which dbus service to call. This seems like side-stepping what xdp is supposed to do.

DemiMarie:
From a security perspective, I’m not a fan of having the string be passed directly to GStreamer without validation.

eeejay:
> From a security perspective, I’m not a fan of having the string be passed directly to GStreamer without validation.

You mean the audio format string voices provide? Yeah, we can validate it in the portal and limit what is allowed (for example only audio mime types).

swick:
Sorry for taking so long to get back to this. A few quick remarks, not a full review:

* I would like to avoid adding new portals which use the old threaded and async model. If you rebase on https://github.com/flatpak/xdg-desktop-portal/pull/1855 you can use libdex instead and make both of our lives easier.
* The libspeechprovider dependency and subproject seems to be left over from when it was required
* We use the Xdp prefix for portals nowadays and the G_DECLARE/G_DEFINE type macros

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+           An array of providers. Each provider in the array is structure with the following members:
+             * A well known name
+             * A human readable name
+    -->
+    <method name="GetProviders">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
```
swick, line 139: This doesn't seem to require user interaction, which means that this should always generate reply before the dbus timeout, which means that this does not require a Request.

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
+          more information about the @handle.
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
+
+        * ``voices`` (``a(ssstas)``)
+
+          An array of voices. Each voice in the array is structure with the following members:
```
swick: I would suggest to use an array of vardicts here; if we ever need to add another field, we'd need to add a `voices2` key otherwise. Same with the `providers` result in `GetProviders` but more likely to happen here.

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+            * A human readable name
+            * A unique identifier
+            * Synthesis output format
+            * A voice features bit field.
+              See here for full list: https://github.com/project-spiel/libspeechprovider/blob/main/interface/org.freedesktop.Speech.Provider.xml#L65
```
swick: I'd rather have this inline here. If the linked xml ever changes it becomes unclear what is actually supported in a particular version of the portal.

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+            * A list of languages the voice support represented as BCP 47 tags
```
swick: Please link to whatever defines `BCP 47 tags`

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+    -->
+    <method name="GetVoices">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="parent_window" direction="in"/>
+      <arg type="s" name="provider_id" direction="in"/>
```
swick: Is there any reason why this is two calls? You could just return all providers and voices from a single call, no?

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In10" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg direction="out" type="o" name="handle" />
+    </method>
+
+    <signal name="ProvidersChanged">
+      <arg type="o" name="session_handle" direction="in"/>
+    </signal>
+
+    <signal name="VoicesChanged">
```
swick, line 307: If the provides and voices can change at any time, do we need to sync this somewhat or is it okay to get an outdated view of the providers and voices, and pass in a `voice_id` to `Synthesize` that is not available anywhere?

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+      <arg direction="in"  type="h" name="pipe_fd" />
+      <arg direction="in"  type="s" name="text" />
+      <arg direction="in"  type="s" name="voice_id" />
+      <arg direction="in"  type="d" name="pitch" />
+      <arg direction="in"  type="d" name="rate" />
+      <arg direction="in"  type="b" name="is_ssml" />
+      <arg direction="in"  type="s" name="language" />
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In10" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg direction="out" type="o" name="handle" />
```
swick, line 287: I didn't see any kind of access control anywhere. Can all apps always use the speech synthesis? Should the user get prompted for anything?

data/org.freedesktop.portal.Speech.xml
swick: Similarly, you also don't need a `parent_window` because this is a non-interactive call and we don't have to show a UI.

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+        @pipe_fd: File descriptor of pipe to write to.
+        @text: The text to be spoken.
+        @voice_id: The voice identifier for the voice that should be spoken.
+        @pitch: The voice pitch in which the text should be spoken.
+        @rate: The rate in which the text should be spoken.
+        @is_ssml: True if the text should be interpretted as an SSML snippet (https://www.w3.org/TR/speech-synthesis/).
+        @language: The language the utterance should be spoken in. Some voices support more than one language.
+
+        This is the basic synthesis method.
+        When called, the speech provider will send the synthesized output to the given file descriptor.
```
swick, line 227: This needs to be more specific about what data comes in over the socket, most likely linking to the speech provider specification.

swick (commented):
Didn't look at the code yet. The interface seems reasonable overall.

src/speech.c
```diff
@@ -0,0 +1,994 @@
...
+          g_variant_new ("(os)", session->id, provider_id), &error))
+    {
+      g_warning ("Failed to emit VoicesChanged signal: %s", error->message);
+    }
+
+  return TRUE;
+}
+
+static void
+collect_providers_in_thread_func (GTask *task,
```
swick: So, this kind of discovery requires any app which provides a speech provider to own multiple bus names. I would suggest to instead use https://specifications.freedesktop.org/desktop-entry/latest/interfaces.html via `g_desktop_app_info_get_implementations` to get all the apps which implement a speech provider. This way, they can implement `org.freedesktop.Speech.Provider` on their main dbus name.

data/org.freedesktop.portal.Speech.xml
```diff
@@ -0,0 +1,176 @@
...
+      <arg type="o" name="handle" direction="out"/>
+    </method>
+
+    <!--
+        Synthesize:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+        @pipe_fd: File descriptor of pipe to write to.
+        @text: The text to be spoken.
```
swick, line 217: One big issue here is that apps are now sending potentially private information off to arbitrary apps which registered a speech provider. A malicious app could simply claim to provide a common speech provider and snoop on other apps like that.

I think the speech provider must be somehow bound to the app id which actually provides it. Even that doesn't completely help, because for example an app might just always use the first speech provider.

If the speech provider is not in the trusted system base (i.e. provided by the system), then the user must somehow be made aware to which app the text is going which claims to do synthesis for them.

jadahl: I would assume that as of now, speech providers can only be assumed to always part of the system. If/when we would allow installing speech providers as flatpaks, we'd have to have something equivalent to an a11y portal, where providers cant just add themselves.

eeejay: Speech providers are currently distributed as flatpaks themselves. That is a big selling point of this system since it allows modularity and installation of providers that cannot be distributed in most distros (non-free licenses, non-free training models, etc).

jadahl: And they are shipped with sandbox holes to make them able to hook up to some otherwise unavailable provider broker service?

eeejay: Not typically. An example of a provider is a Piper provider that just does the inference locally with voices (models) it has. So from a [Flatpak perspective](https://github.com/project-spiel/speech-provider-piper/blob/main/build-aux/ai.piper.Speech.Provider.json), for example it has no holes but just publishes a service with its app ID.

eeejay: That's not to say that a future provider would not rely on a cloud service for speech generation. We should provide add a flag for that to let the user know it is not a local synth. [Similar to the Web Speech API localService](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisVoice/localService).

eeejay: I would also consider not allowing "cloud providers" at all, and requiring providers to be completely sandboxed. This does touch a bit on the voice distribution question I brought up a few times. It would disallow providers to provide their own UI/store where users can download additional voices.

jadahl: > So from a Flatpak perspective, for example it has no holes but just publishes a service with its app ID.

I see, the speech provider broker lets any application be on the receiving end of a synthesis request without the user being really aware of it. That is indeed quite problematic from a privacy perspective.

Users being told what application is the provider in use has the problem of being of the "dismiss annoying dialog to get things working" type of prompts. Having to actually add (or rather, enable), on a system level, what applications can be speech providers would address that, but I suppose with this architecture, it'd have to be something handled in the speech provider broker, and the settings application.

eeejay: what if non-system providers (ie. flatpaks) were fully restricted and disallowed network access. Would that address the privacy concern?

jadahl: Hmm, perhaps. Disallowed network access would have to include localhost as well, and they would also need to be disallowed any IPC accept this "blessed" provider API, to disallow extracting gathered data that way. It's unclear to me how easily that would be enforceable.

eeejay: Also, since this is a portal we are talking about, we can have some UI where a user accepts the fact that their app is sending data to external untrusted services. Just like we have prompts for camera, location, etc.

DemiMarie: > Having to actually add (or rather, enable), on a system level, what applications can be speech providers would address that, but I suppose with this architecture, it'd have to be something handled in the speech provider broker, and the settings application.

This is definitely my preferred approach.

eeejay: installing flatpaks already happens on the system level, so this would more be a UX change. ie. its the difference between a user installing a flatpak via the "software" app or a settings pane.

swick: This is really about creating a system where the user is aware that all their data will be available to a flatpak app. I don't think it's reasonable to assume that a user is aware that simply installing any app gives that app access to everything that uses text to speech. Similarly, I don't believe it's reasonable to just "tell" the user that this is the case at an arbitrary point because users generally do not read messages like that. It really needs some kind of *action* from the user to enable an app as a speech provider!

So my suggestion: add a new RegisterAsProvider method. This calls into a portal impl which then can open e.g. a GNOME Settings page which shows all the available providers and if they are enabled or not, and highlights the speech provider which issued the request.

data/org.freedesktop.portal.Speech.xml
eeejay: I decided to have this because maybe we do want an interactive access dialog in the future? This would allow us to do that. I don't know on what basis that would be decided though

data/org.freedesktop.portal.Speech.xml
eeejay: I'll change it to a link to actual API docs instead of an arbitrary line number in a github repo.

data/org.freedesktop.portal.Speech.xml
eeejay: Yeah, I think that is the question.. Should we add an access dialog?

data/org.freedesktop.portal.Speech.xml
eeejay: Yeah, I guess so. I could change that.

data/org.freedesktop.portal.Speech.xml
eeejay: Sorry, I did `push -f` before I noticed you commented inline here and now I am not sure what you are referring to.

eeejay:
Thanks for your patience. This current patch has taken a lot of the feedback given. I might have missed a few points with the chaos of pushing a change before reading the inline stuff. Please let me know if there was anything I missed.

Here are the high level changes:

1. Ported to libdex using @swick's branch. This makes things so much easier! I needed to add a direct constructor to `XdpRequest` since `xdp_request_init_invocation` isn't called on fiber method handlers. The use of `XdpRequestFuture` is not possible since this portal does not have an implementation. Same with `XdpSessionFuture`.
2. Removed the providers getter and signal and stuck to a voices getter and signal. This simplifies the D-Bus API and allows getting everything in one go. I added a `GetProviderDetails` so that additional info on the providers could be retrieved, but it is not necessary for basic functionality.
3. Changed the provided voices to a vardict instead of a struct for extensibility.
4. I added an access dialog.
5. I improved the documentation, and documented the [speech stream format](https://project-spiel.org/libspeechprovider/overview.html).
6. Last, but not least, I introduced an "allow list" in the permissions store. The list details the speech providers that the host deems as trusted and can be exposed via the portal. The mechanics for altering this list are not in this patch, and may require implementation interfaces as mentioned in the comments, but this is out of scope for this PR.

data/org.freedesktop.portal.Speech.xml
eeejay: done. I added access dialogs.

swick:
Still busy with other things unfortunately, but good to see progress here.

You might want to take a look at XdpRequestDex and XdpSessionDex which should help simplifying things even more. The other thing that I'm a bit unsure about is the enumeration via dbus service names. I really feel like an [interface](https://specifications.freedesktop.org/desktop-entry/latest-single/#interfaces) is the better approach because we can store arbitrary data without having to activate the service.

Either way, I just merged the dex support and I'll also work on making it possible to have experimental portals, so I hope we can get this merged soon.

eeejay:
> You might want to take a look at XdpRequestDex and XdpSessionDex which should help simplifying things even more.

I wasn't sure how they simplify this portal. Since the portal has no implementation interface it seemed like more work using those because there would need to be some modification to their constructors. Happy to switch over to them if I am wrong though.

> The other thing that I'm a bit unsure about is the enumeration via dbus service names. I really feel like an [interface](https://specifications.freedesktop.org/desktop-entry/latest-single/#interfaces) is the better approach because we can store arbitrary data without having to activate the service.

Is `Implements` supported in `.service` files? Not all speech providers will have `.desktop` files because they may not have an app/GUI. Would this still be worthwhile?

> 
> Either way, I just merged the dex support and I'll also work on making it possible to have experimental portals, so I hope we can get this merged soon.

Thanks!

swick:
> I wasn't sure how they simplify this portal. Since the portal has no implementation interface it seemed like more work using those because there would need to be some modification to their constructors. Happy to switch over to them if I am wrong though.

There are a few good reasons to use them:

* They work without threads. You currently use SESSION_AUTOLOCK_UNREF for example because the old session class requires locking because it internally uses threads.
* No subclassing. You can store the session in your own struct/class however you want.
* Eliminates state tracking. The new session class automatically emits close, calls close for the backend, etc when the ref count reaches zero.

> Is Implements supported in .service files? Not all speech providers will have .desktop files because they may not have an app/GUI.

No, but you can still ship desktop files for non-UI applications with `NoDisplay=true`.

eeejay:
Ok, updated request/session to dex versions. Regarding `Implements`, seems like a non-breaking addition we can do in the future. I'll try to tackle that when I have free cycles.

philn:
Thank you @eeejay for working on this! 

We have a libspiel backend in WebKit but it's currently disabled because it runs in the sandboxed WebProcess. Having a portal would definitely be welcome :)

PR #1703, opened 2025-05-01
input-capture: Fix state after a Deactivated signal

whot:
Copy/paste error from the Activated handling: A Deactivated signal changes the input capture session back into ENABLED.

### DIFF
--- a/src/input-capture.c
+++ b/src/input-capture.c
@@ -1107,7 +1107,7 @@ on_deactivated_cb (XdpDbusImplInputCapture *impl,
         break;
       case INPUT_CAPTURE_SESSION_STATE_ACTIVE:
         pass_signal (impl, "Deactivated", session_id, options, data);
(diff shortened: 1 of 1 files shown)


whot:
FTR, I did try to add a test to this but utterly failed because scheduling multiple `Glib.timeouts` with the same callback is apparently not possible?

swick (approved):

jadahl:
> FTR, I did try to add a test to this but utterly failed because scheduling multiple Glib.add_timeout with the same callback is apparently not possible?

You sure about that? Just tested it here, and it worked fine (using C).

jadahl (approved):

swick:
Exceptions inside the glib callbacks do not propagate properly, so you might just have seen that?

whot:
> You sure about that? Just tested it here, and it worked fine (using C).

I used the python bindings, I'm sure it works in C because I've done that in the past. Not sure what exactly happened but yes, probably an exception in the callback and I forgot to try/except that one...

PR #1707, opened 2025-05-10
Load portal configuration from config dirs

superatomic:
This PR modifies `load_installed_portals` to check config directories, in the same exact way that `load_portal_configuration` currently checks config directories first, and then data directories. This makes loading configuration files more consistent.

### DIFF
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -349,7 +349,27 @@ load_installed_portals (gboolean opt_verbose)
       goto out;
     }
 
+  /* $XDG_CONFIG_HOME/xdg-desktop-portal/portals */
+  user_portal_dir = g_build_filename (g_get_user_config_dir (), XDP_SUBDIR, "portals", NULL);
+  load_installed_portals_dir (portals, user_portal_dir, opt_verbose);
+
+  /* $XDG_CONFIG_DIRS/xdg-desktop-portal/portals */
+  dirs = g_get_system_config_dirs ();
+
+  for (iter = dirs; iter != NULL && *iter != NULL; iter++)
+    {
+      g_autofree char *dir = NULL;
+
+      dir = g_build_filename (*iter, XDP_SUBDIR, "portals", NULL);
+      load_installed_portals_dir (portals, dir, opt_verbose);
+    }
+
+  /* ${sysconfdir}/xdg-desktop-portal/portals */
+  portal_dir = SYSCONFDIR "/" XDP_SUBDIR "/portals";
+  load_installed_portals_dir (portals, portal_dir, opt_verbose);
+
   /* $XDG_DATA_HOME/xdg-desktop-portal/portals */
+  g_clear_pointer (&user_portal_dir, g_free);
   user_portal_dir = g_build_filename (g_get_user_data_dir (), XDP_SUBDIR, "portals", NULL);
   load_installed_portals_dir (portals, user_portal_dir, opt_verbose);
 


swick:
Fine with me, but would like to have the opinion of @ebassi and @smcv

smcv (approved):

smcv:
Implementation-wise this seems fine, but what is the use-case for loading e.g. `xdg-desktop-portal/portals/gnome.portal` from configuration locations?

> This makes loading configuration files more consistent.

But these are not configuration files: they are integration files, telling one component (x-d-p) about the existence and capabilities of other installed components (for example x-d-p-gnome). How is that configuration?

smcv:
The commit message doesn't say why either, not even "makes loading configuration files more consistent".

I think this is perhaps a good example of how a PR to change behaviour is not always a great starting point: often it's better to have a solution-neutral problem statement, and then propose a solution to that problem. That gives reviewers much better context for understanding what you're trying to achieve, without which we can't really assess whether a proposed change is the best way to achieve it.

superatomic:
My use case is to implement my own portals to create custom behavior. Specifically, I wrote a custom implementation for `org.freedesktop.impl.portal.AppChooser` which I wanted to load. It makes for since to me to load from a config directory instead of a data one, because it is a custom change I made with a short python script, and also because the systemd unit for the backend is also in a config directory.

Since config paths are loaded before data paths, this also allows for overriding the `DBusName` or `Interfaces` fields for an existing backends as well. This makes it work in the same way that other programs (say `systemd`) load configurations.

Apologies for not creating an issue first. When I looked into the implementation of adding this behavior, it was fairly easy, so I figured it would be helpful to provide the implementation for the change I was requesting. I should have definitely provided more context.

smcv:
> My use case is to implement my own portals to create custom behavior. Specifically, I wrote a custom implementation for org.freedesktop.impl.portal.AppChooser which I wanted to load.

The intended way to achieve this is `~/.local/share/xdg-desktop-portal/portals/myappchooser.portal`.

> It makes for since to me to load from a config directory instead of a data one, because it is a custom change I made with a short python script, and also because the systemd unit for the backend is also in a config directory.

I don't think that follows. Yes your short Python script is a local tweak, but that doesn't make it configuration: in terms of the overall system design, it's a piece of locally-written software that you have installed. If it was an app or script that you wanted to appear in menus or be a MIME-type handler, you would have to register it via `~/.local/share/applications/myapp.desktop`; if it was a D-Bus-activatable service, you'd have to register it via `~/.local/share/dbus-1/services/com.example.MyService.service`; and if it came with an icon theme, you'd have to provide it in `~/.local/share/icons`. Registering it as a possible portal backend is "the same shape" as those use-cases.

Declaring that, out of all the possible portal backends, you want to use yours (and only yours) to implement `AppChooser` *is* configuration, and you can do that via [portals.conf(5)](https://flatpak.github.io/xdg-desktop-portal/docs/portals.conf.html). This is similar to the way locally-written apps that are MIME-type handlers for `image/jpeg` have to declare themselves as such in `~/.local/share/applications`, but if you want to say "when I open an `image/jpeg`, I want it to open in `eog`" the place you do that is `~/.config/mimeapps.list`.

systemd units do treat `~/.config/systemd` as equivalent to `~/.local/share/systemd`, but that's because the format of systemd units is a mixture of integration glue (declarative facts) that is appropriate for `~/.local/share`, and configuration (opinions) that is appropriate for `~/.config`.

> Since config paths are loaded before data paths, this also allows for overriding the `DBusName` or `Interfaces` fields for an existing backends as well.

Why would that be useful? The `DBusName` and `Interfaces` are facts about the backend, not configuration: if they match the real behaviour of the backend, it will work correctly, otherwise it will not work correctly. You can't just arbitrarily change them: you would have to also change the implementation of the backend, and recompile/reinstall it. Of course if your locally-written backend is a small Python script, there is no "recompile", and perhaps you are editing it in-place instead of "installing" anything - but that's a trivial case of the typical design of a backend.

If you have an existing backend (for example x-d-p-gnome) that implements a particular interface (for example `FileChooser`) but you don't want to use it for that interface, writing a `.portal` file that pretends x-d-p-gnome doesn't implement FileChooser is not the right way to achieve that. In the past it was sometimes used as a workaround, but now that we have [portals.conf(5)](https://flatpak.github.io/xdg-desktop-portal/docs/portals.conf.html), instead you can write a `portals.conf` that says `org.freedesktop.impl.portal.FileChooser=otherbackend`, or even `org.freedesktop.impl.portal.FileChooser=none`.

superatomic:
That makes sense! Thank you for your explanation.

PR #1713, opened 2025-05-16
xdp-utils: Try PIDFD_GET_PID_NAMESPACE ioctl first to get pidns

swick:
The ioctl is the less brittle path to get the pid namespace from a pidfd, so let's try it first, and if the kernel doesn't support it, fall back to the previous code path.

### DIFF
--- a/src/xdp-utils.c
+++ b/src/xdp-utils.c
@@ -29,6 +29,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/ioctl.h>
 
 #include <gio/gio.h>
 #include <gio/gunixoutputstream.h>
@@ -38,6 +39,9 @@
 #define DBUS_NAME_DBUS "org.freedesktop.DBus"
 #define DBUS_INTERFACE_DBUS DBUS_NAME_DBUS
 #define DBUS_PATH_DBUS "/org/freedesktop/DBus"
+
+#define PIDFS_IOCTL_MAGIC 0xFF
+#define PIDFD_GET_PID_NAMESPACE _IO(PIDFS_IOCTL_MAGIC, 5)
 
 /* Based on g_mkstemp from glib */
 gint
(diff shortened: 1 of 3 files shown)


swick:
Thanks! I threw this together in a few minutes and didn't bother because I want to add pidfd handling to glib anyway soon-ish. Now I couldn't resist though, so I updated the PR.

It differentiates between pidfds (i.e. the actual kernel concept of a fd that represents a process) and `/proc/$PID/` dirfds which makes things more clear. I think that this is what confused the person who screwed this up (probably me).

EBADBEEF:
These changes look good to me, but it's not working for strange reasons.

Error: `Realtime error: Could not get pidns for pid 23131: PIDFD_GET_PID_NAMESPACE ioctl failed: Invalid argument`
Strace: `[pid 25828] ioctl(19, PIDFD_GET_PID_NAMESPACE, 0x7ffff4ffa748) = -1 EINVAL (Invalid argument)`

If I do the ioctl again it works. If I add usleep(1) before the ioctl it works. If I printf before the ioctl it works.

Tried with kernel 6.14.8 and 6.15.1.

EBADBEEF:
Duh, I figured it out. The argument is not always empty if you don't explicitly pass one in. See the strace above:
```
ioctl(19, PIDFD_GET_PID_NAMESPACE, 0x7ffff4ffa748)
```

The kernel rejects ioctls on pidfd if they have an argument, see kernel/fs/pidfs.c pidfd_ioctl:
```
static long pidfd_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
...
    if (arg)
        return -EINVAL;
```

You will need to explicitly pass an empty one in:
```
ioctl (pidfd, PIDFD_GET_PID_NAMESPACE, 0);
```

src/xdp-utils.c
```diff
@@ -908,17 +913,52 @@ xdp_pidfds_to_pids (const int  *pidfds,
...
+                     ino_t   *ns,
+                     GError **error)
+{
+  g_autofd int pidns_fd = -1;
+  struct stat st;
+
+  g_return_val_if_fail (pidfd >= 0, FALSE);
+  g_return_val_if_fail (ns != NULL, FALSE);
+
+  pidns_fd = ioctl (pidfd, PIDFD_GET_PID_NAMESPACE);
```
EBADBEEF: pidns_fd = ioctl (pidfd, PIDFD_GET_PID_NAMESPACE, 0);

src/xdp-utils.c
```diff
@@ -29,6 +29,8 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/ioctl.h>
+#include <linux/types.h>
```
EBADBEEF: why do you need linux/types.h?

swick:
Didn't really expect this to get merged without tests...

kotauskas:
Building this from source and running yields `PIDFD_GET_PID_NAMESPACE ioctl failed: Inappropriate ioctl for device` on my system. The old `6.6.87-rt54-arch1-1-rt-lts` kernel is surely to blame, but there certainly has to be a changelog entry documenting that kernel minor branches from before [October 2024](<https://lwn.net/Articles/992740/>) don't benefit from this bugfix.

kotauskas:
Just switched to `6.14.0-rt3-arch1-1-rt`, works as expected.

PR #1665, opened 2025-03-13
Load portal config from data dirs with priority

swick:
Continuation of https://github.com/flatpak/xdg-desktop-portal/pull/1561 to resolve https://github.com/flatpak/xdg-desktop-portal/issues/603.

Tried for a few hours to get proper tests for the configuration (unshare, setting xdg base dir env, ...) but gave up in the end.

PR #1716, opened 2025-05-17
configuration: Allow fallbacks from XDG_DESKTOP_PORTAL_DIR.

Apteryks:
When XDG_DESKTOP_PORTAL_DIR is set, it causes the rest of the search mechanism to be aborted.  Instead allow the search to continue.

* src/xdp-portal-impl.c (load_portal_configuration): Return only in case a match was found from XDG_DESKTOP_PORTAL_DIR.

### DIFF
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -501,9 +501,8 @@ load_portal_configuration (gboolean opt_verbose)
 
   if (portal_dir != NULL)
     {
-      load_config_directory (portal_dir, desktops, opt_verbose);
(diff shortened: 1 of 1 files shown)


swick:
No, not happening. XDG_DESKTOP_PORTAL_DIR is an internal environment variable for testing and not searching anywhere else is part of the feature.

smcv:
Now that #1665 has been merged (it's in 1.20.1) please set `XDG_DATA_HOME` or `XDG_DATA_DIRS` instead. They are searched before the path that is hard-coded at build-time.

PR #1627, opened 2025-02-12
Make XdpAppInfo more testable

swick:
See commits

PR #1720, opened 2025-05-19
app-info: Check that g_initable_new succeeds before modifying

swick:


jadahl:
Probably want to test what the effects are when a flatpak app info didn't manage to get created, but need https://github.com/flatpak/xdg-desktop-portal/pull/1627 for that.

smcv (approved):

smcv:
I think it would be a good idea to cherry-pick this into 1.20.x, too.

swick:
Yeah, this and https://github.com/flatpak/xdg-desktop-portal/pull/1721 are regressions we introduced since the last release.

Issue #1719, opened 2025-05-18
g_close(fd:32) failed with EBADF. The tracking of file descriptors got messed up


superatomic:
### Operating System

Fedora Workstation 42

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.20.1

### Desktop Environment

GNOME

### Expected Behavior

No warnings are outputted.

### Current Behavior

The following warnings appear in the system journal a few seconds after the service starts.
```
g_close(fd:31) failed with EBADF. The tracking of file descriptors got messed up
Error closing pidfd: Bad file descriptor
```

### Steps to Reproduce

1. Run `systemctl --user restart xdg-desktop-portal.service`
2. Wait a few seconds. Sometimes it seems like switching the focused window helps it trigger.
3. Check the system journal for the service.

This warning will occur repeatedly, not just once. Restarting the service is not necessary to trigger the bug.

### Anything else we should know?

This warning did not occur on version 1.20.0.

PR #1721, opened 2025-05-19
app-info/flatpak: Steal bwrap pidfd because AppInfo doesn't dup anymore

swick:
Fixes: 2d8c2472 ("app-info: Steal pidfd when constructing a XdpAppInfo")
Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1719

### DIFF
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -823,7 +823,7 @@ xdp_app_info_flatpak_new (int      pid,
                                      "flags", flags,
                                      "id", id,
                                      "instance", instance,
-                                     "pidfd", bwrap_pidfd,
+                                     "pidfd", g_steal_fd (&bwrap_pidfd),
                                      NULL);
 
   app_info_flatpak->flatpak_info = g_steal_pointer (&metadata);


smcv:
In future I would prefer it if we can avoid doing refactoring in a stable branch (in this case 1.20.0 → 1.20.1) because it tends to come with a higher risk of regressions like the one fixed by this PR.

smcv (approved):

swick:
> In future I would prefer it if we can avoid doing refactoring in a stable branch (in this case 1.20.0 → 1.20.1) because it tends to come with a higher risk of regressions like the one fixed by this PR.

Personally I would like if we could not be stuck on a stable series in the main branch, and even if we dropped the stable series concept. It's not great when feature work is slowed down so much by arbitrary time durations which also means we merge some things we probably shouldn't in a stable series. We should just do releases and then backport important fixes.

smcv:
> Personally I would like if we could not be stuck on a stable series in the main branch

Branching immediately after a stable .0 release, so that `main` is always open for new features, seems good - in dbus I even branched for 1.16.x when I was still doing release candidates (1.15.9x), although that does mean more need for cherry-picking.

> and even if we dropped the stable series concept

I have to put *something* in Debian stable releases, and similarly packagers for other LTS distros like Ubuntu need to choose something. As an upstream, if we have a branch with backported bug fixes, we can encourage distros to align on that branch and take its point releases (giving us some sort of influence over what they use), but if we only have feature releases, distro users will be stuck with only the changes that distros backport for them (in practice only security fixes, if at all).

Issue #1726, opened 2025-05-19
fails to compile with GLib 2.74.x (Debian 12)


smcv:
### Operating System

Debian 12

### XDG Desktop Portal version

1.20

### XDG Desktop Portal version (Other)

1.20.2 or contemporary version from git main

### Desktop Environment

GNOME

### Expected Behavior

Either a successful build, or a `meson setup` failure because my GLib is an unsupported version

### Current Behavior

```
../src/xdp-portal-impl.c: In function ‘load_installed_portals’:
../src/xdp-portal-impl.c:374:21: warning: implicit declaration of function ‘g_hash_table_steal_all_values’; did you mean ‘g_hash_table_steal_all’? [-Wimplicit-function-declaration]
  374 |   implementations = g_hash_table_steal_all_values (portals);
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                     g_hash_table_steal_all
../src/xdp-portal-impl.c:374:19: warning: assignment to ‘GPtrArray *’ {aka ‘struct _GPtrArray *’} from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
  374 |   implementations = g_hash_table_steal_all_values (portals);
      |                   ^
../src/xdp-portal-impl.c:375:3: warning: implicit declaration of function ‘g_ptr_array_sort_values’; did you mean ‘g_ptr_array_sort’? [-Wimplicit-function-declaration]
  375 |   g_ptr_array_sort_values (implementations, sort_impl_by_use_in_and_name);
      |   ^~~~~~~~~~~~~~~~~~~~~~~
      |   g_ptr_array_sort
```

### Steps to Reproduce

1. Debian 12 system (in particular this has GLib 2.74.x)
2. `sudo apt build-dep xdg-desktop-portal`
3. `meson setup _build -Dprefix=/usr`
4. `meson compile -C _build`

### Anything else we should know?

`meson.build` declares a dependency on GLib 2.72. `g_hash_table_steal_all_values()` and `g_ptr_array_sort_values()` were new in GLib 2.76.

smcv:
For the 1.20.x branch I think it would be best if we can continue to be buildable with the declared GLib version (2.72). I'll send a PR shortly (but note that I can't run the test suite on Debian 12, because its `dbusmock` is too old to have `dbusmock.BusType`).

For `main`, I think it would be fine to bump the minimum GLib to 2.76 and revert the addition of 2.72 compatibility.

PR #1728, opened 2025-05-19
portal-impl: Add a fallback code path for GLib older than 2.76

smcv:
meson.build still declares 2.72 as our minimum version.

Resolves: #1726

### DIFF
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -327,6 +327,31 @@ load_installed_portals_dir (GHashTable *portals,
     }
 }
 
+#if !GLIB_CHECK_VERSION(2, 76, 0)
+static gboolean
+steal_portal_impl_foreach (gpointer key,
+                           gpointer value,
+                           gpointer user_data)
+{
+  GPtrArray *implementations = user_data;
+
+  g_free (key);
+  g_ptr_array_add (implementations, value);
+  /* declare that we took ownership */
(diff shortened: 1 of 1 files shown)


smcv:
Marked as draft until I figure out a way to run the test suite against this code path (old GLib, new dbusmock).

smcv:
On Debian 12 (GLib 2.72), compiles successfully (but dbusmock is too old to run the tests).

On Debian 13 (GLib 2.84), passes tests with this hack to force it onto the old-GLib code path, either built normally or with ASan and `ASAN_OPTIONS=detect_leaks=0`. The need to turn off leak detection doesn't seem to be a regression, possibly a false positive (for example caused by building against a non-ASan GLib).

```diff
diff --git a/src/xdp-portal-impl.c b/src/xdp-portal-impl.c
index 27ffd221..8ac22cab 100644
--- a/src/xdp-portal-impl.c
+++ b/src/xdp-portal-impl.c
@@ -31,6 +31,9 @@
 #include <glib.h>
 #include <gio/gio.h>
 
+#undef GLIB_CHECK_VERSION
+#define GLIB_CHECK_VERSION(x, y, z) (0)
+
 typedef struct _PortalInterface {
   /* dbus_name is NULL if this is the default */
   char *dbus_name;
```

GeorgesStavracas (approved):

Issue #1718, opened 2025-05-18
x-d-p crashes or does not allow portal operations if Flatpak app does not export APP_ID.desktop


Kimiblock:
### Operating System

Arch Linux

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.20.1

### Desktop Environment

GNOME

### Expected Behavior

Portal runs normally

### Current Behavior

Portal crashes.

### Steps to Reproduce

1. Install [Portable Sandbox](https://github.com/Kraftland/portable): `aur/portable-git`
2. Run `portable-pools test`, which will create App ID `top.kimiblock.test`
3. Notice that Portal crashes immediately even before sandbox is started, just after the D-Bus proxy is started.
...

### Anything else we should know?

coredumpctl:

```
➜  Downloads coredumpctl gdb 669903
           PID: 669903 (xdg-desktop-por)
           UID: 1011 (kimiblock)
           GID: 1011 (kimiblock)
        Signal: 11 (SEGV)
     Timestamp: Mon 2025-05-19 00:01:16 CST (1min 31s ago)
  Command Line: /usr/lib/xdg-desktop-portal
    Executable: /usr/lib/xdg-desktop-portal
 Control Group: /user.slice/user-1011.slice/user@1011.service/session.slice/xdg-desktop-portal.service
          Unit: user@1011.service
     User Unit: xdg-desktop-portal.service
         Slice: user-1011.slice
     Owner UID: 1011 (kimiblock)
       Boot ID: 6d60f9ae66684b579474d272e532f165
    Machine ID: f0d0a842ac1841608a510f64913f0338
      Hostname: kimiblocks-workstation
       Storage: /var/lib/systemd/coredump/core.xdg-desktop-por.1011.6d60f9ae66684b579474d272e532f165.669903.1747584076000000.zst (present)
  Size on Disk: 634.1K
       Message: Process 669903 (xdg-desktop-por) of user 1011 dumped core.
                
                Stack trace of thread 669908:
                #0  0x00005c8a7d67edd4 n/a (/usr/lib/xdg-desktop-portal + 0x8edd4)
                #1  0x00005c8a7d66cb09 n/a (/usr/lib/xdg-desktop-portal + 0x7cb09)
                #2  0x00007583d2c74371 n/a (libgio-2.0.so.0 + 0x73371)
                #3  0x00007583d2bb3812 g_closure_invoke (libgobject-2.0.so.0 + 0x11812)
[...]

PR #1729, opened 2025-05-19
app-info: Don't require a .desktop file for Flatpak and Snap apps

smcv:
* app-info: Don't require a .desktop file for Flatpak apps
    
    Several portal interfaces like Background, Location and Wallpaper
    try to use a "main" .desktop file from the sandboxed app to provide
    a "friendly name" and/or an icon. However, not all Flatpak apps export
    a .desktop file of the same name as their app-ID, as demonstrated by
    org.gnome.NautilusPreviewer. Portal interfaces like Background are
    expected to fall back to displaying the app-ID if they cannot derive a
    "friendly name" from the .desktop file.
    
    Restore the 1.20.0 functionality by not requiring a GAppInfo
    (.desktop file)
    
    Resolves: https://github.com/flatpak/xdg-desktop-portal/issues/1718

* app-info: Don't require a .desktop file for Snap apps
    
    It isn't clear to me whether Snap apps can lack a "main" .desktop file,
    but before 1.20.0 we didn't require it, and we need to be able to cope
    with absence of a .desktop file *anyway* (for host and Flatpak apps),
    so the conservative thing to do here is to not require it.
    
    I've made this its own commit, so that it can be reverted if a Snap
    expert tells us that this is a "can't happen" situation.

----

The practical result of this branch is that the only apps that require a .desktop file are host apps that call into `org.freedesktop.host.portal.Registry`, which is inherently opt-in.

### DIFF
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -811,7 +811,6 @@ xdp_app_info_flatpak_new (int      pid,
 
   /* TODO: we can use pidfd to make sure we didn't race for sure */
 
-  flags |= XDP_APP_INFO_FLAG_REQUIRE_GAPPINFO;
   flags |= XDP_APP_INFO_FLAG_SUPPORTS_OPATH;
   if (has_network)
     flags |= XDP_APP_INFO_FLAG_HAS_NETWORK;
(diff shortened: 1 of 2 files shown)


smcv:
Please consider this for backport to 1.20.x.

GeorgesStavracas (approved):

PR #1734, opened 2025-06-03
Refactor pid status lookup

swick:


EBADBEEF:
FYI I left some comments on #1713 that apply here too. And for linking/tracking, this and #1713 fix issue #1653 .

src/xdp-utils.c
```diff
@@ -1112,6 +991,88 @@ pid_dirfd_parse_status (int      pid_dirfd,
...
+
+  f = fdopen (g_steal_fd (&status_fd), "r");
+  if (f == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, g_io_error_from_errno (errno),
+                   "Could not fdopen pidfd fdinfo: %s", g_strerror (errno));
+      return FALSE;
+    }
+
+  r = parse_status_file (f, pid_out, nspid_out, uid_out);
```
AdrianVovk, line 1031: Fundamentally I've got a tad bit of concern about this approach here, because the fdinfo file is not the same thing as the status file. It's formatted similarly, sure, and even reuses the `Pid` and `NsPid` field names yes, but it doesn't have all the status file fields.

src/xdp-utils.c
```diff
@@ -1112,6 +991,88 @@ pid_dirfd_parse_status (int      pid_dirfd,
...
+
+pid_t
+xdp_pidfd_to_pid (int      pidfd,
+                  GError **error)
+{
+  pid_t pid;
+
+  g_return_val_if_fail (pidfd >= 0, -1);
+
+  if (!pidfd_parse_status (pidfd, &pid, NULL, NULL, error))
```
AdrianVovk, line 1052: Modern kernels actually have an ioctl for this, which we should use instead whenever available

Here's systemd's implementation: https://github.com/systemd/systemd/blob/main/src/basic/pidfd-util.c#L85

src/xdp-utils.c
```diff
@@ -1053,11 +1060,11 @@ parse_status_file (FILE  *f,
       g_warning ("Failed to parse 'status::%s': %s",
                  key, g_strerror (-r));
 
-  } while (r == 0 && (!have_uid || !have_pid));
+  } while (r == 0 && (!have_pid || !have_nspid || !have_uid));
 
   if (r != 0)
     return r;
-  else if (!have_uid || !have_pid)
+  else if (!have_pid || !have_nspid || !have_uid)
```
AdrianVovk, line 946: Actually right here it bites you that the fdinfo file is not the same thing as a status file. The function that looks at status files will return -ENXIO if there's no `Uid` field. The fdinfo file for a pidfd doesn't have a `Uid` field! So this code doesn't actually work. From my system:

```
# cat /proc/1/fdinfo/99
pos:    0
flags:  02000002
mnt_id: 5
ino:    1178
Pid:    1177
NSpid:  1177
```

AdrianVovk (changes requested):
This won't actually work, see comments

I still think there's a way to rework things to clean it up: systemd uses common helpers to extract various fields from procfs files like fdinfo and status, but doesn't attempt to parse the fdinfo file like a status file. ~~On the other hand, it really might just not be worth the effort to rework this~~ (save for calling the ioctl, I think that's a worthwhile change). EDIT: I think I have a different approach to fixing this up that's much more doable, see conversation below

src/xdp-utils.c
AdrianVovk: Hmm, an approach to fixing this might be to just check if the caller is expecting the return value:

`if ((!have_pid && pid_out) || (!have_nspid && nspid_out) || (!have_uid && uid_out))`

Then maybe just rename the function to make clear that you're parsing a status file OR and fdinfo file (maybe something like `parse_procfs_status_fdinfo_common`?), and then actually this would probably be OK

swick:
I think I'll leave it here. We really should have a proper library or direct support in glib for the whole pidfd stuff.

PR #1737, opened 2025-06-10
Add `FileAccess` portal

JakobDev:
Currently, there are two ways for Flatpak apps to access files:

1. Use a `--filesystem` permission in the manifest. This gives the Flatpak constant access to a path, even if it's not always needed.  It also requires the developer to know the exact path when building the app.

 1. Use the `FileChooser` portal. This is great if you want to give the user the ability to select a file, but you have no control over which file the user selects.

This new portal adds a third option: Flatpaks can now request access to a specific path at runtime. This new portal allows sandboxed applications to request access to a specific path at runtime. The user receives a yes/no dialog. If they choose yes, the Flatpak gains access to the path through the document portal.

I have two use cases in mind for this new portal:

 1. Apps can use this portal if they need filesystem access only for a specific feature. For example, if an app provides a feature that requires it to see all installed apps on the system, it can now request access to /usr/share/applications only when needed.

2. The path is not known in advance. Take, for example, an app that works with Steam games. Steam allows storing games in different locations. You need to read the file `~/.steam/steam/config/libraryfolders.vdf` to get a list of all locations where Steam saves games. An app that wants to work with your Steam games could read this file (with a `--filesystem` permission for `~/.steam/steam`) and then request access to all locations.

TODO:
- Save the permission. I think I can use the permissions from the document portal.
- Write tests.
- Handle XDG directories (I'm not sure about that yet).

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -12,6 +12,7 @@ portal_sources = files(
   'org.freedesktop.portal.Documents.xml',
   'org.freedesktop.portal.DynamicLauncher.xml',
   'org.freedesktop.portal.Email.xml',
+  'org.freedesktop.portal.FileAccess.xml',
   'org.freedesktop.portal.FileChooser.xml',
   'org.freedesktop.portal.FileTransfer.xml',
   'org.freedesktop.portal.GameMode.xml',
--- a/data/org.freedesktop.portal.FileAccess.xml
+++ b/data/org.freedesktop.portal.FileAccess.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 20125 JakobDev
+
+ SPDX-License-Identifier: LGPL-2.1-or-later
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2.1 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.FileAccess:
+      @short_description: Portal requesting file access
+
+      This simple interface lets sandboxed applications
+      request access to a path.
+
+      This documentation describes version 1 of this interface.
+  -->
+  <interface name="org.freedesktop.portal.FileAccess">
+    <!--
+        RequestPathAccess:
+        @window: Identifier for the window
+        @path: The path the app wants to access
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
+
+        Gets information about the user.
+
+        Supported keys in the @options vardict include:
+
+        * ``handle_token`` (``s``)
+
+          A string that will be used as the last element of the @handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Request` documentation for
+          more information about the @handle.
+
+        * ``reason`` (``b``)
+
+          If only read only access is needed
+
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
+
+        * ``uri`` (``s``)
+
+          The uri to the path.
+    -->
+    <method name="RequestPathAccess">
+      <arg type="s" name="window" direction="in"/>
+      <arg type="s" name="path" direction="in"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="handle" direction="out"/>
+    </method>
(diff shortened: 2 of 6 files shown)


Mikenux:
Correction for case 2, the "another solution": The user is selecting the file to open, then the app can request the locations. This is the fileset request (from my neighboring files proposals). However, in the UI for this, I wanted to say something like "project" file to provide a hint to users about what this portal is about.

Mikenux:
> The Portal is pretty clear: It ask the user to access a given path. Why should this a problem to understand?

The problem is the path. With your proposal, you are giving apps the power to ask for anything. So what happens if the app asks for access to a path that the user doesn't know what it is or what feature it serves? What will the user do? If they choose "allow" because "the app asks for it", they are giving something without knowing what it is. If they deny because they don't understand, and it turns out to be useful for the app and the user, they will not be able to take advantage of the app feature (they can change the permission, but that would be like denying anything and then changing the permission once they know what the path is for).

> There is currently no portal for this, only a discussion. This portal would make the libraries portal mostly obsolete.

No. The Library Portal does not allow an app to request access to specific locations. It allows the user to select multiple locations representing a library. In my vision for this portal, it would allow an app to request access to a specific type of location (e.g., a music library). The user would have to select these as well. An exception is made for xdg user directories: the user can select a known directory directly in the interface (but the app does not request access to it).

Furthermore, it is not pleasant in terms of user experience to present the user one by one each location to allow or deny access to (although this gives the user the power to not give all locations; however, we may find it reasonable to request access to all locations on the basis that these locations come from a single file that the user opened).

> Why should the user be forces to select a file for which the app already knows the exact path.

Again, the understanding of the request. Also, the fact that users select the file proves (somewhat) that they know what they are doing.

In any case, be aware that there is no portal that is a solution for covering multiple different cases. Portals are designed based on a case or group of cases with similarities and for which we know exactly what "object" they want to access. This object is as precise as possible and is directly understood by the user.

Mikenux:
Globally:
- Launching apps / games: Portal to access launchers and launching them.
- Accessing libraries of others apps: If going with a library portal, maybe a request to access libraries given to a specific app might be added. Folders added via this portal could be automatically tagged in the permission store with their library type. However, this will only work with folders added this way. Also, that’s not possible if the library has been created by the app in its own space.

For game saves, only the file chooser can be used, the user selecting the relevant folder. That said, this isn’t easy if they are in the hidden folders (.var/...). Logically, unless an app has a good reason to do so, its libraries shouldn’t be in their space, but in user space.

JakobDev:
> So what happens if the app asks for access to a path that the user doesn't know what it is or what feature it serves?

If you don't know, why a permission is needed you can always reject.

> they will not be able to take advantage of the app feature

That's the same with every feature. 
If I want to record my screen with OBS and deny the request of OBS to access the Screen, I will not be able to take advantage of the screen recording feature.

> No. The Library Portal does not allow an app to request access to specific locations.

So it's a completely different portal.

> to avoid confusing yes/no allow/deny dialogs

There are multiple examples of such dialog in xdg-desktop-portal.

> not ask the user confusing questions

Why is requesting access to a path a confusing question?

> train application developers to create unlimited requests

What would be the alternative? Use the file chooser portal and tell the user tom select this specific path? Or tell the user to open add a static permission to the Flatpak?

> train users to always accept any request

That's always the problem with permissions. In the end, it's the choice of the user if they want to grant a permission or not.

> If the application needs direct access to /usr/share/applications, then it's just not sandboxed.

Why it's not sandboxed? The purpose of a sandbox is, that Apps just have the permissions they need to work and nothing else.
If a App needs to access `/usr/share/applications` then that's a correct permission.

> Might as well use a static permission?

As already written in the starting post, the goal here is to avoid static permissions, is the access is only needed for a specific feature.
A example for this is [jsSimpelAutostart](https://flathub.org/apps/page.codeberg.JakobDev.jdSimpleAutostart) which allows editing autostart entries.
It also have a feature to add programs from the menu to the autostart. So it only needs permission to access `/usr/share/applications` when the user wants to use this specific feature.

This new portal would allows to request the path only when it's needed which allows Apps to ship with less static permissions by default.

I try to give a better example than Steam, as it is not really understood:
There are Apps which need a list of all Installed Flatpsks (e.g [Flatseal](https://flathub.org/apps/com.github.tchx84.Flatseal) or [jdFlatpakSnapshot](https://flathub.org/apps/page.codeberg.JakobDev.jdFlatpakSnapshot)).
By default Flatpak has 2 installations: `/var/lib/flatpak`and `~/.local/share/flatpak`. So the get a list of all installed Flatpaks, Apps need access to those directories.
However, Users can add [custom installations](https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation).
Apps packaged as Flatpaks are not able to support custom installations, as there is no way to give a static permission, as the custom installation is not known in advance.

This new Portal would allow Apps to read `/etc/flatpak/installations.d` and ask for permission to access the installation.

> Portal to access launchers and launching them

This is also a Portal I want, but keep in mind that I just presented a example.

Mikenux:
I've already explained why and I'm tired of giving the same explanations over and over again.

That said, given the way GNOME and Red Hat are acting again, this is no longer my problem.

Mikenux:
I invite you to reread why we do not ask for paths and why general purpose portals are not desired.

Again, the last examples you gave can each be the object of one or more portals.

It is better to select a case, such as accessing installed apps, and specify the necessary options, such as launching apps, modifying permissions, modifying autostart. Note that these options could each be the object of a portal rather than being options within a portal. It is important to discuss what to do. Once things are defined, you can develop a portal.

PR #1738, opened 2025-06-10
notification-portal: Add missing cell broadcast severities

agx:
It is unclear what standard was initially applied here, https://github.com/flatpak/xdg-desktop-portal/pull/1304#discussion_r1527451517 talks about five categories but the portal currently only lists four and both CMAS/WEA and EU-Alert use at least six relevant categories that are handled on todays mobile devices:

1. Presidential wide alerts, corresponds to EU-ALERT level 1
2. Extreme threat, corresponds to EU-ALERT level 2
3. Severe threat, corresponds to EU-ALERT level 3
4. Public safety, corresponds to EU-ALERT level 4 (former EU-Info)
5. Child Abduction Emergency (Amber Alert), corresponds to EU-Amber
   Message
6. Test alerts

There are several kinds of test alerts and the different channels used to send cell broadcast messages also vary in the certainty that a user is affected by an event but I think we won't get closer than that for notification categories, and everything else should be left to a specific portal, see https://github.com/flatpak/xdg-desktop-portal/pull/1304#discussion_r1536646369

EDIT: note that on most handests the 1. category can't be disabled in the UI, maybe that's why it didn't get added initially?

Reference is 3GPP TS23.041.


swick:
@jsparber this one is for you

jsparber:
Adding the two missing categories seams reasonable. Also, good catch on the missing dash.

+1 from me.

swick:
Thanks everyone

PR #1740, opened 2025-06-18
Replace PipeWire and WirePlumber modules with in-tree modules

tytan652:
Implement of the first part of #1739 which consist  of replacing portal logic from PipeWire and WirePlumber in xdg-desktop-portal tree.

~~Sadly, this PR relies on WirePlumber 0.5 with its configuration fragment support which Ubuntu 24.04 does not have.~~

I plan to do some testing on real hardware with GNOME OS and sysext. DONE

### DIFF
--- a/desktop-portal/camera.c
+++ b/desktop-portal/camera.c
@@ -219,8 +219,8 @@ open_pipewire_camera_remote (const char *app_id,
   struct pw_properties *pipewire_properties;
 
   pipewire_properties =
-    pw_properties_new ("pipewire.access.portal.app_id", app_id,
-                       "pipewire.access.portal.media_roles", "Camera",
+    pw_properties_new ("pipewire.access.xdg-desktop-portal.app_id", app_id,
+                       "pipewire.access.xdg-desktop-portal.media_roles", "Camera",
                        NULL);
   remote = pipewire_remote_new_sync (pipewire_properties,
                                      NULL, NULL, NULL, NULL,
@@ -416,8 +416,8 @@ create_pipewire_remote (Camera *camera,
       return FALSE;
     }
 
-  pipewire_properties = pw_properties_new ("pipewire.access.portal.is_portal", "true",
-                                           "portal.monitor", "Camera",
+  pipewire_properties = pw_properties_new ("pipewire.access.xdg-desktop-portal.is_portal", "true",
+                                           "xdg-desktop-portal.monitor", "Camera",
                                            NULL);
   camera->pipewire_remote = pipewire_remote_new_sync (pipewire_properties,
                                                       global_added_cb,
--- a/meson.build
+++ b/meson.build
@@ -44,8 +44,18 @@ endif
 pipewire_dep_base_name = 'pipewire-0.3'
 pipewire_module_dir = libdir / pipewire_dep_base_name
 
+pipewire_conf_d_dir = get_option('pipewire-drop-in-conf-dir')
+if pipewire_conf_d_dir == ''
+  pipewire_conf_d_dir = prefix / datadir / 'pipewire' / 'pipewire.conf.d'
+endif
+
 wireplumber_dep_name = 'wireplumber-0.5'
 wireplumber_module_dir = libdir / wireplumber_dep_name
+
+wireplumber_conf_d_dir = get_option('wireplumber-conf-fragments-dir')
+if wireplumber_conf_d_dir == ''
+  wireplumber_conf_d_dir = prefix / datadir / 'wireplumber' / 'wireplumber.conf.d'
+endif
 
 dataroot_dir = get_option('datarootdir')
 if dataroot_dir == ''
@@ -61,7 +71,9 @@ summary({
 	'Flatpak interfaces dir': flatpak_intf_dir,
 	'systemd user unit dir': systemd_userunit_dir,
 	'PipeWire module dir': pipewire_module_dir,
+	'PipeWire drop-in configuration dir:': pipewire_conf_d_dir,
 	'WirePlumber module dir': wireplumber_module_dir,
+	'WirePlumber configuration fragments dir': wireplumber_conf_d_dir,
 	'Installed tests dir': installed_tests_dir,
 	},
 	section: 'Directories',
@@ -148,7 +160,7 @@ geoclue_dep = dependency(
   version: '>= 2.5.2',
   required: get_option('geoclue'),
 )
-pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 1.0.2')
+pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 1.3.81')
 libdbus_dep = dependency('dbus-1')
 wireplumber_dep = dependency(wireplumber_dep_name)
 libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
--- a/meson.build
+++ b/meson.build
@@ -41,6 +41,9 @@ if systemd_userunit_dir == ''
     systemd_userunit_dir = prefix / 'lib' / 'systemd' / 'user'
 endif
 
+pipewire_dep_base_name = 'pipewire-0.3'
+pipewire_module_dir = libdir / pipewire_dep_base_name
+
 dataroot_dir = get_option('datarootdir')
 if dataroot_dir == ''
     dataroot_dir = datadir
@@ -54,6 +57,7 @@ summary({
 	'DBus service dir': dbus_service_dir,
 	'Flatpak interfaces dir': flatpak_intf_dir,
 	'systemd user unit dir': systemd_userunit_dir,
+	'PipeWire module dir': pipewire_module_dir,
 	'Installed tests dir': installed_tests_dir,
 	},
 	section: 'Directories',
@@ -140,7 +144,8 @@ geoclue_dep = dependency(
   version: '>= 2.5.2',
   required: get_option('geoclue'),
 )
-pipewire_dep = dependency('libpipewire-0.3', version: '>= 0.2.90')
+pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 0.2.90')
+libdbus_dep = dependency('dbus-1')
 libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 gudev_dep = dependency('gudev-1.0', required: get_option('gudev'))
 umockdev_dep = dependency('umockdev-1.0', required: get_option('tests'))
@@ -237,6 +242,7 @@ subdir('data')
 subdir('shared')
 subdir('desktop-portal')
 subdir('document-portal')
+subdir('pipewire')
 subdir('po')
 subdir('doc')
 
--- a/meson.build
+++ b/meson.build
@@ -44,6 +44,9 @@ endif
 pipewire_dep_base_name = 'pipewire-0.3'
 pipewire_module_dir = libdir / pipewire_dep_base_name
 
+wireplumber_dep_name = 'wireplumber-0.5'
+wireplumber_module_dir = libdir / wireplumber_dep_name
+
 dataroot_dir = get_option('datarootdir')
 if dataroot_dir == ''
     dataroot_dir = datadir
@@ -58,6 +61,7 @@ summary({
 	'Flatpak interfaces dir': flatpak_intf_dir,
 	'systemd user unit dir': systemd_userunit_dir,
 	'PipeWire module dir': pipewire_module_dir,
+	'WirePlumber module dir': wireplumber_module_dir,
 	'Installed tests dir': installed_tests_dir,
 	},
 	section: 'Directories',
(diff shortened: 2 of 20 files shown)


meson.build
```diff
@@ -119,7 +139,13 @@ geoclue_dep = dependency(
   version: '>= 2.5.2',
   required: get_option('geoclue'),
 )
-pipewire_dep = dependency('libpipewire-0.3', version: '>= 0.2.90')
+pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 1.3.81', required: false)
+pw_portal_disableable = pipewire_dep.found()
+if not pipewire_dep.found()
+  pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 1.0.2')
+endif
```
tytan652: TODO: Use one `dependency` call and a version comparison.

.github/workflows/container.yml
```diff
@@ -1,5 +1,5 @@
 env:
-  IMAGE_TAG: 2025-07-21.3
+  IMAGE_TAG: 20251024-1
```
swick: can we please keep the tag readable?

pipewire/pw-module/module-portal.c
```diff
@@ -64,6 +56,12 @@
  *
  */
 
+#include <pipewire/context.h>
```
swick, line 19: Should be above the comment

pipewire/module-portal.c
```diff
@@ -16,53 +19,14 @@
 #include <spa/utils/string.h>
 #include <spa/utils/result.h>
 #include <spa/support/dbus.h>
-#include "pipewire/utils.h"
```
swick: shouldn't this get removed in the prev commit?

pipewire/module-portal.c
```diff
@@ -16,53 +19,14 @@
...
-
-/** \page page_module_portal Portal
```
swick: Any reason to delete the comment? A bit of context what the code here is for would be nice.

wireplumber/meson.build
```diff
@@ -0,0 +1,20 @@
+xdp_wireplumber_config_h = configuration_data()
```
swick: please prefix the commit messages with pipewire as well

meson.build
```diff
@@ -138,13 +142,16 @@ geoclue_dep = dependency(
...
 )
-pipewire_dep = dependency('libpipewire-0.3', version: '>= 0.2.90')
+pipewire_dep = dependency(f'lib@pipewire_dep_base_name@', version: '>= 0.2.90')
 libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 gudev_dep = dependency('gudev-1.0', required: get_option('gudev'))
 umockdev_dep = dependency('umockdev-1.0', required: get_option('tests'))
 libglnx_dep = dependency('libglnx')
 gvdb_dep = dependency('gvdb')
 
+pw_portal_disableable = pipewire_dep.version().version_compare('>= 1.3.81')
```
swick: Without more context the name here is really confusing

pipewire/wireplumber.conf
```diff
@@ -0,0 +1,36 @@
+wireplumber.profiles = {
```
swick, line 4: can we please put the wireplumber stuff into pipewire/ as well?

desktop-portal/camera.c
```diff
@@ -235,8 +235,8 @@ open_pipewire_camera_remote (const char *app_id,
   struct pw_properties *pipewire_properties;
 
   pipewire_properties =
-    pw_properties_new ("pipewire.access.portal.app_id", app_id,
-                       "pipewire.access.portal.media_roles", "Camera",
+    pw_properties_new ("pipewire.access.xdg-desktop-portal.app_id", app_id,
```
swick, line 222: I'm a bit confused. If we only want to support the build-in modules, then this change here is fine, but then some of the meson code can be simplified (i.e. we just require a specific version).

pipewire/wireplumber.conf
tytan652: No. I prefer to keep both module codebase separated in their own folder.
Or do you mean as a subfolder of pipewire which not possible due to the pipewire module being skipped if PipeWire is too old to allow disabling the original module.

wireplumber/meson.build
tytan652: Why ? It's not dependent on the pipewire side.

pipewire/module-portal.c
tytan652: Technically it's PipeWire documentation about the module.

I'll replace it with a comment and try to add architecture documentation about the module.

meson.build
tytan652: Added a comment, is this enough ?

pipewire/wireplumber.conf
swick: > I prefer to keep both module codebase separated in their own folder.

why?

wireplumber/meson.build
swick: Well, at a appropriate prefix.

pipewire/wireplumber.conf
tytan652: Is it insane to keep two different binary source code into two specific folder ?
Because for me the reverse, putting two different binary source code into the same folder (no subdirectories and/or shared code) does not look good.

-----

xdp PipeWire module/folder is entirely skipped if the PipeWire version is not recent enough, unless there is consensus to bump the minimum to 1.3.81 rather than WirePlumber's minimum.

wireplumber/meson.build
tytan652: I may have the wrong commit in my mind (the one adding the wireplumber module), you meant the last one where the prefix can be `pipewire,wireplumber:`.

PS: GitHub's per commit review is a disaster…

desktop-portal/camera.c
tytan652: The PipeWire module only deals with XDP PID, only the WirePlumber one deal with those properties.

tytan652:
I will resume the PipeWire version requirement here:
- 1.0.2 or later is WirePlumber 0.5 requirement
- 1.3.81 is the version that allows to disable the built-in portal module with a drop-in config file, before you had to modify the main config file.

PS: If bumping to 1.3.81 is fine and there is consensus, I will bump. I just tried to not be too aggressive with bumping requirements.

meson.build
swick: Let's just require that version. I'll merge the PR after the next stable release.

pipewire/pw-module/module-portal.c
```diff
@@ -1,4 +1,7 @@
-/* PipeWire */
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+/* SPDX-FileCopyrightText: Copyright © 2025 tytan652 */
```
swick: Let's please not add unnecessary copyright lines. Git keeps track of who wrote what.

wireplumber/meson.build
swick: the wireplumber one is missing a prefix as well

pipewire/wireplumber.conf
swick: it's very much related code and I don't see why we should create toplevels for every small thing. It makes sense to not put it in src because it's about completely different artifacts, but logically it makes no sense to separate the two plugins.

.github/workflows/Containerfile
```diff
@@ -72,12 +72,24 @@ RUN pip install --user --break-system-packages pre-commit
 # Install qdbusxml2cpp
 RUN apt install -y --no-install-recommends qt6-base-dev-tools
 
-# Install WirePlumber 0.5.8 (same as 25.04) since 24.04 only has 0.4
-RUN git clone --depth 1 --branch 0.5.8 \
+# Install PipeWire 1.4 since 24.04 has 1.0.5
+# NOTE: not 1.6 to ensure to keep API compat with 1.x
```
swick: Why does it matter if we install 1.4 and not 1.6? What kind of API compatibility are we talking about here?

tytan652: - I don't know, their changelog was specifying that 1.4 kept compat with 1.x while 1.6 only with 1.6.x.
- I didn't wanted to bump too much and prevent introducing access to 1.6 API while we only require at least 1.3.81.

swick: If the code introduced here does in fact only support 1.2 to 1.4, then meson should ensure that. In my opinion we should try to always support the latest version, so I would prefer the code to be adjusted to work on 1.6 over restricting it to 1.4.

tytan652: Dropped most of container changes for #1988

tytan652:
Drafted since I dropped most of container changes for #1988

swick:
Needs a rebase (sorry about that)

pipewire/pw-module/module-portal.c
```diff
@@ -1,7 +1,5 @@
-/* PipeWire */
```
swick: Why is this commit changing the license?

pipewire/wp-plugin/meson.build
```diff
@@ -20,4 +20,4 @@ xdp_wireplumber_module = shared_library(
   dependencies: [wireplumber_dep, pipewire_dep],
   install: true,
   install_dir: wireplumber_module_dir,
```
swick, line 22: Whitespace change only

pipewire/pw-module/module-portal.c
```diff
@@ -171,13 +173,15 @@ static void on_portal_pid_received(DBusPendingCall *pending,
 		return;
 	}
 
-	spa_auto(DBusError) error = DBUS_ERROR_INIT;
+	dbus_error_init(&error);
```
swick: Probably should use g_autoptr/G_DEFINE_AUTOPTR_CLEANUP_FUNC

pipewire/pw-module/module-portal.c
```diff
@@ -130,7 +125,11 @@ static void module_destroy(void *data)
 	spa_hook_remove(&impl->context_listener);
 	spa_hook_remove(&impl->module_listener);
 
-	cancel_and_unref(&impl->portal_pid_pending);
```
swick: Seems cleaner to just implement the function.

pipewire/wp-plugin/meson.build
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
+
+xdp_wireplumber_module_sources = [
+  'xdp-wp-permission-manager.c',
+  'xdp-wp-plugin.c',
+]
+
+xdp_wireplumber_module_sources += gnome.gdbus_codegen(
```
swick, line 9: We already have `impl_built_sources`.

tytan652: Yes but now with libdex, for WirePlumber sticking with the GLib async is preferred.

pipewire/pw-module/module-portal.c
tytan652: No GLib in the PipeWire module, only PipeWire, SPA APIs and libdbus.

pipewire/module-portal.c
tytan652: Created doc/pipewire.rst

pipewire/pw-module/module-portal.c
tytan652: Continuation: https://github.com/flatpak/xdg-desktop-portal/pull/1740#discussion_r3379781189

pipewire/pw-module/module-portal.c
tytan652: The main purpose is to ensure that future contribution on that file are under the same license as xdp.

pipewire/pw-module/module-portal.c
tytan652: I wanted to avoid copying a private header (which has the SPA+libdbus convenience function) over and PipeWire/SPA only has support for libdbus.

pipewire/wp-plugin/meson.build
tytan652: I fail to see the issue, sorry.

swick: The files `pipewire/pw-module/meson.build` and `pipewire/wp-plugin/meson.build` have changes which are entirely irrelevant to the rest of the commit, and the changes are whitespace changes only.

pipewire/pw-module/module-portal.c
swick: You can't just remove the license. You can add another license if they are compatible, and you can add a note saying that new contributions to xdp in this file will be under the new license.

pipewire/pw-module/module-portal.c
swick: Sure, but you can just implement the function in this file, and the same is true for some of the others.

tytan652: Copied them.

pipewire/wp-plugin/meson.build
tytan652: Found them.

pipewire/pw-module/module-portal.c
tytan652: ~~WIP (the push is for rebuilding the container image with the new date)~~

tytan652: Tried something, feel free to suggest something more fitting.

swick:
Thanks a lot! GNOME OS installs from git nowadays so we should notice if things go really wrong.

PR #1745, opened 2025-06-28
Allow opening files in `/tmp`

### LINKED ISSUE #1744: Cannot open files in Flatpak sandbox's private /tmp
### Operating System

Arch Linux

1.20

_No response_

### Desktop Environment

Plasma

_No response_

### Expected Behavior

Given:
```
flatpak run --command=bash org.freedesktop.Sdk//24.08 
$ echo "banana" > /tmp/myfile.text
$ xdg-open /tmp/myfile.text 
```
(note here I tried to make a simplest common example, this comes from the debugging from another tool. Also same behaviour happens when using QDesktopServices::openUrl, which has a similar functionality to xdg-open)

I'd expect the file to get open, much like if I do the `xdg-open` on something in `/usr` or `/app` or a host path.

The file is indeed available on the host:
```
$ ls /run/user/1000/.flatpak/org.freedesktop.Sdk/tmp/
myfile.text
```

### Current Behavior

The portal silently rejects the process, only leaving this debug line.
```
(/usr/lib/xdg-desktop-portal:156325): xdg-desktop-portal-DEBUG: 03:45:39.777: Rejecting open request: stat /tmp/myfile.text: No such file or directory
```

### Steps to Reproduce

As explained in the expected behaviour

_No response_

JakobDev:
This PR allows opening file sin `/tmp`.

To access the `/tmp` of a Flatpak, the directory `/proc/<pid>/root` is used. This directory gives access to the filesystem of each process.

While Flatpak provide `XDG_RUNTIME_DIR//.flatpak/{ID}/tmp` to access the temp files of a Flatpak, this don't work for a subsandbox. For those who don't know: Flatpak provides a way to create a sanbox inside the sandbox using `flatpak-spawn --sandboxed`. The `/tmp` of those subsandbox can  (as far as I know) only be accessed using the `/proc/<pid>/root` filesystem.

Fixes #1744

[edited for formatting —smcv]

### DIFF
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -933,13 +933,9 @@ xdp_app_info_flatpak_new (int      pid,
                                            NULL, NULL);
 
   if (filesystems)
-    {
-      has_tmp_access = g_strv_contains ((const char * const *)filesystems, "/tmp");
-    }
+    has_tmp_access = g_strv_contains ((const char * const *)filesystems, "/tmp");
   else
-    {
-      has_tmp_access = FALSE;
-    }
+    has_tmp_access = FALSE;
 
   /* flatpak has a xdg-dbus-proxy running which means we can't get the pidfd
    * of the connected process but we can get the pidfd of the bwrap instance


src/xdp-app-info-flatpak.c
```diff
@@ -239,6 +244,11 @@ xdp_app_info_flatpak_remap_path (XdpAppInfo *app_info,
     return g_build_filename (g_get_home_dir (), ".var", "app",
                              app_id, "data",
                              path + strlen ("/var/data/"), NULL);
+  else if (g_strcmp0 (path, "/tmp") == 0 && !app_info_flatpak->host_tmp_access)
+    return g_build_filename (app_info_flatpak->root_path, "tmp", NULL);
+  else if (g_str_has_prefix (path, "/tmp") && !app_info_flatpak->host_tmp_access)
```
swick: This really needs to be `"/tmp/"`, otherwise this would work for `"/tmpx"` etc.

src/xdp-app-info-flatpak.c
```diff
@@ -820,6 +848,10 @@ xdp_app_info_flatpak_new (int      pid,
   if (has_network)
     flags |= XDP_APP_INFO_FLAG_HAS_NETWORK;
 
+  flatpak_instance = flatpak_instance_from_metadata (metadata, error);
```
swick, line 955: FlatpakInstance does a lot of what XdpAppInfoFlatpak does as well. Might make sense to replace the logic in here with calls to FlatpakInstance, and move the rest of the stuff in here to FlatpakInstance.

src/xdp-app-info-flatpak.c
```diff
@@ -805,6 +819,20 @@ xdp_app_info_flatpak_new (int      pid,
...
+                                           NULL, NULL);
+
+  if (filesystems)
+    {
+      has_tmp_access = g_strv_contains ((const char * const *)filesystems, "/tmp");
+    }
+  else
+    {
+      has_tmp_access = FALSE;
+    }
```
swick:
```suggestion
  if (filesystems)
    has_tmp_access = g_strv_contains ((const char * const *)filesystems, "/tmp");
  else
    has_tmp_access = FALSE;
```

src/flatpak-instance.c
```diff
@@ -503,6 +517,30 @@ flatpak_instance_get_all (void)
...
 
+/**
+ * flatpak_instance_from_metadata:
+ *
+ * Gets FlatpakInstance objects from teh flatpak info metadata
+ *
+ * Returns: a #FlatpakInstance object
+ */
+FlatpakInstance *
+flatpak_instance_from_metadata (GKeyFile *metadata,
```
swick, line 528: Should really split the flatpak-instance changes into its own commit.

src/xdp-app-info-flatpak.c
smcv: To avoid falling into this trap every time, perhaps copy `flatpak_has_path_prefix()` from Flatpak's `common/flatpak-utils.c`, or the derived function `_srt_get_path_after()` from https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/HEAD/steam-runtime-tools/utils.c, and use that?

JakobDev:
I tested this a bit more, and it only works for host applications, as Flatpak applications cannot access `/proc/<pid>/root` for some reason. At least we have resolved 50% of the problem now.

> FYI, this can likely be applied to the snap case by mapping it to `/tmp/snap-private-tmp/snap.${SNAP_NAME}`

I don't have any experience with Snap, so I am keeping this PR focused on Flatpak. Please feel free to submit another PR with Snap support once this one is merged.

src/xdp-app-info-flatpak.c
JakobDev: I would like to keep this PR focused on the bug fix rather than a major rewrite.

Issue #1747, opened 2025-07-04
xdp_request_export() crashes when handle token contains "-"


taoky:
### Operating System

Arch Linux

### XDG Desktop Portal version

1.20

### Desktop Environment

GNOME

### Expected Behavior

xdg-desktop-portal reports error but does not crash (as portal would be accessed by sandboxed apps).

### Current Behavior

xdg-desktop-portal crashes.

### Steps to Reproduce

1. Use `org.freedesktop.portal.Inhibit.Inhibit` method with `handle_token: keepawake-rs-xxxxxx` (I locally reproduced this with https://github.com/taoky/keepawake-rs/commit/5dd4d028094e7db2d7ccfcd16a321665a25323e2#diff-fbee4e1f9fe18e74d58032b151a840829e2dc04e205286b61355f67acf34843aR117-R119).
2. `journalctl --user -u xdg-desktop-portal` reports:
    ```
    g_dbus_connection_signal_subscribe: assertion 'object_path == NULL || g_variant_is_object_path (object_path)' failed
    g_dbus_interface_skeleton_export: assertion 'g_variant_is_object_path (object_path)' failed
    Process 661203 (xdg-desktop-por) of user 1000 dumped core.
    ```
3. xdg-desktop-portal crashes and is restarted by user systemd

### Anything else we should know?

gdb:

```
> bt
#0  0x000055e63b0cb19e in xdp_request_export (request=0x7f95e404ca70, connection=<optimized out>) at ../xdg-desktop-portal/src/xdp-request.c:275
#1  0x000055e63b0a6433 in handle_inhibit
    (object=0x55e63d0ad390, invocation=0x7f95ec0156e0, arg_window=<optimized out>, arg_flags=<optimized out>, arg_options=<optimized out>)
    at ../xdg-desktop-portal/src/inhibit.c:240
#2  0x000055e63b05f9e1 in _g_dbus_codegen_marshal_BOOLEAN__OBJECT_STRING_UINT_VARIANT
    (invocation_hint=<optimized out>, closure=0x55e63d0b0ce0, return_value=0x7f95f9b4b990, n_param_values=<optimized out>, param_values=0x7f95e4052270, marshal_data=<optimized out>) at src/xdp-dbus.c:1537
#7  Python Exception <class 'gdb.error'>: There is no member named v_pointer.
#8  0x000055e63b065e3f in _xdp_dbus_inhibit_skeleton_handle_method_call
[...]

PR #1748, opened 2025-07-04
utils: Add xdp_is_valid_token function

taoky:
The token check function is moved from xdp-session to xdp-utils, as xdp-request also needs checking the token validity to avoid crash.

Closes: #1747


src/xdg-desktop-portal.c
```diff
@@ -152,7 +152,17 @@ authorize_callback (GDBusInterfaceSkeleton *interface,
     }
 
   if (method_needs_request (invocation))
-    xdp_request_init_invocation (invocation, app_info);
+    {
+      if (!xdp_request_validate_invocation(invocation))
```
mcatanzaro:
```suggestion
      if (!xdp_request_validate_invocation (invocation))
```

src/xdp-utils.c
```diff
@@ -343,6 +343,23 @@ xdp_is_valid_app_id (const char *string)
...
 
+gboolean
+xdp_is_valid_token (const char *string)
+{
+  g_autofree char *path = NULL;
+  int i;
+
+  for (i = 0; string[i]; i++)
+    {
+      if (G_UNLIKELY( !is_valid_name_character(string[i], FALSE) ))
```
mcatanzaro:
```suggestion
      if (G_UNLIKELY (!is_valid_name_character (string[i], FALSE)))
```

mcatanzaro (commented):
Not a full review: just leaving a couple style fixes...

src/xdg-desktop-portal.c
```diff
@@ -152,7 +152,17 @@ authorize_callback (GDBusInterfaceSkeleton *interface,
...
+    {
+      if (!xdp_request_validate_invocation (invocation))
+        {
+          g_dbus_method_invocation_return_error (invocation,
+                                                 XDG_DESKTOP_PORTAL_ERROR,
+                                                 XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                                 "Invalid request");
+          return FALSE;
+        }
+      xdp_request_init_invocation (invocation, app_info);
```
swick: I think it would be nicer if this was an implementation detail of `xdp_request_init_invocation`. So `xdp_request_init_invocation` would call `xdp_is_valid_token` on `token` and return `FALSE` with an error set.

taoky: OK, I have moved the check inside `xdp_request_init_invocation`.

swick:
Thanks!

PR #1751, opened 2025-07-09
app-info: Check if the root dir is an inaccessible link

### LINKED ISSUE #1750: Cannot capture screens with non-flatpak versions of Gimp and Firefox
### Operating System

Fedora 42

1.20

1.20.3

### Desktop Environment

wlroots

river

### Expected Behavior

Allow screen captures from non-flatpak apps if a wlroots-based compositor is used and if LSM security modules are in use. `open_flatpak_info` should not stop the search for non-flatpak apps if access to an app's `/proc/<app-pid>/root` directory is restricted by the software or a security module. The function ought check whether the symbolic link to an app's root directory exists but is inaccessible.

### Current Behavior

I cannot capture the screen with non-flatpak versions of Gimp and Firefox on Fedora 42 when I use a wlroots-based compositor. Both Firefox and Gimp start the child process to capture the screen with a `/proc/<app-pid>/root` directory that exists but is inaccessible to xdg-desktop-portal. I use  the Linux security modules SELinux, Lockdown and Yama, which could play a role.

`open_flatpak_info` in `xdp-app-info-flatpak.c` calls `openat`, which returns `EACCES` if the root directory is inaccessible. There's already a test for a fuse rootfs, but the function should also check whether `/proc/<app-pid>/root` is an existing symbolic link. Currently, the function assumes that the software is a flatpak app that has died and sets the error to G_IO_ERROR_FAILED instead of XDP_APP_INFO_ERROR_WRONG_APP_KIND. This causes xdg-desktop-portal to stop the search for non-flatpak apps.

### Steps to Reproduce

1. Install Fedora 42, a wlroots-based compositor such as River and non-flatpak versions of Firefox and Gimp.
2. Enable the Linux security modules SELinux, Lockdown and Yama. I use version 6.12 of the Linux kernel.
3. Run the non-flatpak versions of Firefox and Gimp.
4. In Firefox, go to https://mozilla.github.io/webrtc-landing/gum_test.html and try to capture a screen.
5. In Gimp, select File > Create > Screenshot.
6. Get the error message "Unable to open `/proc/<app-pid>/root`".

According to the manual page proc_pid_root(5), permission to dereference or read `/proc/<app-pid>/root` is controlled with ptrace.

I haven't checked whether access to `/proc/<app-pid>/root` is generally unavailable to child processes of Firefox and Gimp on Fedora 42 or whether access is restricted by one of the Linux security modules. The restricted access to the root directory might be some "sandbox" feature in Firefox, Gimp or recent Linux kernel versions or a feature that is enabled by security modules such as Yama. xdg-desktop-portal should work in any case.

voegelas:
Currently, xdg-desktop-portal stops the search for non-flatpak apps if access to `/proc/<app-pid/root>` is restricted. If the root directory is inaccessible, the function `open_flatpak_info` simply assumes that the app is a flatpak app that has died.  There's already an exception for root on a fuse filesystem. This pull request adds a test that checks whether the root directory exists and is a symbolic link.  If this is the case, the error is set to XDP_APP_INFO_ERROR_WRONG_APP_KIND instead of G_IO_ERROR_FAILED. Fixes #1750.

### DIFF
--- a/src/xdp-app-info-flatpak.c
+++ b/src/xdp-app-info-flatpak.c
@@ -654,6 +654,7 @@ open_flatpak_info (int      pid,
       if (errno == EACCES)
         {
           struct statfs buf;
+          struct stat stat_buf;
 
           /* Access to the root dir isn't allowed. This can happen if the root is on a fuse
            * filesystem, such as in a toolbox container. We will never have a fuse rootfs
@@ -665,6 +666,18 @@ open_flatpak_info (int      pid,
             {
               g_set_error (error, XDP_APP_INFO_ERROR, XDP_APP_INFO_ERROR_WRONG_APP_KIND,
                            "Not a flatpak (fuse rootfs)");
+              return -1;
+            }
+
+          /* Permission to dereference or read the symbolic link to the root dir may not
+           * have been granted. See proc_pid_root(5) for more information. This can happen
+           * if non-flatpak versions of Gimp and Firefox start a child process to capture
+           * a screen.
+           */
+          if (lstat (root_path, &stat_buf) == 0 && S_ISLNK (stat_buf.st_mode))
+            {
+              g_set_error (error, XDP_APP_INFO_ERROR, XDP_APP_INFO_ERROR_WRONG_APP_KIND,
+                           "Not a flatpak (restricted link)");
               return -1;
             }
         }


swick:
I'm also not comfortable with this change.

Issue #1648, opened 2025-02-21
xdg-permission-store: Crash on add_permissions


A6GibKm:
### Operating System

Debian 12

### XDG Desktop Portal version

1.16

### Desktop Environment

GNOME

### Expected Behavior

The daemon does not crash.

### Current Behavior

The daemon can crash.

### Steps to Reproduce

Unknown. Happened while trying to save a file with thunderbird (interestingly the crash says firefox).

### Anything else we should know?

Backtrace bellow.
```
[Current thread is 1 (Thread 0x7b66bc3f6d40 (LWP 5629))]
(gdb) bt
#0  __strcmp_evex () at ../sysdeps/x86_64/multiarch/strcmp-evex.S:295
#1  0x0000601d7a52f6ed in add_permissions (app_permissions=app_permissions@entry=0x601d922fb380, permissions=0x601d9230cc40) at ../document-portal/permission-db.c:1158
#2  0x0000601d7a5316f6 in permission_db_entry_set_app_permissions (entry=entry@entry=0x601d922fd040, app=app@entry=0x601d923087e0 "org.mozilla.firefox", permissions=permissions@entry=0x601d92308460) at ../document-portal/permission-db.c:1254
#3  0x0000601d7a52a256 in handle_set_permission (object=0x601d922894b0, invocation=0x7b66a8009cf0, table_name=0x601d9230c890 "notifications", create=1, id=0x601d9230bba0 "notification", app=0x601d923087e0 "org.mozilla.firefox", permissions=0x601d92308460)
    at ../document-portal/xdg-permission-store.c:452
#4  0x00007b66bc6e5f7a in ffi_call_unix64 () at ../src/x86/unix64.S:104
#5  0x00007b66bc6e540e in ffi_call_int (cif=cif@entry=0x7ffc628a4f50, fn=fn@entry=0x601d7a52a200 <handle_set_permission>, rvalue=<optimized out>, rvalue@entry=0x7ffc628a4ed0, avalue=<optimized out>, closure=closure@entry=0x0) at ../src/x86/ffi64.c:673
#6  0x00007b66bc6e5b0d in ffi_call (cif=cif@entry=0x7ffc628a4f50, fn=fn@entry=0x601d7a52a200 <handle_set_permission>, rvalue=rvalue@entry=0x7ffc628a4ed0, avalue=avalue@entry=0x7ffc628a4e30) at ../src/x86/ffi64.c:710
[...]

PR #1761, opened 2025-08-05
permission-db: Prevent crash on NULL child_app_id

A6GibKm:
This prevents the crash described at
https://github.com/flatpak/xdg-desktop-portal/issues/1648.


swick:
I'm closing this PR. This doesn't fix the root issue and there don't seem to be others affected by the same issue.

PR #1762, opened 2025-08-07
camera: Revert using an empty app id for all host apps

swick:
This effectively reverts 6cd99b04 ("Camera: use an empty app id for host apps").

We now have the org.freedesktop.host.portal.Registry portal which allows apps to register their app id instead of relying on the currently somewhat broken auto-detection. The commit this is reverting was introduced to deal with the broken auto-detection in firefox. Firefox now uses the Registry in ESR 140, so it should be safe to revert the behavior here.

### DIFF
--- a/src/camera.c
+++ b/src/camera.c
@@ -186,16 +186,6 @@ handle_access_camera_in_thread_func (GTask *task,
     }
 }
 
-static const char *
-app_id_from_app_info (XdpAppInfo *app_info)
-{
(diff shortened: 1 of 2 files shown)


grulja:
Looks good to me, but I'm not probably the one who should ultimately decided this can go in. I verified this with Firefox and it makes now correct requests for the camera access and says it's "Firefox" who wants to access the camera.

PR #1764, opened 2025-08-08
xdp-app-info-snap: Use the desktop ID as app ID and snap ID for permissions

### LINKED ISSUE #769: parse_app_info_from_snap() sets the app id to "snap.$AppName", but gnome-shell's access dialog compares it to $DesktopFile

3v1n0:
From the main commit:

When using snap applications the permissions are saved using a common ID
that is not matching the desktop ID, since they are per-snap, not per
application.

While that is fine,  we were using the very same ID also as the
app ID, which is instead expected to be matching the desktop-ID (without
extension).

To prevent this to happen, use the desktop ID as the app ID and the snap
ID for permission purposes only.

As per this, we can now use the snap desktop ID as the handle that we
expose everywhere, since that's what all the components (shells, and
portal implementations) are expecting to receive.

This fixes an issue existing for long time, that was causing some
portals not to work at all (like the access one) or to misbehave (like
the notification one), while preserving the permission behavior that we
used to have.

Closes: #769

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -1367,7 +1367,7 @@ portal_list (GDBusMethodInvocation *invocation,
              GVariant *parameters,
              XdpAppInfo *app_info)
 {
-  const char *app_id = xdp_app_info_get_id (app_info);
+  const char *app_id;
   g_auto(GStrv) ids = NULL;
   GVariantBuilder builder;
   int i;
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -106,16 +106,16 @@ persist_entry (PermissionDbEntry *entry)
 static void
 do_set_permissions (PermissionDbEntry    *entry,
                     const char        *doc_id,
-                    const char        *app_id,
+                    const char        *permissions_id,
                     DocumentPermissionFlags perms)
 {
   g_autofree const char **perms_s = xdg_unparse_permissions (perms);
 
   g_autoptr(PermissionDbEntry) new_entry = NULL;
 
-  g_debug ("set_permissions %s %s %x", doc_id, app_id, perms);
+  g_debug ("set_permissions %s %s %x", doc_id, permissions_id, perms);
 
-  new_entry = permission_db_entry_set_app_permissions (entry, app_id, perms_s);
+  new_entry = permission_db_entry_set_app_permissions (entry, permissions_id, perms_s);
   permission_db_set_entry (db, doc_id, new_entry);
 
   if (persist_entry (new_entry))
@@ -124,7 +124,7 @@ do_set_permissions (PermissionDbEntry    *entry,
                                                 TABLE_NAME,
                                                 FALSE,
                                                 doc_id,
-                                                app_id,
+                                                permissions_id,
                                                 perms_s,
                                                 NULL,
                                                 NULL, NULL);
@@ -137,6 +137,7 @@ portal_grant_permissions (GDBusMethodInvocation *invocation,
                           XdpAppInfo            *app_info)
 {
   const char *target_app_id;
+  const char *target_permissions_id;
   const char *id;
   g_autofree const char **permissions = NULL;
   DocumentPermissionFlags perms;
@@ -158,11 +159,12 @@ portal_grant_permissions (GDBusMethodInvocation *invocation,
         return;
       }
 
-    if (!xdp_is_valid_app_id (target_app_id))
+    target_permissions_id = target_app_id;
+    if (!xdp_is_valid_permissions_id (target_permissions_id))
       {
         g_dbus_method_invocation_return_error (invocation,
                                                XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                                               "'%s' is not a valid app name", target_app_id);
+                                               "'%s' is not a valid app permissions ID", target_permissions_id);
         return;
       }
 
@@ -183,12 +185,12 @@ portal_grant_permissions (GDBusMethodInvocation *invocation,
         return;
       }
 
-    do_set_permissions (entry, id, target_app_id,
-                        perms | document_entry_get_permissions_by_app_id (entry, target_app_id));
+    do_set_permissions (entry, id, target_permissions_id,
+                        perms | document_entry_get_permissions_by_app_permissions_id (entry, target_permissions_id));
   }
 
   /* Invalidate with lock dropped to avoid deadlock */
-  xdp_fuse_invalidate_doc_app (id, target_app_id);
+  xdp_fuse_invalidate_doc_app (id, target_permissions_id);
 
   g_dbus_method_invocation_return_value (invocation, g_variant_new ("()"));
 }
@@ -198,8 +200,9 @@ portal_revoke_permissions (GDBusMethodInvocation *invocation,
                            GVariant              *parameters,
                            XdpAppInfo            *app_info)
 {
-  const char *app_id = xdp_app_info_get_id (app_info);
+  const char *permissions_id = xdp_app_info_get_permissions_id (app_info);
   const char *target_app_id;
+  const char *target_permissions_id;
   const char *id;
   g_autofree const char **permissions = NULL;
   GError *error = NULL;
@@ -221,11 +224,12 @@ portal_revoke_permissions (GDBusMethodInvocation *invocation,
         return;
       }
 
-    if (!xdp_is_valid_app_id (target_app_id))
+    target_permissions_id = target_app_id;
+    if (!xdp_is_valid_permissions_id (target_permissions_id))
       {
         g_dbus_method_invocation_return_error (invocation,
                                                XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                                               "'%s' is not a valid app name", target_app_id);
+                                               "'%s' is not a valid app permissions ID", target_permissions_id);
         return;
       }
 
@@ -239,7 +243,7 @@ portal_revoke_permissions (GDBusMethodInvocation *invocation,
     /* Must have grant-permissions, or be itself */
     if (!document_entry_has_permissions (entry, app_info,
                                     DOCUMENT_PERMISSION_FLAGS_GRANT_PERMISSIONS) ||
-        strcmp (app_id, target_app_id) == 0)
+        strcmp (permissions_id, target_permissions_id) == 0)
       {
         g_dbus_method_invocation_return_error (invocation,
                                                XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
@@ -247,12 +251,12 @@ portal_revoke_permissions (GDBusMethodInvocation *invocation,
         return;
       }
 
-    do_set_permissions (entry, id, target_app_id,
-                        ~perms & document_entry_get_permissions_by_app_id (entry, target_app_id));
+    do_set_permissions (entry, id, target_permissions_id,
+                        ~perms & document_entry_get_permissions_by_app_permissions_id (entry, target_permissions_id));
   }
 
   /* Invalidate with lock dropped to avoid deadlock */
-  xdp_fuse_invalidate_doc_app (id, target_app_id);
+  xdp_fuse_invalidate_doc_app (id, target_permissions_id);
 
   g_dbus_method_invocation_return_value (invocation, g_variant_new ("()"));
 }
@@ -575,7 +579,7 @@ metadata_check_file_access (const char *keyfile_path,
  * to (e.g. when the app has a more strict access but the file is
  * still accessible) */
 static gboolean
-app_has_file_access_fallback (const char *target_app_id,
+app_has_file_access_fallback (const char *target_permissions_id,
                               DocumentPermissionFlags target_perms,
                               const char *path)
 {
@@ -596,10 +600,10 @@ app_has_file_access_fallback (const char *target_app_id,
   if (g_str_has_prefix (path, "/usr") || g_str_has_prefix (path, "/app") || g_str_has_prefix (path, "/tmp"))
     return FALSE;
 
-  user_metadata = g_build_filename (user_installation, "app", target_app_id, "current/active/metadata", NULL);
-  system_metadata = g_build_filename (system_installation, "app", target_app_id, "current/active/metadata", NULL);
-  user_override = g_build_filename (user_installation, "overrides", target_app_id, NULL);
-  system_override = g_build_filename (system_installation, "overrides", target_app_id, NULL);
+  user_metadata = g_build_filename (user_installation, "app", target_permissions_id, "current/active/metadata", NULL);
+  system_metadata = g_build_filename (system_installation, "app", target_permissions_id, "current/active/metadata", NULL);
+  user_override = g_build_filename (user_installation, "overrides", target_permissions_id, NULL);
+  system_override = g_build_filename (system_installation, "overrides", target_permissions_id, NULL);
   user_global_override = g_build_filename (user_installation, "overrides", "global", NULL);
   system_global_override = g_build_filename (system_installation, "overrides", "global", NULL);
 
@@ -630,7 +634,7 @@ app_has_file_access_fallback (const char *target_app_id,
 
 
 static gboolean
-app_has_file_access (const char *target_app_id,
+app_has_file_access (const char *target_permissions_id,
                      DocumentPermissionFlags target_perms,
                      const char *path)
 {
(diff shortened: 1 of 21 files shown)


src/xdp-app-info.h
```diff
@@ -47,6 +47,8 @@ gboolean xdp_app_info_is_host (XdpAppInfo *app_info);
 
 const char * xdp_app_info_get_id (XdpAppInfo *app_info);
 
+const char * xdp_app_info_get_permissions_id (XdpAppInfo *app_info);
```
mcatanzaro, line 60: Oh man... I wonder how many "oops we used the wrong ID" CVEs this will eventually result in. :(

Maybe we could rename `xdp_app_info_get_id()` as well to discourage misuse (even though that will bloat this pull request)? Otherwise, it's hard to know which one you really want.

3v1n0: Well, reason why of commit e225c34f6a851be9be99409c03fd1c0b3fef7934 and thus the usage of the API is very limited to that scope or to debug places, but I also was unsure how to name it...

Or we could make the API even more "private", and only include it inside the permission code.

3v1n0:
> I think that would be a good idea - perhaps `xdp_app_info_get_permissions_id` or even `xdp_app_info_get_trust_domain` for the trust domain (snap app ID or Flatpak app ID), and `xdp_app_info_get_desktop_id` for the one that is specifically the .desktop file? That way, anyone calling one of these functions is forced to acknowledge that there is a question about which one they want, and answer it by choosing the appropriate one.

Indeed that is a good proposal, @GeorgesStavracas (or @swick) maybe you can comment here and I could move the refactoring commits to another PR where we just move things around, and then here we only handle the specific snap cases

smcv:
> especially if one of them is "the app" and the other is a sandboxable thumbnailer or something

For comparison, in Flatpak-world, this would be implemented by having the user-facing part of the app talk to portals in the usual way, but preventing the sandboxable thumbnailer from talking to portals at all (`flatpak-spawn --sandbox` doesn't give D-Bus access, AIUI).

src/wallpaper.c
```diff
@@ -190,18 +190,10 @@ handle_set_wallpaper_in_thread_func (GTask *task,
...
-            {
-              name = g_app_info_get_display_name (G_APP_INFO (info));
-              app_id = xdp_get_app_id_from_desktop_id (g_app_info_get_id (info));
-            }
-          else
-            {
-              name = id;
-              app_id = g_strdup (id);
-            }
+            name = g_app_info_get_display_name (G_APP_INFO (info));
```
swick: wrong indentation

src/wallpaper.c
```diff
@@ -212,7 +204,7 @@ handle_set_wallpaper_in_thread_func (GTask *task,
            * apps for which an app ID can't be determined.
            */
           g_assert (xdp_app_info_is_host (request->app_info));
-          app_id = g_strdup ("");
+          app_id = "";
```
swick: isn't app_id and id the same now?

tests/xdp_utils.py
```diff
@@ -382,8 +382,9 @@ def new_snap(
         metadata: bytes | None = None,
     ):
         kind = AppInfoKind.SNAP
-        app_id = f"snap.{snap_name}"
-        desktop_file = f"{snap_name}_{app_name}.desktop"
+        permissions_id = f"snap.{snap_name}"
+        app_id = f"{snap_name}_{app_name}.desktop"
```
swick: is this a typo or is the app_id really supposed to include the .desktop suffix?

swick (commented):
Yay for looking into this! I really like that this fixes up the tests for snap and manages to hide the permission_id detail in `xdp-permissions.h`.

src/xdp-app-info.h
swick: Honestly, I think this is fine. That detail is really hidden in the xdp-permission part of the code.

mcatanzaro:
> Dunno if this actually helps with anything because it's still just the thing we currently call the app id, even if we now start using another value for storing the permissions.

This problem is already solved by changing the xdp permissions functions to take the entire `XdpAppInfo *` rather than a string, so we don't have to worry about it anymore.

swick:
I think to get this merged, what I want is...

* concrete examples in the commit message (what @smcv is asking for)
* an ack from @jhenstridge or someone else from the snap side

jadahl:
@3v1n0, how do you imagine places that check the permission store outside of xdg-desktop-portal doing? E.g. https://gitlab.freedesktop.org/pipewire/wireplumber/-/blob/15f5f96693d6155750db3713d696eb41446fadbc/src/scripts/client/access-portal.lua#L60

Should portal backends be discouraged to handle integrating with the permission store themselves?

swick:
There are a bunch of commits in this PR which would be nice to have, no matter the decision about what to use in the permission store. Could you split them out into its own PR? I think only commits 46b69f7a112e6e04271b12f1a9679bdbfe5720b0 and 0e1f29089a2f2f1a9007f5f1a74d3c1fa39eb96c change the semantics.

3v1n0:
> concrete examples in the commit message (what @smcv is asking for)

Ok, sure... I will try to add also a test with more complex cases.

Do you have a deadline for the re-review and release?

jadahl:
> Well, using desktop ID in these cases should be fine, it's likely not respecting the real permissions but it should good enough (as it's better to show a pointless prompt more, than just not work).

I think for the camera portal (which `access-portal.lua` is currently only about) will just deny access if the app ID added to the permission store in https://github.com/flatpak/xdg-desktop-portal/blob/7af478c5448939c6a99fb5728dd4ff3dee8ecf36/src/camera.c#L150. I see that you're now changing to set it to the "snap ID" however, so I guess it will still work as expected, so as long as wireplumber doesn't assume it can use the "app ID" as anything that can match to a desktop file no one will notice.

3v1n0:
> . I see that you're now changing to set it to the "snap ID" however, so I guess it will still work as expected, so as long as wireplumber doesn't assume it can use the "app ID" as anything that can match to a desktop file no one will notice.

So in terms of permissions it should work, although I noticed that camera portal does not work as expected at the moment, unlike when doing a "simpler" change such as https://github.com/3v1n0/xdg-desktop-portal/commits/snap-initial-fixes/ so let me check it further, because in both branches we're should be sending the same permission ID, while the desktop ID is used by the shell must be different

jadahl:
> They might have the expectation that what's stored there is the app id which maps to a desktop file and not potentially another thing like the permission id we want to introduce here.

Right; seems we don't really do much with the permission store in xdp-gnome, gnome-shell and localsearch (which are the three places I checked), but would there be checks that require GUI, whatever is passed should also usable to resolve a desktop file describing the app.

> Similarly, existing permissions for snaps will become useless. Might even have to check if the new permission id can possibly collide with the previous app id and result in some confused permission.

If this changes what is stored in the permission store for snaps, it'd be useful to have a migration script of sorts, like we've had for similar situations (in other projects) when stored formats change.

3v1n0:
@jadahl so https://github.com/flatpak/xdg-desktop-portal/pull/1764/commits/72bad0f56654537b1292c4fb771dae28fbc0f43b is enough here for pipewire to work, but as said is probably better to generalize it for the permissions portal in general

Sodivad:
> I think Jonas has a point with others components poking at the permission store. They might have the expectation that what's stored there is the app id which maps to a desktop file and not potentially another thing like the permission id we want to introduce here. That actually needs checking.

> Right; seems we don't really do much with the permission store in xdp-gnome, gnome-shell and localsearch (which are the three places I checked), but would there be checks that require GUI, whatever is passed should also usable to resolve a desktop file describing the app.

gnome settings does however https://gitlab.gnome.org/GNOME/gnome-control-center/-/blob/main/panels/applications/cc-applications-panel.c?ref_type=heads#L265

I didn't follow the entire discussion but is that the correct id to use for snaps? I do the same  in Plasma FWIW. It would also good if the we could document which ids will be used for permissions and how they relate to the `app_id` that is passed to the backend. Also for something like https://github.com/flatpak/xdg-desktop-portal/pull/1768 that could require the backend to look into the permission store.

swick:
I think we really need to standardize some new fields in desktop files, namely:

* AppID
* CollectionID
* EngineID
* PermissionID

Where The AppID is the desktop file name if not specified (we should still recommend the desktop file to match this; it's just unfortunate that snap doesn't follow this).

The CollectionID would be the source, so org.flathub for example, or io.snapcraft for snaps.

The EngineID would be org.flatpak or io.snapcraft, or nothing for host apps.

The PermissionID would be the same as the AppID if not specified.

But even if we do that, we have to find a way to deal with previous versions of desktop files.

document-portal/document-portal.c
```diff
@@ -647,8 +647,13 @@ app_has_file_access (const char *target_permissions_id,
 
   if (g_str_has_prefix (target_permissions_id, "snap."))
     {
+      g_auto (GStrv) parts = NULL;
+
+      parts = g_strsplit (target_permissions_id + strlen ("snap."), "_", -1);
+
+      /* Snap does not support permission check for sub-applications */
```
aleasto, line 654: huh what?! which permission does it check then, since permissions are per-app and not per-snap?

src/xdp-app-info-snap.c
```diff
@@ -255,7 +257,19 @@ parse_snap_metadata (GKeyFile         *metadata,
...
+                                         SNAP_METADATA_KEY_APP_NAME,
+                                         NULL);
+
+  if (snap_app_name == NULL || g_str_equal (snap_app_name, snap_name))
+    {
+      permissions_id = g_strconcat ("snap.", snap_name, NULL);
+    }
+  else
+    {
+      permissions_id = g_strconcat ("snap.", snap_name, "_", snap_app_name, NULL);
```
aleasto, line 271: Since `_` is an arbitrary separator AFAICS, it would be best to use a character that cannot be included in snap_name so that it can easily be parsed back, like `.`

`_` can be part of `X-SnapInstanceName` in case of multi-instance snaps

document-portal/document-portal.c
3v1n0: Indeed, but `snap routine` check does not do that, I feel we should avoid using it though, and just use the REST api here...

@swick how would you feel about having libsoup and json-glib as dependencies (or optional ones)? :)

Or we should improve the tool in snapd instead.

swick: Not entirely opposed, especially fine with a JSON dependency. What would you need libsoup for though?

aleasto: for talking to the snapd REST API instead of shelling-out to `snap routine portal-info`

3v1n0: Yeah, basically snapd has a [REST API](https://snapcraft.io/docs/reference/development/snapd-rest-api/) that we can use through an UNIX socket.

This could be done via [snapd-glib](https://github.com/canonical/snapd-glib/) or more manually using libsoup directly and some JSON lib.

swick: Alright, fine with me.

3v1n0: Ok, I'll do that in another PR though, since it would imply adjusting tests, so I'd prefer not to make this even more complex.

document-portal/permission-db.c
```diff
@@ -49,11 +49,11 @@ struct PermissionDb
 
   gboolean   dirty;
 
-  /* Map id => GVariant (data, sorted-dict[appid->perms]) */
+  /* Map id => GVariant (data, sorted-dict[app_permissions_id->perms]) */
```
swick, line 52: Can you please also check the docs to make sure we don't describe an outdated model?

src/camera.c
```diff
@@ -225,15 +225,15 @@ handle_access_camera (XdpDbusCamera *object,
...
+open_pipewire_camera_remote (const char *permissions_id,
                              GError **error)
 {
   PipeWireRemote *remote;
   struct pw_permission permission_items[3];
   struct pw_properties *pipewire_properties;
 
   pipewire_properties =
-    pw_properties_new ("pipewire.access.portal.app_id", app_id,
+    pw_properties_new ("pipewire.access.portal.app_id", permissions_id,
```
swick, line 236: There is a PR which moves the pipewire plugin into the repo so we could rename this appropriately.

document-portal/permission-db.c
```diff
@@ -1248,3 +1249,40 @@ permission_db_entry_print_string (PermissionDbEntry *entry,
 {
   return g_variant_print_string ((GVariant *) entry, string, FALSE);
 }
+
+char *
+permissions_id_for_app_id (const char *app_id)
```
swick, line 1254: I think this is the big pain point for me in this PR. I don't like that this isn't abstracted away in the XdpAppInfo but then again the XdpAppInfo is about a specific instance. Fundamentally the problem here seems to be that we get passed app ids when we would need permissions ids.

I'm willing to accept this code here, *if* it is documented clearly that this is a stop-gap measure, and you sign up to do the work to convert the callers to provide permissions ids.

swick (commented):
This also needs a rebase because I changed the pytest XdpAppInfo again.

swick:
The changes are a bit much to still get them into the 1.22 stable release.

src/xdp-app-info-snap.c
aleasto: Ok, to avoid having to migrate existing permission stores, let's keep using `_`; it's still fine, as to parse back into snap components we can simply split on the last `_` character instead of the first.

document-portal/document-portal.c
aleasto: We found out by talking to the snapd people that while permissions are declared on a per-app basis, they're only actually granted/revoked on a per-snap basis. So this already matches expectations.

document-portal/xdg-permission-store.c
```diff
@@ -397,12 +401,16 @@ handle_set (XdgPermissionStore     *object,
...
       g_autoptr(PermissionDbEntry) old_entry = NULL;
       const char *child_app_id;
+      const char *child_permissions_id;
       g_autofree const char **permissions;
 
       g_variant_get (child, "{&s^a&s}", &child_app_id, &permissions);
 
       old_entry = new_entry;
-      new_entry = permission_db_entry_set_app_permissions (new_entry, child_app_id, (const char **) permissions);
+      child_permissions_id = child_app_id;
```
aleasto, line 410: why are you not translating `app_id` -> `permissions_id` here?

aleasto (changes requested):
if xdg-permission-store implicitly translates app ids to permissions ids, then users calling `Lookup()` will be unable to index the returned map by what they thought would be app ids.

that might be fine to break the interface like this, as the only user shall be x-d-p itself, but then we need to adjust all the callers of `org.freedesktop.impl.portal.PermissionStore.Lookup`

This is causing bug https://bugs.launchpad.net/ubuntu/+source/xdg-desktop-portal/+bug/2152906

aleasto:
> that might be fine to break the interface like this, as the only user shall be x-d-p itself

Actually that's not true. It is also used to implement settings in apps like gnome-control-center

aleasto:
> > that might be fine to break the interface like this, as the only user shall be x-d-p itself
> 
> Actually that's not true. It is also used to implement settings in apps like gnome-control-center

I suppose that's why you have https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/3353
Maybe you can argue this never worked previously for snaps, so it is not a breaking change.

Sodivad:
I have the equivalent code as gnome-settings in plasma as well

swick:
@aleasto I'm a bit lost on the state of this PR. Is this done and you're just waiting for a review or are you still trying to iron out some edge cases?

aleasto:
This needs addressing my earlier comment https://github.com/flatpak/xdg-desktop-portal/pull/1764#pullrequestreview-4561265649 at least, because it's a regression.

I believe @3v1n0 will address that sometime soon.

PR #1766, opened 2025-08-11
tests: Skip test_trash_file_fails if we couldn't get a fd

aleasto:
Tests may be running in an environment not allowed to open /proc/cmdline

### DIFF
--- a/tests/test_trash.py
+++ b/tests/test_trash.py
@@ -5,6 +5,7 @@
 import tests.xdp_utils as xdp
 
 import os
+import pytest
 import tempfile
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


tests/test_trash.py
```diff
@@ -15,8 +16,11 @@ def test_version(self, portals, dbus_con):
 
     def test_trash_file_fails(self, portals, dbus_con):
         trash_intf = xdp.get_portal_iface(dbus_con, "Trash")
-        with open("/proc/cmdline") as fd:
-            result = trash_intf.TrashFile(fd.fileno())
+        try:
+            with open("/proc/cmdline") as fd:
+                result = trash_intf.TrashFile(fd.fileno())
+        except PermissionError as e:
+            pytest.skip(f"Couldn't open file: {e}")
```
swick: Should be a bit more specific about what it's trying to open, or even better, find something that works everywhere.

swick (approved):
Thank you!

PR #1767, opened 2025-08-12
subprojects: add a conftest.py so we don't run pytests

whot:
cc @swick - not sure if running the subproject pytests was intentional


swick:
I never ran into that problem. I guess you called pytest from the project root?

Anyway, LGTM.

PR #1768, opened 2025-08-12
inputcapture: Add support for the permission store

whot:
The "inputcapture" permission table/id now store a per-app-id triplet of
masks. The first one is the mask capabilities will always be clamped to.
The second mask is the set of capabilities approved without user
feedback, the third set is the set of capabilties that require user
approval. Any capability requests that fall outside those will be
automatically denied.

This allows a user to configure e.g.:
- never approve keyboards but proceed with any other capability
- auto-allow keyboard/pointer but ask if touch is requested too
- auto-allow pointer/keyboard, ignore touch requests altogether
- allow/deny everything automatically based on the appid

This should cover all use-cases we need to worry about.

Example usage using the flatpak CLI to allow keyboard+pointer but ask
if touch is requested (with 255 as overeager clamping mask):
```
    $ flatpak permission-set inputcapture inputcapture org.deskflow.deskflow 255 3 7
```
Note that xdg-desktop-portal only handles the denial part of this
process since the "ask" part needs to be handled in the impl.

---

This is the x-d-p counterpart (and near-equivalent) to [xdg-desktop-portal-gnome!228](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/228).

Draft because we're hogging the generic "inputcapture" id so ideally this would be agreed upon by upstream x-d-p so we can have all compositors implement the same thing.

KDE has something similar-ish for the RemoteDesktop portal in [xdg-desktop-portal-kde!326](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/326)

cc @hsitter, @nbolton, @davidedmundson, @conan-kudo, @davidre, @jadahl


src/input-capture.c
```diff
@@ -222,10 +289,6 @@ validate_capabilities (const char  *key,
   return TRUE;
 }
 
-static XdpOptionKey input_capture_create_session_options[] = {
```
swick, line 252: Why is this PR open coding this instead?

whot: `xdp_filter_options()` filters and appends to the builder but there's only one option (`capabilities`) and the ones we pass on are the possibly modified capabilities, not the original ones. So we'd have to build a new options variant or somehow overwrite the existing one in the builder, both seem more work.

Although now that I write this I wonder if `g_variant_builder_add()` overwrites the key-value if it already exists but from a quick test case I tried here it just takes the first value added and ignores subsequent ones.

swick:
Honestly, I don't understand where the user is queried for permissions here at all.

davidre:
Why am i linked to this PR and retrieve notifications??
Perhaps because of a (wrong) match between @Sodivad account name abbreviation and my account name abbreviation?
How to unlink?

swick:
There is an unsubscribe button in the sidebar.

davidre:
yes, i know, but i'm incorrectly mentioned as a participant (see screenshot). And the pr shows up in my gitkraken launchpad. That's a bit annoying. Any idea how to remove that?
image

src/input-capture.c
```diff
@@ -152,6 +156,74 @@ input_capture_session_new (InputCapture     *input_capture,
...
+static gboolean
+get_capability_masks_from_permission_store (XdpAppInfo *app_info,
+                                            uint32_t   *clamp_mask,
+                                            uint32_t   *yes_mask,
+                                            uint32_t   *ask_mask)
+{
+  g_auto(GStrv) permissions = NULL;
+
+  permissions = xdp_get_permissions_sync (app_info, PERMISSION_TABLE, PERMISSION_ID);
+  if (!permissions || g_strv_length (permissions) < 3)
```
jadahl, line 170: Feels like this should be documented somewhere, e.g. in the impl.portal XML file.

src/input-capture.c
```diff
@@ -269,9 +332,37 @@ handle_create_session (XdpDbusInputCapture   *object,
...
+  if (!g_variant_lookup (arg_options, "capabilities", "u", &requested_capabilities))
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                              XDG_DESKTOP_PORTAL_ERROR,
+                                              XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                                              "Missing 'capabilities' in CreateSession");
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+
+  if (!validate_capabilities(requested_capabilities, &error))
```
jadahl: Missing space before (

src/input-capture.c
whot: Tried it, it felt wrong to put it into the xml file so I put it (and for the other portals) onto a separate documentation page now.

whot:
> Should you send the caps somehow back to the backend or an indication if a dialog should be shown? Otherwise the backend needs to do a compare the masks to requested capabilities again.
The screenshot portal has permission_store_checked for example that the backend doesnt need to show a dialog

This has now been added, there's a boolean that tells the impl whether the dialog should be shown or not.

whot:
> Also, is there a reason why this is a draft?

Haven't implemented and tested the xdp-gnome bits for everything yet so there's a chance some bug may still surface.

whot:
Closing in favour of #1898 - that is the better approach

PR #1769, opened 2025-08-12
screenshot: Adjust texts when we are not setting permissions

Sodivad:
If the permission is neither unset, "yes" or "no" we ask but won't set the permission afterwards. Adjusts the text shown to the user for this case to be not misleading about what happens.

### DIFF
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -189,6 +189,111 @@ static XdpOptionKey screenshot_options[] = {
   { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL }
 };
 
+static gboolean
+check_non_interactive_permission_in_thread (XdpRequest *request,
+                                            const char *parent_window,
+                                            gboolean modal)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) access_results = NULL;
+  g_auto(GVariantBuilder) access_opt_builder =
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
+  XdpPermission permission;
+  g_autofree char *subtitle = NULL;
+  g_autofree char *title = NULL;
+  const char *body = "";
+  uint access_response = 2;
+  const char *app_id;
+
+  permission = xdp_get_permission_sync (request->app_info, PERMISSION_TABLE, PERMISSION_ID);
+
+  if (permission == XDP_PERMISSION_YES)
+    return TRUE;
+
+  if (permission == XDP_PERMISSION_NO)
+    return FALSE;
+
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                         "deny_label", g_variant_new_string (_("Deny")));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                         "grant_label", g_variant_new_string (_("Allow")));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                         "icon", g_variant_new_string ("applets-screenshooter-symbolic"));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                         "modal", g_variant_new_boolean (modal));
+
+  app_id = xdp_app_info_get_id (request->app_info);
+
+  if (g_strcmp0 (app_id, "") != 0)
+    {
+      g_autoptr(GDesktopAppInfo) info = NULL;
+      g_autofree gchar *id = NULL;
+      const gchar *name = NULL;
+
+      id = g_strconcat (app_id, ".desktop", NULL);
+      info = g_desktop_app_info_new (id);
+
+      if (info)
+        name = g_app_info_get_display_name (G_APP_INFO (info));
+      else
+        name = app_id;
+
+      if (permission == XDP_PERMISSION_UNSET)
+        {
+          title = g_strdup_printf (_("Allow %s to Take Screenshots?"), name);
+          subtitle = g_strdup_printf (_("%s wants to be able to take screenshots at any time."), name);
+        }
+      else /* permission == XDP_PERMISSION_ASK */
+        {
+          title = g_strdup_printf (_("Allow %s to Take a Screenshot?"), name);
+          subtitle = g_strdup_printf (_("%s wants to take a screenshot."), name);
+        }
+    }
+  else
+    {
+      g_assert (xdp_app_info_is_host (request->app_info));
+      if (permission == XDP_PERMISSION_UNSET)
+        {
+          /* Note: this will set the screenshot permission for all unsandboxed
+           * apps for which an app ID can't be determined.
+           */
+          title = g_strdup (_("Allow Applications to Take Screenshots?"));
+          subtitle = g_strdup (_("An application wants to be able to take screenshots at any time."));
+        }
+      else /* permission == XDP_PERMISSION_ASK */
+        {
+          title = g_strdup (_("Allow to Take a Screenshot?"));
+          subtitle = g_strdup (_("An application wants to take a screenshot."));
+        }
+    }
+
+  if (permission == XDP_PERMISSION_UNSET)
+    body = _("This permission can be changed at any time from the privacy settings.");
+
+  if (!xdp_dbus_impl_access_call_access_dialog_sync (access_impl,
+                                                     request->id,
+                                                     app_id,
+                                                     parent_window,
+                                                     title,
+                                                     subtitle,
+                                                     body,
+                                                     g_variant_builder_end (&access_opt_builder),
+                                                     &access_response,
+                                                     &access_results,
+                                                     NULL,
+                                                     &error))
+    {
+      g_warning ("Failed to show access dialog: %s", error->message);
+      return FALSE;
+    }
+
+  if (permission == XDP_PERMISSION_UNSET)
+    xdp_set_permission_sync (request->app_info, PERMISSION_TABLE, PERMISSION_ID, access_response == 0 ? XDP_PERMISSION_YES : XDP_PERMISSION_NO);
+
+  return access_response == 0;
+}
+
+
 static void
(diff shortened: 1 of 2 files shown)


src/screenshot.c
```diff
@@ -264,20 +264,30 @@ handle_screenshot_in_thread_func (GTask *task,
...
+            }
         }
       else
         {
-          /* Note: this will set the wallpaper permission for all unsandboxed
+          /* Note: this will set the screenshot permission for all unsandboxed
            * apps for which an app ID can't be determined.
            */
           g_assert (xdp_app_info_is_host (request->app_info));
-          title = g_strdup (_("Allow Applications to Take Screenshots?"));
```
swick, line 272: Why is this being removed?

src/screenshot.c
```diff
@@ -264,20 +264,30 @@ handle_screenshot_in_thread_func (GTask *task,
...
+          if (permission == XDP_PERMISSION_UNSET)
+            {
+              title = g_strdup_printf (_("Allow %s to Take Screenshots?"), name);
+              subtitle = g_strdup_printf (_("%s wants to be able to take screenshots at any time."), name);
+            }
+          else
+            {
+              title = g_strdup_printf (_("Allow %s to Take a Screenshot?"), name);
+              subtitle = g_strdup_printf (_("%s wants to take a screenshot."), name);
+            }
```
swick: The cases XDP_PERMISSION_YES, XDP_PERMISSION_NO, and XDP_PERMISSION_UNSET were already handled, so this branch is XDP_PERMISSION_ASK.

We never write XDP_PERMISSION_ASK because it makes no sense to do non-interactive screenshots while also asking every time.

So I'm a bit confused what this PR is trying to do.

src/screenshot.c
```diff
@@ -264,20 +264,30 @@ handle_screenshot_in_thread_func (GTask *task,
...
           g_assert (xdp_app_info_is_host (request->app_info));
-          title = g_strdup (_("Allow Applications to Take Screenshots?"));
           subtitle = g_strdup (_("An application wants to be able to take screenshots at any time."));
         }
 
-      body = _("This permission can be changed at any time from the privacy settings.");
+      if (permission == XDP_PERMISSION_UNSET)
+      {
+        body = _("This permission can be changed at any time from the privacy settings.");
+      }
```
swick:
```suggestion
      if (permission == XDP_PERMISSION_UNSET)
        body = _("This permission can be changed at any time from the privacy settings.");
```

src/screenshot.c
Sodivad: By accident.

src/screenshot.c
Sodivad: Yes the portal currently doesn't write it but the user could set it manually to ask.

My idea is that the system such as GNOME settings could also present such an option. 

> We never write XDP_PERMISSION_ASK because it makes no sense to do non-interactive screenshots while also asking every time.

Maybe. But user might still not want to give an app blanket permissions to take screenshot but still have them work. Apps usually do not have a fallback to interactive if the non-interactive screenshot fails because the user disallowed them for this application.

swick: Got it. The code here is really not easy to follow, so I'm wondering if this could be refactored somehow.

Sodivad: Maybe spelling things out more clearly instead of not equals comparison?

Sodivad: I shuffled the code a bit, maybe it's now clearer.

swick: This is already much better. Can you add one more commit to move it into its own function? That should get rid of the goto.

src/screenshot.c
```diff
@@ -217,29 +217,34 @@ handle_screenshot_in_thread_func (GTask *task,
...
     interactive = FALSE;
 
   if (!g_variant_lookup (options, "modal", "b", &modal))
     modal = TRUE;
 
-  if (!interactive && permission != XDP_PERMISSION_YES)
+  if (interactive)
+    goto query_impl;
+
+  // Non-interactive
```
swick: The comment style we use is still only `/* */`.

src/screenshot.c
```diff
@@ -214,125 +316,23 @@ handle_screenshot_in_thread_func (GTask *task,
...
-          g_warning ("Failed to show access dialog: %s", error->message);
-          send_response (request, 2, g_variant_builder_end (&opt_builder));
-          return;
-        }
-
-      if (permission == XDP_PERMISSION_UNSET)
-        xdp_set_permission_sync (app_id, PERMISSION_TABLE, PERMISSION_ID, access_response == 0 ? XDP_PERMISSION_YES : XDP_PERMISSION_NO);
-
-      if (access_response != 0)
+      if (!interactive && !check_non_interactive_permission(app_id, request, parent_window, modal))
```
swick:
```suggestion
      if (!interactive && !check_non_interactive_permission (app_id, request, parent_window, modal))
```

src/screenshot.c
```diff
@@ -214,125 +316,23 @@ handle_screenshot_in_thread_func (GTask *task,
...
         {
           send_response (request, 2, g_variant_builder_end (&opt_builder));
           return;
         }
+        permission_store_checked = TRUE;
```
swick:
```suggestion
      permission_store_checked = TRUE;
```

src/screenshot.c
```diff
@@ -189,6 +189,109 @@ static XdpOptionKey screenshot_options[] = {
   { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL }
 };
 
+static gboolean
+check_non_interactive_permission (const char *app_id,
```
swick: Let's add the `in_thread` suffix as well, so it becomes clear where this should be called and that sync operations are okay.

src/screenshot.c
```diff
@@ -264,20 +269,37 @@ handle_screenshot_in_thread_func (GTask *task,
...
+          if (permission == XDP_PERMISSION_UNSET)
+            {
+              /* Note: this will set the screenshot permission for all unsandboxed
+               * apps for which an app ID can't be determined.
+               */
+              title = g_strdup (_("Allow Applications to Take Screenshots?"));
+              subtitle = g_strdup (_("An application wants to be able to take screenshots at any time."));
+           }
+          else /* permission == XDP_PERMISSION_ASK */
+          {
```
swick:
```suggestion
          else /* permission == XDP_PERMISSION_ASK */
            {
```

src/screenshot.c
```diff
@@ -189,6 +189,109 @@ static XdpOptionKey screenshot_options[] = {
...
+static gboolean
+check_non_interactive_permission_in_thread (const char *app_id,
+                                            XdpRequest *request,
+                                            const char *parent_window,
+                                            gboolean modal)
+{
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) access_results = NULL;
+  g_auto(GVariantBuilder) access_opt_builder =
+  G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
```
swick:
```suggestion
  g_auto(GVariantBuilder) access_opt_builder =
    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
```

src/screenshot.c
```diff
@@ -189,6 +189,109 @@ static XdpOptionKey screenshot_options[] = {
...
+    return FALSE;
+
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                          "deny_label", g_variant_new_string (_("Deny")));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                          "grant_label", g_variant_new_string (_("Allow")));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                          "icon", g_variant_new_string ("applets-screenshooter-symbolic"));
+  g_variant_builder_add (&access_opt_builder, "{sv}",
+                            "modal", g_variant_new_boolean (modal));
```
swick:
```suggestion
  g_variant_builder_add (&access_opt_builder, "{sv}",
                         "deny_label", g_variant_new_string (_("Deny")));
  g_variant_builder_add (&access_opt_builder, "{sv}",
                         "grant_label", g_variant_new_string (_("Allow")));
  g_variant_builder_add (&access_opt_builder, "{sv}",
                         "icon", g_variant_new_string ("applets-screenshooter-symbolic"));
  g_variant_builder_add (&access_opt_builder, "{sv}",
                         "modal", g_variant_new_boolean (modal));
```

src/screenshot.c
```diff
@@ -189,6 +189,109 @@ static XdpOptionKey screenshot_options[] = {
...
+        }
+    }
+  else
+    {
+      g_assert (xdp_app_info_is_host (request->app_info));
+      if (permission == XDP_PERMISSION_UNSET)
+        {
+          /* Note: this will set the screenshot permission for all unsandboxed
+            * apps for which an app ID can't be determined.
+            */
```
swick:
```suggestion
          /* Note: this will set the screenshot permission for all unsandboxed
           * apps for which an app ID can't be determined.
           */
```

src/screenshot.c
```diff
@@ -189,6 +189,109 @@ static XdpOptionKey screenshot_options[] = {
...
+          subtitle = g_strdup (_("An application wants to take a screenshot."));
+        }
+    }
+
+  if (permission == XDP_PERMISSION_UNSET)
+    body = _("This permission can be changed at any time from the privacy settings.");
+
+  if (!xdp_dbus_impl_access_call_access_dialog_sync (access_impl,
+                                                      request->id,
+                                                      app_id,
```
swick:
```suggestion
  if (!xdp_dbus_impl_access_call_access_dialog_sync (access_impl,
                                                     request->id,
                                                     app_id,
```

swick (commented):
LGTM with the indentation fixed.

PR #1770, opened 2025-08-12
tests: Add support for running tests with valgrind's memcheck

swick:
See commits


tests/meson.build
```diff
@@ -3,7 +3,6 @@ env_tests.set('XDG_DATA_DIRS', meson.current_build_dir() / 'share')
 env_tests.set('G_TEST_SRCDIR', meson.current_source_dir())
 env_tests.set('G_TEST_BUILDDIR', meson.current_build_dir())
 env_tests.set('G_DEBUG', 'gc-friendly')  # from glib-tap.mk
-env_tests.set('LSAN_OPTIONS', 'suppressions=' + meson.current_source_dir() / 'asan.suppression')
```
hfiguiere, line 6: why removing this?

asan and valgrind are mutually exclusive but being able to do either is important (asan performance penalty is much lower)

swick: The commit message has a hint.

src/xdg-desktop-portal.c
```diff
@@ -431,13 +431,13 @@ on_name_lost (GDBusConnection *connection,
               const gchar     *name,
               gpointer         user_data)
 {
-  g_main_loop_quit (loop);
+  exit_with_status (0);
```
jadahl, line 434: I find exiting the "right" way (exiting the loop and returning from `main`) to be generally better, and allows for more proper cleanup.

Why are you changing to a more sudden exit without proper cleanup?

swick: It does call `g_main_loop_quit` and sets the status code, which IMO is better than just exiting without being explicit about the status code.

jadahl: Ah, exit_with_status is an internal helper, I was grepping in the wrong tree..

PR #1771, opened 2025-08-12
Refactor portal-impl into a GObject

swick:


jadahl:
Perhaps `XdpImplRegistry` would have been a better name. It's a bit more than a list of impls, which the name somewhat implies.

swick:
I always thought that it really is the portal config. Registry might provoke a wrong association with the host registry portal.

jadahl:
With `XdpPortalImplementation` more predictably named, `XdpPortalConfig` indeed makes more sense than the alternatives.

PR #1772, opened 2025-08-13
AppInfo: Cleanups permissions and desktop IDs usage

3v1n0:
Various cleanups that were included in #1764 that can be generic fixes


swick:
The last commit seems to break the tests because of this: https://github.com/flatpak/xdg-desktop-portal/blob/main/tests/test_background.py#L99C18-L99C69

3v1n0:
> The last commit seems to break the tests because of this: https://github.com/flatpak/xdg-desktop-portal/blob/main/tests/test_background.py#L99C18-L99C69

Yeah, I noticed promptly, but I take the occasion to make it a bit more general instead of hardcoding the expected path

swick:
Perfect, thank you!

PR #1777, opened 2025-08-17
print: Add two new options

matthiasclasen:
Print dialogs can have a feature that lets the user select 'current page' or 'selection' for what pages to print.
This only makes sense if there *is* a current page or page selection.

This information needs to be provided by the caller, which is what the new options are about.

An implementation for the gnome backends is here: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/229

And this MR makes gtk use the new options: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8866

### DIFF
--- a/data/org.freedesktop.portal.Print.xml
+++ b/data/org.freedesktop.portal.Print.xml
@@ -34,7 +34,7 @@
       to print the formatted document. It is expected that high-level toolkit
       APIs such as GtkPrintOperation will hide most of this complexity.
 
-      This documentation describes version 3 of this interface.
+      This documentation describes version 4 of this interface.
   -->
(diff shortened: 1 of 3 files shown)


data/org.freedesktop.portal.Print.xml
```diff
@@ -248,6 +260,7 @@
 
           Token that can be passed to a subsequent org.freedesktop.portal.Print.Print() call to
           bypass the print dialog.
+
```
swick: Unrelated whitespace change

swick:
Resolved the merge conflicts.

PR #1779, opened 2025-08-19
Change to a new versioning scheme

swick:
To avoid the even/odd confusion and finally make use of the major version.

This is also meant as a starting point for a discussion. While I think this would work, I'm very open to ideas and opinions.

@GeorgesStavracas @jadahl @smcv

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -18,9 +18,9 @@ body:
     label: "XDG Desktop Portal version"
     description: "What version of XDG Desktop Portal are you using?"
     options:
-    - "1.21"
+    - "main"
+    - "2"
     - "1.20"
-    - "1.18"
     - Git
     - Other
   validations:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: Release new version
 on:
   push:
     tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+\.[a-z0-9]+'
 
 jobs:
   build-container:
@@ -54,7 +54,7 @@ jobs:
           echo $releaseChangelog
 
           # Check if version is a pre-release
-          preRelease=$((`echo $releaseVersion | cut -d '.' -f2` % 2))
+          preRelease=$([[ $releaseVersion =~ ^[0-9]+\.[0-9]+$ ]]; echo $?)
           {
             echo -n "preRelease="
             if [ $preRelease = 1 ] || [ $preRelease = "true" ]; then
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
-Changes in 1.21.0
-=================
-Released: Not yet
+Changes in 2.rc0
+================
+Released: 2025-08-19
 
 ...
 
--- a/RELEASE_HOWTO.md
+++ b/RELEASE_HOWTO.md
@@ -2,7 +2,6 @@
 
 ### Prepare the release
 
-- Make sure the version in `meson.build` is correct
 - Create a branch
 ```sh
 $ git checkout -b release-${version}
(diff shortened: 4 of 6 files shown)


GeorgesStavracas:
I had a longer comment written, but @smcv beat me to it. What I had in mind is basically what Simon mentioned in his comment.

swick:
I didn't say we should follow semver. The major version as an indicator that something isn't backwards compatible is useless for portals because the app does not have any say in which version the system has. If we have to do backwards incompatible changes, the versioning has absolutely no relevance to anything whatsoever.

swick:
> it will also create new even/odd confusion for anyone who already knows

This is such a weird system that I've only ever seen in the gnome context. On the other hand, everyone (tm) understands major/minor releases. There might be confusion to people who know the  current scheme, but that's unavoidable when changing things.

> What I've been intending to do in dbus (and probably Flatpak too, https://github.com/flatpak/flatpak/issues/6014) is to add a suffix to the odd branches but not the even ones, like:

That looks even more confusing than the current scheme. If I see `1.17.3-rc` I would expect it to be the rc of the `1.17.3` series. It's not. It's the first time I've ever seen this. I don't understand how introducing an even more complicated scheme is making anything better.

swick:
> This doesn't work for any project that has the concept of a stable branch that is suitable for inclusion in LTS distributions, and a development branch that has higher-risk changes that would be an unacceptable regression risk for LTS distros' release managers.

That is however the core thing I want to get rid of. No more two different modes. Every release is supposed to be stable. If you really want an LTS version, we just decide on one every N years, and cherry-pick fixes.

smcv:
> If I see `1.17.3-rc` I would expect it to be the rc of the `1.17.3` series

That's a feature, sort of. If a consumer of x-d-p (a distro, packager or user) doesn't understand our versioning scheme and doesn't read our documentation, we want that situation to "fail safe" towards the consumer of x-d-p deciding that they are only going to use our stable releases, rather than  starting to use our development snapshots because they have been confused into thinking they're stable.

If the consumer mistakenly thinks they are waiting for a 1.17.3 stable release that will never actually happen, then the likely result is that they'll wait for 1.17.3, never get it, eventually see that 1.18.0 has been released, and upgrade to 1.18.0 instead. That's the right thing happening, even if it was for the wrong reasons!

Meanwhile, if they did read our documentation in the past, then they can make an informed decision about whether they are going to use a development or stable branch, and the right thing happens - regardless of whether they have updated their project knowledge since the change to the versioning scheme.

> If we have to do backwards incompatible changes, the versioning has absolutely no relevance to anything whatsoever.

It has relevance to the distros that ship our code. As a bridge between two worlds (the OS/system/desktop and the app/sandbox), x-d-p has a "surface area" at both ends: yes we provide an interface to the apps, but we also provide an interface to the system/desktop/OS.

If you're Red Hat or Debian or Ubuntu or openSUSE, you need to know at what point an x-d-p update is "business as usual", and at what point it's a potentially disruptive transition that should be expected to have fallout (which those distros wouldn't want to carry out if there isn't enough time left before their next LTS-release deadline). The redesign of how `portals.conf` works is an example of a change that was disruptive for distros and needed some planning ahead, even though it didn't impact apps.

> The major version as an indicator that something isn't backwards compatible is useless for portals because the app does not have any say in which version the system has.

Sure, but is this information even exposed to portals? What we expose to portals is a collection of D-Bus interfaces ("names are ABIs and ABIs are names", and if we created a non-backwards-compatible interface, we'd give it a new name like `org.freedesktop.portal.InputCapture2` that behaves like a major version), together with `Version` properties that behave like a minor version and indicate to the app what compliance-level of functionality we're providing. I agree that the version number seen in tarball releases, git tags, `meson.build` is not very relevant at this end of x-d-p.

smcv:
> No more two different modes. Every release is supposed to be stable

Every release of every project is *supposed to be* stable, but the difference between theory and practice is that in theory, there's no difference :-)

We did try doing a release cycle with no development snapshots and no prereleases (there were no 1.13.x development releases, we skipped directly from 1.12.x stable to 1.14.0 stable) and it's noticeable that we very quickly needed a 1.14.1 to fix a deadlock, which could have been avoided if we had signalled to distros that it was a good time to do their pre-release testing.

swick:
Superseded by #2042

PR #1785, opened 2025-08-27
XdpAppInfo Host pidfd improvements

swick:
Fixes a but that was introduced recently and makes host app pid translation more reliable.


document-portal/document-portal-fuse.c
```diff
@@ -3533,7 +3533,7 @@ xdp_fuse_thread (gpointer data)
     }
 
   fuse_opts = g_new0 (XdpFuseOptions, 1);
-#ifdef WITH_SPLICE
+#ifdef HAVE_SPLICE
```
GeorgesStavracas:
```suggestion
#if HAVE_SPLICE
```

No?

swick: Woops, good catch!

Issue #1742, opened 2025-06-18
Autostart files for non-Flatpak apps are not named correctly


gwuen:
### Operating System

Arch Linux

### XDG Desktop Portal version

1.20

### Desktop Environment

GNOME

### Expected Behavior

When a non-containerized (i.e., non-Flatpak) app requests to be auto-started via a background request to the portal, the resulting autostart desktop file should either get a consistent, at least mildly unique filename (like the executable's name), or throw an error.

### Current Behavior

When launching the requesting app via its desktop icon in GNOME, the resulting autostart file is named just [.desktop](https://github.com/user-attachments/files/20745320/default.desktop.txt) (without any prefix). When the same app is started from the terminal via `gio launch <app>.desktop` or its executable name, the filename is [transient.desktop](https://github.com/user-attachments/files/20745325/transient.desktop.txt).

The `Name=` and `X-XDP-Autostart=` lines in the file are also either empty or contain the string `transient`, respectively.

This does not happen when the Flatpak version of an app is used, presumably because then the desktop portal is able to correctly figure out the app ID to be used as filename.

### Steps to Reproduce

A minimal Python example that only requires PyGObject demonstrates this behavior:

```python
#!/usr/bin/env python3

import gi
gi.require_version("Xdp", "1.0")
from gi.repository import Xdp

xdp = Xdp.Portal.new()

xdp.request_background(
    None,  # parent
    "Testing the creation of autostart files",  # reason
    ["/usr/bin/true"],  # commandline
    Xdp.BackgroundFlags.AUTOSTART, # flags
    None,  # cancellable
    None, # callback
    None,  # user_data
)
```

When executing this script from the terminal, an autostart entry named `transient.desktop` will be created in `$XDG_CONFIG_HOME/autostart/`, without showing any warning or error message about the lack of proper filename.

### Anything else we should know?

In addition to the Python example, I was able to reproduce this behavior with the [gtk4 demo included in libportal](https://github.com/flatpak/libportal/tree/796053d2eebe4532aad6bd3fd80cdf3b197806ec/portal-test/gtk4). Multiple apps using [ASHPD](https://github.com/bilelmoussaoui/ashpd) as a Rust-native alternative to libportal also experience the same issue, including [Packet](https://github.com/nozwock/packet), [Flare](https://gitlab.com/schmiddi-on-mobile/flare), and ASHPD's own demo app.

[...]

PR #1793, opened 2025-09-02
background: Rely on the app id being available to use autostart

swick:
This was the intent anyway, but the check for the app id was wrong (comparing against NULL instead of "").

If apps want to use this, they have to ensure an app id is detected for them, which might require using the host registry portal for host apps.

Closes: https://github.com/flatpak/xdg-desktop-portal/issues/1742


Sodivad:
LGTM

PR #1794, opened 2025-09-02
usb: Error out if the app requests write access to a read-only device

swick:
/cc @Sodivad @hfiguiere 

Should probably have a corresponding test.

### DIFF
--- a/src/usb.c
+++ b/src/usb.c
@@ -1029,6 +1029,35 @@ out:
 }
 
 static gboolean
+filter_access_devices_writable (const char   *key,
+                                GVariant     *value,
+                                GUdevDevice  *device,
+                                GError      **error)
+{
+  const char *device_file;
+  gboolean writable = g_variant_get_boolean (value);
+
+  if (!writable)
+    return TRUE;
+
+  device_file = g_udev_device_get_device_file (device);
+  if (access (device_file, W_OK) != -1)
+    return TRUE;
+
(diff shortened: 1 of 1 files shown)


Sodivad:
Thank,  I wondered that there was no check for this when I skimmed the impl

src/usb.c
```diff
@@ -1028,21 +1028,51 @@ usb_acquire_devices_cb (GObject      *source_object,
...
+               XDG_DESKTOP_PORTAL_ERROR,
+               XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
+               "Requested writable access for read-only device");
+  return FALSE;
+}
+
+typedef struct {
+  const char *key;
+  const GVariantType *type;
+  gboolean (* filter) (const char *key, GVariant *value, GUdevDevice *device, GError **error);
```
AdrianVovk, line 1057: A bit ugly to have what's basically a re-implementation of `xdp_filter_options` here. Maybe there's some way to change `xdp_filter_options` to pass through a generic `void* user_data`?

swick: I'm doing that in https://github.com/flatpak/xdg-desktop-portal/pull/1819. Could try to pull that out of there, but it touches quite a number of places.

AdrianVovk: No that's ok, then let's just land this

PR #1803, opened 2025-09-04
Add clipboard support to the input capture portal

swick:
Loosely based on https://github.com/flatpak/xdg-desktop-portal/pull/1683.

### DIFF
--- a/tests/templates/inputcapture.py
+++ b/tests/templates/inputcapture.py
@@ -40,6 +40,7 @@ class InputcaptureParameters:
     activated_delay: int
     deactivated_delay: int
     zones_changed_delay: int
+    force_clipboard_enabled: bool
 
 
 def load(mock, parameters={}):
@@ -55,6 +56,7 @@ def load(mock, parameters={}):
         activated_delay=parameters.get("activated-delay", 0),
         deactivated_delay=parameters.get("deactivated-delay", 0),
         zones_changed_delay=parameters.get("zones-changed-delay", 0),
+        force_clipboard_enabled=parameters.get("force-clipboard-enabled", False),
     )
 
     mock.current_zones = mock.inputcapture_params.default_zone
@@ -116,6 +118,9 @@ def Start(self, handle, session_handle, app_id, parent_window, options):
 
         capabilities &= params.supported_capabilities
         response = Response(0, {})
+
+        if params.force_clipboard_enabled:
+            response.results["clipboard_enabled"] = True
 
         response.results["capabilities"] = dbus.UInt32(capabilities)
         self.active_session_handles.append(session_handle)
--- a/tests/test_clipboard.py
+++ b/tests/test_clipboard.py
@@ -3,10 +3,12 @@
 # This file is formatted with Python Black
 
 import tests.xdp_utils as xdp
+import tests.test_inputcapture as inputcapture
 
 import dbus
 import pytest
 import os
+from enum import Enum
 
 
 @pytest.fixture
@@ -14,14 +16,26 @@ def required_templates():
     return {
         "Clipboard": {},
         "RemoteDesktop": {"force-clipboard-enabled": True},
+        "InputCapture": {
+            "force-clipboard-enabled": True,
+            "activated-delay": 100,
+        },
     }
 
 
+class SessionType(Enum):
+    REMOTE_DESKTOP = 1
+    INPUT_CAPTURE = 2
+
+
+@pytest.mark.parametrize(
+    "type", (SessionType.REMOTE_DESKTOP, SessionType.INPUT_CAPTURE)
+)
 class TestClipboard:
-    def test_version(self, portals, dbus_con):
+    def test_version(self, portals, dbus_con, type):
         xdp.check_version(dbus_con, "Clipboard", 1)
 
-    def start_session(self, dbus_con):
+    def start_remote_desktop_session(self, dbus_con):
         clipboard_intf = xdp.get_portal_iface(dbus_con, "Clipboard")
         remotedesktop_intf = xdp.get_portal_iface(dbus_con, "RemoteDesktop")
 
@@ -47,32 +61,84 @@ class TestClipboard:
 
         return (session, start_session_response.results.get("clipboard_enabled"))
 
-    def test_request_clipboard_and_start_session(self, portals, dbus_con):
-        _, clipboard_enabled = self.start_session(dbus_con)
+    def start_input_capture_session(self, dbus_con):
+        inputcapture_intf = xdp.get_portal_iface(dbus_con, "InputCapture")
+
+        session = inputcapture.InputcaptureSession(dbus_con)
+        session.create()
+        start_session_response = session.start()
+        session.get_zones()
+        # The default zone is 1920x1080
+        barriers = [
+            {
+                "barrier_id": dbus.UInt32(10, variant_level=1),
+                "position": dbus.Struct(
+                    [0, 0, 1920, 0], signature="iiii", variant_level=1
+                ),
+            },
+        ]
+        session.set_pointer_barriers(barriers)
+        session.connect_to_eis()
+
+        activated_signal_received = False
+
+        def cb_activated(session_handle, options):
+            nonlocal activated_signal_received
+            activated_signal_received = True
+
+        inputcapture_intf.connect_to_signal("Activated", cb_activated)
+
+        session.enable()
+
+        xdp.wait_for(lambda: activated_signal_received)
+
+        return (
+            session.session,
+            start_session_response.results.get("clipboard_enabled"),
+        )
+
+    def start_session(self, dbus_con, type):
+        if type == SessionType.REMOTE_DESKTOP:
+            return self.start_remote_desktop_session(dbus_con)
+        if type == SessionType.INPUT_CAPTURE:
+            return self.start_input_capture_session(dbus_con)
+        raise Exception("Unknown type")
+
+    def test_request_clipboard_and_start_session(self, portals, dbus_con, type):
+        _, clipboard_enabled = self.start_session(dbus_con, type)
 
         assert clipboard_enabled
 
     @pytest.mark.parametrize(
-        "template_params", ({"RemoteDesktop": {"force-clipboard-enabled": False}},)
+        "template_params",
+        (
+            {
+                "RemoteDesktop": {"force-clipboard-enabled": False},
+                "InputCapture": {
+                    "force-clipboard-enabled": False,
+                    "activated-delay": 100,
+                },
+            },
+        ),
     )
-    def test_checks_clipboard_enabled(self, portals, dbus_con):
+    def test_checks_clipboard_enabled(self, portals, dbus_con, type):
         clipboard_intf = xdp.get_portal_iface(dbus_con, "Clipboard")
-        session, clipboard_enabled = self.start_session(dbus_con)
+        session, clipboard_enabled = self.start_session(dbus_con, type)
 
         assert not clipboard_enabled
 
         with pytest.raises(dbus.exceptions.DBusException):
             clipboard_intf.SetSelection(session.handle, {})
 
-    def test_set_selection(self, portals, dbus_con):
+    def test_set_selection(self, portals, dbus_con, type):
         clipboard_intf = xdp.get_portal_iface(dbus_con, "Clipboard")
-        session, _ = self.start_session(dbus_con)
+        session, _ = self.start_session(dbus_con, type)
 
         clipboard_intf.SetSelection(session.handle, {})
 
-    def test_selection_write(self, portals, dbus_con):
+    def test_selection_write(self, portals, dbus_con, type):
         clipboard_intf = xdp.get_portal_iface(dbus_con, "Clipboard")
-        session, _ = self.start_session(dbus_con)
+        session, _ = self.start_session(dbus_con, type)
 
         fd_object: dbus.types.UnixFd = clipboard_intf.SelectionWrite(
             session.handle, 1234
@@ -87,9 +153,9 @@ class TestClipboard:
 
         clipboard_intf.SelectionWriteDone(session.handle, 1234, True)
 
-    def test_selection_read(self, portals, dbus_con):
+    def test_selection_read(self, portals, dbus_con, type):
         clipboard_intf = xdp.get_portal_iface(dbus_con, "Clipboard")
-        session, _ = self.start_session(dbus_con)
+        session, _ = self.start_session(dbus_con, type)
 
         fd_object: dbus.types.UnixFd = clipboard_intf.SelectionRead(
             session.handle, "mimetype"


data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +116,87 @@
...
+
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateSession2">
```
jadahl, line 159: The agreed upon name in the PR this was based on is `CreateDraftSession`, which I think is a better name.

swick: Sorry, but the other portals use CreateSession; even for non-started sessions like is the case here. Inventing new names for everything just makes things more confusing for no good reason.

jadahl: This portal's CreateSession works the way it works, having a CreateSession2 that behaves fundamentally different just makes things more confusing for no good reason. Making the name clear within this portal avoids that confusion.

swick: I strongly disagree. The CreateSession in the remote desktop portal behaves exactly like the CreateSession2 here. If you chose the name this `CreateDraftSession`, then why is the exact same thing in remote desktop called `CreateSession`? Either way, calling it  `CreateDraftSession` or `CreateSession2`, the user of the method has to read the description to understand it anyway, so I don't accept that as an argument.

swick: It's also not fundamentally different. The only difference is that `CreateSession2` starts in the INIT state while the old `CreateSession` starts in the STARTED state. That's it.

jadahl: I don't see "{Method}2" as being a good way to handle it unless it really is doing the same thing with more context or different arguments. I also think having the names being obvious when reading the documentation to have more weight than being consistent with other portals, especially since there are `CreateSession` methods in portals that behave more close to how `CreateSession` in the input capture portal behaves, e.g. global shortcuts, so that argument doesn't hold anyway.

swick: The other thing that I find extremely irritating about `CreateDraftSession` is that it implies that there is a concept of a `DraftSession`. There isn't. It creates a session, the session is represented as a dbus object, it has the same interface. This is literally the same as everywhere else. The behavior of the sessions is different depending on the portal, and some start "doing" something useful straight away, while others need a method to be called before.

But at no point is there a DraftSession.

> I don't see "{Method}2" as being a good way to handle it unless it really is doing the same thing with more context or different arguments.

The way I think about it is: if we don't have to care about backwards compat, would you just change the behavior of $method or would you rename it? In this case, we would obviously just change the behavior and not rename it. Therefor "{Method}2" is the correct way to handle it.

gabeklavans: Disclaimer: I'm not a portal dev and have little context

The name `CreateSession2` is helped by the fact that `CreateSession` seems to be deprecated by this version 2 of the interface. It's fairly reasonable to just increment a version number for a function in an interface library as the version of the interface goes up and you want to retain backwards compat.

`CreateDraftSession` to me implies that `CreateSession` still has use after this version is released (does it?)

sithlord48: Fwiw, That name `CreateSession2` was suggested when i was working on updating the xml.

AdrianVovk: I think I agree with Sebastian here. `CreateDraftSession` sounds like this is creating something called a "draft session" which doesn't exist. And this isn't doing something fundamentally different from `CreateSession` anyway

src/input-capture.c
```diff
@@ -126,47 +124,196 @@ input_capture_session_new (GVariant         *options,
...
+static gboolean
+validate_capabilities (const char  *key,
+                       GVariant    *value,
+                       GVariant    *options,
+                       GError     **error)
 {
-  g_autoptr(XdpRequest) request = data;
+  uint32_t types = g_variant_get_uint32 (value);
+
+  if ((types & ~(1 | 2 | 4 | 8)) != 0)
```
AdrianVovk: I know you didn't write this code like this and just moved it around in the file, but while you're here could you disambiguate what 1, 2, 4, and 8 mean? And maybe put the actual ORing of all the possible values together next to the place where the values are defined so it's easier to remember to maintain? i.e. have CAPABILITIY_* enum and a CAPABILITY_ALL defined so that you can write code like `types & ~CAPABILITIY_ALL` instead of what's there now...

src/input-capture.c
```diff
@@ -126,47 +124,196 @@ input_capture_session_new (GVariant         *options,
...
+      return G_DBUS_METHOD_INVOCATION_HANDLED;
+    }
+  options = g_variant_ref_sink (g_variant_builder_end (&options_builder));
+
+  g_object_set_qdata_full (G_OBJECT (request),
+                           quark_request_session,
+                           g_object_ref (session),
+                           g_object_unref);
+
+  xdp_dbus_impl_input_capture_call_start (impl,
```
AdrianVovk: Backwards compat here doesn't work if the portal impl is using v1 of this interface. We call `Start()` which didn't exist yet in v1. Clearly we do care about this kind of backwards compat, because we do return v1 as the frontend's version if the backend only supports v1. So, this is an oversight

The fix is probably to check `impl_version` and call the impl's `CreateSession()` instead of `Start()`

src/input-capture.c
```diff
@@ -175,6 +322,19 @@ create_session_done (GObject      *source_object,
           goto out;
         }
 
+      xdp_session_register (session);
+    }
+
+  if (request->exported && response == 0)
```
AdrianVovk: I know this isn't entirely new code, but I think this is a good opportunity to port this code to `XdgDesktopPortalResponseEnum`, rather than using the magic response numbers

src/input-capture.c
```diff
@@ -124,6 +112,36 @@ input_capture_session_new (GVariant         *options,
...
+gboolean
+input_capture_session_can_request_clipboard (InputCaptureSession *session)
+{
+  if (session->clipboard_requested)
+    return FALSE;
+
+  if (impl_version < 2)
+    return FALSE;
+
+  switch (session->state)
```
AdrianVovk, line 170: Why a switch case when we're checking just one condition?

`return session->state == INPUT_CAPTURE_SESSION_STATE_INIT`

src/input-capture.h
```diff
@@ -23,5 +23,27 @@
 #include <gio/gio.h>
 #include <stdint.h>
 
+GType input_capture_session_get_type (void);
```
AdrianVovk, line 25: Why aren't we using one of the gobject macros for this?

AdrianVovk (changes requested):
Mostly nits, but there is a backwards compat break: `CreateSession` won't work with a v1 backend

src/input-capture.c
swick: We only ever defined 3 caps, so 1, 2, 4. No idea where the 8 is coming from. I even checked in mutter if we added something but forgot to add it to the docs here. Doesn't seem to be the case.

src/input-capture.h
swick: The classes in old portals do not have a prefix, which is required for the macros. I *should* port them towards having the xdp prefix, but given that this would involve a lot of noise, I think adding this manually should be fine for this PR.

src/input-capture.c
swick: This was copied from remote desktop and there it uses a switch as well. It actually makes sense because if there is a new state, it forces you to look at all the places where the state matters. Using default however completely defeated that purpose, so I added the other states here (just like in remote desktop).

src/input-capture.c
swick: Yup, good point.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +76,66 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <!--
```
jadahl, line 105: Missing the `CreateSession2` variant in the impl.portal.

swick: Is there a need for it? The impl sees a call which creates the session, a call to request clipboard on the session, and a call to start.

jadahl: Yes it's needed; and what call are you referring to? Right now it's the call to `imp.portal...CreateSession` that actually creates the `org.freedesktop.impl.portal.Session` object (indirectly), but can't use that.

swick: Uh, yeah, forgot that the session doesn't call into an impl. Added it, and did a painful rebase as well.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +76,86 @@
...
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @app_id: App id of the application
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Start the remote desktop session. This will typically result in the portal
+        presenting a dialog letting the user decide whether they want to allow the
+        input of the session to be or not, and what capabilities to support.
```
Sodivad: needs rewording, looks like it was copied from remote desktop.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +116,93 @@
...
+    <!--
+        Start:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @options: Vardict with optional further information
+        @handle: Object path for the :ref:org.freedesktop.portal.Request object representing this call
+
+        Start the remote desktop session. This will typically result in the portal
+        presenting a dialog letting the user decide whether they want to allow the
+        input of the session to be or not, and what capabilities to support.
```
Sodivad: dito

data/org.freedesktop.impl.portal.InputCapture.xml
jadahl: Indeed, and the impl session (in xdp-gnome) is typed per session type, so a generic constructor without context wouldn't work anyway.

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +76,80 @@
...
+
+        There are currently no supported keys in the @options vardict.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateSession2">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
```
jadahl, line 100: Missing `results`; might not be useful now, but the others have it, and the session handle is returned via it. The session handle can be derived without it, but its harmless to add, and useful if we need to return some value in the future.

data/org.freedesktop.impl.portal.InputCapture.xml
sithlord48: I maybe did copy it , i was told when i started the (in the pr this is based on) that all that had to happen was copy the parts clipboard parts from remote desktop to input capture . When i did the discussion about new names etc started. sorry if this was incorrect .

swick: No worries, just the name of the portal and a typo. This stuff happens.

data/org.freedesktop.impl.portal.InputCapture.xml
swick: good point

jadahl: What I meant was actually about `CreateSession2`. I.e. this change:

```patch
diff --git a/data/org.freedesktop.portal.InputCapture.xml b/data/org.freedesktop.portal.InputCapture.xml
index 68d2af94..2b927877 100644
--- a/data/org.freedesktop.portal.InputCapture.xml
+++ b/data/org.freedesktop.portal.InputCapture.xml
@@ -137,12 +137,20 @@
           object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
           more information about the session handle.
 
+        Supported keys in the @results vardict may include:
+
+        * ``session_handle`` (``o``)
+
+          The session handle. An object path for the
+          :ref:`org.freedesktop.portal.Session` object representing the created
+          session. Required.
+
         This method was added in version 2 of this interface.
     -->
     <method name="CreateSession2">
       <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
-      <arg type="o" name="session_handle" direction="out"/>
     </method>
 
     <!--
diff --git a/src/input-capture.c b/src/input-capture.c
index c4fbaea3..978fcaf5 100644
--- a/src/input-capture.c
+++ b/src/input-capture.c
@@ -349,6 +349,8 @@ handle_create_session2 (XdpDbusInputCapture   *object,
   g_autoptr(GVariant) options = NULL;
   g_autoptr(GVariant) results = NULL;
+  g_auto(GVariantBuilder) results_builder =
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
 
   if (input_capture->impl_version < 2)
     return G_DBUS_METHOD_INVOCATION_UNHANDLED;
@@ -396,9 +398,11 @@ handle_create_session2 (XdpDbusInputCapture   *object,
 
   xdp_session_register (session);
 
+  g_variant_builder_add (&results_builder, "{sv}",
+                         "session_handle", g_variant_new ("o", session->id));
   xdp_dbus_input_capture_complete_create_session2 (object,
                                                    invocation,
-                                                   session->id);
+                                                   g_variant_builder_end (&results_builder));
 
 }
```

swick:
> Some crashes:

Fixed the autoptr, must have screwed up the rebase.

data/org.freedesktop.impl.portal.InputCapture.xml
swick: All other CreateSession methods return the session handle only, so this introduces irregularity. Added a commit on top which has all the required changes for it. I see the point here though.

jadahl: AFAICS all other but the USB portal responds using a vararg via a request response signal, but since this is just a method, to do the equivalent is to make it return one directly instead.

swick: location portal as well, but yeah, I will squash the changes in

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -62,6 +62,8 @@
         @options: Vardict with optional further information
         @handle: Object path for the :ref:`org.freedesktop.portal.Request` object representing this call
 
+        This method was deprecated in version 2 of this interface.
```
whot:
```suggestion
        This method was deprecated in version 2 of this interface, see CreateSession2 instead.
```

whot (commented):
only one nitpick I could find

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +76,90 @@
...
+        There are currently no supported keys in the @results vardict.
+
+        This method was added in version 2 of this interface.
+    -->
+    <method name="CreateSession2">
+      <arg type="o" name="session_handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In2" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="a{sv}" name="results" direction="out"/>
```
Sodivad, line 102: All other `CreateSession` seem to also return a `u response`

swick: Most do, yes, because they are called in the context of a Request but this one is not.

Sodivad: Right makes sense.

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +116,101 @@
...
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        * ``clipboard_enabled`` (``b``)
+
+          A boolean for whether the clipboard was enabled ('true') or not('false').
+          See the :ref:`org.freedesktop.portal.Clipboard` documentation for more
+          information.
```
Sodivad, line 212: We need to touch up clipboard docs  to mention input capture and/or this otherwise you are stuck in a referencing loop as clipboard currently reads:


> This portal does NOT create it’s own session. Instead, it offers existing sessions created from other portals the option to integrate with the clipboard. For whether this interface is supported for a given session, refer to that portal’s documentation. See Remote Desktop  to integrate clipboard with the remote desktop session.

Sodivad:
Should changing the clipboard via the clipboard portal be allowed as long as the app has an input capture session or should it be restricted to be only be possible when the input capture is active? Or should this be backend policy?

src/input-capture.c
whot: >  No idea where the 8 is coming from

Finally remembered: early version of this portal had absolute/relative events separately, so with keyboard, touch, relative and absolute we had four possible bits.

Then rel/abs got merged into just pointer but this code here didn't get updated.

tests/templates/inputcapture.py
```diff
@@ -40,6 +40,7 @@ class InputcaptureParameters:
     activated_delay: int
     deactivated_delay: int
     zones_changed_delay: int
+    force_clipoboard_enabled: bool
```
whot:
```suggestion
    force_clipboard_enabled: bool
```

tests/templates/inputcapture.py
```diff
@@ -55,6 +56,7 @@ def load(mock, parameters={}):
         activated_delay=parameters.get("activated-delay", 0),
         deactivated_delay=parameters.get("deactivated-delay", 0),
         zones_changed_delay=parameters.get("zones-changed-delay", 0),
+        force_clipoboard_enabled=parameters.get("force-clipboard-enabled", False),
```
whot:
```suggestion
        force_clipboard_enabled=parameters.get("force-clipboard-enabled", False),
```

tests/templates/inputcapture.py
```diff
@@ -96,10 +118,13 @@ def CreateSession(self, handle, session_handle, app_id, parent_window, options):
         capabilities &= params.supported_capabilities
         response = Response(0, {})
 
+        if params.force_clipoboard_enabled:
```
whot:
```suggestion
        if params.force_clipboard_enabled:
```

src/input-capture.c
```diff
@@ -125,92 +125,381 @@ static InputCaptureSession *
...
+    {
+    case INPUT_CAPTURE_SESSION_STATE_INIT:
+      return TRUE;
+     case INPUT_CAPTURE_SESSION_STATE_STARTED:
+     case INPUT_CAPTURE_SESSION_STATE_ENABLED:
+     case INPUT_CAPTURE_SESSION_STATE_ACTIVE:
+     case INPUT_CAPTURE_SESSION_STATE_DISABLED:
+     case INPUT_CAPTURE_SESSION_STATE_CLOSED:
+      return FALSE;
```
whot, line 179:
```suggestion
    case INPUT_CAPTURE_SESSION_STATE_INIT:
      return TRUE;
    case INPUT_CAPTURE_SESSION_STATE_STARTED:
    case INPUT_CAPTURE_SESSION_STATE_ENABLED:
    case INPUT_CAPTURE_SESSION_STATE_ACTIVE:
    case INPUT_CAPTURE_SESSION_STATE_DISABLED:
    case INPUT_CAPTURE_SESSION_STATE_CLOSED:
      return FALSE;
```
(indentation)

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -72,6 +76,90 @@
...
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        * ``clipboard_enabled`` (``b``)
+
+          A boolean for whether the clipboard was enabled ('true') or not('false').
```
whot:
```suggestion
          A boolean for whether the clipboard was enabled ('true') or not ('false').
```

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +116,101 @@
...
+          guarantee a device with that capability is currently available
+          or if one does become available that it will trigger input capture.
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        * ``clipboard_enabled`` (``b``)
+
+          A boolean for whether the clipboard was enabled ('true') or not('false').
```
whot:
```suggestion
          A boolean for whether the clipboard was enabled ('true') or not ('false').
```

tests/templates/inputcapture.py
```diff
@@ -72,19 +74,39 @@ def load(mock, parameters={}):
...
+    out_signature="a{sv}",
+)
+def CreateSession2(self, session_handle, app_id, options):
+    try:
+        logger.debug(f"CreateSession2({session_handle}, {app_id}, {options})")
+
+        assert len(options) == 0
+
+        self.session_handles.append(session_handle)
+        return {}
```
whot: doesn't this need to return the session handle?

whot (commented):
These changes now available in #1897, free for the taking

data/org.freedesktop.portal.InputCapture.xml
swick: I think it's fine to reference clipboard in input capture, and input capture in clipboard. Added input capture to the clipboard docs.

sithlord48:
> > GNOME and KDE have impls, so I think this should be ready to merge. Will do so in a few days if no one speaks up.
> 
> There are no client side implementations yet that I know of, though. Then again, it's the same as remote desktop, so maybe it's worth the risk of merging something even if it it is effectively untested.

we have a stage for this in deskflow you need to build a few parts to get it working https://github.com/deskflow/deskflow/pull/9431 (access is via libportal) it only works server -> client since we have not hooked into the remote desktop clipboard yet but that will be added to the linked pr in time.

Sodivad:
I dont see the problem, you can update clipboard when input switches

> Copy something with the cursor on the host then move "offscreen" to a client to paste.

when you enter "offscreen" (input capture actuvates) you can fetch the latest clipbpoard from the portal. Or only when the past is requested

> Copy something with the cursor on a client then move "onscreen" to the host to paste.

when you are on the client the input capture is active

> Copy something with the cursor on a client to another client (clipboard is synced thru the server)

if one of the clients is using input capture portal then there is no problem as well, the input capture is active on copy and/or after move

data/org.freedesktop.portal.Clipboard.xml
```diff
@@ -35,13 +35,18 @@
...
+        must be made before the session starts. The session must be started,
+        before using any other method in this interface which take a session.
 
         This portal does NOT create it's own session. Instead, it offers existing sessions
-        created from other portals the option to integrate with the clipboard. For whether
-        this interface is supported for a given session, refer to that portal's documentation.
+        created from other portals the option to integrate with the clipboard.
+
+        For whether this interface is supported for a given session, refer to that portal's
+        documentation.
```
Sodivad, line 47:
```suggestion
        documentation. Note that other interfaces might place restriction on when it's possible to interact with the clipboard.
```

data/org.freedesktop.portal.InputCapture.xml
```diff
@@ -114,6 +117,107 @@
...
+
+          It is best to view this set as a negative confirmation - a
+          capability that was requested but is missing is an indication that
+          this application may not capture events of that capability.
+
+        * ``clipboard_enabled`` (``b``)
+
+          A boolean for whether the clipboard was enabled ('true') or not ('false').
+          See the :ref:`org.freedesktop.portal.Clipboard` documentation for more
+          information.
```
Sodivad, line 212:
```suggestion
          information on how to enable clipboard support. When clipboard access has been granted the clipboard may only be accesse when the session is active.
```

PR #1804, opened 2025-09-05
portals: Make the permission dialogs a bit more consistent

swick:


jadahl:
Who might be the resident designer who could put a stamp of approval of the changes? I guess a before & after would be necessary for that as well, since it's a bit hidden in the diff what the actual string changes are.

swick:
Thanks for the review and for linking to the hig. I adjusted everything accordingly.

PR #1805, opened 2025-09-05
Split the entrpoint into main and context

swick:
Includes https://github.com/flatpak/xdg-desktop-portal/pull/1771


src/xdp-main.c
```diff
@@ -0,0 +1,249 @@
...
+  g_setenv ("GIO_USE_VFS", "local", TRUE);
+
+  /* Avoid pointless and confusing recursion */
+  g_unsetenv ("GTK_USE_PORTAL");
+
+  option_context = g_option_context_new ("- desktop portal");
+  g_option_context_set_summary (option_context,
+      "A portal service for flatpak and other desktop containment frameworks.");
+  g_option_context_set_description (option_context,
+      "xdg-desktop-portal works by exposing D-Bus interfaces known as portals\n"
```
jadahl: Wrong indentation.

src/location.c
```diff
@@ -184,12 +183,6 @@ location_updated (GeoclueClient *client,
 
   g_variant_get (ret, "(@a{sv})", &dict);
 
-  if (opt_verbose)
-    {
-      g_autofree char *a = g_variant_print (dict, FALSE);
-      g_debug ("location data: %s\n", a);
-    }
-
```
jadahl, line 192: Unrelated.

src/xdp-main.c
```diff
@@ -0,0 +1,249 @@
...
+                 const gchar     *name,
+                 gpointer         user_data)
+{
+  XdpMain *xdp_main = user_data;
+  g_autoptr(GError) error = NULL;
+
+  if (!xdp_context_register (xdp_main->context, connection, &error))
+    {
+      g_critical ("Starting portals failed: %s", error->message);
+      xdp_main_exit (xdp_main, 1);
```
jadahl: While at it, can replace these with `EXIT_SUCCESS` or `EXIT_FAILURE`.

src/xdp-main.c
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+static void
+on_name_acquired (GDBusConnection *connection,
+                  const gchar     *name,
```
jadahl: Could clean up `gchar` -> `char` while this file is being rewritten.

src/xdp-main.c
swick: What's wrong about it? If the opening bracket of the function call is too far to the right that the arguments all exceed the line length limit, one has to put them there.

src/location.c
swick: Not really. the global verbose option gets dropped and getting the XdpContext to here is not a trivial change.

jadahl: If you mean to keep this change, then I suggest a separate commit that says "location: Remove location data debug logging" or something.

src/xdp-main.c
jadahl: I assume it should be 2 spaces indentation, and not 4.

swick:
Adjusted to the feedback, and rebased on the changed in the PR this one is based on.

PR #1807, opened 2025-09-10
document-portal: use default FUSE implementation for statfs

### LINKED ISSUE #553: df fails on /run/user/1000/doc with "Operation not permitted"
Hi,

this is a followup to issue #512 that has become confused by a root/non-root sub-issue. The problem here is that `df` displays an error message for something that is not an error:
```
$ df
/bin/df: /run/user/1000/doc: Operation not permitted
Filesystem                          1K-blocks      Used Available Use% Mounted on
tmpfs                                 1624464     31012   1593452   2% /run
[...]
```
Here are xdg-desktop-portal processes that run on this system (note that there are no processes owned by root contrary to the other issue I mentioned):
```
$ ps aux | grep xdg
bonnaudl  149811  0.0  0.0  80728  6156 ?        Ssl  09:44   0:00 /usr/libexec/xdg-document-portal
bonnaudl  149814  0.0  0.0  56024  5852 ?        Ssl  09:44   0:00 /usr/libexec/xdg-permission-store
bonnaudl  149834  0.0  0.1 148672 18196 ?        Ssl  09:44   0:00 /usr/libexec/xdg-desktop-portal
bonnaudl  149838  0.0  0.1  95328 20860 ?        Ssl  09:44   0:00 /usr/libexec/xdg-desktop-portal-gtk
bonnaudl  149847  0.0  0.4 301700 73644 ?        Sl   09:44   0:00 /usr/lib/x86_64-linux-gnu/libexec/xdg-desktop-portal-kde
```
Here is another relevant process:
```
root      149819  0.0  0.0   4684  2996 ?        Ss   09:44   0:00 fusermount -o rw,nosuid,nodev,fsname=portal,auto_unmount,subtype=portal -- /run/user/1000/doc
```

Is this problem in xdg-desktop-portal, fuse or df ?  I think that at least one of those 3 pieces of software should be modified so that the error message is not displayed.

ghost:
For a little shy of 5 years, the `xdg-document-portal` implementation of the `statfs` syscall has been broken and returned EPERM. The default FUSE implementation returns a sensible default value, similar to the `/proc` and `/sys` mount points. This merge request comments out the broken implementation in favor of FUSE's default, and fixes #553.

I understand this is not an ideal implementation of this FUSE interface. On the other hand, this change breaks no test, and allows `df -a` (and any other program using the statfs syscall) to run in harmony with `xdg-document-portal`.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
 
   g_debug ("STATFS %" G_GINT64_MODIFIER "x", ino);
 
+  if (xdp_domain_is_virtual_type (inode->domain))
+    {
+      /* statfs data doesn't make much sense for a virtual filesystem, so we reply with
+       * a mostly empty buffer, mimicking an empty .statfs. This is consistent with what
+       * libfuse does if we don't supply a statfs implementation. */
+      buf = (struct statvfs) {
+        .f_namemax = 255,
+        .f_bsize = 512,
+      };
+      fuse_reply_statfs (req, &buf);
+      return;
+    }
+
   if (!xdp_document_inode_checks (op, req, inode, 0))
     return;
 


document-portal/document-portal-fuse.c
```diff
@@ -3458,7 +3458,7 @@ static struct fuse_lowlevel_ops xdp_fuse_oper = {
  .symlink      = xdp_fuse_symlink,
  .link         = xdp_fuse_link,
  .flush        = xdp_fuse_flush,
- .statfs       = xdp_fuse_statfs,
+ /* .statfs       = xdp_fuse_statfs, */
```
smcv: What information does the default implementation provide?

If fuse's default implementation does everything we need, then please delete this instead of commenting it out, and delete the now-unused implementation of `xdp_fuse_statfs()`. There's no point in keeping dead code in-tree, it's just a source of compiler warnings: we can always get it back from the git history if we need it.

smcv: But... would the default implementation allow a sandboxed app to distinguish between the existence and nonexistence of a document that it does not have access to? If it would, then using the default implementation would be an information disclosure vulnerability.

SlonkeyMaster:
The default implementation of fuse uses `memset` too to create the dummy data for `statfs`, so this is concordant, and doesn't break other userspace tools. This should fix #553.

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,12 @@ xdp_fuse_statfs (fuse_req_t req,
 
   g_debug ("STATFS %" G_GINT64_MODIFIER "x", ino);
 
+  if(xdp_domain_is_virtual_type(inode->domain)) {
+	memset(&buf, 0, sizeof(buf));
+	fuse_reply_statfs (req, &buf);
+	return;
+  }
```
smcv: Please indent consistently with the surrounding code:

```suggestion
  if (xdp_domain_is_virtual_type (inode->domain))
    {
      memset (&buf, 0, sizeof (buf));
      fuse_reply_statfs (req, &buf);
      return;
    }
```

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,12 @@ xdp_fuse_statfs (fuse_req_t req,
 
   g_debug ("STATFS %" G_GINT64_MODIFIER "x", ino);
```
smcv, line 3038: I think this deserves a comment as to why this is appropriate, perhaps something like:

```suggestion

  /* Free space, etc. don't make much sense for a virtual filesystem,
   * so reply with a mostly empty buffer. This is consistent with what libfuse does
   * if we don't supply a statfs implementation. */
```

smcv: (but obviously that comment is misleading if what we do here is not consistent with what libfuse would have done without a statfs implementation!)

swick:
Hey @SlonkeyMaster, are you going to pick this PR up again?

document-portal/document-portal-fuse.c
ghost: Added this.

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
...
 
+  if (xdp_domain_is_virtual_type (inode->domain))
+    {
+      struct statvfs buf = {
+	.f_namemax = 255,
+	.f_bsize = 512,
+      };
```
smcv: Something has gone wrong with the indentation here. (Tabs?)

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
...
+      fuse_reply_statfs(req, &buf);
+      return;
+    }
+  /* statfs data doesn't make much sense for a virtual filesystem, so we reply with a
+   * mostly empty buffer, mimicking an empty .statfs. This is consistent with what
+   * libfuse does if we don't supply a statfs implementation. */
```
smcv: This comment is describing what happens for virtual files, so it would make more sense inside the `if (xdp_domain_is_virtual_type)` block.

document-portal/document-portal-fuse.c
ghost: Indeed.

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
 
   g_debug ("STATFS %" G_GINT64_MODIFIER "x", ino);
 
+  if (xdp_domain_is_virtual_type (inode->domain))
+    {
+      /* statfs data doesn't make much sense for a virtual filesystem, so we reply with
+       * a mostly empty buffer, mimicking an empty .statfs. This is consistent with what
+       * libfuse does if we don't supply a statfs implementation. */
+      struct statvfs buf = {
```
swick: Was already declared above.

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
...
+        .f_namemax = 255,
+        .f_bsize = 512,
+      };
+      fuse_reply_statfs(req, &buf);
```
swick:
```suggestion
      fuse_reply_statfs (req, &buf);
```

document-portal/document-portal-fuse.c
swick: Well, now you have uninitialized parts of the struct. Either zero initialize it in the beginning or use a designated initializer here.

swick: Sorry, but now you have disconnected the values from the code path where they are used. Just use

```C
  buf = (struct statvfs) {
    .f_namemax = 255,
    .f_bsize = 512,
  };
```

document-portal/document-portal-fuse.c
```diff
@@ -3036,6 +3036,19 @@ xdp_fuse_statfs (fuse_req_t req,
...
   g_debug ("STATFS %" G_GINT64_MODIFIER "x", ino);
 
+  if (xdp_domain_is_virtual_type (inode->domain))
+    {
+      /* statfs data doesn't make much sense for a virtual filesystem, so we reply with
+       * a mostly empty buffer, mimicking an empty .statfs. This is consistent with what
+       * libfuse does if we don't supply a statfs implementation. */
+      buf = (struct statvfs) {
+	.f_namemax = 255,
+	.f_bsize = 512,
```
GeorgesStavracas: Nitpick here: tabs → spaces

ghost: alright, fixed them

PR #1810, opened 2025-09-12
app-info: Adjust to xdp_app_info_new consuming the pidfd in

swick:
Let's steal the fd so it becomes absolutely clear what's going on there, and then get a new fd for xdp_app_info_host_new_registered.

### DIFF
--- a/src/xdp-app-info.c
+++ b/src/xdp-app-info.c
@@ -1035,6 +1035,7 @@ xdp_invocation_register_host_app_info_sync (GDBusMethodInvocation  *invocation,
   g_autoptr(XdpAppInfo) detected_app_info = NULL;
   g_autoptr(XdpAppInfo) app_info = NULL;
   g_autofd int pidfd = -1;
+  g_autofd int pidfd_dup = -1;
   uint32_t pid;
 
   if (cache_has_app_info_by_sender (sender))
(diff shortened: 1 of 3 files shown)


src/xdp-app-info.c
```diff
@@ -1047,7 +1047,7 @@ xdp_invocation_register_host_app_info_sync (GDBusMethodInvocation  *invocation,
   if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
     return NULL;
 
-  detected_app_info = xdp_app_info_new (pid, pidfd, error);
+  detected_app_info = xdp_app_info_new (pid, g_steal_fd (&pidfd), error);
```
jadahl, line 1063: Why not just dup it instead of getting it twice...

src/xdp-app-info.c
```diff
@@ -1059,6 +1059,9 @@ xdp_invocation_register_host_app_info_sync (GDBusMethodInvocation  *invocation,
       return NULL;
     }
 
+  if (!xdp_connection_get_pidfd (connection, sender, cancellable, &pidfd, &pid, error))
+    return NULL;
+
   app_info = xdp_app_info_host_new_registered (&pidfd, app_id, error);
```
jadahl: Then steal it here instead.

Alternatively do what `.._registerd()` does and take a `int *fd` in `xdp_app_info_new()` that steals the fd from the source only on success paths.

swick: Yeah, changed it to consume the fd.

PR #1814, opened 2025-09-19
secret: Init asynchronously

jadahl:
When run in a nested D-Bus session, the secret portal backend tends to fail to launch due to another secret service running within the same $XDG_RUNTIME_DIR. Handle this without blocking by initializing the secret portal asynchronously.

-----

Not sure we want to do this, but currently the secret portal is causing annoyingly long timeouts causing when the portal is run in a nested D-Bus session using dbus-run-session.

Might be sensible to eventually initialize all portal backends asynchronously, but I suspect that is more feasible with libdex.

I'm not sure it's feasible to fix the secret backend (gnome-keyring), without significantly changing how it works (i.e. without `$XDG_RUNTIME_DIR/keyring`, or muliple ones, I imagine), or switching to a completely different backend.

### DIFF
--- a/src/secret.c
+++ b/src/secret.c
@@ -192,26 +192,50 @@ secret_class_init (SecretClass *klass)
 }
 
 GDBusInterfaceSkeleton *
-secret_create (GDBusConnection *connection,
-	       const char      *dbus_name)
+secret_create_finish (GAsyncResult  *result,
+                      GError       **error)
 {
+  GTask *task = G_TASK (result);
+
+  return g_task_propagate_pointer (task, error);
+}
+
+static void
+proxy_created_cb (GObject      *source_object,
+                  GAsyncResult *result,
+                  gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
   g_autoptr(GError) error = NULL;
 
-  impl = xdp_dbus_impl_secret_proxy_new_sync (connection,
-                                              G_DBUS_PROXY_FLAGS_NONE,
-                                              dbus_name,
-                                              DESKTOP_PORTAL_OBJECT_PATH,
-                                              NULL,
-                                              &error);
-  if (impl == NULL)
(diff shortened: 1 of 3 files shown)


src/secret.c
```diff
@@ -192,26 +192,50 @@ secret_class_init (SecretClass *klass)
...
+secret_create_async (GDBusConnection     *connection,
+                     const char          *dbus_name,
+                     GCancellable        *cancellable,
+                     GAsyncReadyCallback  callback,
+                     gpointer             user_data)
+{
+  g_autoptr(GTask) task = NULL;
 
-  return G_DBUS_INTERFACE_SKELETON (secret);
+  task = g_task_new (NULL, cancellable, callback, user_data);
```
3v1n0: Bind the task to the connection too?

src/secret.c
```diff
@@ -192,26 +192,50 @@ secret_class_init (SecretClass *klass)
...
+
+  return g_task_propagate_pointer (task, error);
+}
+
+static void
+proxy_created_cb (GObject      *source_object,
+                  GAsyncResult *result,
+                  gpointer      user_data)
+{
+  g_autoptr(GTask) task = G_TASK (user_data);
```
3v1n0: Steal it?

src/secret.c
```diff
@@ -192,26 +192,50 @@ secret_class_init (SecretClass *klass)
...
-      g_warning ("Failed to create secret proxy: %s", error->message);
-      return NULL;
+      g_task_return_error (task, g_steal_pointer (&error));
+      return;
     }
 
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (impl), G_MAXINT);
 
   secret = g_object_new (secret_get_type (), NULL);
+  g_task_return_pointer (task, secret, g_object_unref);
```
3v1n0: Steal too l

Issue #1416, opened 2024-08-10
Memory leak


war1025:
### Operating System

Debian

### XDG Desktop Portal version

1.18

### Desktop Environment

GNOME

### Expected Behavior

I would expect this to use a very minimal amount of ram.

### Current Behavior

This is consistently the most memory intensive process on my machine.
Killing it does absolutely nothing bad as far as I can tell, so whatever it is up to can't be that critical for it to be using so much ram, and when it restarts it is back to minimal memory usage.
This is why I assume it's just leaking like a sieve somewhere.

![Screenshot from 2024-08-10 01-36-11](https://github.com/user-attachments/assets/58fa05f0-1148-4a9b-94dc-af3c7ab42a5c)

### Steps to Reproduce

Use my computer for a day.

GeorgesStavracas:
Lots of memory leaks were fixed in the `main` branch, can you please test it?

hfiguiere:
is this `xdg-desktop-portal` or any of the implementations like `xdg-desktop-portal-gtk` or `xdg-desktop-portal-gnome`

`xdg-desktop-portal-gnome` has some known issues, but it's not this project.

A6GibKm:
Does this issue persist? You can use

    ps aux |grep xdg-desktop-por
    
 To see to which `xdg-desktop-por`... the PID comes from.

swick:
We support asan builds `-Db_sanitize=address -Db_lundef=false`. You can try running such a build and capture the output.

aliaksandr-trush:
Hi @swick ,
I've tried to run with asan build, but didn't receive any useful information: https://gist.github.com/aliaksandr-trush/03001b3c8095011cc471253635c751f2 (I am not sure that I've run build correctly. Let me know if I need to do it in another way).
And xdg-desktop-portal process was killed by OOM Killer (process consumer > 12 Gb or memory)

hfiguiere:
You only get the asan leaks report when xdg-desktop-portal terminates. `kill -HUP` on the xdg-desktop-portal process should trigger it.

swick:
You can introduce a problem on purpose (e.g. in main `char *xxx = (char*)malloc(10 * sizeof(char*));`) and see if you get the asan output.

If that's the case, then maybe `cat /proc/$PID/maps` or `ls -la /proc/$PID/fd/` might help, because it does sound like some other resource is getting lost.

swick:
[...]

PR #1815, opened 2025-09-21
Use XdpContext in more places and add a XdpAppInfoRegistry

swick:


src/xdp-app-info.c
```diff
@@ -1000,8 +1000,6 @@ xdp_connection_create_app_info_sync (GDBusConnection  *connection,
 
   cache_insert_app_info (sender, app_info);
 
-  xdp_connection_track_name_owners (connection, on_peer_died);
```
swick, line 1030: Turns out that this is a pretty bad thing to do. It subscribes to a signal every time a new client connects and never disconnects it. It fixes https://github.com/flatpak/xdg-desktop-portal/issues/1416.

tests/asan.suppression
```diff
@@ -8,6 +8,10 @@ leak:pw_context_load_module
 # Gio? (fixed in latest)
 leak:g_dbus_message_new_from_blob
 
+# libpx
```
AdrianVovk: Had to look up what `libpx` is. It's `libproxy`

src/xdp-request.h
```diff
@@ -101,4 +91,6 @@ auto_lock_helper (GMutex *mutex)
   return mutex;
 }
 
-#define REQUEST_AUTOLOCK(request) G_GNUC_UNUSED __attribute__((cleanup (auto_unlock_helper))) GMutex * G_PASTE (request_auto_unlock, __LINE__) = auto_lock_helper (&request->mutex);
+#define REQUEST_AUTOLOCK(request) \
+  G_GNUC_UNUSED __attribute__((cleanup (auto_unlock_helper))) \
```
AdrianVovk: Another cleanup you can do: there's actually a helper for this in glib: https://docs.gtk.org/glib/func.MUTEX_AUTO_LOCK.html

src/xdp-context.c
```diff
@@ -82,9 +83,18 @@ xdp_context_dispose (GObject *object)
...
+      g_assert (context->connection);
+      xdp_connection_untrack_peer_disconnect (context->connection,
+                                              context->peer_disconnect_handle_id);
+      context->peer_disconnect_handle_id = 0;
+    }
+
   g_clear_object (&context->portal_config);
   g_clear_object (&context->connection);
   g_clear_object (&context->lockdown_impl);
+  g_clear_object (&context->connection);
```
AdrianVovk: Extraneous, you clear the connection two lines above already

src/xdp-app-info.h
```diff
@@ -45,6 +46,19 @@ G_DECLARE_DERIVABLE_TYPE (XdpAppInfo,
 
 void xdp_app_info_delete_for_sender (const char *sender);
 
+XdpAppInfo * xdp_app_info_new (uint32_t   pid,
```
AdrianVovk: The other two I get, but why's this one public?

src/xdp-app-info-registry.c
```diff
@@ -0,0 +1,174 @@
...
+}
+
+gboolean
+xdp_app_info_registry_has_sender (XdpAppInfoRegistry *registry,
+                                  const char         *sender)
+{
+  gboolean has_app_info = FALSE;
+
+  g_mutex_lock (&registry->app_infos_mutex);
+  has_app_info = !!g_hash_table_lookup (registry->app_infos, sender);
```
AdrianVovk: There's a purpose-built `g_hash_table_contains` you can use instead

src/xdp-app-info-registry.c
```diff
@@ -0,0 +1,174 @@
...
+
+void
+xdp_app_info_registry_delete (XdpAppInfoRegistry *registry,
+                              const char         *sender)
+{
+  XdpAppInfo *app_info = NULL;
+
+  g_mutex_lock (&registry->app_infos_mutex);
+  app_info = g_hash_table_lookup (registry->app_infos, sender);
+  if (app_info)
```
AdrianVovk: You can save the lookup by just putting `if (g_hash_table_remove (..., sender))` directly in the condition

src/xdp-context.c
```diff
@@ -153,10 +163,14 @@ authorize_callback (GDBusInterfaceSkeleton *interface,
                     GDBusMethodInvocation  *invocation,
                     gpointer                user_data)
 {
+  XdpContext *context = user_data;
```
AdrianVovk: `XDP_CONTEXT(user_data)`? Like you did below

src/account.c
```diff
@@ -244,29 +245,40 @@ account_class_init (AccountClass *klass)
...
 
-  g_debug ("using %s at %s\n", "org.freedesktop.impl.portal.Account", dbus_name);
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (impl), G_MAXINT);
 
   account = g_object_new (account_get_type (), NULL);
 
-  return G_DBUS_INTERFACE_SKELETON (account);
+  xdp_context_export_portal (context, G_DBUS_INTERFACE_SKELETON (account));
+
+  g_object_set_data_full (G_OBJECT (context),
```
AdrianVovk: This `g_object_set_data_full` bit is duplicated everywhere, and I wonder if it can be made part of `xdp_context_export_portal` instead?

Really all you need is a unique name, right? Maybe you could get it from something like `G_OBJECT_CLASS_NAME`? Or maybe it's possible to pull out the interface name from the interface skeleton?

src/xdp-app-info-registry.c
swick: Can't print the debug thing then though. With `g_hash_table_steal_extended` the lookup is done once and it gets removed, but it means having to manually free the key and value.

src/account.c
swick: Yeah, I thought about it. What I like about having it in the init function is that the lifetime becomes obvious by just looking at the current file, and the context stays just a helper.

Not a strong opinion, so if you feel strongly about it, I would merge a follow up PR that does this.

src/xdp-request.h
swick: Added a new commit which also cleans this up for session and other places.

src/xdp-app-info.h
swick: Oversight, probably from a previous version.

src/xdp-types.h
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+typedef struct _XdpContext XdpContext;
+typedef struct _XdpAppInfo XdpAppInfo;
+
```
jadahl: Trailing newline.

src/xdp-context.c
```diff
@@ -69,6 +69,7 @@ struct _XdpContext
   gboolean verbose;
```
jadahl, line 69: commit message: -> registration

src/xdp-app-info-registry.c
```diff
@@ -0,0 +1,174 @@
...
+  return registry;
+}
+
+XdpAppInfo *
+xdp_app_info_registry_lookup_sender (XdpAppInfoRegistry *registry,
+                                     const char         *sender)
+{
+  XdpAppInfo *app_info = NULL;
+
+  g_mutex_lock (&registry->app_infos_mutex);
```
jadahl: There is also `g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&registry->app_info_mutex);` or `G_MUTEX_AUTO_LOCK (&registry->app_info_mutex, locker);` that avoids having to manually manage unlocking.

It'd make at least this and one other function below a bit nicer.

src/xdp-app-info-registry.h
```diff
@@ -0,0 +1,50 @@
...
+
+XdpAppInfoRegistry * xdp_app_info_registry_new (void);
+
+XdpAppInfo * xdp_app_info_registry_lookup_sender (XdpAppInfoRegistry *registry,
+                                                  const char         *sender);
+
+gboolean xdp_app_info_registry_has_sender (XdpAppInfoRegistry *registry,
+                                           const char         *sender);
+
+void xdp_app_info_registry_insert (XdpAppInfoRegistry *registry,
```
jadahl, line 41: On the nit side, but I find "add" vs "remove" to be better names for this, as "insert" to some degree is overloaded with "insert vs replace" semantics in other API.

src/xdp-app-info-registry.c
```diff
@@ -0,0 +1,174 @@
...
+G_DEFINE_FINAL_TYPE (XdpAppInfoRegistry,
+                     xdp_app_info_registry,
+                     G_TYPE_OBJECT)
+
+static void
+xdp_app_info_registry_dispose (GObject *object)
+{
+  XdpAppInfoRegistry *registry = XDP_APP_INFO_REGISTRY (object);
+
+  if (registry->app_infos)
```
jadahl, line 43: This condition is redundant.

swick: Is it? The way I understand it `g_mutex_clear` doesn't write NULL/init to the variable; it just frees the thing, so removing the condition would lead to double frees.

jadahl: Ah, it's not, it's just releasing, not clearing it. Unfortunate naming..

src/account.c
jadahl: Could make it `xdp_context_take_and_export_portal()` and it'll be obvious ownership is transferred / tied to the context.

src/xdp-app-info-registry.h
swick: Tried it, didn't like it much. This is quite literally an insert operation on the hash table which could be replacing. We also use delete instead of remove for other things, so it would just become inconsistent.

src/xdp-app-info-registry.c
```diff
@@ -0,0 +1,167 @@
...
+
+  return g_hash_table_contains (registry->app_infos, sender);
+}
+
+void
+xdp_app_info_registry_insert (XdpAppInfoRegistry *registry,
+                              XdpAppInfo         *app_info)
+{
+  const char *sender = xdp_app_info_get_sender (app_info);
+  G_MUTEX_AUTO_LOCK (&registry->app_infos_mutex, locker);
```
jadahl, line 108:
```suggestion
 
  G_MUTEX_AUTO_LOCK (&registry->app_infos_mutex, locker);
```

I think it'd be better to move it out of the variable declarations, since its side effect is rather crucial. It's obvious enough if it's only one declaration, but when there are many, things tend to get "hidden", so better set a good starting point for consistency.

src/account.c
swick: Did that in a new commit.

src/xdp-app-info-registry.c
swick: It's a bit weird because we must only declare variables at the beginning of a block, and the   `G_MUTEX_AUTO_LOCK` and friends are variable declarations; but in some places of xdp, they are mixed with code. I suspect it is because this looks like a statement when it really is a declaration, and moving it out of the declaration block just enforces this bad pattern. So both of those things are not great.

src/xdp-app-info-registry.c
AdrianVovk: Yeah almost left this comment myself... Certainly a gotcha

src/xdp-app-info-registry.c
AdrianVovk: > we must only declare variables at the beginning of a block

I assume it's safe to say that we're not targeting C89 in this project right? As of C99 variables don't need to be declared at the beginning of a block

src/xdp-context.c
```diff
@@ -239,16 +248,24 @@ xdp_context_export_portal (XdpContext             *context,
...
-      g_warning ("Error: %s", error->message);
-      return;
-    }
+  if (g_dbus_interface_skeleton_export (skeleton,
+                                        context->connection,
+                                        DESKTOP_PORTAL_OBJECT_PATH,
+                                        &error))
+    g_debug ("Providing portal %s", name);
+  else
+    g_warning ("Error: %s", error->message);
```
AdrianVovk: Probably needs a better error message. Otherwise we'll end up debugging something nice an incomprehensible like `Error: Permission Denied` in the future

PR #1818, opened 2025-09-25
Introduce save/restore portal

AdrianVovk:
Implements the portal proposed in https://github.com/flatpak/xdg-desktop-portal/discussions/1698

Links to other MRs:
- Backend impl in gnome-session: Nothing yet
- GTK client for this portal: Nothing yet


data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,164 @@
...
+            content inside of each tab, instead showing a placeholder page and
+            prompting the user to reload the tab. To support this feature
+            properly, apps should periodically save their state while taking care
+            to avoid unnecessary I/O. For instance: apps can flush their state
+            to disk every 15 seconds, but only while the app is focused.
+
+          - ``RESTORE``: 3. The app should attempt to restore as much state as
+            possible.
+
+        * ``discrard-instance-ids`` (``as``)
```
AdrianVovk: typo

Mikenux:
About the naming...

Why `SaveHint `instead of `Save`, given that it's `Save ` on the `SaveRestore `interface? Or `SaveState`?

AdrianVovk:
> Why SaveHint instead of Save

I renamed it to make it a bit more clear

> given that it's `Save` on the SaveRestore interface

Sorry, where? It's SaveHint in all the dbus interfaces here

Mikenux:
> Sorry, where? It's SaveHint in all the dbus interfaces here

Sorry, mistake.

> I renamed it to make it a bit more clear

Relative to? If you are looking for something meaningful, I think `SaveState` or `SaveAppState` are better options. Same for the interface name.

AdrianVovk:
> Relative to? If you are looking for something meaningful, I think SaveState or SaveAppState are better options. Same for the interface name

Whoops didn't finish typing my comment. `SaveHint` makes it a bit more clear that it's just a hint and that the app shouldn't be using it as the primary source for deciding to save state. If it was just `Save` it might seem like the portal will emit the signal regularly or whatever. Anyway, not the most important thing in the world, but that's why I renamed it

AdrianVovk:
CC @davidedmundson and @technobaboo as the two folks from other desktops who I've interacted with in the discussion

This is the real API I ended up implementing. PTAL and let me know if there's some blocker to getting something like this working in your desktop, or you see something missing, or etc.

For your reference, here's GNOME's implementation of this portal's backend: https://gitlab.gnome.org/GNOME/gnome-session/-/blob/avovk/session-save-portal/gnome-session/gsm-session-save.c

technobaboo:
wait how does the app get the instance ID and know to restore its data using that ID? afaik there's no standard key name for it or such in the registration map it gets as a return arg for registering

technobaboo:
okay that's nice to know, so if it's required why isn't it an argument instead of inside the map?

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,179 @@
...
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.portal.SaveRestore:
+      @short_description: Portal for coordinating app state save and restore
+
+      The SaveRestore interface allows the session to coordinate with applications
+      to implement a session-wide state save/restore feature. Apps that can save
+      their state to disk should register with this portal. In exchange, the system
+      will present the user with the option to restore session state after their
+      session ends (i.e. after a reboot). It is not a portal in the strict sense,
```
Mikenux: Not sure if all DEs will present the option after session end or reboot (also, they are not the same events).

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,179 @@
...
+  <interface name="org.freedesktop.portal.SaveRestore">
+    <!--
+        Register:
+        @options: Vardict with optional further information
+        @session_handle: Object path for the created :ref:`org.freedesktop.portal.Session` object
+        @registration: Vardict directing the app how to restore its state
+
+        Registers this app with the save/restore API. If the user logs out while
+        the app is registered, the system will remember that this app was running
+        and relaunch it on next log-in. The app can then register itself again
```
Mikenux: Relaunch after logout: not sure that all DEs will act the same way.

technobaboo: agreed, in stardust's case we relaunch things by distance (since it's AR/VR)

data/org.freedesktop.portal.SaveRestore.xml
technobaboo: stardust definitely won't, it'll just automatically do it

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,179 @@
...
+          A list of instance IDs (as above) that the system is done with and will
+          no longer return to the app. The app should clean up any state it has
+          associated with these IDs, and then report completion via
+          :ref:`org.freedesktop.portal.SaveRestore.DiscardInstanceIdsResponse`.
+    -->
+    <method name="Register">
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QVariantMap"/>
+      <arg type="a{sv}" name="options" direction="in"/>
+      <arg type="o" name="session_handle" direction="out"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
```
davidedmundson, line 113: So the idea in general is the desktop starts the applications that we are session restored using the Exec line from the dekstop file.
Then on application startup the app queries the portal and says "why was I started?".

This is going to be very racey to manage with multi-instance applications. 

For example:
If I have a user-configured autostart script that runs 'kwrite ~/journal.txt' 
and I have kwrite with no arguments that's being loaded by session restore it'll be 50/50 on which process ends up thinking it's restored and that'll lead to inconsistency.

data/org.freedesktop.impl.portal.SaveRestore.xml
```diff
@@ -0,0 +1,126 @@
...
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Adrian Vovk <avovk@redhat.com>
```
swick: I would really like to enforce a policy of not adding author information to files.

data/org.freedesktop.impl.portal.Session.xml
```diff
@@ -31,25 +31,51 @@
       object, which will stay alive for the duration of the session.
 
       The portal can abort the interaction by calling
-      org.freedesktop.impl.portal.Session.Close() on the Session object.
+      org.freedesktop.impl.portal.Session.CloseEx() on the Session object.
```
swick: Let's try to avoid multiple ways to do the same thing. In another PR, the succeeding method is just called `${method}2`, here it is `${method}Ex`.

src/xdp-session.c
```diff
@@ -199,8 +200,23 @@ xdp_session_close (XdpSession *session,
...
+          g_variant_builder_add (&options_builder, "{sv}", "voluntary",
+                                 g_variant_new_boolean (voluntary));
+
+          xdp_dbus_impl_session_call_close_ex_sync (session->impl_session,
+                                                    g_variant_builder_end (&options_builder),
+                                                    NULL, &error);
+        }
+      else
+        xdp_dbus_impl_session_call_close_sync (session->impl_session,
+                                               NULL, &error);
```
swick:
```suggestion
      else
        {
          xdp_dbus_impl_session_call_close_sync (session->impl_session,
                                                 NULL, &error);
        }
```

swick: Same issue in multiple places

src/xdp-session.c
```diff
@@ -199,8 +200,23 @@ xdp_session_close (XdpSession *session,
...
+
+      if (error != NULL)
         g_warning ("Failed to close session implementation: %s",
                    error->message);
```
swick: Not a fan of checking the error itself instead of the return value of the function, but I don't really see a much better way in this specific case.

src/save-restore.h
```diff
@@ -0,0 +1,28 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Matthias Clasen <mclasen@redhat.com>
```
swick: 🙃

src/save-restore.c
```diff
@@ -0,0 +1,438 @@
...
+static void
+xdp_save_restore_class_init (XdpSaveRestoreClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->finalize = xdp_save_restore_finalize;
+  object_class->set_property = xdp_save_restore_set_property;
+  object_class->get_property = xdp_save_restore_get_property;
+
+  g_object_class_install_property (object_class, PROP_IMPL,
```
swick: FWIW, I don't think there is a good reason to use gobject props here at all. The portals are always final, they never will need to be bindable, etc. Just use a `xdp_save_restore_new` instead.

src/save-restore.c
```diff
@@ -0,0 +1,438 @@
...
+}
+
+static void
+xdp_save_restore_session_class_init (XdpSaveRestoreSessionClass *klass)
+{
+  GObjectClass *object_class;
+  XdpSessionClass *session_class;
+
+  object_class = G_OBJECT_CLASS (klass);
+  object_class->finalize = xdp_save_restore_session_finalize;
```
swick: don't have to add this if it doesn't do anything

src/save-restore.c
```diff
@@ -0,0 +1,438 @@
...
+
+  xdp_session_register (session);
+
+  g_variant_builder_init (&options_builder, G_VARIANT_TYPE_VARDICT);
+
+  data = g_new (RegisterData, 1);
+  *data = (RegisterData) {
+    .save_restore = self,
+    .invocation = invocation,
+    .session = session,
```
swick: When the client disconnects, the session will get closed, which unexports it, which drops the ref-count to 0. You need to take a ref here.

src/save-restore.c
```diff
@@ -0,0 +1,438 @@
...
+} RegisterData;
+
+static void
+impl_register_done (GObject      *source_object,
+                    GAsyncResult *result,
+                    gpointer      user_data)
+{
+  XdpDbusImplSaveRestore *impl_proxy = XDP_DBUS_IMPL_SAVE_RESTORE (source_object);
+  g_autofree RegisterData *data = user_data;
+  g_autoptr (GError) error = NULL;
```
swick: In xdp and flatpak for some reason the autoptr style is `g_autoptr(GError)`.

swick (commented):
Just looked through the PR to understand what is being proposed here and found a bunch of things.

data/org.freedesktop.portal.SaveRestore.xml
AdrianVovk: We discussed this on a call, and IIRC the conclusion was that you were just going to try it and see how it works for you. So, marking resolved for now but if it's ultimately a problem we can come back to it

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,177 @@
...
+        Quit:
+        @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object
+
+        The Quit signal is sent to an app whenever the system decides to
+        terminate the app to free up system resources (i.e. memory). The app
+        should immediately save its state to disk and then exit cleanly, within
+        3 seconds of receiving this signal. The app may be transparently restarted
+        by the system at a later date.
+    -->
+    <signal name="Quit">
```
AdrianVovk, line 171: Hmm so I'm not really sure that this is actually necessary...

Instead of asking the app to quit, we can just put it in a cgroup freezer. This removes it from the run queue. We could then adjust the memory limits on the cgroup, which as far as I understand will force the app's memory out to swap and out of RAM. At that point, is there really any point to actually shutting down the app? We can leave the app "running", just frozen in a cgroup freezer and swapped out to disk. If the goal is to free up resources like memory and CPU, we've achieved that.

We can use something like `SaveHint` to ask the app to save its state just in case it somehow gets killed while in the freezer.

Alternatively, apps should just save their state immediately before quitting in response to something like SIGTERM. Then we could handle this simply by sending SIGTERM to the apps, rather than having them subscribe to this signal.

src/save-restore.c
```diff
@@ -0,0 +1,403 @@
...
+  XdpDbusSaveRestoreSkeleton parent_instance;
+
+  XdpDbusImplSaveRestore *impl;
+  GCancellable *cancellable;
+};
+
+#define XDP_TYPE_SAVE_RESTORE (xdp_save_restore_get_type ())
+G_DECLARE_FINAL_TYPE (XdpSaveRestore,
+                      xdp_save_restore,
+                      XDP, SAVE_RESTORE,
```
AdrianVovk, line 44: Hmm so the USB portal names itself XdpUsb, but apparently most every other portal doesn't have an Xdp prefix. And apparently that was intentional

So, IDK am I supposed to rename this portal to just SaveRestore?

Perhaps if we want to maintain a separation, we could have a `Portal` prefix, so that all the portals are `PortalAccounts`, `PortalSaveRestore`, `PortalUsb`, etc and the utility classes/functions can keep `XdpSession`, `XdpContext`, etc

src/save-restore.c
```diff
@@ -0,0 +1,403 @@
...
+
+#define XDP_TYPE_SAVE_RESTORE_SESSION (xdp_save_restore_session_get_type ())
+G_DECLARE_FINAL_TYPE (XdpSaveRestoreSession,
+                      xdp_save_restore_session,
+                      XDP, SAVE_RESTORE_SESSION,
+                      XdpSession)
+
+G_DEFINE_TYPE (XdpSaveRestoreSession,
+               xdp_save_restore_session,
+               xdp_session_get_type ())
```
AdrianVovk, line 70: XDP_TYPE_SESSION?

src/save-restore.c
```diff
@@ -0,0 +1,403 @@
...
+
+static void
+xdp_save_restore_session_close (XdpSession *session)
+{
+  XdpSaveRestoreSession *save_restore_session = XDP_SAVE_RESTORE_SESSION (session);
+
+  save_restore_session->closed = TRUE;
```
AdrianVovk, line 77: I think a couple of the session implementations do nothing but keep track of this closed variable.

Maybe that should move up into the `XdpSession` object?

On the other hand, this class separation actually lets us check that the caller is giving us a session path to a session of a type we expect. So we'd need to have a subclass anyway. Or at the very least some sort of accounting of what portal created any given session

src/save-restore.c
```diff
@@ -0,0 +1,403 @@
...
+
+static void
+impl_register_done (GObject      *source_object,
+                    GAsyncResult *result,
+                    gpointer      user_data)
+{
+  XdpDbusImplSaveRestore *impl = XDP_DBUS_IMPL_SAVE_RESTORE (source_object);
+  g_autoptr(RegisterData) data = user_data;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GVariant) out_restore = NULL;
```
AdrianVovk, line 160: I renamed this to `registration`

src/save-restore.c
```diff
@@ -0,0 +1,403 @@
...
+{
+  g_autoptr(XdpSaveRestore) save_restore = NULL;
+  GDBusConnection *connection = xdp_context_get_connection (context);
+  XdpPortalConfig *config = xdp_context_get_config (context);
+  XdpImplConfig *impl_config;
+  g_autoptr(XdpDbusImplSaveRestore) impl = NULL;
+  g_autoptr(GError) error = NULL;
+
+  impl_config = xdp_portal_config_find (config, SAVE_RESTORE_DBUS_IMPL_IFACE);
+  if (impl_config != NULL)
```
davidedmundson, line 383: This is backwards.

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,177 @@
...
+
+          - ``LAUNCH``: 1. This is a normal launch, so the app should restore
+            some state (i.e. previous window size and position, default view on
+            the home page) but it shouldn't attempt a full state restore. The
+            exception is apps that have a preference that allows the user to
+            decide if the app should always save and restore its state: if the
+            user opts into this behavior, then it's permissible for the app to
+            treat ``LAUNCH`` as equivalent to ``RESTORE``.
+
+          - ``RECOVER``: 2. This is the first launch after the app unexpectedly
```
AdrianVovk, line 87: Split this into two different reasons: recover after compositor crash, and recover after app crash

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,177 @@
...
+
+        Supported keys in the @options vardict include:
+
+        * ``session_handle_token`` (``s``)
+
+          A string that will be used as the last element of the session handle. Must be a valid
+          object path element. See the :ref:`org.freedesktop.portal.Session` documentation for
+          more information about the session handle.
+
+        Supported keys in the @registration vardict include:
```
AdrianVovk, line 61: Some apps have built-in crash monitoring. Maybe it's useful to communicate to the compositor whether the app has its own crash monitoring

Apps that don't have their own crash monitoring may be auto-restarted by the session (or the session might offer to restart the app). But if apps have their own builtin crash monitoring then the session would just do nothing, under the assumption that the app will restart itself.

Builtin crash monitoring might seem niche, but in practice I suspect it'll increasingly become the norm: systemd has a `Restart=on-failure` feature for instance, and many apps (especially single instance apps) natrually map to systemd user services (especially on desktops like KDE, which uses transient systemd units for all apps). And desktops implementing a bespoke solution sounds suspiciously like reimplementing parts of systemd.

Perhaps this is better suited as a setting in the app's .desktop file? KDE would use the setting to decide whether `Restart=on-failure` is passed to the systemd transient unit

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,177 @@
...
+        would be a good moment for the app to save its state. In response, apps
+        should immediately save their state and then respond with
+        :ref:`org.freedesktop.portal.SaveRestore.SaveHintResponse`.
+
+        Note that apps should not be using this signal as the only reason to
+        save their state. This signal will be emitted relatively rarely. Each
+        app should periodically save its own state at moments where it makes
+        sense for the app.
+    -->
+    <signal name="SaveHint">
```
AdrianVovk, line 146: After talking to David, we decided to drop the `SaveHint` and `Quit` methods. Instead, I'd extend the inhibit portal with signals that tell apps about upcoming system suspend and resume

data/org.freedesktop.portal.SaveRestore.xml
Mikenux: Isn't launching an app supposed to be handled by the system? If so, a system can then ask what to do when an app has crashed. This could also be useful if someone wants to propose a standardized crash reporting tool in the future.

Mikenux: > But if apps have their own builtin crash monitoring then the session would just do nothing, under the assumption that the app will restart itself.

What about proper restoration? (position and size) Wayland?

data/org.freedesktop.portal.SaveRestore.xml
Mikenux: The inhibit portal allows an app to prevent changes in session status, including its suspension. It has nothing to do with saving and restoring an app's session. Furthermore, `SaveHint` is vague, implying to desktop environment developers that they can send it whenever they want. Moreover, it provides no information about a user's action (the Settings portal provide some appearance settings, but the less, the better).

Therefore, I don't understand the purpose of this change…

data/org.freedesktop.portal.SaveRestore.xml
```diff
@@ -0,0 +1,177 @@
...
+          well, because the system will return different identifiers when running
+          different graphical sessions. For instance: if a user switches between
+          GNOME and KDE regularly, even single-instance apps will need to save
+          and load their state under different instance IDs.
+
+        * ``restore-reason`` (``u``)
+
+          Tells the app which state it needs to restore. Supported values:
+
+          - ``PRISTINE``: 0. The app should not attempt to restore any state.
```
AdrianVovk, line 77: Pristine isn't the most obvious name in the world. Maybe "none" or "disabled" is better?

PR #1819, opened 2025-09-25
portals: Stop using global state

swick:
Based on https://github.com/flatpak/xdg-desktop-portal/pull/1815.

Unfortunately it's a single mega-commit but it's really hard to split this up into smaller chunks, except for one commit per-portal, but that would not be useful either.

### DIFF
--- a/src/xdp-context.c
+++ b/src/xdp-context.c
@@ -285,10 +285,10 @@ on_peer_disconnect (const char *name,
 
   xdp_usb_delete_for_sender (context, name);
   notification_delete_for_sender (context, name);
-  xdp_app_info_registry_delete (context->app_info_registry, name);
   close_requests_for_sender (name);
   close_sessions_for_sender (name);
   xdp_session_persistence_delete_transient_permissions_for_sender (name);
+  xdp_app_info_registry_delete (context->app_info_registry, name);
 }
 
 gboolean


src/account.c
```diff
@@ -155,10 +156,11 @@ get_user_information_done (GObject *source,
...
 static gboolean
-validate_reason (const char *key,
-                 GVariant *value,
-                 GVariant *options,
-                 GError **error)
+validate_reason (const char  *key,
+                 GVariant    *value,
+                 GVariant    *options,
+                 gpointer     user_data,
+                 GError     **error)
```
jadahl, line 163: Unrelated style cleanup.

src/account.c
```diff
@@ -193,11 +196,13 @@ handle_get_user_information (XdpDbusAccount *object,
...
-                                                       G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
-                                                       g_dbus_proxy_get_name (G_DBUS_PROXY (impl)),
-                                                       request->id,
-                                                       NULL, &error);
+  impl_request = xdp_dbus_impl_request_proxy_new_sync (
+    g_dbus_proxy_get_connection (G_DBUS_PROXY (account->impl)),
+    G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
+    g_dbus_proxy_get_name (G_DBUS_PROXY (account->impl)),
+    request->id,
+    NULL, &error);
```
jadahl, line 204: Unrelated style cleanup.

src/usb.c
```diff
@@ -812,7 +812,7 @@ handle_create_session (XdpDbusUsb            *object,
                            &options_builder,
```
AdrianVovk, line 823: Does it become possible to replace the bespoke option filtering implementation in the USB portal's `filter_access_devices` function with the standard `xdp_filter_options` now that `xdp_filter_options` has a userdata pointer?

src/account.c
```diff
@@ -183,6 +184,7 @@ handle_get_user_information (XdpDbusAccount *object,
                              const gchar *arg_parent_window,
                              GVariant *arg_options)
 {
+  Account *account = (Account *) object;
```
AdrianVovk, line 187: Any reason you're using normal casts instead of checked casts? `ACCOUNT (object)` rather than `Account *object`? Same applies to the rest of the casts in the PR

I assume it's because we don't actually ever DECLARE the gobject, only DEFINE it. But why not also DECLARE it? Then you get type-checked casts, you get the autoptr for free, and you can drop the manual declaration of the objects' classes too

src/account.c
```diff
@@ -236,27 +239,55 @@ account_iface_init (XdpDbusAccountIface *iface)
...
 static void
 account_class_init (AccountClass *klass)
 {
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->dispose = account_dispose;
+}
+
+static Account *
+account_new (XdpDbusImplAccount *impl)
```
AdrianVovk: Hmm any particular reason to split things out into a separate `*_new` method? Seems like a somewhat arbitrary split

src/background.c
```diff
@@ -634,58 +643,95 @@ check_background_apps (void)
...
-  while (TRUE)
-    {
-      g_main_context_iteration (monitor_context, TRUE);
-      /* We check twice, to avoid killing unlucky apps hit at a bad time */
-      sleep (5);
-      check_background_apps ();
-      sleep (5);
-      check_background_apps ();
-    }
+  fprintf(stderr, "WHHHHHHEEEEEEEEEEEEEEEEEEEE\n\n\n\n");
```
AdrianVovk: WHHHHHHEEEEEEEEEEEEEEEEEEEE!

I suspect this was some debug logging you forgot to get rid of?

src/background.c
```diff
@@ -634,58 +643,95 @@ check_background_apps (void)
...
+  g_return_if_fail (g_task_is_valid (result, background));
 
-  monitor_context = g_main_context_new ();
+  g_clear_object (&background->monitor_task);
 
-  thread = g_thread_new ("background monitor", background_monitor, NULL);
+  if (!g_task_propagate_boolean (task, NULL))
+    return;
+
+  /* FIXME: schedule idle? */
```
AdrianVovk: New FIXME, did you mean to address it now?

src/camera.c
```diff
@@ -36,19 +36,16 @@
...
-
 typedef struct _Camera Camera;
 typedef struct _CameraClass CameraClass;
 
 struct _Camera
 {
   XdpDbusCameraSkeleton parent_instance;
 
+  XdpContext *context;
+  XdpDbusImplAccess *access_impl;
```
AdrianVovk, line 46: Looks like the access impl is used enough that it should get the same treatment as the lockdown impl: construct one copy and keep it in the XdpContext. Instead of each portal constructing a dedicated proxy

src/camera.c
```diff
@@ -526,20 +501,60 @@ camera_class_init (CameraClass *klass)
...
+camera_new (XdpContext        *context,
+            XdpDbusImplAccess *access_impl)
+{
+  Camera *camera;
+  g_autofree char *pipewire_socket_path = NULL;
+  g_autoptr(GFile) pipewire_socket = NULL;
+  g_autoptr(GError) error = NULL;
+
+  camera = g_object_new (camera_get_type (), NULL);
+  camera->context = context;
```
AdrianVovk: I'm not super comfortable with the pseudo-weak-ref we've got here. Especially since we handle dbus method calls on threads. What happens when the XdpContext gets destroyed, unrefs the `Camera` object, but the Camera object is still reffed by gdbus because we're busy handling a method call? _Maybe_ the implementation is actually safe as-written, but convincing myself of that requires some careful thought I don't really feel like doing. Plus, a future change might break this anyway

But what I noticed is that we're storing the context only to get the lockdown impl. So we can just get the lockdown impl from the context, ref it, and store that instead. No more problem, and no more pseudo-weak-ref to worry about

src/clipboard.c
```diff
@@ -187,15 +192,18 @@ handle_set_selection (XdpDbusClipboard *object,
                            &options_builder,
                            clipboard_set_selection_options,
                            G_N_ELEMENTS (clipboard_set_selection_options),
-                           NULL, &error))
+                           NULL,
+                           &error))
```
AdrianVovk: hmm why?

src/dynamic-launcher.c
```diff
@@ -625,11 +660,13 @@ handle_prepare_install (XdpDbusDynamicLauncher *object,
 
   REQUEST_AUTOLOCK (request);
 
-  impl_request = xdp_dbus_impl_request_proxy_new_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (impl)),
-                                                       G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
-                                                       g_dbus_proxy_get_name (G_DBUS_PROXY (impl)),
-                                                       request->id,
-                                                       NULL, &error);
+  impl_request = xdp_dbus_impl_request_proxy_new_sync (
```
AdrianVovk, line 663: For a future PR, I'm seeing this a lot. Looks like a good candidate for a utility function...

src/file-chooser.c
```diff
@@ -807,29 +816,57 @@ file_chooser_iface_init (XdpDbusFileChooserIface *iface)
...
+}
+
+static FileChooser *
+file_chooser_new (XdpContext             *context,
+                  XdpDbusImplFileChooser *impl)
+{
+  FileChooser *file_chooser;
+
+  file_chooser = g_object_new (file_chooser_get_type (), NULL);
+  file_chooser->context = context;
```
AdrianVovk: Same story here as the camera portal regarding the context, and using a lockdown impl instead

src/global-shortcuts.c
```diff
@@ -635,7 +657,12 @@ handle_configure_shortcuts (XdpDbusGlobalShortcuts *object,
       return G_DBUS_METHOD_INVOCATION_HANDLED;
     }
 
-  xdp_dbus_impl_global_shortcuts_call_configure_shortcuts (impl,
+  g_object_set_data_full (G_OBJECT (invocation),
+                          "-xdp-global-shortcuts",
```
AdrianVovk: In the previous dynamic launchers commit you just named this `dynamic_launchers`, so probably best to pick a style and stick to it?

src/input-capture.c
```diff
@@ -99,8 +97,43 @@ IS_INPUT_CAPTURE_SESSION (gpointer ptr)
...
+{
+  InputCaptureSession *input_capture_session = INPUT_CAPTURE_SESSION (session);
+
+  input_capture_session->state = INPUT_CAPTURE_SESSION_STATE_CLOSED;
+
+  g_debug ("screen cast session owned by '%s' closed", session->sender);
+}
+
+static void
+input_capture_session_finalize (GObject *object)
```
AdrianVovk: This does nothing but chain up, so unnecessary override?

src/input-capture.c
```diff
@@ -1154,74 +1202,74 @@ on_zones_changed_cb (XdpDbusImplInputCapture *impl,
...
+
+  xdp_dbus_input_capture_set_version (XDP_DBUS_INPUT_CAPTURE (input_capture), 1);
+
+  g_object_bind_property (G_OBJECT (input_capture->impl), "supported-capabilities",
+                          G_OBJECT (input_capture), "supported-capabilities",
+                          G_BINDING_SYNC_CREATE);
+
+  g_signal_connect (input_capture->impl, "disabled",
+                    G_CALLBACK (on_disabled_cb),
+                    input_capture->impl);
```
AdrianVovk: The old code passed in `input_capture` as userdata to the signal callbacks, this code now passes in `input_capture->impl`. But at quick glance it looks like the userdata is never actually used (it's passed into the callbacks, which pass it into `pass_signal`, which never uses it). So really you should just be passing in `NULL` as the userdata

src/location.c
```diff
@@ -714,31 +727,57 @@ location_iface_init (XdpDbusLocationIface *iface)
...
 }
 
+static Location *
+location_new (XdpContext        *context,
+              XdpDbusImplAccess *access_impl)
+{
+  Location *location;
+
+  location = g_object_new (location_get_type (), NULL);
+  location->context = context;
```
AdrianVovk: Another instance of pseudo-weak-pointer storage of context

src/notification.c
```diff
@@ -1164,68 +1185,47 @@ notification_handle_remove_notification (XdpDbusNotification   *object,
...
+}
 
-  if (name[0] == ':' &&
-      strcmp (name, from) == 0 &&
-      strcmp (to, "") == 0)
-    {
-      GHashTableIter iter;
-      Pair *p;
+void
+notification_delete_for_sender (const char *sender)
```
AdrianVovk: I assume that this is the reason we keep the `active` hashtable as global state. I think in a previous PR somewhere you mentioned that you plan to make `XdpContext` emit a signal that portals can listen for instead? Would it make sense to just do it here so that we can fully get rid of the global state? On the other hand this PR is already chunky enough...

src/open-uri.c
```diff
@@ -1134,29 +1145,60 @@ open_uri_iface_init (XdpDbusOpenURIIface *iface)
...
+}
+
+static OpenURI *
+open_uri_new (XdpContext            *context,
+              XdpDbusImplAppChooser *app_chooser_impl)
+{
+  OpenURI *open_uri;
+
+  open_uri = g_object_new (open_uri_get_type (), NULL);
+  open_uri->context = context;
```
AdrianVovk: Another instance of the not-reffed context

src/print.c
```diff
@@ -331,29 +333,57 @@ print_iface_init (XdpDbusPrintIface *iface)
...
+}
+
+static Print *
+print_new (XdpContext       *context,
+           XdpDbusImplPrint *impl)
+{
+  Print *print;
+
+  print = g_object_new (print_get_type (), NULL);
+  print->context = context;
```
AdrianVovk: Another context without ref

src/realtime.c
```diff
@@ -256,9 +263,9 @@ load_all_properties (GDBusProxy *proxy)
       GVariant *parameters;
       g_autoptr (GError) error = NULL;
 
-      parameters = g_variant_new ("(ss)", "org.freedesktop.RealtimeKit1", properties[i]);
+      parameters = g_variant_new ("(ss)", RTKIT_DBUS_IFACE, properties[i]);
       result = g_dbus_proxy_call_sync (proxy,
-                                       "org.freedesktop.DBus.Properties.Get",
+                                       DBUS_DBUS_IFACE ".Properties.Get",
                                         g_steal_pointer (&parameters),
```
AdrianVovk, line 269: Spacing issue, but overall I think the bigger question is why the code works this way and doesn't just use the cached properties already loaded by the proxy object? (i.e. bind the impl proxy's properties to the Realtime object and call it a day?)

There would be one difference in behavior: the code as-written ignores property change events, but binding the properties wouldn't do that. But I have a sneaking suspicion that these properties never actually change

src/registry.c
```diff
@@ -147,21 +153,33 @@ registry_iface_init (XdpDbusHostRegistryIface *iface)
...
 
-void
-init_registry (XdpContext *context)
+static Registry *
+registry_new (XdpContext *context)
 {
+  Registry *registry;
+
   registry = g_object_new (registry_get_type (), NULL);
   registry->context = context;
```
AdrianVovk: Another context we're storing with a pseudo-weak-pointer. Looks like I missed this in the last PR

This time, we need the app info registry rather than the lockdown impl. But basically the same idea

src/remote-desktop.c
```diff
@@ -216,18 +264,23 @@ remote_desktop_session_new (GVariant *options,
                             "impl-connection", impl_connection,
                             "impl-dbus-name", impl_dbus_name,
                             NULL);
+  if (session == NULL)
+    return NULL;
+
+  rd_session = REMOTE_DESKTOP_SESSION (session);
+  rd_session->remote_desktop = remote_desktop;
```
AdrianVovk: Iffy for much the same reason that I've been saying with the `XdpContext` throughout this PR. Though in this case you can probably just ref the `RemoteDesktop` object. There wouldn't be a ref loop created, since it's the `XdpContext` that ultimately decides when to destroy the session object, right?

src/screenshot.c
```diff
@@ -488,37 +499,71 @@ screenshot_iface_init (XdpDbusScreenshotIface *iface)
...
+
+  /* Before there was a version property, the version was hardcoded to 2, so
+   * make sure we retain that behaviour */
+  version = g_dbus_proxy_get_cached_property (G_DBUS_PROXY (screenshot->impl),
+                                              "version");
+  screenshot->impl_version =
+    (version != NULL) ? g_variant_get_uint32 (version) : 2;
+
+  xdp_dbus_screenshot_set_version (XDP_DBUS_SCREENSHOT (screenshot),
+                                   screenshot->impl_version);
```
AdrianVovk: You're missing the `MAX` check to make sure that we don't report a newer version than we actually support if the backend happens to be newer than the frontend

src/screenshot.c
```diff
@@ -488,37 +499,71 @@ screenshot_iface_init (XdpDbusScreenshotIface *iface)
...
+  g_autoptr(GVariant) version = NULL;
+
+  screenshot = g_object_new (screenshot_get_type (), NULL);
+  screenshot->access_impl = g_object_ref (access_impl);
+  screenshot->impl = g_object_ref (impl);
+
+  g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (screenshot->impl), G_MAXINT);
+
+
+  /* Before there was a version property, the version was hardcoded to 2, so
```
AdrianVovk, line 540: This was confusing so I went on a deep dive. See https://github.com/flatpak/xdg-desktop-portal/pull/766#issuecomment-3430191515

TL;DR: We shouldn't be reporting the impl's version as the frontend's version. These versions don't correspond to each other on the screenshot portal. Instead we should just hardcode the frontend to v2, and leave a comment pointing at the link above explaining that the frontend should never report a version less than 2.

src/settings.c
```diff
@@ -137,13 +138,15 @@ settings_handle_read_all (XdpDbusSettings       *object,
                                   g_free,
                                   (GDestroyNotify) g_variant_dict_unref);
 
-  for (j = n_impls - 1; j >= 0; j--)
+  for (j = self->n_impls - 1; j >= 0; j--)
```
AdrianVovk: You switched to `size_t` in the loops below but not up here. Ditto with the definition of `n_impls` in the struct itself

src/usb.c
```diff
@@ -491,9 +496,12 @@ xdp_usb_session_new (GDBusConnection  *connection,
   if (!session)
     return NULL;
 
+  usb_session = XDP_USB_SESSION (session);
+  usb_session->usb = usb;
```
AdrianVovk: Another object not being reffed

src/usb.c
```diff
@@ -1548,45 +1569,59 @@ xdp_usb_init (XdpUsb *self)
 }
 
 void
-xdp_usb_delete_for_sender (const char *sender)
+xdp_usb_delete_for_sender (XdpContext *context,
+                           const char *sender)
 {
+  XdpUsb *usb = g_object_get_data (G_OBJECT (context), "-xdp-portal-usb");
```
AdrianVovk: This doesn't work anymore right? I think you added a lookup function for this, where you can look up a portal impl by name?

src/usb.c
```diff
@@ -1516,8 +1539,6 @@ xdp_usb_init (XdpUsb *self)
 
   g_debug ("[usb] Initializing USB portal");
```
AdrianVovk, line 1519: On one hand, looks like all the other portals have an empty `_init` function and instead moved everything into `_new`. On the other hand, the USB portal seems to have a lot more logic in its `_init` function than the other portals. So IDK what the best course of action here is. Just wanted to point out that this diverges from the pattern, and maybe it shouldn't

src/notification.c
AdrianVovk: It looks like in a later commit (the one that updates the USB portal), you pass in the XdpContext to a `_delete_for_sender` method, and then use the XdpContext to look up the implementation of the portal. That was a great idea!  You could do the same here to get back your `Notification` object. Then you could move the `active` hashtable into the `Notification` object to entirely get rid of global state for this portal.

src/realtime.c
AdrianVovk: Checked: and no none of them change at runtime. They only change via CLI args on the command-line of rtkit itself

So it should be fine to just bind to the properties

src/account.c
swick: Because none of the portals use a `MODULE` (i.e. `xdp` or whatever), which makes it impossible to `DECLARE` it. It's definitely on my list of things to clean up, but last time I tried that, there was resistance.

src/background.c
swick: lol, yes

src/background.c
swick: I do not remember ever adding this, but I clearly had to debug something...

src/account.c
jadahl: I had the same thought, wondering why GBM needs to be involved in GPU copies at all when EGL exists. But I'm not confident it would be a real solution. Still, something nice to try in a separate branch some day.

src/usb.c
swick: Yup, added a new commit.

src/account.c
swick: Did a comment about portals end up in mutter as well? :P

src/usb.c
swick: Just slipped through. Moved it to `_new` as well.

src/account.c
swick: This is not unrelated; it uses the `account->impl` instead of the global `impl`.

src/account.c
jadahl: Hrm, that looks like I copied a comment from a mutter MR that I was going to quote...

jadahl: What I meant to say was: maybe we should have a `XDP_DECLARE_PORTAL()` & `XDP_DEFINE_PORTAL()` macros that handle setting up the types, so that we can continue to clearly distinguish between "plumbing" (`Xdp*`) and actual portal implementations.

src/camera.c
swick: Wasn't sure if there is a semantic difference between creating multiple proxies vs using a single proxy so I left it as it was before. I checked it now and it seems like it should be fine to use a single proxy.

src/account.c
swick: Either way something for a different PR.

src/screenshot.c
swick: Let's address that in another PR.

src/background.c
```diff
@@ -634,58 +643,92 @@ check_background_apps (void)
...
+
 static void
-start_background_monitor (void)
+on_monitor_background_done (GObject      *object,
+                            GAsyncResult *result,
+                            gpointer      user_data)
 {
-  g_autoptr(GThread) thread = NULL;
+  Background *background = (Background *) object;
+  g_autoptr(GTask) task = user_data;
```
AdrianVovk: Don't we pass `NULL` as the `user_data` when we construct the task below?

src/global-shortcuts.c
```diff
@@ -737,23 +777,50 @@ shortcuts_changed_cb (XdpDbusImplGlobalShortcuts *impl,
...
 
+static GlobalShortcuts *
+global_shortcuts_new (XdpDbusImplGlobalShortcuts *impl)
+{
+  GlobalShortcuts *global_shortcuts;
+
+  global_shortcuts = g_object_new (global_shortcuts_get_type (), NULL);
+  global_shortcuts->impl = g_object_ref (impl);
+
+  g_signal_connect (global_shortcuts->impl, "activated",
```
AdrianVovk: `g_signal_connect_object` here? Or disconnect the signal handlers in `_dispose`?

I think using plain `g_signal_connect` was fine before because we'd simply construct the portal once, store it in a global, and then leak it. But now that we're trying to not leak things, and actually destroying our gobjects here, suddenly we need to be careful of the fact that the impls or whatever else might still emit signals _after_ our object got destructed. Otherwise we might crash with a use-after-free

src/clipboard.c
```diff
@@ -549,24 +576,48 @@ selection_owner_changed_cb (XdpDbusImplClipboard *impl,
...
+  Clipboard *clipboard;
+
+  clipboard = g_object_new (clipboard_get_type (), NULL);
+  clipboard->impl = g_object_ref (impl);
+
+  g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (clipboard->impl), G_MAXINT);
+
+  xdp_dbus_clipboard_set_version (XDP_DBUS_CLIPBOARD (clipboard), 1);
+
+  g_signal_connect (clipboard->impl, "selection-transfer",
```
AdrianVovk: Here too, `g_signal_connect_object`

src/camera.c
```diff
@@ -526,39 +500,65 @@ camera_class_init (CameraClass *klass)
...
+  pipewire_socket = g_file_new_for_path (pipewire_socket_path);
+  camera->pipewire_socket_monitor =
+    g_file_monitor_file (pipewire_socket, G_FILE_MONITOR_NONE, NULL, &error);
+  if (!camera->pipewire_socket_monitor)
+    {
+      g_warning ("Failed to track cameras: %s", error->message);
+    }
+  else
+    {
+      g_signal_connect (camera->pipewire_socket_monitor,
```
AdrianVovk: `g_signal_connect_object`, but also it looks like we never actually deallocate the `pipewire_socket_monitor` anywhere?

src/background.c
```diff
@@ -1221,50 +1267,115 @@ background_iface_init (XdpDbusBackgroundIface *iface)
...
+  xdp_dbus_background_set_version (XDP_DBUS_BACKGROUND (background), 2);
+
+  g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (background->impl),
+                                    G_MAXINT);
+
+  g_mutex_init (&background->applications_lock);
+  background->applications = g_hash_table_new_full (g_str_hash, g_str_equal,
+                                                    g_free, instance_data_free);
+
+  g_signal_connect (background->impl, "running-applications-changed",
```
AdrianVovk: g_signal_connect_object here too, and below

src/camera.c
```diff
@@ -194,7 +194,7 @@ handle_access_camera (XdpDbusCamera *object,
                       GVariant *arg_options)
```
AdrianVovk, line 198: This commit probably needs a better message? `context: Rename get_lockdown to get_lockdown_impl` or something like that?

src/xdp-context.c
```diff
@@ -71,6 +71,7 @@ struct _XdpContext
   XdpPortalConfig *portal_config;
```
AdrianVovk, line 71: Ditto with the commit message here too. `context: Store a shared proxy to the Access impl` or something like that

src/camera.c
swick: So, for now I'm just taking a ref and never store the context in the portals.

However, I do think that it should be possible to do so. The context owns the portals, and when the context goes away, the portals get cleaned up before. So this kind of "unowned" reference is actually fine and *should* be used.

However, this is currently probably not working correctly because of the async nature of the dbus dispatches, so I'll keep it like this, but we should clean up the shutdown anyway.

src/input-capture.c
```diff
@@ -1154,74 +1192,74 @@ on_zones_changed_cb (XdpDbusImplInputCapture *impl,
...
+
+  g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (input_capture->impl), G_MAXINT);
+
+  xdp_dbus_input_capture_set_version (XDP_DBUS_INPUT_CAPTURE (input_capture), 1);
+
+  g_object_bind_property (G_OBJECT (input_capture->impl), "supported-capabilities",
+                          G_OBJECT (input_capture), "supported-capabilities",
+                          G_BINDING_SYNC_CREATE);
+
+  g_signal_connect (input_capture->impl, "disabled",
```
AdrianVovk: Also `g_signal_connect_object`, or if you end up passing NULL as the userdata you'd need to disconnect the signal handlers in `_dispose`

src/location.c
```diff
@@ -252,7 +257,8 @@ location_session_start (LocationSession *loc_session)
                 NULL);
   
   g_signal_connect (loc_session->client, "location-updated",
-                    G_CALLBACK (location_updated), loc_session);
+                    G_CALLBACK (on_location_updated),
+                    loc_session);
```
AdrianVovk: Another case for `g_signal_connect_object`? Perhaps it'd be worthwhile to grep through the whole project, to find all the instances of `g_signal_connect` and convert them?

Otherwise I think it's possible for a use-after-free bug if we get unlucky timing and a client closes their session at the same time a location update is emitted by geoclue...

src/memory-monitor.c
```diff
@@ -109,12 +103,33 @@ memory_monitor_class_init (MemoryMonitorClass *klass)
...
+static MemoryMonitor *
+memory_monitor_new (void)
+{
+  MemoryMonitor *memory_monitor;
+
+  memory_monitor = g_object_new (memory_monitor_get_type (), NULL);
+
+#ifdef HAS_MEMORY_MONITOR
+  memory_monitor->monitor = g_memory_monitor_dup_default ();
+  g_signal_connect (memory_monitor->monitor, "low-memory-warning",
```
AdrianVovk: Here too

src/network-monitor.c
```diff
@@ -217,34 +217,64 @@ network_monitor_iface_init (XdpDbusNetworkMonitorIface *iface)
...
+static NetworkMonitor *
+network_monitor_new (void)
+{
+  NetworkMonitor *network_monitor;
+
+  network_monitor = g_object_new (network_monitor_get_type (), NULL);
+
+  network_monitor->monitor = g_network_monitor_get_default ();
+
+  g_signal_connect (network_monitor->monitor, "network-changed",
```
AdrianVovk: You know what I'm gonna say

src/notification.c
```diff
@@ -1235,31 +1260,78 @@ notification_iface_init (XdpDbusNotificationIface *iface)
...
+                          G_OBJECT (notification), "supported-options",
+                          G_BINDING_SYNC_CREATE);
+
+  notification->active =
+    g_hash_table_new_full (pair_hash, pair_equal,
+                           pair_free,
+                           g_free);
+  g_mutex_init (&notification->active_mutex);
+
+  g_signal_connect (notification->impl, "action-invoked",
```
AdrianVovk: Another signal

src/power-profile-monitor.c
```diff
@@ -109,12 +103,36 @@ power_profile_monitor_class_init (PowerProfileMonitorClass *klass)
...
+static PowerProfileMonitor *
+power_profile_monitor_new (void)
+{
+  PowerProfileMonitor *power_profile_monitor;
+
+  power_profile_monitor = g_object_new (power_profile_monitor_get_type (), NULL);
+
+#ifdef HAS_POWER_PROFILE_MONITOR
+  power_profile_monitor->monitor = g_power_profile_monitor_dup_default ();
+  g_signal_connect (power_profile_monitor->monitor,
```
AdrianVovk: Another signal

src/settings.c
```diff
@@ -273,51 +285,70 @@ settings_class_init (SettingsClass *klass)
...
+
+  settings = g_object_new (settings_get_type (), NULL);
+  settings->n_impls = impls->len;
+  settings->impls = (XdpDbusImplSettings **) g_ptr_array_steal (impls, NULL);
+
+  xdp_dbus_settings_set_version (XDP_DBUS_SETTINGS (settings), 2);
+
+  for (size_t i = 0; i < settings->n_impls; i++)
+    {
+      g_signal_connect (settings->impls[i],
```
AdrianVovk: signal

src/usb.c
```diff
@@ -1486,25 +1518,30 @@ xdp_usb_init (XdpUsb *self)
...
+                                                g_free, g_free);
+  usb->sessions = g_hash_table_new (g_direct_hash, g_direct_equal);
+  usb->sender_infos =
     g_hash_table_new_full (g_str_hash, g_str_equal,
                            g_free, (GDestroyNotify) usb_sender_info_unref);
 
-  self->gudev_client = g_udev_client_new (subsystems);
-  g_signal_connect (self->gudev_client,
+  usb->gudev_client = g_udev_client_new (subsystems);
+  g_signal_connect (usb->gudev_client,
```
AdrianVovk: signal

src/usb.c
```diff
@@ -1212,7 +1231,11 @@ handle_acquire_devices (XdpDbusUsb            *object,
   xdp_request_set_impl_request (request, impl_request);
   xdp_request_export (request, g_dbus_method_invocation_get_connection (invocation));
 
-  xdp_dbus_impl_usb_call_acquire_devices (usb_impl,
+
+  g_object_set_data_full (G_OBJECT (request), "-xdp-request-usb",
```
AdrianVovk: Just a nit: looks like you missed this case when renaming (i.e. doesn't match "dynamic-launcher" and "global-shortcuts")

src/background.c
swick: Yikes, good catch! The task is actually the GASyncResult or Background::monitor_task.

src/global-shortcuts.c
swick: Funnily I remember doing exactly that, but somehow I must have lost the changes. Anyway, grepped through it and changed it where possible; the remaining instances are connecting to a signal on themselves.

src/camera.c
swick: Good catch about the `pipewire_socket_monitor`.

src/xdp-context.c
swick: You know the PR is getting too big when you don't even notice your own placeholder commit messages...

src/input-capture.c
swick: passing input_capture as the object with `g_signal_connect_object` now, but ignoring it in the callbacks.

PR #1823, opened 2025-09-27
Notification: Fix and document some v1 backward incompatbilities

3v1n0:
See commit for details, but in short:
 - Expose to applications the actual portal version they should target
 - Fix backward incompatibility handling of icon property
 - Document the new properties in v2

/cc @jsparber


data/org.freedesktop.portal.Notification.xml
```diff
@@ -94,11 +94,14 @@
...
           be wrapped or/and truncated.
 
+          Since: version 2
+
         * ``icon`` (``v``)
 
-          A serialized icon to add to the notification. The icon must pass
-          `icon validation <icons.html>`_ in order to be used. The format for
-          serialized icon is a tuple (sv) with the following supported keys:
+          Since version 2, a serialized icon to add to the notification.
```
swick: Shouldn't this not also be `Since: version 2` below?

src/notification.c
```diff
@@ -1268,12 +1269,12 @@ notification_create (GDBusConnection *connection,
 
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (impl), G_MAXINT);
 
+  version = g_dbus_proxy_get_cached_property (G_DBUS_PROXY (impl), "version");
```
swick: we do have the codegen'ed property accessors

e: noticed this is just moved, but still...

data/org.freedesktop.portal.Notification.xml
```diff
@@ -116,9 +116,7 @@
             Since version 2, this is deprecated and should not be used.
             Please use the `themed` or `file-descriptor` option to set an icon.
 
-            This is the same format as a serialized `GBytesIcon
-            <https://docs.gtk.org/gio/class.BytesIcon.html>`_ at the moment,
-            but this interoperability may change in the future.
+            This is the icon bytes into a serialized variant bytes array.
```
swick: I now know nothing about the particular encoding of it. Isn't there somewhere we can link to for it?

swick:
Thanks a lot! Mostly nitpicks.

data/org.freedesktop.portal.Notification.xml
3v1n0: Yeah, it was a bit problematic to include since the paragraph is quite long, however all the fields are actually supported, so so we can just keep the version as it is.

data/org.freedesktop.portal.Notification.xml
3v1n0: Not that I'm aware of, since technically the implementation is also private, it just _happens_ to match the protocol implementation, reason why I am also changing the tests so that we don't rely on GIcon there, since it's very much platform specific.

3v1n0: The best we have is https://docs.gtk.org/gio/method.Icon.to_string.html though

swick: Well, linking there would still be better than nothing, no?

3v1n0: done

PR #1825, opened 2025-09-27
main: Initialize option booleans

AdrianVovk:
Otherwise, if the uninitialized memory isn't zero'd out for whatever reason, these booleans will magically become true!

On my system, opt_show_version consistently initializes to something non-zero and thus xdg-desktop-portal just prints its version number and exits


swick:
Yikes, my fault. They were static before and that guaranteed initialization to 0.

swick:
I really have to figure out why the screenshot test is to damn flaky in CI

PR #1828, opened 2025-10-05
ProxyResolver 2.0, with Proxy portal backend

### LINKED ISSUE #554: Improving Proxy support
While working on implementing proxy support to respect system settings for the [Chromium flatpak](https://github.com/flathub/org.chromium.Chromium), I've noticed a few things that are currently missing/not covered by the portal APIs and that could be improved.

Afaics (please correct me if I'm wrong), the current way to use system proxy settings from a portal is through the [ProxyResolver portal interface](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.ProxyResolver).
The current version of the `ProxyResolver` interface (v1) only supports one method (`GetProxyForURL`) and has some limitations:
- No way to check if proxy is enabled in the system, which could be used to avoid invoking the portal to resolve proxies altogether if disabled
- No change notification signal, making it hard to implement a proper cache mechanism client side
- No access to actual settings (enabled/disabled, automatic/manual, pac_url, manual urls, ignore-hosts), forcing the client to use the portal for proxy resolution
- No support to launch the system settings UI and no write access to settings ^

Not having access to the actual settings could arguably be fine, as long as we have a way to reliably implement client caching and to skip the (D-Bus) calls to `ProxyResolver.GetProxyForURL` altogether when proxy is disabled in the system.

Not being able to launch the system settings UI is also Ok, the client could always rely on the user using the system settings UI directly and simply not support launching it, but that lowers the user experience. For example regular non-flatpak chromium linux builds have a link in its settings to launch the system settings UI, but that just spawns a command directly (e.g. `gnome-control-center ...` when running on GNOME). There are possible workarounds to make this work, but maybe we should consider an API for it?

With that in mind, I have a few ideas on how to improve here:
1. Add API to access actual settings, with change notification
   - This could be done either as part of the ProxyResolver interface or as a new interface 
2. Add a method/property to the ProxyResolver interface to check if proxy is enabled and also a change notification signal trigerred when any of the settings change (including enabled/disabled)

In the chromium case, ideally we would have access to 1) so chromium could use its internal proxy mechanism based on the system settings retrieved using the portal, but 2) would already be a great improvement and allow for a proper cache mechanism and to avoid unnecessary D-Bus calls when the proxy is disabled. We could also provide both worlds, but that may be overkill...

Also worth considering possible privacy concerns if we end up implementing something along the lines of 1) above.

Thoughts?

orowith2os:
This is a skeleton to test the waters to resolve https://github.com/flatpak/xdg-desktop-portal/issues/554.

I experimented with implementing this full API in libproxy, and determined that it would be easier to have portal backends provide the information via the Proxy backend portal, and have libproxy build on top of that. Browsers and other semi-advanced use cases can read the Portal properties directly if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 __pycache__
+.buildconfig
--- a/data/org.freedesktop.portal.ProxyResolver.xml
+++ b/data/org.freedesktop.portal.ProxyResolver.xml
@@ -48,7 +48,9 @@
         protocol. ``direct://`` is used when no proxy is needed.
 
         This is a convenience wrapper over reading the PAC file or the other
-        portal settings, in that order of priority from highest to lowest.
+        portal settings, in that order of priority from highest to lowest. It
+        is provided to ensure backwards compatibility with older versions
+        of GLib.
     -->
     <method name="Lookup">
       <arg type="s" name="uri" direction="in"/>
@@ -76,14 +78,9 @@
     -->
     <property name="http_proxy" type="as" access="read"/>
     <!--
-      socks_proxy:
-      An integer value of 0 is equivalent to SOCKS; of 4 is SOCKS4,
-      and 5 is SOCKS5. Any other value is invalid, but if present, can
-      be assumed to be equivalent to 0.
-
       This property was added in version 2.
     -->
-    <property name="socks_proxy" type="a(su)" access="read"/>
+    <property name="socks_proxy" type="as" access="read"/>
     <!--
       This property was added in version 2.
     -->
--- a/data/org.freedesktop.portal.ProxyResolver.xml
+++ b/data/org.freedesktop.portal.ProxyResolver.xml
@@ -1,6 +1,7 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
  Copyright (C) 2016 Red Hat, Inc.
+ Copyright (C) 2025 Dallas Strouse
 
  SPDX-License-Identifier: LGPL-2.1-or-later
 
@@ -18,8 +19,9 @@
  License along with this library. If not, see <http://www.gnu.org/licenses/>.
 
  Author: Matthias Clasen <mclasen@redhat.com>
+ Author: Dallas Strouse @ v2 <dallas.strouse2007@mail.com>
 -->
-<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+<node xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd" name="/">
   <!--
       org.freedesktop.portal.ProxyResolver:
       @short_description: Proxy information
@@ -29,7 +31,10 @@
       user interaction. Applications are expected to use this interface indirectly,
       via a library API such as the GLib GProxyResolver interface.
 
-      This documentation describes version 1 of this interface.
+      This documentation describes version 2 of this interface.
+
+      If pac_url exists, it should be the only setting read;
+      if it doesn't, the associated dbus properties should be utilized.
   -->
   <interface name="org.freedesktop.portal.ProxyResolver">
     <!--
@@ -41,11 +46,53 @@
         proxy uri are of the form ``protocol://[user[:password] AT host:port``.
         The protocol can be ``http``, ``rtsp``, ``socks`` or another proxying
         protocol. ``direct://`` is used when no proxy is needed.
+
+        This is a convenience wrapper over reading the PAC file or the other
+        portal settings, in that order of priority from highest to lowest.
     -->
     <method name="Lookup">
       <arg type="s" name="uri" direction="in"/>
       <arg type="as" name="proxies" direction="out"/>
     </method>
+    <!--
+    The following values are valid:
+      0: Proxy should not be used.
+      1: Automatic proxy mode; use pac_url.
+      2: Manual proxy mode; read associated settings manually.
+
+      This property was added in version 2.
+    -->
+    <property name="use_proxy" type="u" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="pac_url" type="s" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="https_proxy" type="as" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="http_proxy" type="as" access="read"/>
+    <!--
+      socks_proxy:
+      An integer value of 0 is equivalent to SOCKS; of 4 is SOCKS4,
+      and 5 is SOCKS5. Any other value is invalid, but if present, can
+      be assumed to be equivalent to 0.
+
+      This property was added in version 2.
+    -->
+    <property name="socks_proxy" type="a(su)" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="ftp_proxy" type="as" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="ignored_hosts" type="as" access="read"/>
     <property name="version" type="u" access="read"/>
   </interface>
 </node>
+
--- a/src/xdp-context.c
+++ b/src/xdp-context.c
@@ -308,6 +308,11 @@ xdp_context_register (XdpContext *context,
     export_portal_implementation (connection,
                                   print_create (connection, impl_config->dbus_name, lockdown));
 
+  impl_config = xdp_portal_config_find (portal_config, "org.freedesktop.impl.portal.Notification");
+  if (impl_config != NULL)
+    export_portal_implementation (connection,
+                                  proxy_resolver_create (connection, impl_config->dbus_name));
+
   impl_config = xdp_portal_config_find (portal_config, "org.freedesktop.impl.portal.Proxy");
   if (impl_config != NULL)
     export_portal_implementation (connection,
--- a/src/xdp-context.c
+++ b/src/xdp-context.c
@@ -19,7 +19,6 @@
 
 #include "config.h"
 
-#include "xdp-utils.h"
 #include "xdp-call.h"
 #include "xdp-dbus.h"
 #include "xdp-documents.h"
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
 __pycache__
+.buildconfig
```
mcatanzaro, line 2: Life pro tip: put your IDE's stuff into `~/.config/git/ignore` :)

src/xdp-context.c
```diff
@@ -309,10 +308,13 @@ xdp_context_register (XdpContext       *context,
     export_portal_implementation (connection,
                                   print_create (connection, impl_config->dbus_name, lockdown));
 
-  impl_config = xdp_portal_config_find (portal_config, "org.freedesktop.impl.portal.Notification");
```
mcatanzaro, line 312: Removing org.freedesktop.impl.portal.Notification was an accident, right?

src/xdp-context.c
```diff
@@ -239,9 +239,9 @@ on_peer_died (const char *name)
 }
 
 gboolean
-xdp_context_register (XdpContext       *context,
-                      GDBusConnection  *connection,
-                      GError          **error)
+xdp_context_register (XdpContext *context,
+                      GDBusConnection *connection,
+                      GError **error)
```
mcatanzaro, line 244: Is removing the spaces normal code style for xdg-desktop-portal? Probably not? Here you are getting rid of GNOME's standard style for parameter lists.

src/xdp-context.c
```diff
@@ -207,7 +207,7 @@ export_host_portal_implementation (GDBusConnection        *connection,
    * method calls must see the modified value.
    */
 
-  g_autoptr(GError) error = NULL;
+  g_autoptr (GError) error = NULL;
```
mcatanzaro, line 210: GNOME is not consistent on this, but xdg-desktop-portal is: no space in the g_autoptr. I don't know why tbh; everywhere else, we _do_ use a space before opening parentheses....

src/xdp-context.c
```diff
@@ -44,21 +44,21 @@
...
-#include "xdp-request.h"
 #include "screen-cast.h"
 #include "screenshot.h"
 #include "secret.h"
 #include "settings.h"
 #include "trash.h"
 #include "usb.h"
 #include "wallpaper.h"
+#include "xdp-permissions.h"
+#include "xdp-request.h"
```
mcatanzaro, line 61: Since you're reordering these, why can't they go into the list of other xdp headers?

data/org.freedesktop.portal.ProxyResolver.xml
```diff
@@ -29,7 +31,10 @@
       user interaction. Applications are expected to use this interface indirectly,
       via a library API such as the GLib GProxyResolver interface.
 
-      This documentation describes version 1 of this interface.
+      This documentation describes version 2 of this interface.
+
+      If pac_url exists, it should be the only setting read;
+      if it doesn't, the associated dbus properties should be utilized.
```
mcatanzaro, line 37:
```suggestion
      if it doesn't, the associated D-Bus properties should be utilized.
```

data/org.freedesktop.portal.ProxyResolver.xml
```diff
@@ -41,11 +46,53 @@
...
+    -->
+    <property name="ftp_proxy" type="as" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="ignored_hosts" type="as" access="read"/>
     <property name="version" type="u" access="read"/>
   </interface>
 </node>
+
```
mcatanzaro, line 95: I strongly suspect some bug is Builder is causing it to sometimes add blank lines at the ends of files; perhaps there is some bug in its code to ensure the file ends in a newline character, which is actually useful. We shouldn't be committing blank lines at end of file.

data/org.freedesktop.portal.ProxyResolver.xml
```diff
@@ -41,11 +46,53 @@
...
+    -->
+    <property name="use_proxy" type="u" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="pac_url" type="s" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="https_proxy" type="as" access="read"/>
```
mcatanzaro, line 75: Why is this an array? Same question for: http_proxy, socks_proxy, and ftp_proxy. The underlying GNOME settings are not arrays. Is your expectation that apps will try one and fall back to the next if it fails? I think applications won't be prepared to do this.

data/org.freedesktop.portal.ProxyResolver.xml
```diff
@@ -41,11 +46,53 @@
...
+      1: Automatic proxy mode; use pac_url.
+      2: Manual proxy mode; read associated settings manually.
+
+      This property was added in version 2.
+    -->
+    <property name="use_proxy" type="u" access="read"/>
+    <!--
+      This property was added in version 2.
+    -->
+    <property name="pac_url" type="s" access="read"/>
```
mcatanzaro, line 71: This isn't enough to implement GNOME proxy settings because `pac_url` is optional in GNOME, to allow web proxy autodiscovery. You could use a maybe type `ms` to allow for this.

mcatanzaro (changes requested):
I'm not sure about adding direct access to all the settings. I predict that all or nearly all apps that try to use them will implement them improperly or incompletely, since there are subtleties. E.g. libproxy does not interpret the settings the same way that glib-networking itself does; if you look at glib-networking/proxy/tests/common.c you'll find that libproxy only passes about half the `ignored_hosts` tests. So already the two libraries that are supposed to be our sources of truth for how to interpret the settings do not agree; what hope do individual applications have?

In https://github.com/flatpak/xdg-desktop-portal/issues/554 Chrome wants to build a client-side cache, presumably keyed on hosts, but to be correct it would have to be keyed on _URLs_ since PAC allows you to configure different proxies for different URLs, and in practice corporations do actually depend on this. I think it's pretty unlikely that application developers would try to build a cache that's _actually_ keyed on URLs. This raises the question of whether providing direct settings access is truly useful. CC @andrunko

We should at least document very clearly how these settings are all expected to work, so applications (and also portal backends!) have at least a chance to implement them consistently. E.g. [this blog post](https://about.gitlab.com/blog/we-need-to-talk-no-proxy/) describes considerations for parsing proxy config environment variables. Fortunately we don't have to work with environment variables. And GSimpleProxyResolver has enough documentation; you could copy from that.

We should also make sure this is good enough to implement KDE proxy settings. (Probably it is?)

A simpler counterproposal: you could expose just `use_proxy` property. This seems lke the most important part by far, since it allows applications to skip over all other use of the portal if false, and normally it will be false. GProxyResolverPortal (and also Chrome) could rely on that to know that it doesn't need to use the portal at all, and subscribe to a standard property change notification to know when this changes.

mcatanzaro:
> In https://github.com/flatpak/xdg-desktop-portal/issues/554 Chrome wants to build a client-side cache, presumably keyed on hosts, but to be correct it would have to be keyed on URLs since PAC allows you to configure different proxies for different URLs, and in practice corporations do actually depend on this. I think it's pretty unlikely that application developers would try to build a cache that's actually keyed on URLs. This raises the question of whether providing direct settings access is truly useful. CC @andrunko

Actually, this comment is wrong, for two reasons:

 * PAC can depend on _anything_, e.g. time of day. Any attempt to build a cache if PAC is enabled would be incorrect.
 * But Chrome -- and GProxyResolverPortal -- would presumably just check the settings to see whether PAC is enabled or not, and use a cache if not. So this could save a lot of D-Bus activity in the non-PAC case.

So, yeah, I change my mind already: I like exposing all the settings after all. I still wouldn't want to _encourage_ applications to attempt to interpret the settings themselves, but allowing GProxyResolverPortal to cache the proxy settings so it doesn't have to keep doing D-Bus again and again and again is definitely useful.

src/xdp-context.c
orowith2os: That's automatic formatting on behalf of Builder. I'll apply only my code changes when this is ready for merging.

src/xdp-context.c
orowith2os: Formatting by Builder.

src/xdp-context.c
orowith2os: Builder formatting. Not going to be present in the final draft.

data/org.freedesktop.portal.ProxyResolver.xml
orowith2os: An empty string is the same as the string not being present. That's how it plays out in dconf. I'll keep it the same here. There's no semantic difference between NULL and "" here, I don't think.

mcatanzaro:
> As for pac_url: an empty string is fine, no? It's the same result in the end: don't display it to the user, and a pac url isn't available. It makes handling a bit simpler. 

Well, as long as you document it carefully, then it should work.

But I would use a maybe type. Empty strings in APIs are confusing and prone to error.

mcatanzaro:
> As written in Matrix channel i don't think it is a good idea to share all information. Just the tristate should be enough as you cannot create a cache for some proxy types, but Michael already wrote it here. So IMHO the API should be just one information about the state, if ever.

Maybe four states would be better:

0. No proxy enabled (no need for further D-Bus calls)
1. Proxy enabled, results not cacheable (PAC)
2. Proxy enabled, results cacheable per-host (e.g. standard configuration with ignored hosts)
3. Proxy enabled, results cacheable per protocol (standard configuration without ignored hosts)

That should provide enough information to implement caching _without_ exposing the underlying settings, which would encourage applications to inadvisably attempt to interpret the settings.

mcatanzaro:
> Proxy enabled, results cacheable per-host (e.g. standard configuration with ignored hosts)

Most applications would probably not bother to do anything with this state, but they could if they wanted to.

mcatanzaro:
> Most applications would probably not bother to do anything with this state, but they could if they wanted to.

And notably, if GProxyResolverPortal grows such a cache, that would cover most GNOME applications all at once.

PR #1830, opened 2025-10-07
Launch service after graphical session

jadahl:
This means e.g. $WAYLAND_DISPLAY and/or $DISPLAY will reliably be something valid, which is necessary to make e.g. OpenURI work when launching non-dbus-activated apps.

This fixes an issue where Firefox would not open links due to no display environment variable being set, if xdg-desktop-portal launched before the display server.

### DIFF
--- a/src/xdg-desktop-portal.service.in
+++ b/src/xdg-desktop-portal.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=Portal service
 PartOf=graphical-session.target
-Requires=dbus.service
-After=dbus.service
+Requisite=graphical-session.target
+After=graphical-session.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


rmoehn:
This broke screensharing in my labwc setup in a way that would have been hard to diagnose without Claude Code. (IOW screensharing suddenly didn't work anymore and I had no clue why.) The fix is to add this to `autostart`:

```
systemctl --user --no-block start labwc-session.target
```

And, for hygiene, this to `shutdown`, I suppose:

```
systemctl --user stop graphical-session.target
```

Credits: https://labwc.github.io/labwc.1.html

smcv:
@rmoehn, that sounds like another version of #1983. See also #2027.

PR #1831, opened 2025-10-09
Revert "notification: Add platform-data and response to ActionInvoked signal"

3v1n0:
Sadly reusing the old activation API for v2 by adding more values to the
parameters array was not working well enough for various reasons:
 - We used a DBus return value as a container for multiple (unrelated)
   and generic return values without being able to reuse any kind of type
   check provided by the platform.
 - Despite the API documented these return values as an ordered list of
   items with (ideally) a fixed position, it was not possible to
   maintain such promise in case of an empty parameter was provided by
   the app or - in future - an empty (!= no) response was sent.
   Note how `response` was referenced as the third parameter of such
   array, while this might have not been possible.
 - It was confusing for both applications using it and portal
   implementing the API, as it was not clear for the portals
   implementations what to pass back to the application when no parameter
   was received or what the application should expect to receive in such
   case as the first array value.

So, let's go back to the v1 state, as this will allow us to provide a
proper ActionInvoked2 implementation.

This also implies breaking the `im.reply-with-text` category behavior
though, since that's affected by the very same problem.

We're doing this without changing the portal version though, as this
won't break applications expecting a v2 portal anyways.

This reverts commit 3aedb5fbba0a1ef7ee49b5fdbe33cd13508024e9.
This partially reverts commit def2ad7412045fe3382f06d9220916b705624eaa

--- 

That said, this is not as easy as it should since `im.reply-with-text` opens the gate to another problematic sitaution, reason why I think we should also modify `ActionInvoked2` to not send the parameters as `av` at all, but rather as `s{av}` and same for the exported actions: we just send them back an `a{sv}` with `target` and `response` for now.

### DIFF
--- a/data/org.freedesktop.portal.Notification.xml
+++ b/data/org.freedesktop.portal.Notification.xml
@@ -290,7 +290,7 @@
           Additional categories and button purposes may be defined by notification servers
           using ``x-vendor`` prefix e.g. ``x-gnome.class.specific``
 
-          The following categories are standarized so far:
+          The following categories are standardized so far:
 
           * ``im.received``
 


3v1n0:
> the code that handles org.freedesktop.Application is completely different from the code that might use the portal to send a notification.

> I don't see how this could work.

Well, it would just work in the same way it's already working, no?
See: https://gitlab.gnome.org/3v1n0/glib/-/commits/gapplication-dict-action-parameter

With [GLib's commit `0144feb41`](https://gitlab.gnome.org/GNOME/glib/-/commit/0144feb41fc1819cec6415483bd8a61e64373aca) the signature of that value already changed from being just a singla `GVariant` value to a tuple in case multiple values are passed, so while that was more conservative it was already making such parameter value "unstable".

Now, IMHO we could just switch to something else that is a bit more safer in this regards than returning a tuple of undefined lenght whose elements are defined by their index

swick:
We could maybe make it easier for us by having a new method to set the "used" version. Currently only the portal will tell the client its version, but what we need here is essentially the client opting-in a specific version, so we can just say that if the client opts-in to version 2, the `av` in `ActivateAction` contains a single variant which is a `a{sv}`.

Then it becomes rather easy because if we operate on a version 2 client, we always have a `a{sv}`, and the docs can equally just say assume that.

3v1n0:
> The problem I see is that the portal can't just change something in a new version because some clients will only support older versions. So you'd have to switch the signature based on if the notification contains something from version 2. That seems like a recipe for disaster, and people will get it wrong.
> 

Well... Don't think it's a portal problem, if we document things properly TBH. Not to mention that "disaster" is very unlikely to happen as really I'm not aware of any implementation taking care of the "target" parameter.

> We could maybe add a version field to the `notification` vardict instead?

That's maybe better... But, we may then have the other problem: applications not setting the version and thus they'll stick to version 1, that's probably not a big deal if they don't care about the new fields we send them back though.

PR #1840, opened 2025-10-24
Add reduced motion setting to the portal

ebassi:
A setting for minimising the amount of non-essential motion in animations, to avoid discomfort for people with vestibular motion disorders, or distraction for those with attention deficits.

The setting key is modelled on the contrast key: an unsigned value with two states.

See also: https://www.w3.org/TR/mediaqueries-5/#prefers-reduced-motion


danirabbit:
Adding my ack for Pantheon. Thank you!

PR #1848, opened 2025-10-29
Globalshortcuts clarification

Sodivad:
Add some implementation recommendations based on what GNOME and Plasma do to not show endless dialogs.

Clarify ListShortcuts as discussed with @swick

### DIFF
--- a/data/org.freedesktop.portal.GlobalShortcuts.xml
+++ b/data/org.freedesktop.portal.GlobalShortcuts.xml
@@ -155,7 +155,10 @@
         @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object representing the session
         @options: Vardict with optional further information
 
-        Lists all shortcuts.
+        Lists shortcuts of a session.
+
+        If `BindShortcuts` was called for `session` all active shortcuts for `session` are returned.
+        Otherwise returns the shortcuts that were sucessfully bound in a previous session by this application.
 
         Supported keys in the @options vardict include:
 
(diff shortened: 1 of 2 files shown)


data/org.freedesktop.portal.GlobalShortcuts.xml
```diff
@@ -155,7 +155,10 @@
         @session_handle: Object path for the :ref:`org.freedesktop.portal.Session` object representing the session
         @options: Vardict with optional further information
 
-        Lists all shortcuts.
+        Lists shortcuts of a session.
+
+        If `BindShortcuts` was called for `session` all active shortcuts for `session` are returned.
+        Otherwise returns the shortcuts that were sucessfully bound in a previous session by this application.
```
garnacho, line 161: I notice I am late for feedback, but I think this deviates from the original intent, where BindShortcuts is not an optional step. Per that, the client must always submit the requested bindings, and the response may change (i.e. be different subsets) depending on runtime conditions between runs (e.g. the user remapped overlapping shortcuts at the compositor level).

I gather the clarifications are overall in order to allow the xdp implementation to avoid showing the dialog if the bound shortcuts did not change, but IMO this is a corner that sandboxed applications should not cut.

Sodivad: The clarification to ListShortcuts was not intended to drop the need for BindShortcuts, just to define what happens if it's called before BindShortcuts. 

We noticed that it was not defined and there was diverging behavior.

PR #1849, opened 2025-10-29
print: add two new capabilities (n-copies and number-up)

lbaudin:
Both the app and the printer may be able to do copies. If the printer can do it, then it is likely more efficient and it should not be done in the app, so the app needs to know that the printer will do it.

Since the app may save print settings, we cannot just hide number-up and n-copies from the app, so we need both the app and the printer to advertise such capabilities if they are available.

Currently, this is not handled and n-copies is applied twice (so in GNOME documents are printed n-copies^2 times).

Related GNOME MR: https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/233

### DIFF
--- a/data/org.freedesktop.portal.Print.xml
+++ b/data/org.freedesktop.portal.Print.xml
@@ -34,7 +34,7 @@
       to print the formatted document. It is expected that high-level toolkit
       APIs such as GtkPrintOperation will hide most of this complexity.
 
-      This documentation describes version 4 of this interface.
+      This documentation describes version 5 of this interface.
   -->
   <interface name="org.freedesktop.portal.Print">
     <!--
@@ -85,6 +85,19 @@
           Whether it makes sense to return "selection" for the ``print-pages`` setting.
 
           This option was added in version 4.
+
+        * ``has_n_copies`` (``b``)
+
+          Whether the app is able to manually implement the ``n-copies`` and ``collate`` settings.
+
+          This option was added in version 5.
+
+        * ``has_number_up`` (``b``)
+
+          Whether the app is able to manually implement the ``number-up`` and ``number-up-layout`` setting.
+
+          This option was added in version 5.
+
 
         The @settings and @page_setup vardict contain hints for the initial state of the print dialog.
 
@@ -255,6 +268,23 @@
           Print settings as set up by the user in the print dialog. The same keys as in @settings
           are supported (see above).
 
+        * ``printer-info`` (``a{sv}``)
+
+          Printer information with the following keys:
+
+          * ``has-n-copies`` (``b``)
+
+            Whether the printer has the n-copies capability. In this case, the app should not provide
+            a document with multiple copies, even if it has set has_n_copies to true in @options since it
+            is going to be handled directly by the printer.
+           
+          * ``has-number-up`` (``b``)
+
+            Whether the printer has the n-up capability. In this case, the app should not provide
+            a document with pages per side if it has set has_n_up to true in @options.
+            
+          This option was added in version 5.
+           
         * ``page-setup`` (``a{sv}``)
 
           Page setup as set up by the user in the print dialog. The same keys as in @page_setup
--- a/src/print.c
+++ b/src/print.c
@@ -210,6 +210,7 @@ handle_print (XdpDbusPrint *object,
 static XdpOptionKey response_options[] = {
   { "settings", G_VARIANT_TYPE_VARDICT, NULL },
   { "page-setup", G_VARIANT_TYPE_VARDICT, NULL },
+  { "printer-info", G_VARIANT_TYPE_VARDICT, NULL },
   { "token", G_VARIANT_TYPE_UINT32, NULL }
 };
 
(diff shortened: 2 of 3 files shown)


swick:
have to bump the version of the frontend `xdp_dbus_print_set_version` (which we apparently forgot to do for v4).

data/org.freedesktop.portal.Print.xml
```diff
@@ -86,6 +86,19 @@
 
           This option was added in version 4.
 
+        * ``has_n_copies`` (``b``)
+
+          Whether the app is able to manually implement the n-copies and collate settings.
```
jadahl:
```suggestion
          Whether the app is able to manually implement the ``n-copies`` and ``collate`` settings.
```

data/org.freedesktop.portal.Print.xml
```diff
@@ -86,6 +86,19 @@
...
+
+          This option was added in version 5.
+
+        * ``has_number_up`` (``b``)
+
+          Whether the app is able to manually implement the number-up and number-up-layout setting.
```
jadahl:
```suggestion
          Whether the app is able to manually implement the ``number-up`` and ``number-up-layout`` setting.
```

data/org.freedesktop.portal.Print.xml
```diff
@@ -254,7 +267,22 @@
 
           Print settings as set up by the user in the print dialog. The same keys as in @settings
           are supported (see above).
```
jadahl, line 269: Add an empty line above `* ``printer-info``

data/org.freedesktop.portal.Print.xml
```diff
@@ -254,7 +267,22 @@
...
+        * ``printer-info`` (``a{sv}``)
+
+          Printer information with the following keys:
+
+          * ``has-n-copies`` (``b``)
+
+            Whether the printer has the n-copies capability. In this case, the app should not provide
+            a document with multiple copies if it has set has_n_copies to true in @options.
```
jadahl: I suppose it's clear enough this is not something that can be assumed by the backend? The backend can't assume the app will know about these new properties.

lbaudin: Thanks for the review.

This is backwards compatible in the sense that for an app which ignore the new properties (and then is currently already broken):

- if the app makes copies when n-copies > 1:
   - if the printer does not support n-copies, then it worked before and still works
   - if the printer does support n-copies, then we had n-copies^2 copies, and we will still have that number (so it is broken in the same way, see https://gitlab.gnome.org/GNOME/gtk/-/issues/7605)
- if the app does not make copies:
   - if the printer does not support n-copies, then it is broken but it was broken before
   - if the printer does support n-copies, then it worked before and still works

For an app which supports the new properties, then: 

- if the app makes copies when n-copies > 1, it sets `has_n_copies` to true in options, so:
   - if the printer does not support n-copies, `has-n-copies` is equal to false and it should copy
   - if the printer supports it, then `has-n-copies` is equal to true and it should not copy
- if it does not, it sets `has_n_copies` to false, so:
   - if the printer does not support n-copies, then the portal dialog should not allow for n-copies to be greater than 1
   - if it does, then the app is not going to copy anyway, I just slightly edited the documentation to remove any ambiguity

cjao: Do you know if any other printer capabilities should be forwarded back to the app for it to render its pages correctly? For instance, starting from [here](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/print/backends/gtkprintbackendcups.c#L6660), the Gtk CUPS print backend manipulates the following properties of the print job which in turn [propagate back to the print operation](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/print/gtkprintoperation-unix.c#L638-L646) before the app starts rendering pages:
- collate
- reverse
- copies (already handled by your PR)
- scale
- page_set
- n_up (handled by your PR)

lbaudin: For now I would like to solve n-copies (including collate) and number-up. These are especially important in order to not send too many pages to the printer.
For the others: scale needs more thought anyway, as we discussed it in the papers issue. ATM the app is responsible for page set. Reverse should be the same I suppose, but I think there are issues in the GNOME portal. This is for another day though.

cjao: Wouldn't it be better to collect as many related fixes as possible into a single portal API revision instead of spreading them out across multiple portal API versions and increasing the test matrix for every Flatpak app? The problem you're solving is that apps in Flatpak sandboxes can't render printed pages correctly if they aren't aware of the printer's native capabilities. So why stop at `"n-copies"` and `"number-up"`?

Btw I just checked in Papers -- "reverse" also doesn't work through the portal. The document is reversed first by the app and then [by the cups backend](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/print/backends/gtkprintbackendcups.c#L6676); just as with n-copies, in the portal situation the cups backend can't directly mutate the app's print job to unset the "reverse" property if the printer can natively reverse its input..

lbaudin: Well, as I said: these others arguments do not necessarily need a change in the protocol or need to thought about carefully. There is zero overhead with reverse so the problem can be solved differently. Scale is more complicated (more difficult to get everyone on board with shrink/expand). Given that there is not exactly a crowd of maintainers for the printing stack, let's keep this simple first.

Furthermore this is a first step to see if everyone agrees on the approach, we could do scale+autorotate next.

jadahl: Also the version isn't fixed until a release, so while this PR might not add all possible cases, one can still add more of them before needing to bump the version again.

cjao: Fair enough! Printing has long been an awkward aspect of the portals and I'm excited for these improvements.

lbaudin:
Hi, this is my monthly ping about that! :smile: 
More seriously, sorry for the noise and I know that everyone is busy, but it would really be great to get this in so as I can then try to get the two fixes in the GNOME portal and in GTK (hopefully for GNOME 50) for the very annoying printing bugs of n-copies^2 and number-up^2.

PR #1853, opened 2025-11-04
tests: Only run tests with the host and flatpak app info by default

swick:
Every AppInfo kind multiplies the number of tests. Let's run the tests with the two which will give us the biggest coverage by default.

By setting `XDP_TEST_RUN_LONG`, all tests run with all AppInfo kinds.

In CI, we run the tests 3 times (test in the build dir, installed tests, dist). Let's run "main" tests with XDP_TEST_RUN_LONG=1 to get the full coverage while keeping the other tests short.


swick:
Going to merge in a few days if there are no objections.

PR #1854, opened 2025-11-05
app-info-registry: Only ever create one app info per connection

swick:
We dispatch invocations in threads, including the authorize-method which creates the app info for a connection. A call to
xdp_app_info_registry_ensure_for_invocation_sync however is only locking the hashtable when reading or writing to it; this means that the sequence of lookup, create new, insert is racy, and another thread can create and insert a new app info between our lookup and insert.

So far this has not been a problem because there was nothing hanging off the app info and equality checks went via the app id.

This will change, so we really have to make sure the whole lookup, create, insert sequence is atomic.

We could just hold the global lock for it, but it would mean any other method invocation, even from other connection, will block until the app info has been created. Creating it requires a roundtrip through the dbus broker.

So this takes another approach and uses a lock per connection.

---

I'm not very happy with the implementation, so if anyone has good ideas, I'm all ears.

### DIFF
--- a/src/xdp-app-info-registry.c
+++ b/src/xdp-app-info-registry.c
@@ -23,17 +23,70 @@
 
 #include "xdp-app-info-registry.h"
 
+typedef struct _XdpAppInfoRegistryItem
+{
+  gatomicrefcount ref_count;
+  XdpAppInfo *app_info;
+  GMutex mutex;
+} XdpAppInfoRegistryItem;
+
 struct _XdpAppInfoRegistry
 {
   GObject parent_instance;
 
-  GHashTable *app_infos; /* unique dbus name -> app info */
+  GHashTable *app_infos; /* unique dbus name -> XdpAppInfoRegistryItem */
   GMutex app_infos_mutex;
 };
 
 G_DEFINE_FINAL_TYPE (XdpAppInfoRegistry,
                      xdp_app_info_registry,
                      G_TYPE_OBJECT)
+
+static XdpAppInfoRegistryItem *
+xdp_app_info_registry_item_new (void)
+{
+  XdpAppInfoRegistryItem *item;
+
+  item = g_new0 (XdpAppInfoRegistryItem, 1);
+  g_mutex_init (&item->mutex);
+  g_atomic_ref_count_init (&item->ref_count);
+
+  return item;
+}
+
+static XdpAppInfoRegistryItem *
+xdp_app_info_registry_item_ref (XdpAppInfoRegistryItem *item)
+{
+  g_atomic_ref_count_inc (&item->ref_count);
+
+  return item;
+}
+
+static void
+xdp_app_info_registry_item_unref (XdpAppInfoRegistryItem *item)
+{
+  if (g_atomic_ref_count_dec (&item->ref_count))
+    {
+      g_clear_object (&item->app_info);
+      g_mutex_clear (&item->mutex);
+      g_free (item);
+    }
+}
+
+static void
+xdp_app_info_registry_item_lock (XdpAppInfoRegistryItem *item)
+{
+  g_mutex_lock (&item->mutex);
+}
+
+static void
+xdp_app_info_registry_item_unlock (XdpAppInfoRegistryItem *item)
+{
+  g_mutex_unlock (&item->mutex);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (XdpAppInfoRegistryItem,
+                               xdp_app_info_registry_item_unref)
 
 static void
 xdp_app_info_registry_dispose (GObject *object)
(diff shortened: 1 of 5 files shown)


src/xdp-app-info-registry.h
```diff
@@ -30,16 +30,26 @@ G_DECLARE_FINAL_TYPE (XdpAppInfoRegistry,
                       XDP, APP_INFO_REGISTRY,
                       GObject)
 
+typedef struct _XdpAppInfoRegistryLocker XdpAppInfoRegistryLocker;
```
AdrianVovk: You can `typedef void` rather than declaring a struct you never mention again

src/xdp-app-info-registry.c
```diff
@@ -69,71 +122,102 @@ xdp_app_info_registry_new (void)
...
-           sender);
+  match = g_hash_table_lookup (registry->app_infos, sender);
+  if (!match)
+    return NULL;
 
-  g_hash_table_insert (registry->app_infos,
-                       g_strdup (sender),
-                       g_object_ref (app_info));
+  if (!g_mutex_trylock (&match->mutex))
+    return NULL;
```
AdrianVovk: Is this correct? If thread 2 looks up the app info while thread 1 is busy creating it, shouldn't thread 2 block until thread 1 completes and then return the populated app info? Rather than returning NULL?

Seems like this might cause spurious failures. App makes two portal method calls. This creates two threads. If we're unlucky with the scheduling, both threads will try to look up the caller's appinfo at the same time. One gets the mutex and returns an appinfo, the other returns NULL for.... some reason.

src/xdp-app-info-registry.c
```diff
@@ -69,71 +122,102 @@ xdp_app_info_registry_new (void)
...
+
+  return item->app_info ? g_object_ref (item->app_info) : NULL;
+}
+
+XdpAppInfo *
+xdp_app_info_registry_lookup_sender (XdpAppInfoRegistry *registry,
+                                     const char         *sender)
+{
+  g_autoptr(XdpAppInfo) app_info = NULL;
+  XdpAppInfoRegistryItem *match;
```
AdrianVovk: Weird that `lookup_unlocked` ensures the existence of an item, but `lookup_sender` doesn't

src/xdp-app-info-registry.h
```diff
@@ -30,16 +30,26 @@ G_DECLARE_FINAL_TYPE (XdpAppInfoRegistry,
...
-                                           const char         *sender);
+XdpAppInfoRegistryLocker * xdp_app_info_registry_lock (XdpAppInfoRegistry *registry,
+                                                       const char         *sender);
+
+void xdp_app_info_registry_insert_unlocked (XdpAppInfoRegistry *registry,
+                                            XdpAppInfo         *app_info);
 
-void xdp_app_info_registry_insert (XdpAppInfoRegistry *registry,
-                                   XdpAppInfo         *app_info);
+XdpAppInfo * xdp_app_info_registry_lookup_unlocked (XdpAppInfoRegistry *registry,
```
AdrianVovk, line 51: Weird that `lookup_sender` and `lookup_unlocked` don't match name-wise. We should have either `lookup`/`lookup_unlocked` or `lookup_sender`/`lookup_sender_unlocked`

src/xdp-app-info-registry.c
```diff
@@ -23,18 +23,71 @@
...
                      G_TYPE_OBJECT)
 
+static XdpAppInfoRegistryItem *
+xdp_app_info_registry_item_new (void)
+{
+  XdpAppInfoRegistryItem *item;
+
+  item = g_new0 (XdpAppInfoRegistryItem, 1);
+  g_mutex_init (&item->mutex);
+  g_atomic_ref_count_init (&item->ref_count);
```
AdrianVovk, line 52: Trying to think: why the ref count? Doesn't this ultimately have single ownership in the registry hash table?

The only time we give this away somewhere else is when we want to use it as a mutex locker. But we could instead just return a `GMutexLocker` and drop our custom type entirely.

swick: This is all about concurrent deletes. If one thread tries to create the AppInfo for a peer which just disconnected, and got removed from the hash table in another thread, we need to ensure the item is still around. That's why `xdp_app_info_registry_ensure_item` returns an owned item, and it's also why `XdpAppInfoRegistryLocker` exists in the first place; when you call `xdp_app_info_registry_lock`, the item must be kept alive until the locker gets cleaned up.

src/xdp-app-info-registry.c
swick: Yeah, not sure what I was thinking there. I changed it so it waits until it gets the lock.

src/xdp-app-info-registry.c
AdrianVovk: Doesn't this just boil down to grabbing the item's mutex before you remove the item from the hash table? A thread that tries to remove the item from the hash table will just sit blocked until everything operating on the item is done

swick: I tried to not have to lock the item because it is usually locked for a long time (at least one dbus roundtrip) and that would stall the main thread for that time.

PR #1855, opened 2025-11-05
Port a few portals to libdex

swick:
Just an experiment for now.

There are currently two main pain points:

* Requests/Sessions use threads, so we still need take a lock when dealing with them, but with even more pitfalls because keeping the lock while suspended is not great at all
* The `g-authorize-method` mechanism is currently blocking the main thread when ran in a fiber. This should use async, but also needs locking across suspension points.

The request/session thing could be solved by making it async, because they are local to a portal which will always either use threads or use fibers, but never both.

The async lock thing is a bit more trick /cc @chergert.


src/meson.build
```diff
@@ -1,39 +1,45 @@
 ###### xdg-desktop-portal
+libdex_codegen_ext = libdex_dep.get_variable(pkgconfig: 'gdbus_codegen_extension')
 
 portal_built_sources = gnome.gdbus_codegen(
   'xdp-dbus',
   sources: portal_sources + flatpak_intf_sources,
   interface_prefix: 'org.freedesktop.portal',
   namespace: 'XdpDbus',
   docbook: 'portal',
+  extra_args: [f'--extension-path=@libdex_codegen_ext@'],
```
AdrianVovk: Looks like we can factory out the fstring? Or even the array entirely?

swick:
@AdrianVovk this one is ready for some high level reviewing

PR #1866, opened 2025-11-20
github: Remove the disfunc open collective

swick:
It is no more so we should not point to it.


jadahl:
Cc @mwleeds

jadahl:
Fair point.

PR #1879, opened 2025-12-15
Preparations for libdex support

swick:
A bunch of changes that will become important for libdex support which also do not depend on libdex.

### DIFF
--- a/src/xdp-context.c
+++ b/src/xdp-context.c
@@ -62,6 +62,14 @@
 
 #include "xdp-context.h"
 
+enum
+{
+  PEER_DISCONNECT,
+  N_SIGNALS,
+};
+
+static guint signals[N_SIGNALS] = { 0 };
+
 struct _XdpContext
 {
   GObject parent_instance;
@@ -122,6 +130,14 @@ xdp_context_class_init (XdpContextClass *klass)
   GObjectClass *object_class = G_OBJECT_CLASS (klass);
 
   object_class->dispose = xdp_context_dispose;
+
+  signals[PEER_DISCONNECT] =
+    g_signal_new ("peer-disconnect",
+                  G_TYPE_FROM_CLASS (object_class),
+                  G_SIGNAL_RUN_LAST,
+                  0, NULL, NULL, NULL,
+                  G_TYPE_NONE, 1,
+                  G_TYPE_STRING);
 }
 
 static void
@@ -301,6 +317,9 @@ on_peer_disconnect (const char *name,
 
   xdp_usb_delete_for_sender (context, name);
   notification_delete_for_sender (context, name);
+
+  g_signal_emit (context, signals[PEER_DISCONNECT], 0, name);
+
   close_requests_for_sender (name);
   close_sessions_for_sender (name);
   xdp_session_persistence_delete_transient_permissions_for_sender (name);


src/xdp-session.h
```diff
@@ -54,6 +54,7 @@ typedef struct _XdpSessionClass
 } XdpSessionClass;
 
 GType xdp_session_get_type (void);
+#define XDP_TYPE_SESSION (xdp_session_get_type ())
```
AdrianVovk, line 59: Hmm, why aren't we using `G_DECLARE_FINAL_CLASS` or whatever to generate this boilerplate?

I know many of the portals don't use this, because they don't have a library prefix. But this does have a prefix

src/xdp-context.c
```diff
@@ -75,6 +75,8 @@ struct _XdpContext
   guint peer_disconnect_handle_id;
   XdpAppInfoRegistry *app_info_registry;
   GHashTable *exported_portals; /* iface name -> GDBusInterfaceSkeleton */
+  GHashTable *registerd_object_paths; /* char *object_path set */
```
AdrianVovk: misspelled: registerd -> registered

src/xdp-request.c
```diff
@@ -239,7 +243,7 @@ xdp_request_init_invocation (GDBusMethodInvocation *invocation,
 
   G_LOCK (requests);
 
-  while (g_hash_table_lookup (requests, id) != NULL)
```
AdrianVovk, line 242: I assume this is the global state that's gotten rid of? We still have the state though. I assume that not relying on it here means that you'll be able to do other things later. But the commit message should reflect that

src/xdp-session.c
```diff
@@ -346,6 +353,16 @@ xdp_session_initable_init (GInitable     *initable,
   id = g_strdup_printf (DESKTOP_DBUS_PATH "/session/%s/%s",
                         sender_escaped, session->token);
 
+  while (!xdp_context_claim_object_path (session->context, id))
```
AdrianVovk, line 359: This check wasn't here before, and really most of this commit seems to be about introducing this check than anything else

The previous commit also isn't super clear on its own: it sets up plumbing that doesn't immediately seem useful. So here's what I propose: move the bits that make xdp-request call `xdp_context_claim_object_path` into the previous commit and make it about that. Then this commit can be about introducing the anti-collision check

src/xdp-session.c
```diff
@@ -346,6 +353,16 @@ xdp_session_initable_init (GInitable     *initable,
...
+    {
+      uint32_t r = g_random_int ();
+      g_free (id);
+      id = g_strdup_printf (DESKTOP_DBUS_PATH "/session/%s/%s/%u",
```
AdrianVovk, line 363: Also, when would a collision happen? And how would a caller be expected to react to it?

I guess we trust the caller to correctly predict a token (for the purposes of connecting to signals and whatnot). But if they predict _wrong_, then what can you do? We end up returning a different object path, and most callers won't even look at the return value (meaning things are broken). And even if they do look at the return value and re-do their signal hookups, then things are still broken because now it's racy and the caller might have missed signals.

In other words, I'm not understanding why we don't just hard error here and give up. Then at least the caller isn't going to end up in some inconsistent state because they should already be prepared to handle errors.

src/xdp-context.c
```diff
@@ -301,6 +317,9 @@ on_peer_disconnect (const char *name,
 
   xdp_usb_delete_for_sender (context, name);
   notification_delete_for_sender (context, name);
+
+  g_signal_emit (context, signals[PEER_DISCONNECT], 0, name);
```
AdrianVovk, line 321: Seems to me that we should also rearrange things so that instead of having XdpContext call into `xdp_usb_delete_for_sender` and `notification_delete_for_sender` and so on, those should just be handlers for this peer-disconnect signal

src/xdp-session.h
swick: This is an abstract class but `G_DECLARE_DERIVABLE_TYPE` tells you (https://docs.gtk.org/gobject/func.DECLARE_DERIVABLE_TYPE.html):

> the GtkFrobber struct is created with GtkWidget as the first and only item. You are expected to use a private structure from your .c file to store your instance variables.

IOW, we'd have to turn `XdpSession` into `XdpSessionPrivate` and expose the fields as methods, which is a bigger change. We should do it, but not in this PR.

src/xdp-session.c
swick: The short story is: because XdpRequest does so. Having this inconsistency isn't helping anyone, even if I agree that we should always have errored out in the case of a collision.

src/xdp-context.c
swick: That's a lot more complicated than it seems. I prepared a follow up PR for it though.

src/xdp-session.c
swick: All the plumbing in here is to get the context to the sessions and requests. Without claiming the object name we don't need it and it becomes even less obvious why this is done. As soon as you claim the object you have to deal with the fact that claiming it can fail, and have to handle it somehow.

src/xdp-request.c
swick: Adjusted the commit message.

src/xdp-context.c
swick: renamed

src/xdp-context.c
swick: https://github.com/flatpak/xdg-desktop-portal/pull/1911

src/xdp-session.c
AdrianVovk: I don't think anything is built to expect us to change the path around. So we could rip that behavior out and just start returning an error, no?

src/xdp-session.c
AdrianVovk: Seems that we're talking past each other. The status quo is two commits:

1) Introduces the plumbing for xdp_context_claim_object_path
2) Uses this new plumbing in XdpRequest, plumbs the Context into the sessions and requests, and introduces an anti-collision check in XdpSession

I was suggesting splitting up these changes differently:

1) Introduces the plumbing for xdp_context_claim_object_path, plumbs the context into requests, and makes the requests use the new plumbing in its anti-collision check
2) Plumbs the context into the sessions, and introduces the anti-collision check in XdpSession

AdrianVovk: Anyway, I don't think this is too important. Feel free to mark as resolved and merge as-is

src/xdp-session.c
AdrianVovk: I'm also happy for us to continue this in a follow-up. I do agree that if one thing does this anti-collision object path change then the other should do, so I don't mind too much merging as-is. As long as we discuss changing the behavior entirely, in both places

src/xdp-request.c
AdrianVovk: Follow-up question: I'm not understanding part of the commit message:

> We also want to be able to have different implementations for requests and sessions without colliding object paths in the future

What does this mean?

swick: In the libdex PR I introduce the RequestFuture and SessionFuture classes which also create Request and Session dbus objects but are otherwise independent. The collision detection has to be shared though and that's what the object path claiming on the context provides.

src/xdp-session.c
swick: This is long existing behavior and there will be apps depending on it, so changing this now is essentially impossible.

PR #1882, opened 2026-01-03
po: Update Georgian translation

EkaterinePapava:


swick:
Seems like this one needs to be rebased.

EkaterinePapava:
Hi,
Is anything needed from my side, please?

swick:
There was a release recently which updated the PO files, so the PR here does not apply cleanly anymore. I don't know how the translation tools work, but I assume you have to do the translations on the latest version.

PR #1889, opened 2026-01-23
Introduce Credentials portal (experimental)

iinuwa:
:warning: Still a work in progress and is not ready for review yet.

This seems to work so far. Some work to be done:
- add a mock backend implementation and client for unit testing
- convert to using libdex
- Review whether the signatures make sense, or if they need to be flattened some more.
- add `DiscoverCredential(services as)` method to frontend and backend.

### DIFF
--- a/data/org.freedesktop.impl.portal.experimental.Credential.xml
+++ b/data/org.freedesktop.impl.portal.experimental.Credential.xml
@@ -1,0 +1,455 @@
+<?xml version="1.0"?>
+<!--
+ Copyright (C) 2026 Isaiah Inuwa <dev@iinuwa.xyz>
+
+ SPDX-License-Identifier: LGPL-2.1-or-later
+
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2.1 of the License, or (at your option) any later version.
+
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Isaiah Inuwa <dev@iinuwa.xyz>
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.experimental.Credential:
+      @short_description: Credential portal backend interface
+      @stability: Unstable
+
+      **Interface**: ``org.freedesktop.impl.portal.experimental.Credential``
+
+      **Version**: Experimental 🧪
+
+      The Credential portal allows sandboxed applications to retrieve
+      application-scoped web credentials. The backend interface guides the user
+      through the interactive authentication ceremony.
+
+  -->
+  <interface name="org.freedesktop.impl.portal.experimental.Credential">
+
+    <!--
+        CreateSession:
+        @since: 1
+        @session_handle: Object path for the :ref:`org.freedesktop.impl.portal.Session` to create.
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`.
+        @origin: The origin of the request.
+        @type: Type of the credential to create. Current supported types are ``PublicKeyCreate = 0``, and ``PublicKeyGet = 1``.
+        @sources: Initial list of credential sources. Map from source ID to transport. Valid transports are ``Ble``, ``HybridLinked``, ``HybridQr``, ``Internal``, ``Nfc``, ``Usb``. Sources may be added or removed during the ceremony.
+        @app_id: App ID of the application.
+        @app_pid: PID of the application.
+        @options: Vardict with optional further information.
+
+        Creates a session for an authentication ceremony with the client context
+        and an initial list of credential sources. During the ceremony, the
+        backend sends user interaction events to the frontend via signals in
+        response to background events sent from the frontend via the other
+        methods on this interface.
+
+        Before calling this method, the frontend MUST subscribe to all
+        signals on this interface, filtering for the given @session_handle
+        argument. A :ref:`org.freedesktop.impl.portal.Session` object will be
+        exported on the @session_handle path, which can be used by the frontend
+        to cancel the ceremony or receive notification that the
+        user cancelled it. Frontends MUST use the same D-Bus connection
+        throughout the ceremony.
+
+        Upon receiving this method call, backends MUST send
+        :ref:`org.freedesktop.impl.portal.experimental.Credential::DiscoveryRequested`
+        when the user is ready to perform the authentication ceremony.
+        Backends MUST send all signals as unicast signals to the original sender
+        calling this method.
+
+        Supported keys in the @options vardict include:
+
+        * ``activation_token`` (``s``)
+
+          A token that can be used to activate the credential selection dialog.
+
+        * ``top_origin`` (``s``)
+
+          Top-level origin of the request if different from the origin.
+
+        * ``rp_id`` (``s``)
+
+          WebAuthn relying party ID.
+
+          Required if @type is ``PublicKeyCreate`` or ``PublicKeyGet``.
+    -->
+    <method name="CreateSession">
+      <arg name="session_handle" type="o" direction="in"/>
+      <arg name="parent_window" type="s" direction="in"/>
+      <arg name="origin" type="s" direction="in"/>
+      <arg name="type" type="u" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In4" value="QList&lt;QVariantMap&gt;"/>
+      <arg name="sources" type="a(ss)" direction="in"/>
+      <arg name="app_id" type="s" direction="in"/>
+      <arg name="app_pid" type="u" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In9" value="QVariantMap"/>
+      <arg name="options" type="a{sv}" direction="in"/>
+    </method>
+
+    <!--
+        NotifyNeedsPin:
+        @since: 1
+        @session_handle: The :ref:`org.freedesktop.impl.portal.Session` for this authentication ceremony.
+        @attempts_left: The number of PIN attempts remaining before the device is locked out and must be reset/wiped. ``0xffffffff`` means the number of attempts left is unknown.
+        @options: Vardict with optional further information.
+
+        The device needs PIN user verification: prompt the user to enter the
+        PIN. The backend must emit the collected value as a
+        :ref:`org.freedesktop.impl.portal.experimental.Credential::ClientPinEntered`
+        signal. See the documentation on that signal for PIN requirements.
+
+        The backend should consider rate-limiting requests, or requiring the
+        user to power-cycle (e.g., unplug and plug back in) the device after
+        repeated incorrect attempts to prevent the device from being locked out
+        due to user error or abuse.
+
+        The backend should warn the user that the credentials may be permanently
+        lost if they continue to enter incorrect PINs.
+
+        There are currently no supported keys in the @options vardict.
+
+
+        .. seealso::
+           * `CTAP 2.3 documentation on PIN and user verification retries`_
+    -->
+    <method name="NotifyNeedsPin">
+      <arg name="session_handle" type="o" direction="in"/>
+      <arg name="attempts_left" type="u" direction="in"/>
+      <arg name="options" type="a{sv}" direction="in"/>
+    </method>
+
+    <!--
+        NotifyNeedsUserVerification:
+        @since: 1
+        @session_handle: The :ref:`org.freedesktop.impl.portal.Session` for this authentication ceremony.
+        @attempts_left: The number of user verification attempts remaining before the built-in user verification methods are disabled until the next successful PIN verification. ``0xffffffff`` means the number of attempts left is unknown.
+        @options: Vardict with optional further information.
+
+        The authenticator needs on-device user verification (likely biometrics, or
+        can be on-device PIN entry). Prompt the user to interact with the
+        device.
+
+        The backend should display the number of built-in user verification
+        attempts remaining before falling back to client PIN verification.
+
+        .. note::
+
+           This is distinct from
+           :ref:`org.freedesktop.impl.portal.experimental.Credential.NotifyNeedsPin`,
+           which uses a PIN captured on the client device rather than directly
+           on the authenticator.
+
+        There are currently no supported keys in the @options vardict.
+
+        .. seealso::
+           * `CTAP 2.3 documentation on PIN and user verification retries`_
+
+        .. _CTAP 2.3 documentation on PIN and user verification retries: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#authnrClientPin-globalState-retries
+    -->
+    <method name="NotifyNeedsUserVerification">
+      <arg name="session_handle" type="o" direction="in"/>
+      <arg name="attempts_left" type="u" direction="in"/>
+      <arg name="options" type="a{sv}" direction="in"/>
+    </method>
+
+    <!--
+        NotifyNeedsUserPresence:
+        @since: 1
+        @session_handle: The :ref:`org.freedesktop.impl.portal.Session` for this authentication ceremony.
+        @options: Vardict with optional further information.
+
+        Device requires user interaction (e.g. pressing a button) to release credentials; prompt the user
+        to do so. The device may require additional user verification, but that
+        might not be known until after the user taps the device.
+
+        There are currently no supported keys in the @options vardict.
+    -->
+    <method name="NotifyNeedsUserPresence">
+      <arg name="session_handle" type="o" direction="in"/>
+      <annotation name="org.qtproject.QtDBus.QtTypeName.In1" value="QVariantMap"/>
+      <arg name="options" type="a{sv}" direction="in"/>
+    </method>
+
+    <!--
+        NotifySelectingCredential:
+        @since: 1
+        @session_handle: The :ref:`org.freedesktop.impl.portal.Session` for this authentication ceremony.
+        @credentials: A list of metadata for credentials that are able to fulfill the request.
+        @options: Vardict with optional further information.
+
+        The device is ready to release credentials, but multiple credentials
+        match the request. The backend must prompt the user to select one of the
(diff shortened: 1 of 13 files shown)


data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+
+      This documentation describes a pre-release version of this
+      interface.
+  -->
+  <interface name="org.freedesktop.portal.CredentialsX">
+
+    <!--
+        CreateCredential:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @origin: The origin of the request. If specified, must be a valid HTTPS origin; otherwise, the app ID will be used as the origin.
```
swick: Isn't the origin something with a scheme? how would the app id origin get formed?

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+        @top_origin: The top-level origin of the client window. Pass an empty string to denote a same-origin request.
```
swick: I'm unsure what this is for.

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+        @request: The credential request context and parameters.
```
swick: Not sure if you really want to call it request given that we already have a Request interface which is not trivial to understand.

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+          string of JSON that corresonds to the WebAuthn
+          ``PublicKeyCredential`` type with the ``response`` field set as
+          an ``AuthenticatorAttestationResponse``.
+    -->
+    <method name="CreateCredential">
+      <arg name="parent_window" type="s" direction="in"/>
+      <arg name="origin" type="s" direction="in"/>
+      <arg name="top_origin" type="s" direction="in"/>
+      <!--
+        CreateCredentialRequest:
```
swick: Should be described in the doc comment above really.

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+
+        The following results get returned via the :ref:`org.freedesktop.portal.Request::Response` signal:
+
+        * ``type`` (``s``)
+          Type of the created credential.
+          Currently, ``publicKey`` is the only supported type.
+
+        * ``registration_response_json`` (``s``)
+          If ``type`` is ``publicKey``, a
+          string of JSON that corresonds to the WebAuthn
```
swick: any reason why the type and json result are in the same dict in the response, but the json is in another dict in the "request"?

data/org.freedesktop.portal.CredentialsX.xml
swick: you could even merge it with the options vardict if we don't find a better name

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+        CreateCredential:
+        @parent_window: Identifier for the application window, see :doc:`window-identifiers`
+        @origin: The origin of the request. If specified, must be a valid HTTPS origin; otherwise, the app ID will be used as the origin.
+        @top_origin: The top-level origin of the client window. Pass an empty string to denote a same-origin request.
+        @request: The credential request context and parameters.
+        @options: Vardict with optional further information
+
+        Requests a new credential to be stored.
+
+        Supported keys in the @options vardict include:
```
swick: you might already want to add an activation token here

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+        GetCredentialRequest:
+        @origin: the web origin at which the request is made.
+        @is_same_origin: Whether the top-level origin of the client
+        is the same as the origin for which the request is made.
+        @type: the type of credential to create. One of
+        ["publicKey"].
+        @public_key: if @type is "publicKey", this field contains
+        the a dictionary, with a single field, `request_json`, which
+        is a string of JSON that corresponds to the WebAuthn
+        PublicKeyCredentialRequestOptions type.
```
swick: I think what's missing here is a reference to where this is defined

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+      <!--
+          GetClientCapabilities:
+
+          Returns a set of boolean flags indicating what features
+          this API supports.
+
+          See the WebAuthn spec for meanings of the client
+          capability keys.
+      -->
+    <method name="GetClientCapabilities">
```
swick: we have dbus properties for this

data/org.freedesktop.portal.CredentialsX.xml
```diff
@@ -0,0 +1,196 @@
...
+      <!--
+          Response:
+          @conditional_create: Whether conditional mediation is
```
swick: I'm not sure what this actually means. I assume it's about what you can put into the json? Probably good again to make that more clear and again add a reference for to the definition of this.

I guess that's true for most things in the list below.

swick (commented):
Went through the frontend interface and added comments for everything that stood out to me. It does look very reasonable!

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: OK, I wasn't sure merging with `options`, since most of the descriptions of that field in other portals say that it contains optional information, but the request would always  be required. For that reason, I pulled it out as a separate field, but if it feels natural to you to put it in `options`, I don't have a problem with that.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Right now, I'm just prefixing the app ID with `app:` in the backend. (This should probably be moved to the frontend though to require less coordination if it ever changes.)

I'm somewhat deliberately making the app ID format unspecified. Right now, we have not specified how app IDs should map to HTTPS origins for sharing—we need the equivalent of Android's Digital Asset Links or Apple App Site Association. That would allow us to define, for example that `app:com.discordapp.Discord` is allowed to request credentials for `https://discord.com` and `https://discordapp.com` while still being considered "same origin."

We may make it more specific in the future, e.g. `flatpak:com.discordapp.Discord` and `snap:com.discordapp.Discord`, or even more specifically, `flathub:com.discordapp.Discord` and `snapcraft:com.discordapp.Discord`. I think that eventually these app IDs should be bound to the app verification process, which probably means it would be namespaced by distributor. (Unless some collaboration process forms between Linux app distributors that unifies application builds but allows separate bundle formats... I think something like that would be beneficial for the ecosystem, but is unlikely.)

But that's a much larger conversation and I think should be considered outside scope for now.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: In a browser, if you request WebAuthn from within an iframe, the `origin` would be the iframe's origin, and the `top_origin` would be the top-level window. If the origins differ, these would be considered cross-origin.

Typically WebAuthn is restricted to same-origin requests, but some contexts require cross-origin requests for better user experience. (For example, in ecommerce, a merchant site may open a payment processor's site in an iframe. Inside the iframe, the payment processor could request a WebAuthn credential within the iframe.) Adding `top_origin` allows the WebAuthn credential signature to be bound to both the origin and `top_origin`, which would give the relying party (the payment processor, in this example) more context on exactly where the request came from. E.g., they could reject the request if it is nested on a site they don't know about.

For other types of credentials, like passwords and OTPs, this parameter doesn't really help much, since the context isn't passed back to the RP.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Yeah, these have specific meanings imported from the [WebAuthn spec](https://www.w3.org/TR/webauthn-3/#enumdef-clientcapability) that I failed to reference here. Is there a specific format I should use to reference the specification (inline link, reference to a footnote)

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Yeah, this definitely makes sense to be properties. Some of these values will eventually be dynamic based on what third-party credential providers can offer, but I can calculate that when the credential providers are enabled/disabled and just update the property and emit the property changed signal. I'll make that change.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: I think I was leaving it open for extending the `publicKey` request type with options that are not specified in WebAuthn, but would be needed in. I think it'd be fine to use the `options` parameter for this, especially if we're going to merge `request` and `options` into a single field.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
bilelmoussaoui: I would avoid pulling vscode specific files in-tree

data/org.freedesktop.impl.portal.CredentialsX.xml
```diff
@@ -0,0 +1,133 @@
...
+ Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public
+ License along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+ Author: Isaiah Inuwa <dev@iinuwa.xyz>
+-->
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.CredentialsX:
```
bilelmoussaoui: why is this named CredentialsX?

data/org.freedesktop.impl.portal.CredentialsX.xml
```diff
@@ -0,0 +1,133 @@
...
+        @request: The credential request parameters.
+        @options: Vardict with optional further information, reserved for future use.
+        @response: Vardict with the results of the call
+
+        Requests a credential to be created for a particular origin.
+    -->
+    <method name="CreateCredential">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
+      <arg type="s" name="app_display_name" direction="in"/>
```
bilelmoussaoui: Can't we derive this from the app_id?

swick:
As for the interface name, let's call it something like `org.freedesktop.portal.experimental.Credentials`. It should also not be exposed by default and have a environment variable to turn it on (e.g. `XDG_DESKTOP_PORTAL_ENABLE_EXPERIMENTAL=credentials`).

I also noticed that there is no code here which determines if access to an origin is allowed or not. I believe that xdp is the right place to handle it. Here we have access to the app metadata and the permission store. From quickly skimming the credentialsd code I believe this is currently hardcoded?

data/org.freedesktop.impl.portal.CredentialsX.xml
iinuwa: We had discussed previously adding a suffix while it was still experimental. But @swick said we should use an environment variable as a feature flag instead, so I'll do that.

swick: We would like to merge this in an experimental state first. But I agree that adding just an X at the end is not ideal :)

data/org.freedesktop.impl.portal.CredentialsX.xml
iinuwa: Yeah. In the current implementation, the frontend looks up the display name by app ID and forwards to the backend. That way backend doesn't have to reimplement the display name lookup if it trusts the frontend.

I'm ignorant of other considerations for doing the lookup from the backend; do you have more?

swick: it's a bit of an issue because the app id doesn't always map properly to a desktop file (e.g. in the snap case), but it's true that we currently only give backends the app id and pretend that you can get the corresponding desktop file.

swick: Backends usually use something like https://docs.gtk.org/gio-unix/ctor.DesktopAppInfo.new.html. I'm unsure what rust provides here.

swick:
We discussed things on matrix, and the conclusion is that we to depend on credentialsd, so icredentialsd is *not* the impl. Instead, the API that credentialsd calls to create the UI should be the impl.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: I went back on this and decided to just use HTTPS origins for apps too, and we'll gate access control locally on the machine. (In the original scheme above, this would have required every website to add support for all the Linux app IDs for their app, which most would never comply, so sending the HTTPS origin just makes it seem like a browser.)

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Merged with `options`

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Merged with `options`.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Split into individual properties in the portal interface. I removed the method from credentialsd, since it can all be set statically for now. If we want to do any dynamic updates in the future, we can add it back in.

data/org.freedesktop.portal.CredentialsX.xml
iinuwa: Added links to all the WebAuthn definitions.

data/org.freedesktop.impl.portal.CredentialsX.xml
iinuwa: I changed the "impl" to be called "handler" instead, since it's not a backend impl. I'm fine with changing nomenclature, but wanted that to be clearer while we talk about it.

iinuwa:
I pushed some updates, but this isn't ready for a full review still since I haven't addressed the other comments from Matrix, namely:
- we need to flip the direction of the control flow between the backend and frontend (e.g., the backend should export an object where we can call methods on rather than the frontend exporting a method that the backend calls.)
- use libdex for async handling here (I'm not super familiar with libdex, so I'm leaving that conversion until last).
- we need to enforce access control in xdp, probably using entitlements.

(I'm mostly pushing so that I can paste a reference point for the work in progress on the credentialsd side.)

data/org.freedesktop.impl.portal.CredentialsX.xml
iinuwa: Removed this from the backend interface and used [`DesktopAppInfo::display_name()`](https://gtk-rs.org/gtk-rs-core/stable/0.22/docs/gio_unix/prelude/trait.AppInfoExt.html#method.display_name) in the reference implementation. Let me know if this is the right direction we should go.

iinuwa:
## spec review
@swick, @bilelmoussaoui, I think I've either addressed or responded to all the review comments above, and the spec portion is ready for re-review. (I'd be happy to split the implementation into a separate PR if that helps. If so, I can split this and unmark this as draft.)

## implementation prereqs
As far as libdex, it looks like that's still a bit in flux? (I'm looking at #2036, which was opened recently)

It looks like entitlements (#1924) is in a similar state.

If I cleaned up the code to fix the goblint errors, would it be acceptable to move toward merging this with callback-style code and rewrite with libdex and add the entitlements feature in a subsequent PR?

swick:
I finished up both libdex and entitlements. I'll look at this PR in the coming days, try to get it running and port it over.

swick:
The history here is a big of a mess. I tried my best to get the relevant changes on top of my branches with libdex and entitlements. It builds but I might have missed some changes.

https://github.com/swick/xdg-desktop-portal/tree/wip/credentials-portal

iinuwa:
Oh, yeah, that's bogus. Sorry about that. I was avoiding rebases to preserve GitHub comments, but I jacked up that merge commit. I squashed and rebased.

I confirmed that all of the changes from this branch are present in your branch.

Edit: the `desktop-portal/xdp-context.c` changes are missing from that new branch

iinuwa:
[This patch](https://github.com/user-attachments/files/29402052/credential-portal-dex.patch) on top of your branch seems to do the trick. I think I did it right. With the async/await pattern and the extra macros, it does simplify the code quite nicely. :)

swick:
Not a full review by any means, but I looked at the API and it's smaller than I expected from previous conversations. That's good, but it also means that I'm questioning the design decision to let credentialsd call into the impls. It looks like we could just do what we do in any other portal and have credentials talk with the frontend, and the frontend then talks to the impl. It would also make credentialsd entirely independent of xdp again.

What are your thoughts on that?

iinuwa:
I guess I have two concerns:
- I'd really like to get something in xdp behind a feature flag to make it easier for users to test. If we can defer this architecture decision until later, I would appreciate that.
- my other concern is long-term maintenance. I am not really qualified to write C, so as the portal expands, I'm not sure I'll be able to keep up with the C part.
 
I'm willing to try if you think that's the right way to go though. Also, I had noticed that the docs in your other PR dropped one of the pages that documents the signals sent from the backend to frontend. We did simplify some things, so you might be seeing that effort paying off, but just wanted to make sure we're looking at the right docs

data/org.freedesktop.impl.portal.experimental.Credential.xml
```diff
@@ -0,0 +1,281 @@
...
+        @options: Vardict with optional further information.
+
+        Initializes a credential ceremony with the client context and an initial list
+        of credential sources. Before calling this method, an application should
+        subscribe to the ``UserInteracted`` signal on the
+        ``org.freedesktop.impl.portal.experimental.Credential.Ceremony``
+        interface for the given
+        @handle. A
+        :ref:`org.freedesktop.impl.portal.Request` object will also be exported on
+        the same path, which can be used to cancel the request.
```
iinuwa: I have received feedback that this should use the Session object to fit this pattern. Changes to make:

- Move the individual tag-value structs on Ceremony into separate signals
- Probably rename Initialize() to CreateSession().
- Export and return a handle to a Session object on Initialize.
- Merge the signals and methods from the Ceremony object into the backend interface by adding a `session_handle` parameter to each of them.
- Update the documentation to let the developer know they should subscribe to the methods on the session handle.

iinuwa:
I updated the backend spec to use the Session pattern used in other portals. I think this follows the same convention, but let me know if anything needs to be tweaked.

Once https://github.com/linux-credentials/credentialsd/pull/194 lands, the credentialsd project will implement the frontend and backend APIs according to this spec.

I think this is ready for review to be merged behind a feature flag, but the first few commits are a placeholder for another solution for #2066, and this will need be rebased on that solution.

iinuwa:
Updated formatting and ported to libdex

PR #1895, opened 2026-01-28
Fix checking for inputcapture capabilities

whot:
These two are easy-to-merge ones from #1803, with commit message typos fixed and rebase on current `main`, cc @swick


whot:
there seems to be some race condition with the document portal hanging during the test teardown, causing the tests to randomly time out. I can reproduce this locally irregularly every few runs, not sure of the cause and fix yet.

PR #1896, opened 2026-01-29
Fix deadlock on document-portal shutdown

whot:
`g_main_loop_quit()` is not signal-safe, calling it inside the signal handler may cause deadlocks. This seems to be the cause for the CI failures in #1895 - the test succeed but then hang and time out during test teardown.

The fix is to switch to `g_unix_signal_add()` but this requires the fuse bits to no longer use `SIGHUP`. I'm not 100% sure on that code[^1] but it seems correct and it works.

Reproducible locally with the `test_screenshot_basic` (fairly reliably, every 7-15 runs).
```console
$ while ./tests/run-test.sh -k test_screenshot_basic --verbose --timeout=7 --log-level=DEBUG -x
do
 ;
done
```

[^1]: after too long debugging Claude helped with that

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -36,6 +36,7 @@
 
 #include <gio/gio.h>
 #include <gio/gunixfdlist.h>
+#include <glib-unix.h>
 #include "document-portal-dbus.h"
 #include "document-store.h"
 #include "src/xdp-app-info.h"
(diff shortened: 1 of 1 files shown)


swick:
Thank you very much! Managed to reproduce locally now and the fix makes sense to me. Let's see if this fixes all the deadlocks in the tests.

PR #1897, opened 2026-01-29
Draft: Add inputcapture clipboard access support (v3)

whot:
This sits on top of #1896 and #1895. It's a re-shuffled version of #1803 with various typos in commit messages fixed, patches split and squashed in as needed. Rebased on top of current `main`. There are no real functional changes over #1803.

@swick - maybe you want to grab this one and force-push it to #1803 so we keep that history?
(I needed this as base for adding persistence in the inputcapture session.)


swick:
Thanks, pulled back.

PR #1898, opened 2026-01-30
InputCapture support for session persistence

whot:
~~Currently based on #1897 which itself is a rebased version of #1803.~~

Replaces #1768 because persistence is the more correct thing to do.

This adds a new `persist_mode` and `restore_token` arguments which work identical to those in the existing RemoteDesktop API (except there it's on `SelectDevices`, here we only have `Start`).

Note that this bumps the interface version to 3 (after #1803/#1897 bump it to 2). That's not required if we merge them in the same cycle.

Corresponding components:
- https://github.com/flatpak/libportal/pull/214
- [xdg-desktop-portal-gnome](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/252)
- https://github.com/deskflow/deskflow/pull/9415

cc @jadahl, @swick, @nbolton, @sithlord48

### DIFF
--- a/tests/templates/inputcapture.py
+++ b/tests/templates/inputcapture.py
@@ -3,7 +3,7 @@
 # This file is formatted with Python Black
 # mypy: disable-error-code="misc"
 
-from tests.templates.xdp_utils import Response, init_logger
+from tests.templates.xdp_utils import Response, ImplSession, init_logger
 
 from collections import namedtuple
 from itertools import count
@@ -89,6 +89,7 @@ def load(mock, parameters={}):
     mock.session_handles = []
     mock.active_session_handles = []
     mock.restore_datas = []
+    mock.sessions = {}
 
 
 @dbus.service.method(
@@ -102,6 +103,8 @@ def CreateSession2(self, session_handle, app_id, options):
 
         assert len(options) == 0
 
+        session = ImplSession(self, BUS_NAME, session_handle, app_id).export()
+        self.sessions[session_handle] = session
         self.session_handles.append(session_handle)
 
         return {"session_handle": session_handle}


sithlord48:
I've noticed that in the input capture session the token of of type `XdpSession*` and in remote desktop the token is a `char*` . is this intended ?

data/org.freedesktop.impl.portal.InputCapture.xml
```diff
@@ -147,6 +147,35 @@
           See the :ref:`org.freedesktop.portal.Clipboard` documentation for more
           information.
 
+        * ``restore_data`` (``(suv)``)
```
Sodivad, line 179: I think this needs to move a bit higher to options?

tests/templates/inputcapture.py
```diff
@@ -119,6 +132,31 @@ def Start(self, handle, session_handle, app_id, parent_window, options):
...
+                raise Exception(f"Invalid restore_data tuple {restore_data}")
+
+        if persist_mode == SessionPersistenceMode.NONE:
+            pass
+        elif persist_mode == SessionPersistenceMode.TRANSIENT:
+            restore_data = make_restore_data("restore-transient")
+        elif persist_mode == SessionPersistenceMode.PERSISTENT:
+            restore_data = make_restore_data("restore-persistent")
+        elif persist_mode is not None:
+            raise Exception(f"Invalid session persistence mode: {persist_mode}")
```
Sodivad, line 158: There is something off, the persistence option always overrides the passed restore_data, 
not sure it matters for the test.

Sodivad:
xdg-desktop-portal-kde impl https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/496

data/org.freedesktop.impl.portal.InputCapture.xml
whot: fixed, thanks

tests/templates/inputcapture.py
whot: fixed, thanks

whot:
Side-note: happy to remove the version bump, shouldn't be needed if we get this out in the same release as the clipboard support we just merged.

whot:
Dropped version to 2, resolved conflicts, should be good to go.

I have now implemented this (too) many times, once in https://github.com/flatpak/libportal/pull/214, then in https://github.com/deskflow/deskflow/pull/9415, also in https://github.com/bilelmoussaoui/ashpd/pull/364 and in my [lan-mouse branch](https://github.com/whot/lan-mouse/tree/wip/inputcapture-clipboard) (got stuck for other reasons, the persistence bits work) and finally in [schengen](https://github.com/schengen-project/) though that work is also stuck for other reasons, the persistence bits work.

So I reckon that's good enough  :)

src/input-capture.c
```diff
@@ -1552,7 +1642,7 @@ input_capture_new (XdpDbusImplInputCapture *impl)
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (input_capture->impl), G_MAXINT);
 
   input_capture->impl_version =
-    MIN (xdp_dbus_impl_input_capture_get_version (impl), 2);
+    MIN (xdp_dbus_impl_input_capture_get_version (impl), 3);
```
swick: Still v3 here and probably also in the test

src/xdp-session-persistence.c
```diff
@@ -366,8 +366,9 @@ xdp_session_persistence_replace_restore_data_with_token (XdpSession *session,
                                                                *in_out_persist_mode,
                                                                in_out_restore_token,
                                                                in_out_restore_data);
-      g_variant_builder_add (&results_builder, "{sv}", "restore_token",
-                             g_variant_new_string (*in_out_restore_token));
+      if (*in_out_restore_token)
+        g_variant_builder_add (&results_builder, "{sv}", "restore_token",
+                               g_variant_new_string (*in_out_restore_token));
```
swick:
```suggestion
      if (*in_out_restore_token)
        {
          g_variant_builder_add (&results_builder, "{sv}", "restore_token",
                                 g_variant_new_string (*in_out_restore_token));
        }
```

PR #1906, opened 2026-02-17
Handle autostart errors

justinrdonnelly:
I've made some changes to `background.c`, that I think make it a little nicer to use. I discovered that in error cases, `libportal`'s [`request_background_finish`](https://libportal.org/method.Portal.request_background_finish.html) doesn't always return `false` or set the `Gerror`. This is because `xdg-desktop-portal`'s `background.c` almost always emits with `portal_response` set to `XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS` (`0`). Looking at the [request documentation](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html), it looks like `portal_response` should be `2` (`XDG_DESKTOP_PORTAL_RESPONSE_OTHER`) in case of errors. The changes in these patches are meant to ensure that the correct response is sent for error cases.

I tested by building GNOME OS 50 (xdg-desktop-portal 1.20.3) with my patches applied, and using a flatpak app in GNOME Builder to test [libporal request_background](https://libportal.org/method.Portal.request_background.html) calls. Then I ran `chmod 000 ~/.config/autostart` before testing in order to force an error. The flatpak app used for testing is available on [my GitHub](https://github.com/justinrdonnelly/portal-background-test-libportal).

I'm very new to the portal code (and not really a C dev), so it's certainly possible that this is off base. But it seems to me like an improvement.

## Test Results

Below are some results from my testing to help understand what the changes do. The tables show the different values used for for the background flags to request (or not request) autostart, as well as the return value from `xdp_portal_request_background_finish`, and whether it set the `GError`.

The part to focus on is the how the patches change what happens in the error case (`chmod 000 ~/.config/autostart`). With the patches, the error case always returns `false`, and sets the `GError`.

### Baseline (xdg-desktop-portal 1.20.3)

`chmod 755 ~/.config/autostart`

| Background Flag               | Return | Error    |
| ----------------------------- | ------ | -------- |
| XDP_BACKGROUND_FLAG_AUTOSTART | TRUE   | NULL     |
| XDP_BACKGROUND_FLAG_NONE      | TRUE   | NULL     |

`chmod 000 ~/.config/autostart`

| Background Flag               | Return | Error    |
| ----------------------------- | ------ | -------- |
| XDP_BACKGROUND_FLAG_AUTOSTART | FALSE  | NULL     |
| XDP_BACKGROUND_FLAG_NONE      | TRUE   | NULL     |

### Patches

`chmod 755 ~/.config/autostart`

| Background Flag               | Return | Error    |
| ----------------------------- | ------ | -------- |
| XDP_BACKGROUND_FLAG_AUTOSTART | TRUE   | NULL     |
| XDP_BACKGROUND_FLAG_NONE      | TRUE   | NULL     |

`chmod 000 ~/.config/autostart`

| Background Flag               | Return | Error    |
| ----------------------------- | ------ | -------- |
| XDP_BACKGROUND_FLAG_AUTOSTART | FALSE  | Not NULL |
| XDP_BACKGROUND_FLAG_NONE      | FALSE  | Not NULL |

### DIFF
--- a/src/background.c
+++ b/src/background.c
@@ -894,6 +894,7 @@ handle_request_background_in_thread_func (GTask *task,
   gboolean autostart_requested = FALSE;
   gboolean autostart_enabled;
   gboolean allowed;
+  XdgDesktopPortalResponseEnum portal_response;
   g_autoptr(GError) error = NULL;
   g_autofree const char **autostart_exec = { NULL };
   gboolean activatable = FALSE;
@@ -988,7 +989,6 @@ handle_request_background_in_thread_func (GTask *task,
 
   if (request->exported)
     {
-      XdgDesktopPortalResponseEnum portal_response;
       g_auto(GVariantBuilder) results =
         G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
 
--- a/src/background.c
+++ b/src/background.c
@@ -821,8 +821,12 @@ enable_autostart_sync (XdpAppInfo          *app_info,
 
   if (!enable)
     {
-      unlink (path);
-
+      if (unlink (path) != 0 && errno != ENOENT)
+        {
+          g_set_error (error, G_FILE_ERROR, g_file_error_from_errno (errno),
+                       "Could not remove Autostart file %s", path);
+          return FALSE;
+        }
       *out_enabled = FALSE;
       return TRUE;
     }
--- a/src/background.c
+++ b/src/background.c
@@ -960,31 +960,56 @@ handle_request_background_in_thread_func (GTask *task,
         {
           g_warning ("AccessDialog call failed: %s", error->message);
           g_clear_error (&error);
+          allowed = FALSE;
+          autostart_enabled = FALSE;
+          portal_response = XDG_DESKTOP_PORTAL_RESPONSE_OTHER;
         }
-
-      allowed = response == 0;
+      else
+        {
+          allowed = (response == XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS);
+          if (!allowed)
+            {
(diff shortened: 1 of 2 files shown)


justinrdonnelly:
Just giving this a little bump. Not sure if it got missed the first time around. Thanks!

src/background.c
```diff
@@ -957,31 +957,56 @@ handle_request_background_in_thread_func (GTask *task,
         {
           g_warning ("AccessDialog call failed: %s", error->message);
           g_clear_error (&error);
+          allowed = FALSE;
+          autostart_enabled = FALSE;
+          portal_response = XDG_DESKTOP_PORTAL_RESPONSE_OTHER;
+        }
+      else
+        {
+          allowed = response == 0;
```
swick: `allowed = (response == XDG_DESKTOP_PORTAL_RESPONSE_SUCCESS);`

src/background.c
```diff
@@ -818,8 +818,13 @@ enable_autostart_sync (XdpAppInfo          *app_info,
 
   if (!enable)
     {
-      unlink (path);
-
+      if (unlink (path) != 0) {
+        if (errno != ENOENT) {
```
swick: The code style is off here, and this should be a single if statement.

swick:
Please also check how we handle commit messages, especially the prefix.

justinrdonnelly:
Thanks for the feedback! I'll update the PR soon.

justinrdonnelly:
I've updated the PR based on the feedback. Let me know if there's anything else, and thanks again!

swick:
Thanks!

PR #1909, opened 2026-02-19
tests: Add simple mocks for flatpak and snap file access checks

3v1n0:
During tests we may actually try to run `flatpak` or `snap` binaries to check if the document portal can access directly to the resources it needs.

This is not great as it these tools may affect the test results or break the ability to run the tests at all.
Other than that, they make it impossible to simulate the portal behavior when we need different access policies.

To handle this, add fixtures that creates very simple wrapper scripts that just return the file-access mode based on the tests fixtures.

By default we return the "hidden" mode, but this can be overridden in each test using different fixtures.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -126,7 +126,12 @@ def xdp_mocked_executables(xdp_app_info: xdp.AppInfo) -> list[xdp.ExecutableMock
     else:
         return []
 
-    return [xdp.ExecutableMock(executable=exe)]
+    return [
+        xdp.ExecutableMock(
+            executable=exe,
+            access_mode=xdp.FileAccessMode.HIDDEN,
+        )
+    ]
 
 
 @pytest.fixture
--- a/tests/xdp_utils.py
+++ b/tests/xdp_utils.py
@@ -898,9 +898,16 @@ class GDBusIface:
         return GDBusIfaceSignal(signal_id, self._proxy)
 
 
+class FileAccessMode(Enum):
+    READ_WRITE = "read-write"
+    READ_ONLY = "read-only"
+    HIDDEN = "hidden"
+
+
 class ExecutableMock(object):
-    def __init__(self, executable: str):
+    def __init__(self, executable: str, access_mode: FileAccessMode):
         self.executable = executable
+        self.access_mode = access_mode
 
     def create(self, path: Path):
         self.path = path / self.executable
@@ -908,4 +915,4 @@ class ExecutableMock(object):
         self.path.chmod(0o755)
 
     def get_executable(self):
-        return f"#!/usr/bin/env sh".encode("utf8")
+        return f"#!/usr/bin/env sh\necho {self.access_mode.value}".encode("utf8")


tests/run-test.sh
```diff
@@ -41,4 +41,8 @@ export XDG_DESKTOP_PORTAL_PATH="$BUILDDIR/src/xdg-desktop-portal"
 export XDG_DOCUMENT_PORTAL_PATH="$BUILDDIR/document-portal/xdg-document-portal"
 export XDG_PERMISSION_STORE_PATH="$BUILDDIR/document-portal/xdg-permission-store"
 
+if [ ! -v UNDER_MESON ]; then
+  meson compile -C "${BUILDDIR}"
```
swick, line 45: Not so sure about this one. I like that it just runs the tests on whatever binaries you set it to. The changes to meson make a lot of sense to me instead because there you expect meson test to ensure it run on the source files.

tests/conftest.py
```diff
@@ -85,6 +87,17 @@ def xdg_document_portal_path() -> Path:
     return Path(os.environ["XDG_DOCUMENT_PORTAL_PATH"])
 
 
+@pytest.fixture
+def xdp_bin_path(tmp_path: Path) -> Path:
+    (tmp_path / "true").touch(mode=0o755)
+
+    bwrap = shutil.which("bwrap")
```
swick: I think we can improve this a bit by having a fixture which returns an array of binaries that should be visible from the host.

tests/conftest.py
```diff
@@ -660,6 +663,68 @@ def unmounted():
...
 
+def create_file_access_wrapper_script(path: Path, name: str, mode: str) -> None:
+    """
+    Creates a wrapper script for the file access check which just returns
+    the given mode. This is used to simulate different file access modes.
+    """
+    if not path.exists():
+        raise FileNotFoundError(f"{path} does not exist")
+
+    script = path / name
```
swick: I also think that we might just want to make this a bit more extendable and instead of having either the flatpak_dummy_binary or the flatpak binary here, we create a python script which passes the args back to pytest so we can mock it here however we want.

tests/run-test.sh
3v1n0: Yeah, sure, but doing the logic in the script for that is a bit problematic... We could do maybe python side, by calling meson compile with a fixture needed for the test that is being run, but it looked a bit an overkill

3v1n0: I mean, compiling the project takes an instant, so recompiling unneded binaries it's not too bad when developing

tests/conftest.py
3v1n0: Ok

3v1n0: Also bwrap is only needed in few tests, but I thought wasn't too bad to just expose it all the times.

Or we can have a fixture and make it explicit, but it can be annoying to maintain since in CI we do not check bwrap

tests/conftest.py
3v1n0: Yeah, I guess you go too far with my pytest knowledge... :D

swick: I think something like this is a good start: https://github.com/swick/xdg-desktop-portal/commit/080a07a46bc3b188a50ad8254839e830ec5ed776

Separates the actual mocking from the conftest, and allows overriding and adjusting via the fixtures.

tests/run-test.sh
swick: Alright, let's go with it.

tests/conftest.py
3v1n0: Nice, I've included it here, splitting it a bit.

However, the only thing that I'm not sure about (that was also preventing me to do something of this kind) was the fact that if i'm not wrong this would bring `xdp_app_info` in all the times, even when not needed, so causing all the tests (and not only the ones we explicitly want) being repeated for each kind, no?

swick: We already run each test with each AppInfoKind:
```
@pytest.fixture(
    params=[
        xdp.AppInfoKind.HOST,
        xdp.AppInfoKind.FLATPAK,
        xdp.AppInfoKind.SNAP,
        xdp.AppInfoKind.LINYAPS,
    ]
)
def xdp_app_info(request) -> xdp.AppInfo:
...
```

The `xdp_mocked_executables` is not "pulling in" anything, but just gets the AppInfo which was created for the current test.

Issue #1838, opened 2025-10-22
Refactoring tracking issue


AdrianVovk:
In my reviews of #1819 and others I keep running into refactors we should do "in another PR", but it's getting rather difficult to keep track of them. So this issue is here to at least have them written down somewhere so we can track them.

- [ ] Find all the instances of `g_signal_connect` and make sure we replace them with `g_signal_connect_object` or otherwise find some way to ensure that we disconnect the signal handlers when the userdata is destroyed (otherwise unlucky timing might cause a user-after-free)

- [ ] Figure out what to do about the various portal implementations not having a library PREFIX and so it being impossible for us to `DEFINE_FINAL_CLASS`. This causes us to do gobject boilerplate manually, and prevents us from having type-checked casts and so on

- [ ] Once we figure out how to define the classes, we need to go through each portal and replace all the C casts with type-checked casts

- [ ] We need to go through all the portals and fix up their version number logic. See last few comments on https://github.com/flatpak/xdg-desktop-portal/pull/766

- [ ] We need to figure out how to shut down portals so that disposing the context is done only afterwards and it becomes safe to keep unowned references to the context

- [ ] Requests and Sessions are currently global but really belong to a XdpAppInfo

- [ ] A client disconnecting should be a signal on the XdpAppInfo

- [ ] D-Bus method call get dispatched in threads but a lot of the code seems to not take that into account. We need to either make things thread safe or remove the usage of threads.

- [ ] There are a lot raw D-Bus calls, property accesses, etc. which should be handled by a proxy

- [ ] Helper functions to obtain proxies to common `impl` objects, like `Request`s

- [ ] Unify the code style/formatting, and add a linter for it

- [ ] Use quarks/`g_object_get_qdata`/`g_object_set_qdata` instead of string access via `g_object_get_data`/`g_object_set_data`

- [ ] Unify error handling: use XDG_DESKTOP_PORTAL_ERROR consistently, return a generic XDG_DESKTOP_PORTAL_ERROR_FAILED on internal and impl errors

- [ ] Logging/warning should use a system where the API call that resulted in it is recorded. Something along the lines of traces with spans, possibly with journalctl structured logging.

PR #1911, opened 2026-02-20
Use peer-disconnect signal and attaching to XdpAppInfo

swick:
Includes https://github.com/flatpak/xdg-desktop-portal/pull/1879

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -1506,6 +1506,7 @@ on_peer_disconnect (const char *name,
                     gpointer    user_data)
 {
   stop_file_transfers_for_sender (name);
+  xdp_app_info_registry_delete (app_info_registry, name);
 }
 
 static void


src/xdp-request.c
```diff
@@ -320,59 +347,3 @@ xdp_request_set_impl_request (XdpRequest         *request,
...
-                                   GCancellable *cancellable)
-{
-  const char *sender = (const char *)task_data;
-  GSList *list = NULL;
-  GSList *l;
-  GHashTableIter iter;
-  XdpRequest *request;
-
-  G_LOCK (requests);
-  if (requests)
```
AdrianVovk, line 337: Is this not the last real use of the `requests` global? So we can drop it?

src/xdp-session.c
```diff
@@ -316,6 +270,18 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
...
 }
 
+static void
+on_peer_disconnect (XdpContext *context,
+                    const char *peer,
+                    gpointer    user_data)
+{
+  XdpSession *session = XDP_SESSION (user_data);
+
+  REQUEST_AUTOLOCK (session);
```
AdrianVovk: SESSION_AUTOLOCK

src/xdp-session-persistence.c
```diff
@@ -29,6 +29,37 @@ static GHashTable *transient_permissions;
 
 #define RESTORE_DATA_TYPE "(suv)"
 
+static void
+xdp_session_persistence_delete_transient_permissions_for_sender (const char *sender_name)
```
AdrianVovk: why keep this as a separate function?

src/usb.c
```diff
@@ -1563,23 +1545,6 @@ usb_new (XdpContext     *context,
   return usb;
 }
 
-void
-xdp_usb_delete_for_sender (XdpContext *context,
-                           const char *sender)
-{
-  GDBusInterfaceSkeleton *usb_skeleton =
-    xdp_context_get_portal (context, USB_DBUS_IFACE);
```
AdrianVovk, line 1568: Seems like this is the last use of xdp_context_get_portal, so we can drop it?

src/xdp-app-info-registry.c
```diff
@@ -72,6 +82,14 @@ xdp_app_info_registry_new (void)
                                                g_object_unref);
   g_mutex_init (&registry->app_infos_mutex);
 
+  if (context)
```
AdrianVovk: tbh, it's probably fine to leave this one as it was, given the weird special casing here. (Except, of course, keep the change above that calls `xdp_app_info_registry_delete` in the document-portal, that's a solid bug fix)

The previous approach, unlike most of the portals, wasn't using any global state. The context already keeps around a reference to the app_info_registry and will continue to do so

src/usb.c
```diff
@@ -251,20 +251,23 @@ usb_sender_info_from_app_info (XdpUsb     *self,
...
+      g_autoptr(UsbSenderInfo) sender_info_owned = NULL;
+
+      sender_info = sender_info_owned = usb_sender_info_new (sender, app_info);
+      g_hash_table_insert (self->sender_infos,
+                           g_strdup (sender),
+                           g_steal_pointer (&sender_info_owned));
     }
 
   g_assert (sender_info != NULL);
-  g_atomic_ref_count_inc (&sender_info->ref_count);
```
AdrianVovk, line 265: Looks like UsbSenderInfo doesn't need a reference count anymore, since it's always just gonna have one direct owner?

src/usb.c
```diff
@@ -72,7 +73,6 @@ struct _XdpUsb
   GHashTable *syspaths_to_ids;
 
   GHashTable *sessions;
-  GHashTable *sender_infos;
```
AdrianVovk, line 75: nit about the commit message: this wasn't using a global before

src/usb.c
```diff
@@ -247,36 +239,29 @@ usb_sender_info_new (const char *sender_name,
...
-  if (!sender_info)
-    {
-      g_autoptr(UsbSenderInfo) sender_info_owned = NULL;
-
-      sender_info = sender_info_owned = usb_sender_info_new (sender, app_info);
-      g_hash_table_insert (self->sender_infos,
-                           g_strdup (sender),
-                           g_steal_pointer (&sender_info_owned));
-    }
+  sender_info = g_object_get_data (G_OBJECT (app_info), "-xdp-usb-sender-info");
```
AdrianVovk, line 247: I was gonna suggest quarks, but it seems that most of the x-d-p codebase isn't using them so maybe it goes onto the "to refactor later" pile (i.e. #1838). What do you think?

src/usb.c
```diff
@@ -1563,23 +1545,6 @@ usb_new (XdpContext     *context,
   return usb;
 }
 
-void
-xdp_usb_delete_for_sender (XdpContext *context,
```
AdrianVovk, line 1564: You forgot to delete the prototype

src/usb.c
swick: We could, the question is if we should. Probably going to come in handy at some point.

src/usb.c
swick: Did that before but undid it again because I don't want to introduce unrelated changes and it certainly continues working, being ref-counted.

src/usb.c
swick: Yup!

swick:
@AdrianVovk ping

src/usb.c
AdrianVovk: Doesn't seem too unrelated, and saves having to clean it up later. Or confusion about why there's a refcount. But sure fine doesn't really matter in the grand scheme of things

src/usb.c
AdrianVovk: IDK, it's kinda global-state-y in a way that we're trying to move away from. But fine I don't think it's super important to drop it

PR #1913, opened 2026-02-23
A few small fixes for the remote desktop portal

whot:


data/org.freedesktop.impl.portal.RemoteDesktop.xml
```diff
@@ -162,6 +162,12 @@
 
           The device types that can be used for remote control. See the
           AvailableDeviceTypes property.
+
+        * ``restore_data`` (``(suv)``)
```
swick, line 175: I was just checking if this actually exists, and noticed that `handle_start` just ignores `arg_options` and always passes an empty vardict to the impl.

swick: Opps, looked at the options and not the results. So never mind this.

However, there are two `devices` here, and `stream` is documented in the impl but not the frontend. The impl also doesn't document `restore_data` and `persist_mode`.

whot: done, I think

PR #1914, opened 2026-02-23
Some minor fixes to session persistence

whot:
The first three commits from #1898, can be merged independently and there's no new protocol involved.


src/xdp-session-persistence.c
```diff
@@ -157,7 +157,7 @@ xdp_session_persistence_get_persistent_permissions (XdpSession *session,
   g_autoptr(GVariant) perms = NULL;
   g_autoptr(GVariant) data = NULL;
   g_autoptr(GError) error = NULL;
-  const char **permissions;
```
GeorgesStavracas, line 160: How about `g_auto(GStrv) permissions = NULL`  instead?

swick: `GStrv` gives you `g_strfreev` which is a deep free, which is wrong in this case.

PR #1915, opened 2026-02-24
doc: Add a list of specialists for different portal topics

swick:
The immediate reason for wanting this list is that the printing portal is constantly lacking reviewers who can judge protocol changes properly, but I'd be happy to add more people for more topics.

### DIFF
--- a/doc/merge-requirements.rst
+++ b/doc/merge-requirements.rst
@@ -68,3 +68,15 @@ Linyaps representative:
 - He YuMing
 - Iceyer
 - reddevillg
+
+Specialists
+-----------
+
+Portals cover a lot of different specialized topics. Committers might not be
+familiar with all of them, so it can make sense to defer to specialists in those
+topics, or in general get feedback from them. The list here should help you to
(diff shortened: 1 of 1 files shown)


doc/merge-requirements.rst
```diff
@@ -68,3 +68,15 @@ Linyaps representative:
...
 - Iceyer
 - reddevillg
+
+Specialists
+-----------
+
+Portals cover a lot of different specialized topics. Committers might not be
+familiar with all of them, so it can make sense to defer to specialists in those
+topics, or in general get feedback from them. The list here should help you to
+find and contact those specialists:
```
AdrianVovk, line 78: Should the specialist have to sign off on non-trivial changes to the portals they specialize in?

swick: I think the committers should use their own good judgement on when this makes sense. You'd have to rely on them for judging what a non-trivial change is anyway. Let's not add more bureaucracy without it proving necessary.

PR #1917, opened 2026-02-25
xdp: Use the same, strict check for valid dbus fd indexes

swick:
The file descriptor index we get from dbus callers can be negative as well, so let's always make sure we get an index which is actually in the GUnixFdList. While at it, let's use the same check for negative and positive checks (minus the inversion) so we don't even have to think about it.

### DIFF
--- a/src/clipboard.c
+++ b/src/clipboard.c
@@ -255,39 +255,27 @@ selection_write_done (GObject *source_object,
     {
       g_dbus_error_strip_remote_error (error);
       g_warning ("A backend call failed: %s", error->message);
+
+      g_dbus_method_invocation_return_error (invocation,
+                                             XDG_DESKTOP_PORTAL_ERROR,
+                                             XDG_DESKTOP_PORTAL_ERROR_FAILED,
+                                             "Internal error");
+      return;
     }
 
   out_fd_list = g_unix_fd_list_new ();
-
-  if (fd_handle)
+  if (!xdp_copy_fd_to_lists (fd_list, out_fd_list,
+                             g_variant_get_handle (fd_handle),
+                             &out_fd_id,
+                             &error))
     {
-      int fd_id = g_variant_get_handle (fd_handle);
+      g_warning ("Passing a fd from impl to frontend failed: %s",
+                 error->message);
 
-      if (fd_id < g_unix_fd_list_get_length (fd_list))
-        {
-          g_autofd int fd = -1;
-
-          fd = g_unix_fd_list_get (fd_list, fd_id, &error);
-
-          if (fd >= 0)
-            out_fd_id = g_unix_fd_list_append (out_fd_list, fd, &error);
-        }
-      else
-        {
-          g_set_error_literal (&error, XDG_DESKTOP_PORTAL_ERROR,
-                               XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
-                               "Bad file descriptor index");
-        }
-    }
-
-  if (out_fd_id == -1)
-    {
-      g_dbus_method_invocation_return_error (
-        invocation,
-        XDG_DESKTOP_PORTAL_ERROR,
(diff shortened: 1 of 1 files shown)


src/clipboard.c
```diff
@@ -263,7 +263,7 @@ selection_write_done (GObject *source_object,
     {
       int fd_id = g_variant_get_handle (fd_handle);
 
-      if (fd_id < g_unix_fd_list_get_length (fd_list))
+      if (fd_id >= 0 && fd_id < g_unix_fd_list_get_length (fd_list))
```
orowith2os: This is a really small and stupid nit, but why not `if (fd_id && fd_id < g_unix...)`? would it not work the same?

orowith2os: Scratch that. I thought if would check for nonnegative values, not just if it's not zero.

src/gamemode.c
```diff
@@ -386,7 +386,7 @@ handle_call_in_thread_fds (XdpDbusGameMode       *object,
   g_autoptr(GTask) task = NULL;
   CallData *call_data;
 
-  if (fdlist == NULL || g_unix_fd_list_get_length (fdlist) != 2)
```
AdrianVovk, line 389: Why was this check here before? Do we know for sure that the `handle-*` signals will always give us a non-null GUnixFDList?

src/clipboard.c
AdrianVovk: This logic is repeated over and over and over again. Good candidate for a macro or inline function or something

AdrianVovk (changes requested):
Otherwise, LGTM

src/gamemode.c
swick: dbus call handlers and calls which have a UnixFdList as argument always get you a valid, non-NULL UnixFdList. But honestly, I don't want to touch gamemode, so I dropped it.

src/clipboard.c
swick: Put it in a `xdp_is_fd_list_index_valid` inline function.

src/clipboard.c
```diff
@@ -257,37 +258,24 @@ selection_write_done (GObject *source_object,
       g_warning ("A backend call failed: %s", error->message);
     }
 
-  out_fd_list = g_unix_fd_list_new ();
-
-  if (fd_handle)
```
AdrianVovk, line 262: Hmm, if the call above fails `fd_handle` will be NULL but this code will execute anyways. I don't think it's actually save to pass NULL into `g_variant_get_handle`

src/clipboard.c
```diff
@@ -422,30 +411,18 @@ selection_read_done (GObject *source_object,
...
```
AdrianVovk, line 427: Ditto here. Again lost NULL check that seems required

src/xdp-utils.c
```diff
@@ -1444,3 +1444,34 @@ xdp_map_tids (ino_t    pidns,
```
AdrianVovk, line 1444: minor: commit message is out of date. This commit no longer uses the new helper

src/clipboard.c
```diff
@@ -255,39 +255,27 @@ selection_write_done (GObject *source_object,
...
 
-  if (out_fd_id == -1)
-    {
-      g_dbus_method_invocation_return_error (
-        invocation,
-        XDG_DESKTOP_PORTAL_ERROR,
-        XDG_DESKTOP_PORTAL_ERROR_FAILED,
-        "Failed to append fd: %s",
-        error->message);
+      g_dbus_method_invocation_return_error (invocation,
```
AdrianVovk, line 275: I wonder if we should have a "log warning and return generic error" helper function. Or just a "return generic error" helper...

swick: I already went way off target here. I agree that it would make sense. Let's save it for when we rewrite the portals with libdex.

PR #1921, opened 2026-03-01
access: remove DE-specific messages

### LINKED ISSUE #1593: Screenshot portal permission request is misleading
### Operating System

Fedora 41

1.18

_No response_

### Desktop Environment

KDE

_No response_

### Expected Behavior

The portal implementation should be able to create a prompt that matches the desktop environment it's on, ideally completely custom, alternatively at least one that has text from the implementation.

### Current Behavior

The permissions request tells users that "This permission can be changed at any time from the privacy settings", but there's nothing called "privacy settings"...

### Steps to Reproduce

Open Flameshot for the first time.

_No response_

xi:
fixes #1593

If needed, DE-specific messages can be added in the respective access portal backends.

### DIFF
--- a/src/background.c
+++ b/src/background.c
@@ -919,7 +919,6 @@ handle_request_background_in_thread_func (GTask *task,
       g_autofree char *app_id = NULL;
       g_autofree char *title = NULL;
       g_autofree char *subtitle = NULL;
-      g_autofree char *body = NULL;
       guint32 response = 2;
       g_autoptr(GVariant) results = NULL;
       g_autoptr(GError) error = NULL;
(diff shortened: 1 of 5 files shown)


danirabbit:
I agree with this in theory. I have a branch for pantheon's portal to override text in location access dialogs, but it uses the icon name to determine what's being requested which is kinda jank.

I think if we had some kind of enum of access portal types it would indeed make sense to let frontends handle this kind of text theirselves

swick:
I'm closing this. If you have a change which does not introduce a regression, please re-open it or open a new PR.

PR #1922, opened 2026-03-02
Draft: Add parental controls to the Accounts portal

davidedmundson:
Applications need to filter content to match the age rating of the user.

The rating restrictions tend to be location and domain specific without
a common ground for where these groupings should be. Instead
applications provide a range of gates and the implementation returns
which group they are between.
    
This change is somewhat influenced by the new California law to make public API available but it's something we want to be pushing for anyway to allow applications to perform relevant filtering.

---

Draft as I haven't tested the complete stack. Implementation is mostly a copy/paste of GetUserDetails.


data/org.freedesktop.impl.portal.Account.xml
```diff
@@ -75,5 +75,35 @@
...
+        The following results get returned via the @results vardict:
+
+        * ``age`` (``u``)
+
+          The user's age in years. Values above 18 are
+          normalized to ``19`` to indicate "18+".
+    -->
+    <method name="GetParentalControls">
+      <arg type="o" name="handle" direction="in"/>
+      <arg type="s" name="app_id" direction="in"/>
```
davidedmundson: I don't feel I need this, I was just copying the parent.

davidedmundson:
One TODO is that the point of Apple's 3 arg API (rather than 2 args N times) is presumably to rate limit it so that can't just bisect an exact value. That's probably the job of this portal?

EtchedPixels:
Even a daily rate limit on the query means an app will discover child dates of birth on the day they transition between age classes. That means it does disclose date of birth exactly (and I think unfixably while in compliance with California as it stands). That means things like GDPR apply in Europe, and as it may be data from a minor must also be handled in accordance with other (eg US state) laws on processing data for children under 13.

I'm not saying it's a bad idea to have parental control - but that you have compliance and regulatory issues to deal with that are rather more complicated than just "it's an API" and the API has to be able to deal with those compliance issues (eg returning a "not permitted" status). The Apple approach doesn't appear to fix the problem.

The API also needs geolocation or some kind of regulatory locale information in order to figure out what answer ranges are permitted. California for example requires

"Send only the minimum amount of information necessary to comply with this title and shall not share the digital signal information with a third party for a purpose not required by this title."

So sending age information outside of the four categories in 1798.501 (a)(2) appears tbe be a violation of 1798.501 (a)(3) because the law was rather badly drafted.

data/org.freedesktop.impl.portal.Account.xml
```diff
@@ -75,5 +74,40 @@
...
+        @response: Numeric response
+        @results: Vardict with the results of the call
+
+        Gets parental control related age information about the user.
+
+        Supported keys in the @options vardict include:
+
+        The following results get returned via the @results vardict:
+
+        * ``low`` (``i``) and ``high`` (``i``)
```
AdrianVovk: Hmm, should the portal backend just return the real age? And then the frontend can handle all the range calculation?

AdrianVovk:
> The other way to approach this would be to turn it on its head, and instead of having a portal which tells apps what age the user is, instead have a portal which apps can query to tell them whether content which has a certain rating should be shown to the user.

This would be non-compliant with the law. The law specifically talks about age ranges

danirabbit:
> The other way to approach this would be to turn it on its head, and instead of having a portal which tells apps what age the user is, instead have a portal which apps can query to tell them whether content which has a certain rating should be shown to the user.

@pwithnall thats what this does. The app sends up to 3 age gates and then the portal responds. You can send just one gate :)

data/org.freedesktop.impl.portal.Account.xml
```diff
@@ -42,11 +42,10 @@
...
-
-        * ``reason`` (``s``)
+        The method takes three input age gates (``gate1``, ``gate2``, ``gate3``)
+        as unsigned integers from the application (e.g. 13, 16, 18).
 
-          A string that can be shown in the dialog to explain why the information is needed.
+        Supported keys in the @options vardict include:
           This should be a complete sentence that explains what the application will do with
           the returned information, for example: Allows your personal information to be included
           with recipes you share with your friends.
```
ArrayBolt3, line 52: This looks like it got slightly mangled; the 'reason' key and the first line of its documentation were lost but the remaining three lines of documentation were kept.

data/org.freedesktop.impl.portal.Account.xml
ArrayBolt3: The real age might not be known by any application on the system, frontend, backend, or otherwise. Some portal implementations will want to (need to) store only just enough information for age bracket signals to work legally while not storing any information that could be used to fingerprint the user if the device becomes infected by malware. (Whonix is the primary system that will need this functionality, and it is essential enough that we will potentially have to make our own custom implementation if a standard implementation does not take this into account.)

data/org.freedesktop.portal.Account.xml
```diff
@@ -39,6 +39,10 @@
 
         Gets information about the user.
 
+        The method takes three input age gates (``gate1``, ``gate2``, ``gate3``)
+        as unsigned integers. The backend compares the user's age against these
+        gates to determine a range.
+
```
ArrayBolt3: This is no longer true since the functionality was broken into a separate method.

data/org.freedesktop.impl.portal.Account.xml
```diff
@@ -42,11 +42,10 @@
 
         Gets information about the user.
 
-        Supported keys in the @options vardict include:
-
-        * ``reason`` (``s``)
```
dylanmtaylor, line 47: The reason key docs were removed as part of this edit and should be restored for GetUserInformation. The age gate text should be on the new GetParentalControls method. There is no reason (heh) to drop the keys for options, including the reason key from this documentation block.

src/account.c
```diff
@@ -233,10 +303,81 @@ handle_get_user_information (XdpDbusAccount *object,
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
+static gboolean
+handle_get_parental_controls (XdpDbusAccount *object,
+                              GDBusMethodInvocation *invocation,
+                              const gchar *arg_parent_window,
+                              guint arg_gate1,
```
dylanmtaylor: Consider using an array of integers instead of 3 fixed gate arguments. As discussed in the thread, different jurisdictions have different numbers of legally relevant age boundaries. A fixed size of 3 bakes Apple's specific API shape into a freedesktop standard.

dylanmtaylor: Also, just noticed, but these are _unsigned_ values but the spec calls for using -1 as a signal. I think gates need to be signed integers.

ArrayBolt3 (commented):
I don't have the understanding of glib code needed to understand all of what's happening here, but the concept looks good to me. A lot is left implementation-defined, but that's arguably a good thing.

src/account.c
```diff
@@ -233,10 +303,81 @@ handle_get_user_information (XdpDbusAccount *object,
...
+  Account *account = (Account *) object;
+  XdpRequest *request = xdp_request_from_invocation (invocation);
+  const char *app_id = xdp_app_info_get_id (request->app_info);
+  XdpPermission permission;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(XdpDbusImplRequest) impl_request = NULL;
+  g_auto(GVariantBuilder) options =
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
+
+  g_debug ("Handling GetParentalControls");
```
dylanmtaylor: Naming: the discussion seems to be converging on "age verification" rather than "parental controls". These are distinct concepts and as the goal is legal compliance and not content filtering here, I think this should be renamed.

src/xdp-types.h
```diff
@@ -54,6 +48,9 @@ typedef struct _XdpDbusImplAccess XdpDbusImplAccess;
 
 #define ACCOUNT_DBUS_IFACE DESKTOP_DBUS_IFACE ".Account"
 #define ACCOUNT_DBUS_IMPL_IFACE DESKTOP_DBUS_IMPL_IFACE ".Account"
+/* Permissions for Account portal */
+#define ACCOUNT_PERMISSION_TABLE "account"
+#define ACCOUNT_PERMISSION_PARENTAL_CONTROLS "parental-controls"
```
dylanmtaylor: I'd also change to age verification here.

tests/templates/account.py
```diff
@@ -62,3 +62,31 @@ def GetUserInformation(self, handle, app_id, window, options, cb_success, cb_err
...
+    in_signature="ossuuua{sv}",
+    out_signature="ua{sv}",
+    async_callbacks=("cb_success", "cb_error"),
+)
+def GetParentalControls(self, handle, app_id, window, gate1, gate2, gate3, options, cb_success, cb_error):
+    logger.debug(f"GetParentalControls({handle}, {app_id}, {window}, {options})")
+    params = self.account_params
+
+    # Default to no information known
+    results = params.results if params.results else {"low": dbus.Int32(-1), "high": dbus.Int32(-1)}
```
dylanmtaylor: The default result of (-1, -1) for "no information known" should be documented in the spec. Right now there's no specification of what values the backend should return when it can't determine the user's age, or what the semantics are.

tests/test_account.py
```diff
@@ -31,7 +31,7 @@ def required_templates():
 
 class TestAccount:
     def test_version(self, portals, dbus_con):
-        xdp.check_version(dbus_con, "Account", 1)
+        xdp.check_version(dbus_con, "Account", 2)
```
dylanmtaylor: The version of this test file is being incremented without any changes. GetParentalControls isn't even being exercised here.

src/account.c
```diff
@@ -155,6 +156,75 @@ get_user_information_done (GObject *source,
...
+    G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
+  g_autoptr(GVariant) lowv = NULL;
+  g_autoptr(GVariant) highv = NULL;
+
+  REQUEST_AUTOLOCK (request);
+
+  response = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (request), "response"));
+  results = (GVariant *)g_object_get_data (G_OBJECT (request), "results");
+
+  if (response != 0)
```
dylanmtaylor: Consider adding a NULL check on results before the lookups.

src/account.c
```diff
@@ -233,10 +303,81 @@ handle_get_user_information (XdpDbusAccount *object,
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
+static gboolean
+handle_get_parental_controls (XdpDbusAccount *object,
+                              GDBusMethodInvocation *invocation,
+                              const gchar *arg_parent_window,
```
dylanmtaylor: What is the parent window being used for here?

data/org.freedesktop.impl.portal.Account.xml
AdrianVovk: Hmm fair enough. But then the age gates requested by the frontend should be passed into the backend as well, and the backend should be totally responsible for implementing all of the logic here. This weird hybrid situation where the backend returns a range and then the portal tries to mangle it into another range won't actually work.

Like if the backend says "the user is between 15 and 17". Then the app tries to query with gates of (13, 16, 18). What should the portal do? The user might be 15, and the portal should return (13,16). Or the user might be almost 17, and then the portal should return (16,18).

src/account.c
```diff
@@ -233,10 +303,81 @@ handle_get_user_information (XdpDbusAccount *object,
...
+
+  g_debug ("Handling GetParentalControls");
+
+  /* Validate application permission before proceeding */
+  permission = xdp_get_permission_sync (request->app_info,
+                                        ACCOUNT_PERMISSION_TABLE,
+                                        ACCOUNT_PERMISSION_PARENTAL_CONTROLS);
+  if (permission != XDP_PERMISSION_YES)
+    {
+      g_dbus_method_invocation_return_error (invocation,
```
dylanmtaylor: Should a failure to look up the permission be treated the same as "not allowed"?

tests/templates/account.py
```diff
@@ -62,3 +62,31 @@ def GetUserInformation(self, handle, app_id, window, options, cb_success, cb_err
...
+
+
+@dbus.service.method(
+    MAIN_IFACE,
+    in_signature="ossuuua{sv}",
+    out_signature="ua{sv}",
+    async_callbacks=("cb_success", "cb_error"),
+)
+def GetParentalControls(self, handle, app_id, window, gate1, gate2, gate3, options, cb_success, cb_error):
+    logger.debug(f"GetParentalControls({handle}, {app_id}, {window}, {options})")
```
dylanmtaylor: Consider outputting the gate values for debugging.

data/org.freedesktop.impl.portal.Account.xml
AdrianVovk: Alternatively, we just drop the ability for the app to provide gates that it's looking for, as I suggested before. They seem to complicate the design here, and I don't think there's actually a good idea of how they'd be used?

It's just an opportunity for apps to be abusive. Like: we need to consider an app that would also send gates `[13, 14, 15, 16, 17, 18]` to just get the user's exact age and use that for whatever illicit purpose. How would we deal with that? How would we even detect it?

So if we ultimately must ignore the app's age gates, then I don't really see a reason to include them in the spec. After all: unlike Apple here, our API takes an `a{sv}` so we can add arbitrary arguments to the function call later, if we decide that it's necessary

AdrianVovk: Also: the backend might only know an age range, but it could just return the bottom of that range to the frontend when asked. If your distro only knows that the user is between 16-18, then you can just safely tell the frontend that the user is 16 (you know that they are at least 16, so you won't ever over-report their age).

Basically I think it might be useful to centralize the actual compliance bits (the map of location -> mandated age gates) in the frontend, so that desktops don't have to worry about getting it right individually. This means that the backend giving the frontend a range will have the same issues as I described above: what if the range spans multiple legally-required ranges?

If you report a single age that's the bottom age of your range, then you'll always fall into a range that's the age of your user, or younger. Which is the only thing you can do given the lack of information.

src/account.c
AdrianVovk: This is what portal APIs look like. The backend may choose to show UI in response to the request, and that UI might be a dialog that's parented to the app's window.

data/org.freedesktop.impl.portal.Account.xml
ArrayBolt3: Valid point. The backend probably has to be responsible for doing all of it then. If we did stick with the "turn one range into another range", the only legally safe thing to do would be to stick with the lower bound and treat that as a canonical age value. That's not great for the reasons you pointed out above (unfair restriction of some app features), but for some distros that's an acceptable compromise because of the importance of other goals.

Edit: This was originally in reply to https://github.com/flatpak/xdg-desktop-portal/pull/1922#discussion_r2886751476

ArrayBolt3: > Alternatively, we just drop the ability for the app to provide gates that it's looking for, as I suggested before. They seem to complicate the design here, and I don't think there's actually a good idea of how they'd be used?

That sort of removes the "parental controls" aspect from this and leaves it as little more than a compliance mechanism. It leaves the age brackets entirely up to the desktop, and while some desktops may want to specify the age brackets exactly, that might not make sense in every scenario.

If it was to be nothing but a compliance mechanism, it might be good to pair it with a standardized way of tracking what age brackets are required by each jurisdiction. I have a concept for that, but I think it's orthogonal to the portal implementation and probably not something go into here. That would be more suited for a tzdata-like project.

ArrayBolt3: Heh, I read further and see you thought of a centralized database already. I have a non-xdg-desktop-portal implementation spec draft I was going to send to the XDG mailing list before I ran into this. Maybe it's worth posting anyway in case some of the ideas are useful here?

data/org.freedesktop.impl.portal.ParentalControls.xml
```diff
@@ -0,0 +1,46 @@
...
+<!--
+ SPDX-License-Identifier: LGPL-2.1-or-later
+-->
+
+<node name="/" xmlns:doc="http://www.freedesktop.org/dbus/1.0/doc.dtd">
+  <!--
+      org.freedesktop.impl.portal.ParentalControls:
+      @short_description: Backend for the ParentalControls portal
+  -->
+  <interface name="org.freedesktop.impl.portal.ParentalControls">
```
danirabbit, line 11: Just wanted to note here that there was some discussion that seemed to indicate a consensus around not calling this `ParentalControls` and instead using specifically `AgeVerification` since we might want a more content-focused parental controls API and there's hope of age verification maybe being deprecated someday

davidedmundson: There's a few too many discussions!

I know there was consensus to split from Accounts which I've done and I know there was consensus that OARS permissions is something we want to forward and should be a separate method.

I'm not sure what we gain from effectively going for one method per portal. They can still have separate permissions and I can't think of a non ParentalControl reason to be interested in age. 

It's not a hill I'll fight over if there is consensus from others.

One key point I would push against is we should steer clear of the word "verification" given we don't really promise that.

AdrianVovk: > I'm not sure what we gain from effectively going for one method per portal. They can still have separate permissions and I can't think of a non ParentalControl reason to be interested in age.

I'd put it in its own portal, distinct from parental controls, because if the laws go away we can deprecate the API and the portal entirely. Also, we don't know what kind of laws will exist in the future, and we may want to add additional functionality to the portal that's only really usable for age verification.

For instance: Apple has APIs to present a confirmation dialog. This dialog would explain some sort of major changes that happened to an app, and would require a parent to confirm that the app should still be allowed to run for their child. If we need to implement an API like this, that'd be another method to add to the age verification portal.

Also, IMO, if you're interested in parental controls you're not actually interested in age. If you're interested in parental controls, you want to know what kind of content restrictions the parent has set up for their child. Or the user has set for themself. Neither has to do with age. Really the only reason you'd be interested in age is to comply with these laws.

> One key point I would push against is we should steer clear of the word "verification" given we don't really promise that

We can call it `AgeRange` or `AgeSignals`? Apple calls it the "Declared Age Range API". Google's API is called "Play Age Signals API". Windows calls it the "Age Consent API"

pedropaulosuzuki: Agreed, there is no point in calling this "Parental control" if there is no control and only age is being passed, instead of flags for what should or not be allowed for that account. Parents should be able to disable contents based on flags, like "Violence", "Online chat", "Gambling", and so on.

While age-range makes sense, maybe we should call this what it really is: age-telemetry/age-surveillance.

davidedmundson: I'll move it so we have the most flexibility moving forwards.

swick:
I don't think anything constructive can be achieved in this PR anymore, so I'm closing it, but feel free to open new PRs.

PR #1924, opened 2026-03-06
XdpAppInfo Entitlements

swick:
/cc @AdrianVovk @davidedmundson


shared/xdp-app-info.c
```diff
@@ -903,3 +903,17 @@ xdp_app_info_get_usb_queries (XdpAppInfo *app_info)
 
   return XDP_APP_INFO_GET_CLASS (app_info)->get_usb_queries (app_info);
 }
+
+gboolean
```
AdrianVovk: Should entitlements take arguments rather than being strictly strings?

src/xdp-app-info-flatpak.c
```diff
@@ -482,6 +482,22 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
 
+static gboolean
+xdp_app_info_flatpak_has_entitlement (XdpAppInfo *app_info,
+                                      const char *entitlement)
+{
+  XdpAppInfoFlatpak *app_info_flatpak = XDP_APP_INFO_FLATPAK (app_info);
+  g_auto(GStrv) values = NULL;
+  size_t len = 0;
+
+  values = g_key_file_get_string_list (app_info_flatpak->flatpak_info,
```
AdrianVovk: Interesting. Why a string list?

src/trash.c
```diff
@@ -109,6 +109,17 @@ handle_trash_file (XdpDbusTrash *object,
 
   g_debug ("Handling TrashFile");
 
+  if (!xdp_app_info_has_entitlement (app_info,
+                                     "org.freedesktop.portal.Trash.Foo"))
+    {
+      g_dbus_method_invocation_return_error (invocation,
```
AdrianVovk: It'd probably be useful to have a helper that does the has_entitlement check, and returns the portal error, all in one go. Since this would probably be the most common usage.

```c
if (!xdp_app_info_ensure_entitlement (app_info, invocation,
                                      "org.freedesktop.portal.Trash.Foo"))
```

src/xdp-app-info-flatpak.c
swick: That is the format we get from flatpak:
```
[Policy subsystem]
key=v1;v2;
```

and the command line is `--add-policy=subsystem.key=v1`.

So the current implementation would mean:

```
[Policy entitlement]
org.freedesktop.portal.Trash.Foo=true
```

but it also means you can have:

```
[Policy entitlement]
org.freedesktop.portal.Trash.Foo=true;false;foobar
```

and it would mean the same thing.

We could also do:

```
[Policy entitlement]
granted=org.freedesktop.portal.Trash.Foo
```

or something like that. I'm very unsure about it.

Maybe that is better though because it allows us to do the backwards compatible way to enforce more entitlements on existing portals like this:

```
[Policy entitlement]
enforce=strict
granted=org.freedesktop.portal.Trash.Foo;org.freedesktop.portal.OpenURI.asd
```

On the other hand, entitlements are then strictly either granted or not granted; no further metadata. Maybe that's fine.

AdrianVovk: Talked about it a little on Matrix. New idea: maybe we make this like the xdg-intents spec:

```
[Policy entitlement]
grant=org.freedesktop.Something;org.freedesktop.SomethingElse
org.freedesktop.SomethingElse.param1=foo;bar
org.freedesktop.SomethingElse.param2=baz
```

Or

```
[Policy entitlement]
grant=org.freedesktop.Something;org.freedesktop.SomethingElse

[Policy org.freedesktop.SomethingElse]
param1=foo;bar
param2=baz
```

I think I like the latter a little more than the former. Overall this is more flexible than my single argument approach because we can store multiple keys and values this way. And it's simpler than my approach in the common case where we just care if an entitlement is there or not

swick: Right, let's go with
```
[Policy entitlement]
grant=org.freedesktop.Something;org.freedesktop.SomethingElse
```
then.

AdrianVovk:
Another thing I thought of. Some apps might be using portals without knowing it. Let's assume for a moment we gate all portals behind entitlements. Without access to the inhibit portal, GTK will not be notified of an impending logout. Thus the app will receive a surprise SIGTERM rather than going through an orderly session end. In short: the platform might change in ways that requires new entitlements, and apps might not know it.

_On the other hand_, libraries like GTK can log and complain about missing entitlements (or other unexpected portal errors). This way, if your app is missing an entitlement, you'll see a warning about it and can fix it. Additionally, GTK would be able to list the required entitlements for various APIs in the API docs. So perhaps it's not a huge problem... Especially if we make sure to keep track of required entitlements in "porting to new versions of the runtime" documentation

ComixHe:
@swick Thanks for the CC. We’ve actually been thinking about a similar approach to tighten up the capability model, so it’s great to see this landing. We haven't finalized our own technical spec yet, but this is a perfect opportunity for us to sync up and discuss how to align. Will get back to you once we've had a chance to chat internally.

swick:
@3v1n0 do you have any opinions, or do you know anyone working on snap who might?

@ComixHe do you have an update?

ComixHe:
> @ComixHe do you have an update?

We'd like to support this, but we don't have a firm ETA at the moment.

3v1n0:
> @3v1n0 do you have any opinions, or do you know anyone working on snap who might?

Just a bit busy with the release ATM, but either me or @aleasto can handle this I think.

swick:
Depends on #1973 now.

All portals now have a legacy entitlement for their portal unless they opt-out of that behavior.

I think this is pretty much ready at this point.

@3v1n0 If we don't get a response here I'll take it as no objection and will go ahead.

swick:
@matthiasclasen We should talk about how this impacts GTK. If apps want to enable strict mode they need to know what portal functionality GTK uses internally.

aleasto:
Snaps already make deep use of statically-granted permissions to access interfaces outside of x-d-p, enforced through kernel LSM. There the "snap store" controls whether or not the app gets access to the interface by default, but an administrator user can always override and explicitly allow/disallow the interface after install. We even expose this in gnome-control-center.

We can probably have a chat with the snap and snap-store people to see if we can extend the same metadata already used in snaps to instead delegate certain static permission enforcement to x-d-p, without any LSM involvement.

The proposed interface here seems nicely extensible to integrate well with snapd in a future where the above actually happens, so _shy-lgtm_.

swick:
@aleasto let me know if you need anything from me

aleasto:
We'll be meeting with snapd people. Please allow me till Wednesday for more feedback re: snaps.

src/xdp-app-info-flatpak.c
```diff
@@ -482,13 +484,63 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
+  for (size_t i = 0; entitlements && entitlements[i]; i++)
+    g_hash_table_add (app_info->entitlements, g_strdup (entitlements[i]));
+
+  enforce = g_key_file_get_string_list (app_info->flatpak_info,
+                                        "Policy entitlement",
+                                        "enforce",
+                                        NULL,
+                                        NULL);
+
+  if (enforce && g_strv_contains ((const char * const *) enforce, "strict"))
```
AdrianVovk, line 516: Thinking ahead: should we immediately make this a version number of some kind? What happens if we want to add other entitlements down the line that would be backwards incompatible with apps targeting `enforce=strict`? If it's going to devolve into `enforce=strict-v2` then we might as well just treat it as a version number from the get go. Thoughts?

AdrianVovk (approved):
Overall LGTM!

src/xdp-app-info-flatpak.c
swick: It's a string list already so we could just add more fields, like "strict,more-strict,whatever"

AdrianVovk: Should we allow apps to opt into "more-strict" enforcement without also opting into "strict" enforcement?

src/xdp-app-info-flatpak.c
AdrianVovk: Should we implement helper functions for looking up the additional parameters now, so that they're there already when we have reason to use them? It'll take time for the other backends to impement this and it's best they build it with support for arbitrary keys from day 0.

If we need a use-case to implement it, perhaps we should have one here? We can add a way to filter what URI schemes you can pass into the OpenURI portal, for instance

src/xdp-app-info-flatpak.c
swick: I'm not sure why we even wanted to add something else here, so it's kind of hard to reason about that. All I'm saying is that this is definitely already extendable in a way. Maybe it would help if you have a scenario in mind.

shared/xdp-app-info-flatpak.c
```diff
@@ -482,13 +484,63 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
+  app_info->entitlements =
+    g_hash_table_new_full (g_str_hash, g_str_equal,
+                           g_free,
+                           NULL);
+
+  entitlements = g_key_file_get_string_list (app_info->flatpak_info,
+                                             "Policy entitlement",
+                                             "grant",
+                                             NULL,
+                                             NULL);
```
3v1n0: I would still get the error and warn in case, for debugging purposes.

src/xdp-app-info-flatpak.c
```diff
@@ -482,13 +484,63 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
+                           NULL);
+
+  entitlements = g_key_file_get_string_list (app_info->flatpak_info,
+                                             "Policy entitlement",
+                                             "grant",
+                                             NULL,
+                                             NULL);
+
+  for (size_t i = 0; entitlements && entitlements[i]; i++)
+    g_hash_table_add (app_info->entitlements, g_strdup (entitlements[i]));
```
3v1n0: This is leaking if `g_hash_table_add` fails.

shared/xdp-app-info-flatpak.c
```diff
@@ -482,13 +484,63 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
+
+  for (size_t i = 0; entitlements && entitlements[i]; i++)
+    g_hash_table_add (app_info->entitlements, g_strdup (entitlements[i]));
+
+  enforce = g_key_file_get_string_list (app_info->flatpak_info,
+                                        "Policy entitlement",
+                                        "enforce",
+                                        NULL,
+                                        NULL);
```
3v1n0: ditto

src/xdp-app-info-flatpak.c
```diff
@@ -482,13 +484,63 @@ xdp_app_info_flatpak_validate_dynamic_launcher (XdpAppInfo  *app_info,
...
+}
+
+static gboolean
+xdp_app_info_flatpak_has_entitlement (XdpAppInfo         *app_info,
+                                      const char         *entitlement,
+                                      XdpEntitlementKind  kind)
+{
+  XdpAppInfoFlatpak *app_info_flatpak = XDP_APP_INFO_FLATPAK (app_info);
+
+  ensure_entitlements (app_info_flatpak);
```
3v1n0: Why not doing this on init?

Prefer a "lazy", but sync load instead?

src/xdp-context.c
```diff
@@ -243,6 +247,16 @@ authorize_callback (GDBusInterfaceSkeleton *interface,
 
   g_object_set_data (G_OBJECT (invocation), "xdp-app-info", app_info);
 
+  if ((flags & XDP_CONTEXT_EXPORT_FLAGS_NO_ENTITLEMENT) == 0 &&
```
3v1n0: Nit, it's more readable with:

 `!(flags & XDP_CONTEXT_EXPORT_FLAGS_NO_ENTITLEMENT) &&`

desktop-portal/registry.c
```diff
@@ -196,5 +196,6 @@ init_registry (XdpContext *context)
 
   xdp_context_take_and_export_portal (context,
                                       G_DBUS_INTERFACE_SKELETON (g_steal_pointer (&registry)),
-                                      XDP_CONTEXT_EXPORT_FLAGS_HOST_PORTAL);
+                                      XDP_CONTEXT_EXPORT_FLAGS_HOST_PORTAL |
+                                      XDP_CONTEXT_EXPORT_FLAGS_NO_ENTITLEMENT);
```
3v1n0: I would explain better the rationale of this in the commit message

src/xdp-context.c
```diff
@@ -266,6 +280,11 @@ xdp_context_take_and_export_portal (XdpContext             *context,
 
   g_return_if_fail (XDP_IS_CONTEXT (context));
   g_return_if_fail (G_IS_DBUS_INTERFACE_SKELETON (skeleton));
+  if (flags & XDP_CONTEXT_EXPORT_FLAGS_HOST_PORTAL)
+    g_return_if_fail (flags & XDP_CONTEXT_EXPORT_FLAGS_NO_ENTITLEMENT);
+
+  g_object_set_data (G_OBJECT (skeleton), "-xdp-skeleton-export-flags",
```
3v1n0: Not a fan of data object, but it's hard to avoid here, so can we at least use a `#define` for this?

tests/test_entitlements.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def entitlements_good():
+    yield (False, [])
+    yield (False, ["org.foo.bar"])
+    yield (True, ["org.freedesktop.portal.Print"])
+    yield (True, ["org.foo.bar", "org.freedesktop.portal.Print"])
+
+
+def entitlements_bad():
```
3v1n0, line 19: Maybe a docstring or typed return values could help in future to understand that they stand for `strict` and `entitlement-id` (well the boolean one is the more unclear :)).

Or use an enum instead.

3v1n0 (commented):
Code only review...

On the change itself we're going to discuss this with the snapd team soon, however, from an high-level prospective, I guess it would imply for it to support new snap file fields that allow to define entitlements and finally expose this in the rest API.

Now, for now I'd assume that the snap implementation would be "allow all", or use the already supported interfaces to expose whether some portal should be allowed.

src/xdp-app-info-flatpak.c
aleasto: actually what if entitlements change on an app update? it would require a restart of xdg-desktop-portal to pick up the new entitlements

3v1n0:
Just wondering, but there won't be a case in which (non-strict?) entitlements could be allowed only after asking to the user, right?

shared/xdp-app-info-flatpak.c
swick: What would this help with? Entitlements being absent is completely valid, and usually you'd only be interested in knowing which entitlements an app has, and while it currently is just a list, entitlements can have arbitrary metadata.

src/xdp-app-info-flatpak.c
swick: The docs say it either replaces and existing one or doesn't, but it gets always inserted. If the existing is replaced, the free func of the hash table frees it.

src/xdp-context.c
swick: Define the string? At that point this could also just use qdata.

src/xdp-app-info-flatpak.c
3v1n0: Yep exactly this was my follow-up question too.

swick: I did it like this to avoid doing the work at all if no entitlements were used, but now every portal call uses entitlements, so there isn't a good reason anymore.

> actually what if entitlements change on an app update?

FWIW, the entitlements are static metadata, so they only change when the app gets updated. The portal looks at the instance app metadata, so if an app gets updated and has different entitlements, the currently running instances correctly still have the old entitlements. New instances, have the new ones. The XdpAppInfo objects are per connection, so never cross-instances, so this all works out very nicely.

I don't know how snap works, but that's definitely something to keep in mind.

swick:
> Just wondering, but there won't be a case in which (non-strict?) entitlements could be allowed only after asking to the user, right?

The intent with entitlements is that they are all app metadata, not user choices (with the caveat that users can override this locally with e.g. flatseal). No matter if enforcement is done in strict mode or not.

src/xdp-app-info-flatpak.c
aleasto: > XdpAppInfo objects are per connection

Ah, ok. Makes sense.

3v1n0:
> > Just wondering, but there won't be a case in which (non-strict?) entitlements could be allowed only after asking to the user, right?
> 
> The intent with entitlements is that they are all app metadata, not user choices (with the caveat that users can override this locally with e.g. flatseal). No matter if enforcement is done in strict mode or not.

Ok, so basically similar to what we have with `snap connections` to [interfaces](https://snapcraft.io/docs/supported-interfaces), despite the fact that these are global and not user-bound, but I also think that it would be the same here for entitlements.

aleasto:
So we had a chat with snapd people and this interface seems adequate to implement it in snapd, so LGTM.

3v1n0:
> So we had a chat with snapd people and this interface seems adequate to implement it in snapd, so LGTM.

I guess for now the snap implementation could just be "allow all", then we'll need to adjust the `snap routine` or the snap REST API to get such information.

src/xdp-context.c
3v1n0: Yeah, even better.

src/xdp-app-info-flatpak.c
3v1n0: Well one argument for having them more dynamic for the snapd folks was to  potentially prompt to user whether they wanted to enable an entitlement

aleasto:
> > So we had a chat with snapd people and this interface seems adequate to implement it in snapd, so LGTM.
> 
> I guess for now the snap implementation could just be "allow all", then we'll need to adjust the `snap routine` or the snap REST API to get such information.

I think the current default already makes sense: allow all LEGACY and disallow other entitlements.

src/xdp-app-info-flatpak.c
AdrianVovk: Let's say we introduce this as-is and it gets released. Next we decide that we want to make it so that the `OpenURI` portal actually filters out the scheme types you can pass in. So for instance, an app has to declare "I might open http://, https://, and foobar:// URIs". But that's a breaking change! So we need to make the app opt into this stricter enforcement. We can't tie this into `enforce=strict` because that's a backwards-incompatible change.

So the point is we need a mechanism to let apps opt into this stricter enforcement profile. And whatever we come up with should be applicable to all the different packaging systems, and also scalable for future changes where we continue to maintain backwards compat forever

src/xdp-app-info-flatpak.c
AdrianVovk: Entitlements are inherently static permissions, not dynamic permissions. So you can prompt at install/update time but not at runtime.

Since entitlements gate access to portals, prompting at runtime would be pretty silly in general. "Can I ask for permission to use the camera?"

src/xdp-app-info-flatpak.c
iinuwa: If we still need a usecase, the credentials portal will need something like (names TBD)[^1]:

```
[Policy entitlement]
# only allow accessing credentials for restricted origins
grant=org.freedesktop.credentials.RestrictedOrigin

[Policy org.freedesktop.credentials.RestrictedOrigin]
origins=https://example.com;https://example.org
```

[Policy 
```
[Policy entitlement]
# allow accessing credentials for any origins
grant=org.freedesktop.credentials.PrivilegedClient
```

[^1]: The specific organization of entitlements can be discussed elsewhere; another way we could do this is two have 3 entitlements: one that generically associates an app with a website and includes a "can request credentials" claim for that, and another "web browser" entitlement with a claim to request credentials.

src/xdp-app-info-flatpak.c
swick: I thought about this a bit and I don't think it's a good idea. We're essentially adding another permission system for the entitlements which are a permission system for the permission system.

Maybe we should just not have enforce=strict for now, or call it "experimental-strict", and give us a bit more time to go through existing portals and see which ones might benefit from more fine grained entitlements?

src/xdp-app-info-flatpak.c
iinuwa: I think it still can make sense to prompt even with an entitlement. The static nature of the entitlement means that the app distributors can reason about it, but the user is still in control if they want specific features of the app but don't want to use the features that require more sensitive access.

Android does this, for example. An app declares that it needs access to the camera for specific things, but these permissions are not effective until the user is prompted by the system:

```
Do you want to allow Foo App
to access your location?

-------------------------
|       This time       |
-------------------------

-------------------------
| While the app is open |
-------------------------

-------------------------
|        Always         |
-------------------------
```

The app is able to query the state of the permission so that it can explain to the user why the app needs X permission.

I think this can still be accomplished in a portal-specific way by requiring entitlements _and_ saving any persistent user consent to the permissions portal.

(If I'm talking about something different than you are, then ignore me. :) )

src/xdp-app-info-flatpak.c
AdrianVovk: It's a backwards compatibility mechanism, not a permission system. I'm just saying that one day someone somewhere is going to come and find some feature they can add as an entitlement to lock some portal down further in a nice way. And when that day comes, we'll need to make an API breaking change. Which means that we're going to have to treat this like a version number: "When this app was packaged, we were targeting v5 entitlements. There's now a set of v6 entitlements that lock things down further, but when the app was packaged only the v5 entitlements existed so we can't enforce the v6 entitlement checks"

It's like android and its API versions. The new scoped storage rules applied only to apps that were compiled after the requirements were introduced. And Google could tell because the app has a "target API version" field in its configuration

src/xdp-app-info-flatpak.c
AdrianVovk: You're misunderstanding the purpose of entitlements a little.

On Android, an app declares that it needs access to the camera. Then when it wants to use the camera, it asks the OS for permission, the OS prompts the user, and if the user accepts the camera feed becomes available to the app.

With Flatpak (before this PR), an app doesn't declare anything. It can, at will, ask the OS for permission to use the camera. The OS prompts the user, and if the user accepts the camera feed becomes available to the app. Any app can ask for camera permission. We have no way to know if an app will do so.

This PR introduces the "entitlement" mechanism to allow apps to declare that they will talk to certain portals. The app declares "I may prompt for camera permission". If the app doesn't have the entitlement, it can try to ask the OS for permission to use the camera and the OS will immediately say "no, you are not entitled to ask me for camera permission, go away". This way, all apps that wish to access the camera must declare that they might ask for permission to access the camera

That's the primary use-case for entitlements. That's why I'm saying that prompting for an entitlement would be an app asking for permission to ask for permission to take an action. "Can I ask for camera permission? Y/N" (Y => app gets the entitlement), followed by "Can I access the camera? Y/N" (Y => app gets access to the camera)

There's a secondary use-case for entitlements. That's as a mechanism to grant additional special treatment to the app by the portal, at the discretion of the repository (i.e. Flathub's reviewers). For instance, let's use the credentials portal as an example. An app without any entitlement can't ask for any credential. We might have an entitlement that lets you ask for your own credentials (after a prompt from the system!). Then we might have another, stronger, entitlement that lets an app ask for any credential (since it's a web browser!).

Another example of the secondary use-case: let's say you have a "wallpaper changer" app that gives you a new background every night. That's all it does. Problem is, each time the app tries to change the wallpaper in the background, the portal prompts the user to confirm the change. Kinda defeats the purpose of the app. So, instead let's say we introduce an additional entitlement for this portal, that allows an app to change the wallpaper _without_ prompting the user? Now the app can ask the portal, the portal sees that "the app is entitled to do this without me asking for confirmation", and so the portal does it without asking for confirmation

src/xdp-app-info-flatpak.c
swick: Alright, if we go down that route, I don't think enforce=strict is a good idea and we should instead have...

* version 0 (the default if everything is absent) which implicitly grants you full access to the currently existing portals
* version 1 which is the same as strict except that this is only for the currently existing portals
* version 2 which will include entitlements for new portals or more specific parts of existing portals
* and so on

swick:
Did another pass over the entire thing, adding docs as well. This is now in a state I consider mergeable.

shared/xdp-entitlements.c
```diff
@@ -0,0 +1,175 @@
...
+{
+  unsigned int entitlement_version;
+
+  g_return_val_if_fail (XDP_IS_ENTITLEMENTS (self), FALSE);
+  g_return_val_if_fail (entitlement > XDP_ENTITLEMENT_NONE, FALSE);
+
+  entitlement_version = xdp_entitlement_get_version (entitlement);
+
+  if (self->version < entitlement_version)
+    return TRUE;
```
AdrianVovk, line 200: I'd add an extra field to the entitlement here, that specifies what happens when the version is too old. For some entitlements it makes sense to fail open (i.e. the backwards-compatible thing would be to give the app access to the functionality). For other entitlements we must fail closed (i.e. entirely new functionality shouldn't be granted to apps that predate its existence)

shared/xdp-app-info-flatpak.c
```diff
@@ -11,6 +11,7 @@
```
AdrianVovk, line 11: The commit message is somewhat misleading here, though technically correct:

> For the host case, we grant all entitlements, for snaps and linyaps we grant none.

Like sure, you don't grant any entitlements here. But you _do_ set the version to 0, which means that linyapps and snaps do implicitly receive (some, given my other comment) newer entitlements for backwards compat. it sounds in the commit message like snaps and linyapps get no entitlements at all

shared/xdp-app-info-host.c
```diff
@@ -214,14 +215,19 @@ xdp_app_info_host_new_registered (const char  *sender,
                                   GError     **error)
 {
   g_autoptr(XdpAppInfoHost) app_info_host = NULL;
+  g_autoptr(XdpEntitlements) entitlements = NULL;
   g_autofd int pidfd_owned = pidfd;
 
   maybe_get_pidfd_from_pid (pid, &pidfd_owned);
 
+  entitlements = xdp_entitlements_new (0);
```
AdrianVovk, line 189: host apps can't be v0, they should be the latest available version of entitlements but with all entitlements granted

shared/xdp-entitlements.c
AdrianVovk: Actually it has to be even a little more complicated than that because some entitlements change behavior depending on how old the app is. Here's an example:

Let's say in version 5 we introduce an Accessibility portal with lots of dangerous powers. Since it's a new feature, we set it to fail closed. Then we discover that we didn't slice the functionality up fine enough, and so in version 6 we introduce additional finer-grained entitlements (i.e. Accessibility.ScreenShare) that fail open for backwards compatibility with version 5. Let's say we have an app that's version 4. If it asks for the Accessibility entitlement it'll get denied. But if it asks the Accessibility.ScreenShare entitlement it'll get accepted! Uh oh!

So actually we need an additional thing, other than the fail open/closed boolean. We need a list of prerequisite entitlements that must be granted for a given entitlement to also be granted. So the Accessibility.ScreenShare entitlement would have a prereq of Accessibility, and Accessibility.ScreenShare can only be granted of Accessibility is also granted.

We can check by simply recursively calling `xdp_entitlements_is_granted` for the prereqs. We want to apply the same version checking and fail open/closed logic as the main entitlement we're actually trying to check.

shared/xdp-entitlements.c
```diff
@@ -31,9 +32,14 @@ typedef struct _XdpEntitlementInfo
 
 static const XdpEntitlementInfo xdp_entitlements[] =
 {
-  [XDP_ENTITLEMENT_NONE] =
```
AdrianVovk: erm, I think you didn't mean to remove XDG_ENTITLEMENT_NONE here right? We should keep it around

shared/xdp-types.h
```diff
@@ -15,6 +15,8 @@ typedef enum {
 typedef enum _XdpEntitlement
 {
   XDP_ENTITLEMENT_NONE = 0,
+  XDP_ENTITLEMENT_USB,
+  XDP_ENTITLEMENT_USB_DEVICES,
```
AdrianVovk, line 45: Why's Usb.Devices a separate entitlement from Usb? I don't think it makes sense to ever have Usb without Usb.Devices, right? You'd be allowed to talk to the portal but do literally nothing with it? So let's just put the list of devices into the Usb entitlement?

desktop-portal/account.c
```diff
@@ -292,6 +292,6 @@ init_account (XdpContext *context)
```
AdrianVovk, line 292: Commit message here still talks about "strict mode", leftover from previous versions of this PR

doc/entitlements.rst
```diff
@@ -0,0 +1,393 @@
...
+* If the app's entitlement version is **equal to or higher** than an
+  entitlement's minimum version, the entitlement must be **explicitly granted**
+  in the app's metadata.
+
+For example, most current entitlements have minimum version 1. An app with
+entitlement version 0 (or no entitlement metadata at all) is implicitly granted
+all of them. An app that opts in to version 1 must explicitly list every
+entitlement it needs.
+
+Entitlement Reference
```
AdrianVovk: I think we should instead put the docs about what entitlements are required into each portal's docs, rather than having one big document of all the entitlements.

So like, if I go look at the Account portal docs page, there should be a section describing what entitlements are there and what I need

AdrianVovk (changes requested):

shared/xdp-entitlements.c
swick: I used `XDP_ENTITLEMENT_NONE` only because it would otherwise produce a lot of compiler unhappiness. I don't see why we would want `XDP_ENTITLEMENT_NONE` here?

shared/xdp-types.h
swick: If we ever add an entitlement for something that gives device access via some other way, having it on the main/portal-level entitlement won't be fitting anymore. It probably makes sense to use separate entitlements for everything.

shared/xdp-entitlements.c
swick: Maybe I'm missing something but this should all work nicely in the current state (one version on the entitlement, one version on the app, failing open):

* New functionality which is guarded with an entitlement can be added as version 0, which means we always enforce it (it gets denied without the app having the entitlement), and it is backwards compatible because it's a new and no one could have depended on it
* Entitlements guarding existing functionality get a version of "current max version + 1", which means they only get enforced when the app updates its version (failing open), but if the app updated the version, they now require that additional entitlement

I think the only thing you can't model with that is "removing" an entitlement in a later version (so going from fine grained to more coarse permissions). But that can be modeled in code with `has_entitlement(coarse) || has_entitlement(fine)`.

shared/xdp-app-info-host.c
swick: I don't think there is a difference between the two? We fail open so it grants all the entitlements in later versions, but because we `grant_all`, it also grants all entitlements in version 0.

shared/xdp-entitlements.c
AdrianVovk: ah ok, I think you're right. That sounds reasonable to me then. I wasn't thinking of adding new entitlements under v0

shared/xdp-app-info-host.c
AdrianVovk: Right, I was thinking about this the wrong way around. Sounds good to me :+1:

shared/xdp-entitlements.c
AdrianVovk: I gave it a second look and you're right, nothing actually blows up if XDP_ENTITLEMENT_NONE isn't in this array.

shared/xdp-types.h
AdrianVovk: OK. I don't feel too strongly about this. It's an entitlement that's implicitly granted so it probably does no harm.

swick:
Includes https://github.com/flatpak/xdg-desktop-portal/pull/2044 now

swick:
I intend to merge this in a few days. If there are still concerns, now is the time to raise them.

doc/for-app-developers.rst
```diff
@@ -35,6 +36,15 @@ loosen their sandbox restrictions.
 
    :doc:`Reasons to Use Portals </reasons-to-use-portals>`
 
+*  .. image:: _static/img/tiles/APIs-l.png
```
AdrianVovk: We probably need to ask for new art, right?

doc/entitlements.rst
```diff
@@ -0,0 +1,124 @@
...
+* If the app's entitlement version is **equal to or higher** than an
+  entitlement's minimum version, the entitlement must be **explicitly granted**
+  in the app's metadata.
+
+For example, most current entitlements have minimum version 1. An app with
+entitlement version 0 (or no entitlement metadata at all) is implicitly granted
+all of them. An app that opts in to version 1 must explicitly list every
+entitlement it needs.
+
+When adding a new entitlement, choose the minimum version as follows:
```
AdrianVovk: I think this is useful documentation, but "when adding a new entitlement" doesn't belong in `for-app-developers`. Either we should rephrase it to something along the lines of "new entitlements will be added with the following versions" or something like that? Or move it out into the for-backend-developers section?

doc/for-app-developers.rst
swick: Either that or we remove it; it's quite awkward to have such a dependency when adding docs which really should be as low friction as possible.

shared/xdp-types.h
swick: I removed the tiles. The sidebar navigation is now the only way to navigate.

doc/entitlements.rst
swick: We also really need some guidance for versioning portal interfaces in general. Dropped the paragraph. It's on my list for further improvements.

doc/for-app-developers.rst
swick: Removed the tiles and now the sidebar is the only navigation.

data/org.freedesktop.portal.Usb.xml
```diff
@@ -17,7 +17,96 @@
...
+      for the user to allow or deny the app from accessing specific devices.
+
+      If permission is granted, xdg-desktop-portal tries to open the device
+      file on the behalf of the requesting app, and pass the file
+      descriptor down. [#usb-logind]_
+
+      The caller must then call ``FinishAcquireDevices()`` until it indicates it
+      is finished. It is only necessary to call it more than once if there
+      are too many file descriptors to return. This is a D-Bus
+      limitation. Check the ``finished`` return argument.
```
Sodivad, line 75: I was confused by the it in the first sentence, thinking at first read it refers to "the caller". Maybe rewriting helps a bit.

>    The caller must then call FinishAcquireDevices() to retrieve the file descriptor(s).
    Due to a D-Bus it maybe be necessary to call this more than once to retrieve all file descriptors. The finished return argument indicates if additional calls are needed.

data/org.freedesktop.background.Monitor.xml
```diff
@@ -9,8 +9,13 @@
...
 
-    This interface provides APIs related to applications
-    that are running in the background.
+    This interface is provided by the XDG Desktop Portal frontend to the host
+    system, if the portal backend provides an implementation of the Background
+    portal.
+
+    This API provides information related to running applications without an
+    active window, which can be used by host system services to provide rich
+    interfaces to manage those background applications.
```
Sodivad, line 18: Should this have went into another commit?

shared/xdp-app-info-flatpak.c
```diff
@@ -30,6 +31,9 @@
 #define FLATPAK_METADATA_KEY_SHARED "shared"
 #define FLATPAK_METADATA_CONTEXT_SHARED_NETWORK "network"
 #define FLATPAK_METADATA_GROUP_RUNTIME "Runtime"
+#define FLATPAK_METADATA_GROUP_ENTITLEMENT "Policy entitlement"
+#define FLATPAK_METADATA_KEY_GRANT "grant"
+#define FLATPAK_METADATA_KEY_VERSION "version"
```
Sodivad, line 36: Could you put version in the commit message example so one sees where it should go while looking at the code?

data/org.freedesktop.background.Monitor.xml
swick: That information was somewhere else before, and I remove it in this commit, so I put it back here. Split it out in its own commit. Hopefully that is less confusing.

btw, this is all #2044

data/org.freedesktop.portal.Usb.xml
swick: I'm trying to just move things around and making the smallest changes possible to keep everything consistent. There is another comment from Adrian about the USB docs. Let's continue improving unrelated docs in follow up PRs please.

swick:
Note to self: ensure that we parse the entitlement data only if the entitlement is granted.

AdrianVovk:
> ensure that we parse the entitlement data only if the entitlement is granted

I liked the behavior that `org.freedesktop.portal.Usb.Devices` is auto-granted. But it should be auto-granted only if `org.freedesktop.portal.Usb` is granted, of course (and we shouldn't parse `org.freedesktop.portal.Usb.Devices` unless `org.freedesktop.portal.Usb` is granted too)

Since the separation of `org.freedesktop.portal.Usb.Devices` is kinda arbitrary (it's there on the _off chance_ we'll come up with some new way to allowlist/denylist USB devices, though I somehow doubt that'll ever happen) so it should be as convenient as possible to use.

swick:
Another thing to consider: how should entitlement children behave, e.g.: `org.freedesktop.portal.Usb` and `org.freedesktop.portal.Usb.Devices`. What if only `org.freedesktop.portal.Usb.Devices` was granted? Currently the behavior of the relation comes from how we implement the entitlements (you can get .Usb.Devices granted but it will be useless without .Usb). We could be more systematic here and not allow `.Usb.Devices` without `.Usb`.

swick:
> I liked the behavior that org.freedesktop.portal.Usb.Devices is auto-granted

I guess you mean if the ini section `[Policy org.freedesktop.portal.Usb.Devices]` exists? It would also be possible to get rid of the `granted` key and use the (potentially empty) sections for everything.

AdrianVovk:
> It would also be possible to get rid of the granted key and use the (potentially empty) sections for everything

Not a bad idea. Maps nicely to json for appd, as well

```
"entitlements": {
   "version": 2,
   "foo.bar.baz": {},
   "something.with.args": {
     "arg1": "value1"
   }
}
```

swick:
Except that --add-policy doesn't really allow us to add empty sections; there always needs to be a key and a value.

PR #1931, opened 2026-03-12
doc: terminology: Specify that URIs follow RFC3986

A6GibKm:
The OpenURI and DynamicLauncher portals call g_uri_is_valid, as per GUris docs:

> Note that the scope of GUri is to help manipulate URIs in various
> applications, following RFC 3986. In particular, it doesn’t intend to
> cover web browser needs, and doesn’t implement the WHATWG URL
> standard.

One thing to note, however, is that other portals like the wallpaper portal also take URIs as input but there is no validation on xdp's side and saying that all URIs follow this spec might over over-restrictive. An alternatively would be to reference the RFC when actually is used in the two examples above.


swick:
Thanks!

PR #1933, opened 2026-03-16
doc: Add docs for reviewing xdp code

swick:
The intent is that those files should be useful for humans and LLM agents at the same time.

This is an experiment, and keeping those files up-to-date will require some work and the value of them is not entirely clear. Still worth a try in my opinion.

### DIFF
--- a/doc/reviewing/API_DESIGN.md
+++ b/doc/reviewing/API_DESIGN.md
@@ -1,0 +1,538 @@
+# D-Bus API Design and Compatibility
+
+XDG Desktop Portal exposes two distinct D-Bus APIs with different compatibility
+requirements. Understanding which API you're changing is critical for
+determining compatibility constraints.
+
+## The Two APIs
+
+### Frontend API (org.freedesktop.portal.*)
+
+**What it is:**
+- D-Bus interface between **applications** and **xdg-desktop-portal**
+- Exposed at bus name `org.freedesktop.portal.Desktop`
+- Used by thousands of sandboxed applications (Flatpak, Snap, etc.)
+
+**Who uses it:**
+- Applications installed on user systems
+- Applications that won't all update simultaneously
+- Third-party apps outside system control
+
+**Compatibility requirement: MUST be backwards compatible**
+- Applications depend on these interfaces working forever
+- Must not break existing apps with portal updates
+- Must support old and new clients simultaneously
+
+**Example interfaces:**
+- `org.freedesktop.portal.FileChooser`
+- `org.freedesktop.portal.Screenshot`
+- `org.freedesktop.portal.Clipboard`
+
+### Backend/Impl API (org.freedesktop.impl.portal.*)
+
+**What it is:**
+- D-Bus interface between **xdg-desktop-portal** and **backend implementations**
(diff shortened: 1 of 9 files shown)


doc/reviewing/API_DESIGN.md
```diff
@@ -0,0 +1,538 @@
...
+- `xdp_get_portal_call_fd()`
+- `XdpRequest`, `XdpSession` structures
+
+## Frontend API Compatibility Rules
+
+These rules apply to `org.freedesktop.portal.*` interfaces.
+
+### Absolute Rules (NEVER Break)
+
+#### 1. Cannot Remove or Rename Existing Methods
```
whot: I'd go with "must not" in all these which is stronger than "cannot" (which has the ambiguity of "technically we can!")

doc/reviewing/API_DESIGN.md
```diff
@@ -0,0 +1,538 @@
...
+### ✅ Using Options for New Parameters
+
+```xml
+<!-- CORRECT - old clients work, new clients can use new feature -->
+<method name="OpenFile">
+  <arg type="s" name="parent_window" direction="in"/>
+  <arg type="a{sv}" name="options" direction="in"/>
+  <!-- new_parameter goes in options with default -->
+</method>
+```
```
whot, line 480: bit odd that checkboxes of things to do are in "common mistakes"

doc/reviewing/SECURITY.md
```diff
@@ -0,0 +1,397 @@
...
+│  - Limited filesystem access            │                   │
+│  - No direct hardware access            │                   │
+│  - Restricted system calls              │                   │
+│  - Must use portals for privileged ops  │                   │
+│                                         │                   │
+│  ┌──────────────────────────────────┐   │                   │
+│  │  Application uses D-Bus API:     │   │                   │
+│  │  org.freedesktop.portal.Desktop  │   │                   │
+│  └──────────────────────────────────┘   │                   │
+└─────────────────────────────────────────────────────────────┘
```
whot, line 54: dunno if this is just github's rendering but the `|` on the right are not well-aligned

whot (approved):
a few minor nitpicks but i think this is a pretty good start, thanks

tytan652:
~~Feel free to mark it as off-topic, but I would have preferred a ban on AI/LLM tools (alike to some GNOME projects).~~

tytan652:
I think before merging this, the AI policy of the project needs to be discussed (and not implicitly decided by one committer with an `AGENTS.md`).

https://github.com/flatpak/xdg-desktop-portal/pull/1963#issuecomment-4236579722:
> If anyone want to add a LLM policy, it needs consensus from committers.

I think this PR falls under the same consensus requirement.

swick:
I don't see why documentation on what we consider good practices for reviewing code requires a policy on LLM generated content, and neither do I see why an AGENTS.md file does. All of those things can be useful no matter if you use LLMs to generate content or not.

tytan652:
All the actual contributor documentation is in RST alongside of the rest of doc (which turns into the documentation website, even the CONTRIBUTING.md redirects to it).

With that in mind, it can make the choice of markdown look like it is mainly related to be used alongside the AGENTS.md.

PR #1938, opened 2026-03-18
libdex preparations part 2

swick:
Those changes are self-contained and tested by re-implementing some portals.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -51,7 +51,7 @@ jobs:
             -Dwerror=true
 
       - name: Run xdg-desktop-portal tests
-        run: XDP_TEST_RUN_LONG=1 timeout --signal=KILL -v ${TESTS_TIMEOUT}m meson test -C _build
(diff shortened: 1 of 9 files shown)


src/meson.build
```diff
@@ -6,34 +6,39 @@ portal_built_sources = gnome.gdbus_codegen(
   interface_prefix: 'org.freedesktop.portal',
   namespace: 'XdpDbus',
   docbook: 'portal',
+  extra_args: [f'--extension-path=@libdex_codegen_ext@'],
```
AdrianVovk: Nit: this is repeated a bunch, should it be factored out I wonder? Or maybe not worth it?

AdrianVovk (approved):
Seems straightforward. Didn't think too hard about the CI/test changes, we'll notice if they end up causing failures. LGTM

swick:
Thanks, I will hold off merging this until we actually want to merge something that uses libdex so we buy a bit more time until the dependencies' versions get bumped.

PR #1940, opened 2026-03-19
print: add auto-scale api

lbaudin:
Add an auto-scale api so as apps can scale to the printable area. This is commonly used when printing, either for images, or for documents that have different paper size (such as printing slides to A4 paper).

See [papers#564](https://gitlab.gnome.org/GNOME/papers/-/issues/564) for instance.

Related to https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/256

### DIFF
--- a/data/org.freedesktop.portal.Print.xml
+++ b/data/org.freedesktop.portal.Print.xml
@@ -18,7 +18,7 @@
       to print the formatted document. It is expected that high-level toolkit
       APIs such as GtkPrintOperation will hide most of this complexity.
 
-      This documentation describes version 4 of this interface.
+      This documentation describes version 5 of this interface.
   -->
   <interface name="org.freedesktop.portal.Print">
     <!--
@@ -69,6 +69,13 @@
           Whether it makes sense to return "selection" for the ``print-pages`` setting.
 
           This option was added in version 4.
+
+        * ``has_automatic_scale`` (``b``)
+
+          Whether the app can deal with automatic scale settings, i.e. ``fit`` or ``shrink``
+          to printable area, as specified in the auto-scale parameter.
+
+          This option was added in version 5.
 
         The @settings and @page_setup vardict contain hints for the initial state of the print dialog.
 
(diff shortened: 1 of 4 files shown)


lbaudin:
ping? @Sodivad @matthiasclasen

jadahl:
Cc: @tillkamppeter

lbaudin:
Updated with latest repository structure, still up for review :)

data/org.freedesktop.portal.Print.xml
```diff
@@ -135,6 +142,10 @@
 
           The scale in percent.
 
+        * ``auto-scale`` (``s``)
```
tillkamppeter, line 145: Best is to allow here the settings possible for the IPP attribute `print-scaling`: `auto`, `auto-fit`, `fill`, `fit`, `none`. Or what is actually reported back from the printer (can be sub-set).

tillkamppeter:
I do not know exactly what is needed to add a parameter to a porta API, but it makes all sense to add this parameter so that the print dialog displays the option on how to scale a document for being printed into the output page frame, and displays the option only if the print job can actually get scaled, cropped, and padded.

lbaudin:
Thanks @tillkamppeter. Do you think that scaling should only be done by the printer then? Or should it be done by the app in case the printer does not support it? (I guess the answer is that it should be done by the app if the printer does not support it, but I have no idea whether it happens in practice.)

karlrelton:
> Thanks @tillkamppeter. Do you think that scaling should only be done by the printer then? Or should it be done by the app in case the printer does not support it? (I guess the answer is that it should be done by the app if the printer does not support it, but I have no idea whether it happens in practice.)

I would argue that scaling should be performed by the App, because:
- the App may not be sending direct to the physical printer, but _could_ instead be sending to a layer such a Cups which would apply its own processing before sending the job to the physical printer (meaning you get the Cups implementation rather than the physical device implementation)
- it gives the App control over the eventual layout, enabling it to do as seems best for the App developers (e.g. an image viewer may have different needs to a document viewer)
- when number-up is also in play, it becomes even more useful for the App to determine what to do with the scaling/rotating/tiling ... rather than relying on the printer implementation

PR #1941, opened 2026-03-20
Make graphical-session a weak dependency of the portal service.

nteodosio:
Not all desktops will reach graphical-session.target; That however should not keep the portal service from running.

C.f. https://launchpad.net/bugs/2144855, where the unit will fail to start with result 'dependency'.

This affects Lxqt and stand-alone window managers and is observed in Ubuntu 26.04.

### DIFF
--- a/src/xdg-desktop-portal.service.in
+++ b/src/xdg-desktop-portal.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=Portal service
 PartOf=graphical-session.target
-Requisite=graphical-session.target
+Wants=graphical-session.target
 After=graphical-session.target
 
 [Service]


aleasto:
A random service starting `graphical-session.target` without the guarantee of an actual desktop running breaks all assumptions.

I don't see an immediate solution on how to properly support desktop sessions that do not start `graphical-session.target` without reverting [4d284de](https://github.com/flatpak/xdg-desktop-portal/commit/4d284de29d1d0740c9b80b634631a8f253287680), but this one cannot be it.

swick:
@AdrianVovk opinions?

AdrianVovk:
For this PR specifically, I totally agree with Alessandro. This service is part of a graphical session, it does not in and of itself constitute a graphical session.

In fact this PR regresses the issue that caused us to add this dependency in the first place. The dependency was put into place because `graphical-session.target` is only reached _after_ the Wayland compositor and other core session services are running. Thus, xdg-desktop-portal can start and confidently know that $WAYLAND_DISPLAY and/or $DISPLAY are set in the service activation environment. However, by turning this into a `Wants=`, we've suddenly lost the guarantee that any display server is even running, let alone done uploading environment variables into the activation environment.

More broadly: I do not think it is a supported configuration for any graphical session to be running without `graphical-session.target` also running. `xdg-desktop-portal` is not the only service that will tie itself to `graphical-session.target`. systemd also doesn't permit running more than one graphical session at a time, so a well-behaved session should also check to see if `graphical-session.target` is already running and refuse to start if so. For things to work right the display server also needs to make sure the environment variables end up in systemd's activation environment

> But I do not think it would be healthy to require every desktop under the sun, in particular small window managers, to provide a service in order to use the portal.

If you're assembling a desktop environment for yourself from parts, you take on the responsibility to pick up the pieces.

People running their own window managers have been building scripts to do things like uploading variables into the dbus activation environment, and then later the systemd activation environment, since forever. This is more of the same.

Erick555:
Ubuntu has [patched](https://salsa.debian.org/debian/xdg-desktop-portal/-/blob/ubuntu/latest/debian/patches/ubuntu/allow-no-graphical-session-target.patch?ref_type=heads) it themselves

Issue #979, opened 2023-03-02
Investigate using PipeWire serial id, instead of node id


GeorgesStavracas:
The ScreenCast portal uses PipeWire node ids, but apparently that's being deprecated in favour of serial ids. We should eventually update that.

columbarius:
In case this requires to change the signature of the api, we should drop the buffer size and position from the reply of `org.freedesktop.portal.ScreenCast.Start` and maybe use PipeWire properties on the stream instead, since this static information is insufficient since the mode and position of the output can change without interruption of the stream.

PR #1942, opened 2026-03-22
screen-cast: add pipewire-serial stream property

glamberson:
Add optional `pipewire-serial` (uint64) to the stream properties dict
returned by `ScreenCast.Start()` and `RemoteDesktop.Start()`. This
exposes the PipeWire `object.serial`, which unlike node_id is never
reused after node destruction.

## Motivation

PipeWire node_ids are reusable. When a ScreenCast stream is destroyed
and recreated (monitor hotplug, resolution change, dock/undock),
consumers holding the old node_id can silently connect to the wrong
PipeWire node if the ID has been reassigned. PipeWire deprecated
`node.target` (node_id based) in 0.3.64 in favor of `target.object`
(serial based) for this reason.

The portal currently returns only node_id, encouraging all consumers to
use the deprecated `pw_stream_connect(target_id=node_id)` pattern.

## Changes

- **Spec** (`ScreenCast.xml`): document `pipewire-serial` as an optional
  `t` (uint64) in the stream properties dict, version 6
- **Frontend** (`screen-cast.c`): parse `pipewire-serial` from backend
  responses, store in `ScreenCastStream`, add getter function
- **Header** (`screen-cast.h`): declare
  `screen_cast_stream_get_pipewire_serial()`

The existing node_id `u` in the tuple is preserved. Old consumers ignore
the new dict key. New consumers prefer serial when present.

## Backend availability

The wlr backend can provide serial immediately (direct PipeWire access).
GNOME and KDE backends need compositor changes to pass serial through
their intermediate protocols and can add support independently.

Ref: #979

### DIFF
--- a/data/org.freedesktop.portal.ScreenCast.xml
+++ b/data/org.freedesktop.portal.ScreenCast.xml
@@ -25,7 +25,7 @@
 
       The Screen cast portal allows to create screen cast sessions.
 
-      This documentation describes version 5 of this interface.
+      This documentation describes version 6 of this interface.
   -->
   <interface name="org.freedesktop.portal.ScreenCast">
     <!--
@@ -185,6 +185,16 @@
           node ID (the first element in the tuple, and a Vardict of
           properties.
 
+          Since version 6 of this interface, the node ID in the stream
+          tuple is deprecated for stream targeting. PipeWire node IDs can
+          be reused after node destruction, so active sessions may
+          misidentify streams across monitor hotplug, resolution or
+          refresh-rate changes, or suspend/resume cycles. Consumers should
+          prefer the ``pipewire-serial`` property from the stream
+          properties dict and ``PW_KEY_TARGET_OBJECT`` for connecting to
+          streams. Backends must continue to populate the node ID for
+          compatibility with clients that do not support version 6.
+
           The array will contain a single stream if 'multiple' (see
           org.freedesktop.portal.ScreenCast.SelectSources())
           was set to 'false', or at least one stream if
@@ -233,6 +243,16 @@
             device.
 
             This property was added in version 5 of this interface.
+
+          * ``pipewire-serial`` (``t``)
+
+            The PipeWire ``object.serial`` for this stream's node. Unlike the
+            node ID in the stream tuple (which can be reused after node
+            destruction), the serial is a monotonically increasing 64-bit
+            identifier that is never reused. Consumers should prefer the serial
+            for stream targeting via ``PW_KEY_TARGET_OBJECT`` when available.
+
+            This property was added in version 6 of this interface.
 
         * ``restore_token`` (``s``)
 
--- a/src/screen-cast.c
+++ b/src/screen-cast.c
@@ -70,6 +70,8 @@ struct _ScreenCastStream
   uint32_t id;
   int32_t width;
   int32_t height;
+  uint64_t pipewire_serial;
+  gboolean has_pipewire_serial;
 };
 
 G_DEFINE_TYPE_WITH_CODE (ScreenCast, screen_cast,
@@ -775,6 +777,9 @@ collect_screen_cast_stream_data (GVariantIter *streams_iter)
       stream->id = stream_id;
       g_variant_lookup (stream_options, "size", "(ii)",
                         &stream->width, &stream->height);
+      stream->has_pipewire_serial =
+        g_variant_lookup (stream_options, "pipewire-serial", "t",
+                          &stream->pipewire_serial);
 
       streams = g_list_prepend (streams, stream);
     }
@@ -1126,7 +1131,8 @@ screen_cast_new (XdpContext            *context,
 
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (screen_cast->impl), G_MAXINT);
 
-  xdp_dbus_screen_cast_set_version (XDP_DBUS_SCREEN_CAST (screen_cast), 5);
+  xdp_dbus_screen_cast_set_version (XDP_DBUS_SCREEN_CAST (screen_cast),
+                                    MIN (xdp_dbus_impl_screen_cast_get_version (screen_cast->impl), 6));
 
   g_object_bind_property (G_OBJECT (screen_cast->impl), "available-source-types",
                           G_OBJECT (screen_cast), "available-source-types",
--- a/data/org.freedesktop.impl.portal.ScreenCast.xml
+++ b/data/org.freedesktop.impl.portal.ScreenCast.xml
@@ -25,7 +25,7 @@
 
       The Screen cast portal allows to create screen cast sessions.
 
-      This documentation describes version 5 of this interface.
+      This documentation describes version 6 of this interface.
   -->
   <interface name="org.freedesktop.impl.portal.ScreenCast">
     <!--
@@ -148,6 +148,12 @@
           node ID (the first element in the tuple, and a Vardict of
           properties.
 
+          Since version 6 of this interface, the node ID in the stream
+          tuple is deprecated for stream targeting. PipeWire node IDs can
+          be reused after node destruction; consumers should prefer the
+          ``pipewire-serial`` property from the stream properties dict
+          and ``PW_KEY_TARGET_OBJECT`` for connecting to streams.
+
           The array will contain a single stream if 'multiple' (see
           SelectSources) was set to 'false', or at least one stream if
           'multiple' was set to 'true' as part of the SelectSources method.
@@ -188,6 +194,16 @@
             device.
 
             This property was added in version 5 of this interface.
+
+          * ``pipewire-serial`` (``t``)
+
+            The PipeWire ``object.serial`` for this stream's node. Unlike the
+            node ID in the stream tuple (which can be reused after node
+            destruction), the serial is a monotonically increasing 64-bit
+            identifier that is never reused. Consumers should prefer the serial
+            for stream targeting via ``PW_KEY_TARGET_OBJECT`` when available.
+
+            This property was added in version 6 of this interface.
 
         * ``persist_mode`` (``u``)
 


data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -234,6 +234,16 @@
...
 
+          * ``pipewire-serial`` (``t``)
+
+            The PipeWire ``object.serial`` for this stream's node. Unlike the
+            node ID in the stream tuple (which can be reused after node
+            destruction), the serial is a monotonically increasing 64-bit
+            identifier that is never reused. Consumers should prefer the serial
+            for stream targeting via ``PW_KEY_TARGET_OBJECT`` when available.
+
+            This property was added in version 6 of this interface. Optional.
```
Sodivad: You make this optional so no backend changes are required. Because of the id reuse problem I would prefer the following:
- also version bump the backend, if the backend implements the new version we can also expose the new version  and the app can then rely on `pipewire-serial` being present
-  deprecate the node_id also our side

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -234,6 +234,16 @@
 
             This property was added in version 5 of this interface.
 
+          * ``pipewire-serial`` (``t``)
```
Sodivad, line 247: Also needs documentation in the backend API

data/org.freedesktop.portal.ScreenCast.xml
glamberson: Agreed on both points.

For the backend version bump: looking at how input-capture.c, notification.c, and screenshot.c handle this, the frontend version is derived from the backend version (e.g. `MIN(impl_version, N)`). screen-cast.c currently hardcodes it. Should I switch screen-cast.c to the backend-derived pattern so the frontend only advertises v6 when the backend actually reports v6? That way apps get a hard guarantee rather than a maybe-present optional field.

For node_id deprecation: the stream tuple type `(u, a{sv})` can't change without breaking ABI, so this would be documentation-level deprecation on the `u` field, directing consumers to use `pipewire-serial` from the dict via `PW_KEY_TARGET_OBJECT`. Something along the lines of what Notification does ("Since version N, this is deprecated..."). Does that match what you had in mind?

Sodivad: > ould I switch screen-cast.c to the backend-derived pattern so the frontend only advertises v6 when the backend actually reports v6? That way apps get a hard guarantee rather than a maybe-present optional field.

Yes!

> ("Since version N, this is deprecated..."). Does that match what you had in mind?

Exactly !

jadahl: Deprecation will only ever be "documentation-level" since it must still continue to work for backward compatibility. As in, applications should prefer to use the serial, but clients that doesn't update should remain functional.

data/org.freedesktop.portal.ScreenCast.xml
glamberson: Will do (was already planning to follow the mapping_id pattern from ea06555). I'll add `pipewire-serial` to the `org.freedesktop.impl.portal.ScreenCast` Start() stream properties and bump the backend version to 6.

data/org.freedesktop.portal.ScreenCast.xml
glamberson: Understood, backward compatibility preserved. Just documentation guiding new consumers toward serial.

data/org.freedesktop.portal.ScreenCast.xml
glamberson: This is been resolved in my opinion. Thanks.

data/org.freedesktop.portal.ScreenCast.xml
glamberson: I've resolved this I believe. Please reviw for completeness. Thanks.

src/screen-cast.c
```diff
@@ -675,6 +677,19 @@ screen_cast_stream_get_pipewire_node_id (ScreenCastStream *stream)
...
+{
+  if (stream->has_pipewire_serial)
+    {
+      if (out_serial)
+        *out_serial = stream->pipewire_serial;
+      return TRUE;
+    }
+  return FALSE;
+}
+
```
tytan652: Unused function,, so removal is to be considered

tytan652 (changes requested):

src/screen-cast.c
glamberson: Sounds good. The backend response passes through to the client as-is, so the frontend doesn't need to read the serial back. 

I originally added it to mirror the node_id getter pattern in case the frontend later needs the serial for internal operations like permission setup, but nothing uses it today. 

Removed per your suggestion.

glamberson:
> Looks fine to me, maybe a test could be added to verify the property appears properly (heh) if the backend implements v6

  As there aren't any screencast tests in the repo today, adding a test for this would mean building the full mock backend template and test infrastructure from scratch. 

I'd be happy to do that as a follow-up PR rather than expanding the scope of this one.

src/screen-cast.c
```diff
@@ -1126,7 +1131,8 @@ screen_cast_new (XdpContext            *context,
 
   g_dbus_proxy_set_default_timeout (G_DBUS_PROXY (screen_cast->impl), G_MAXINT);
 
-  xdp_dbus_screen_cast_set_version (XDP_DBUS_SCREEN_CAST (screen_cast), 5);
+  xdp_dbus_screen_cast_set_version (XDP_DBUS_SCREEN_CAST (screen_cast),
+                                    MIN (xdp_dbus_impl_screen_cast_get_version (screen_cast->impl), 6));
```
tytan652, line 1135: Portal and portal impl interface versions are not meant to always match (e.g. Settings is 2 where impl is still 1).

And this PR adds an additional property in the optional argument, simply set it to 6.

Edit: And we are not pretty consistent about it.

glamberson: Thanks for the review. This pattern was specifically requested by @Sodivad in the earlier review round to provide a presence guarantee: when the frontend advertises v6, apps can rely on pipewire-serial actually being present in the stream properties rather than having to check for the optional key.  

I understand that dict key additions do not strictly require version bumps and that the portal and backend versions do not always match. The MIN(impl_version, N) pattern is used this way in input-capture.c and notification.c for exactly this purpose, letting the frontend version act as a capability signal derived from backend support. 

I am happy to switch to unconditional v6 if maintainers prefer that, but I would want @Sodivad or @jadahl to confirm, since it would reverse their earlier feedback.

Thanks.

tytan652: No problem, those are things that are not properly defined (yet).

tytan652: I made some digging while working on #1964.

ScreenCast version unlike other portals does not represent the feature-set but the version documented in the interface definition.
So it should be set to 6 to not break the expectation based on the portal own precedence.

InputCapture was still version 1 until 2 arrived, so its version did not have a behavior set by precedence unlike ScreenCast.

Edit: ~~GlobalShortcuts has the same behaviour as ScreenCast since it is set to 2 whatever the impl version (even if I propose to change it in another PR since it's still not in a stable release).~~ And there is a likely hood that another one is like ScreenCast.
Edit 2: RemoteDesktop, Print share the same behaviour as ScreenCast.

data/org.freedesktop.portal.ScreenCast.xml
jadahl, line 259: This needs updating.

data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -185,6 +185,12 @@
           node ID (the first element in the tuple, and a Vardict of
           properties.
 
+          Since version 6 of this interface, the node ID in the stream
+          tuple is deprecated for stream targeting. PipeWire node IDs can
+          be reused after node destruction; consumers should prefer the
+          ``pipewire-serial`` property from the stream properties dict
+          and ``PW_KEY_TARGET_OBJECT`` for connecting to streams.
```
jadahl, line 192: Could make it clear that while its deprecated, it's still required, for backward compat.

jadahl (commented):
Needs squashing, but serials indeed should be preferred over node IDs. I'd like to see at least one backend implementing this before landing.

data/org.freedesktop.portal.ScreenCast.xml
glamberson: Thanks @jadahl, got it. Bumping the interface-level version declaration at the top of the file (currently "This documentation describes version 5 of this interface.") to version 6, and addressing the backward-compat clarification on the node ID paragraph. Both land in the squashed commit I just pushed.

glamberson:
A bit of background on the user-visible motivation (for posterity if for nothing else), since the PR description has been light on the *why*.

The node-id reuse problem shows up whenever the compositor destroys and recreates PipeWire capture nodes during a session:

- Monitor hotplug (unplug one display, plug another in)
- Resolution, refresh-rate, or scale changes
- Suspend/resume cycles
- Cursor-mode or output-rotation transitions (on some compositors)

Under these, a restore-token-backed session or an active remote-desktop stream can silently re-bind to the wrong node, because old node IDs can be reused for unrelated streams. The serial is monotonic and never reused, so clients can either reconnect robustly or detect that their stream is genuinely gone.

KDE is converging on the same fix in its own stack. `kde-screencasting` bumps its Wayland protocol from v5 to v6 adding a `serial` event, with the existing node-id-based `created` event marked `deprecated-since="6"` ([plasma-wayland-protocols!129](https://invent.kde.org/libraries/plasma-wayland-protocols/-/merge_requests/129)). The portal-kde backend that would satisfy the "at least one backend implementing this" gate is tracked in [!539](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/539), wired through [kwin!9024](https://invent.kde.org/plasma/kwin/-/merge_requests/9024) and [kpipewire!244](https://invent.kde.org/plasma/kpipewire/-/merge_requests/244). Same problem framing, same solution, same version number. Happy to coordinate timing with @Sodivad if that would help land the portal spec and the backend together.

I've just force-pushed a squashed commit addressing @jadahl's three notes (interface-level version bump, backward-compat clarification, single commit for the whole change).

Sodivad:
> KDE is converging on the same fix in its own stack. kde-screencasting bumps its Wayland protocol from v5 to v6 adding a serial event, with the existing node-id-based created event marked deprecated-since="6" ([plasma-wayland-protocols!129](https://invent.kde.org/libraries/plasma-wayland-protocols/-/merge_requests/129)). The portal-kde backend that would satisfy the "at least one backend implementing this" gate is tracked in [!539](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/539), wired through [kwin!9024](https://invent.kde.org/plasma/kwin/-/merge_requests/9024) and [kpipewire!244](https://invent.kde.org/plasma/kpipewire/-/merge_requests/244). Same problem framing, same solution, same version number. Happy to coordinate timing with @Sodivad if that would help land the portal spec and the backend together.

Thanks for bringing this problem to my attention btw!

> The node-id reuse problem shows up whenever the compositor destroys and recreates PipeWire capture nodes during a session

Could even just be window open/close I guess.

jadahl (approved):

src/screen-cast.c
```diff
@@ -775,6 +777,9 @@ collect_screen_cast_stream_data (GVariantIter *streams_iter)
       stream->id = stream_id;
       g_variant_lookup (stream_options, "size", "(ii)",
                         &stream->width, &stream->height);
+      stream->has_pipewire_serial =
```
swick, line 780: So, this data is saved here, but then never used. Shouldn't we also switch over to the serial in xdp? And if not (or not yet in this PR) then why is this code here?

swick:
@Sodivad this is a protocol change so I need an ack/approve from KDE

Sodivad:
Acked by KDE

src/screen-cast.c
glamberson: Two parts to this. On why the code is here: The backend already computes and sends the serial in the stream properties dict, so storing it on the frontend stream struct is essentially free, one uint64_t plus a presence flag. It means a follow-up that switches xdp's internal stream lookups from node id to serial does not need a second round trip through the backend or another protocol bump. The data is already in hand at the point we'd want it.

On whether xdp should also switch over: Yes, that's the natural next step, and it belongs in its own PR. The protocol addition stands on its own value, since clients can use pipewire-serial today regardless of what xdp does internally. Switching the frontend's internal lookups touches permission and lifecycle paths that deserve separate review.

src/screen-cast.c
glamberson: Thanks for the precedent dig. You're right that ScreenCast, RemoteDesktop, and Print treat the interface version as the version of the interface as documented rather than a capability signal.

The reason MIN(impl_version, 6) is here anyway is the presence guarantee Sodivad asked for in the first review round. A client seeing frontend v6 can rely on pipewire-serial actually being in the stream properties dict, rather than probing for the optional key and falling back. Setting the frontend to 6 unconditionally would mean a v6 frontend running against a v5 backend silently drops the new property, which is exactly the case the MIN pattern catches.

Given the approval is on the MIN version and the precedent question is about interface-version semantics rather than correctness, I'd prefer to leave it as is for this PR. If the project wants to standardize ScreenCast on unconditional version setting later, that seems like a separate cleanup across all the affected portals rather than a per-PR decision.

src/screen-cast.c
swick: > Yes, that's the natural next step, and it belongs in its own PR

Okay, then let's add the code in that PR. In this PR, the code is just dead.

src/screen-cast.c
swick: Both approaches are backwards compatible. The advantage of using `MIN(impl_version, 6)` is that this way we can *require* the new property to be available in v6, whereas otherwise, v6 would only indicate that it might be available. We *want* everything to switch over to the new property, so making it optional isn't a great idea.

So yeah, I think this is the right decision.

swick:
Will merge after the dead code is removed.

src/screen-cast.c
tytan652: If we keep `MIN(impl_version, 6)` we should safeguard against impl_version = 0.
Like #1966: `impl_version = MAX (impl_get_version , 1)` then `MIN(impl_version, 6)`.

swick: Good point.

src/screen-cast.c
swick: For some reason (I suspect the PR source repo being an org) I can't push changes to the PR, so I'm going to merge this as is, but I do expect follow up patches for this.

grulja:
I spotted a small inconsistence in the naming. I think we should have used `pipewire_serial` (underscore vs dash) following the rest of the properties. I see KDE already adopted it and we already made a release so probably too late to do something about it.

Sodivad:
FWIW the KDE change is not released yet.

PR #1950, opened 2026-03-28
Use name_to_handle_at file handles instead of dev+ino in the document portal

swick:
The dev is specifically not stable across reboots, which can result in granted documents not being accessible anymore.

This adds another field to the document permission table to store the file handle returned by name_to_handle_at with AT_HANDLE_FID. The code transparently handles the old format, updates existing entries with the handle and prefers the handle over the dev+ino for lookup.

I did a bit of testing already, but it definitely could use more.

### DIFF
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -190,6 +190,7 @@ struct _XdpDomain {
   guint64 doc_dir_device;
   guint64 doc_dir_inode;
   guint32 doc_flags;
+  GBytes *doc_dir_handle;
 
   /* Below is mutable, protected by mutex */
   GMutex  tempfile_mutex;
@@ -564,6 +565,7 @@ xdp_domain_unref (XdpDomain *domain)
       g_free (domain->app_id);
       g_free (domain->doc_path);
       g_free (domain->doc_file);
+      g_clear_pointer (&domain->doc_dir_handle, g_bytes_unref);
       if (domain->inodes)
         g_assert (g_hash_table_size (domain->inodes) == 0);
       g_clear_pointer (&domain->inodes, g_hash_table_unref);
@@ -648,6 +650,7 @@ xdp_domain_new_document (XdpDomain         *parent,
   domain->doc_flags = document_entry_get_flags (doc_entry);
   domain->doc_dir_device = document_entry_get_device (doc_entry);
   domain->doc_dir_inode =  document_entry_get_inode (doc_entry);
+  domain->doc_dir_handle = document_entry_dup_handle (doc_entry);
 
   db_path = document_entry_get_path (doc_entry);
   if (xdp_document_domain_is_dir (domain))
@@ -975,9 +978,20 @@ xdp_domain_get_doc_dir (XdpDomain   *doc_domain,
   if (fstat (dirfd, buf) != 0)
     return -errno;
 
-  if (buf->st_ino != doc_domain->doc_dir_inode ||
-      buf->st_dev != doc_domain->doc_dir_device)
-    return -ENOENT;
+  if (doc_domain->doc_dir_handle != NULL)
+    {
+      /* If we have a handle, use it exclusively — st_dev is not stable across reboots */
+      g_autoptr(GBytes) handle = xdp_file_handle_for_fd (dirfd);
+
+      if (handle == NULL || !g_bytes_equal (handle, doc_domain->doc_dir_handle))
+        return -ENOENT;
+    }
+  else
+    {
+      if (buf->st_ino != doc_domain->doc_dir_inode ||
+          buf->st_dev != doc_domain->doc_dir_device)
+        return -ENOENT;
+    }
 
   if (dirfd_out)
     *dirfd_out = g_steal_fd (&dirfd);
--- a/document-portal/document-portal-fuse.c
+++ b/document-portal/document-portal-fuse.c
@@ -960,16 +960,27 @@ xdp_inode_kernel_unref (XdpInode *inode, unsigned long count)
 }
 
 static int
-verify_doc_dir_devino (int dirfd, XdpDomain *doc_domain)
+xdp_domain_get_doc_dir (XdpDomain   *doc_domain,
+                        int         *dirfd_out,
+                        struct stat *buf_out)
 {
-  struct stat buf;
+  g_autofd int dirfd = -1;
+  struct stat local_buf;
+  struct stat *buf = buf_out ? buf_out : &local_buf;
 
-  if (fstat (dirfd, &buf) != 0)
+  dirfd = open (doc_domain->doc_path, O_PATH | O_DIRECTORY);
+  if (dirfd < 0)
     return -errno;
 
-  if (buf.st_ino != doc_domain->doc_dir_inode ||
-      buf.st_dev != doc_domain->doc_dir_device)
+  if (fstat (dirfd, buf) != 0)
+    return -errno;
+
+  if (buf->st_ino != doc_domain->doc_dir_inode ||
+      buf->st_dev != doc_domain->doc_dir_device)
     return -ENOENT;
+
+  if (dirfd_out)
+    *dirfd_out = g_steal_fd (&dirfd);
 
   return 0;
 }
@@ -987,11 +998,7 @@ xdp_nonphysical_document_inode_opendir (XdpInode *inode)
   g_assert (domain->type == XDP_DOMAIN_DOCUMENT);
   g_assert (inode->physical == NULL);
 
-  dirfd = open (domain->doc_path, O_PATH | O_DIRECTORY);
-  if (dirfd < 0)
-    return -errno;
-
-  res = verify_doc_dir_devino (dirfd, domain);
+  res = xdp_domain_get_doc_dir (domain, &dirfd, NULL);
   if (res != 0)
     return res;
 
@@ -2382,12 +2389,8 @@ xdp_fuse_opendir (fuse_req_t             req,
 
               d = xdp_dir_new_buffered (req);
 
-              if (stat (domain->doc_path, &buf) == 0 &&
-                  buf.st_ino == domain->doc_dir_inode &&
-                  buf.st_dev == domain->doc_dir_device)
-                {
-                  xdp_dir_add (d, req, domain->doc_file, buf.st_mode);
-                }
+              if (xdp_domain_get_doc_dir (domain, NULL, &buf) == 0)
+                xdp_dir_add (d, req, domain->doc_file, buf.st_mode);
             }
         }
       else
@@ -3842,11 +3845,14 @@ xdp_fuse_lookup_id_for_inode (ino_t      ino,
     }
   else
     {
+      struct stat buf;
+
       /* directory document */
 
       /* Only return entire doc for main dir */
-      if (file_devino.dev == domain->doc_dir_device &&
-          file_devino.ino == domain->doc_dir_inode)
+      if (xdp_domain_get_doc_dir (domain, NULL, &buf) == 0 &&
+          buf.st_dev == file_devino.dev &&
+          buf.st_ino == file_devino.ino)
         return g_strdup (domain->doc_id);
 
       /* But maybe its a subfile of the document */
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -1328,10 +1328,8 @@ get_app_permissions (PermissionDbEntry *entry)
 static GVariant *
 get_path (PermissionDbEntry *entry)
 {
-  g_autoptr (GVariant) data = permission_db_entry_get_data (entry);
-  const char *path;
+  const char *path = document_entry_get_path (entry);
 
-  g_variant_get (data, "(^&ayttu)", &path, NULL, NULL, NULL);
   return g_variant_new_bytestring (path);
 }
 
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -39,6 +39,7 @@
 #include <gio/gio.h>
 #include <gio/gunixfdlist.h>
 #include <glib-unix.h>
+#include <libglnx.h>
 
 #include "document-portal-dbus.h"
 #include "document-portal-fuse.h"
@@ -299,6 +300,26 @@ portal_delete (GDBusMethodInvocation *invocation,
 
   /* Now fuse view is up-to-date, so we can return the call */
   g_dbus_method_invocation_return_value (invocation, g_variant_new ("()"));
+}
+
+GBytes *
+xdp_file_handle_for_fd (int fd)
+{
+  g_autofree struct file_handle *handle = NULL;
+  g_autofd int fd_owned = -1;
+
+  if (!(fcntl (fd, F_GETFL) & O_PATH))
+    fd = fd_owned = glnx_fd_reopen (fd, O_PATH, NULL);
+
+  if (fd < 0 ||
+      !glnx_name_to_handle_at (fd, "",
+                               AT_EMPTY_PATH | AT_HANDLE_FID,
+                               &handle,
+                               NULL,
+                               NULL))
+    return NULL;
+
+  return g_bytes_new (handle->f_handle, handle->handle_bytes);
 }
 
 static char *
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -489,6 +489,7 @@ validate_fd (int fd,
              ValidateFdType ensure_type,
              struct stat *st_buf,
              struct stat *real_dir_st_buf,
+             GBytes **real_dir_handle_out,
              char **path_out,
              gboolean *writable_out,
              GError **error)
@@ -528,6 +529,9 @@ validate_fd (int fd,
   dir_fd = open (dirname, O_CLOEXEC | O_PATH);
   if (dir_fd < 0 || fstat (dir_fd, real_dir_st_buf) != 0)
     goto errout;
+
+  if (real_dir_handle_out)
+    *real_dir_handle_out = xdp_file_handle_for_fd (dir_fd);
 
   if (name != NULL)
     {
@@ -887,6 +891,7 @@ document_add_full (int                      *fd,
   const char *app_id = xdp_app_info_get_id (app_info);
   g_autoptr(GPtrArray) ids = g_ptr_array_new_with_free_func (g_free);
   g_autoptr(GPtrArray) paths = g_ptr_array_new_with_free_func (g_free);
+  g_autoptr(GPtrArray) handles = g_ptr_array_new_with_free_func ((GDestroyNotify) g_bytes_unref);
   g_autofree struct stat *real_dir_st_bufs = NULL;
   struct stat st_buf;
   g_autofree gboolean *writable = NULL;
@@ -894,6 +899,7 @@ document_add_full (int                      *fd,
 
   g_ptr_array_set_size (paths, n_args + 1);
   g_ptr_array_set_size (ids, n_args + 1);
+  g_ptr_array_set_size (handles, n_args + 1);
   real_dir_st_bufs = g_new0 (struct stat, n_args);
   writable = g_new0 (gboolean, n_args);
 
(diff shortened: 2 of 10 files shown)


document-portal/document-portal.c
```diff
@@ -310,6 +311,30 @@ portal_delete (GDBusMethodInvocation *invocation,
...
+{
+  g_autofree struct file_handle *handle = NULL;
+  g_autofd int path_fd = -1;
+
+  if (!(fcntl (fd, F_GETFL) & O_PATH))
+    {
+      path_fd = glnx_fd_reopen (fd, O_PATH, NULL);
+      if (path_fd < 0)
+        return g_bytes_new (NULL, 0);
+      fd = path_fd;
```
A6GibKm: This makes it a bit hard to reason about whats happening (semantically), will both `fd` and `path_fd` be cleaned by g_autofd? Perhaps some g_clear_fd/g_steal_fd can make this cleaner?

swick: simplified this a bit

meson.build
```diff
@@ -124,6 +124,9 @@ libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 gudev_dep = dependency('gudev-1.0', required: get_option('gudev'))
 umockdev_dep = dependency('umockdev-1.0', required: get_option('tests'))
 
+libglnx_proj = subproject('libglnx', default_options: 'tests=false')
+libglnx_dep = libglnx_proj.get_variable('libglnx_dep')
```
tytan652: Why not using a wrap file (or a git submodule) ?
I don't see the point of keeping a copy to maintain.

smcv: libglnx doesn't have a stable API/ABI guarantee, so it doesn't make sense for distributions to package it separately as an external project: it's a "copylib".

`git submodule` is [really bad at](https://blog.timhutt.co.uk/against-submodules/) [what it does](https://diziet.dreamwidth.org/14666.html), and a wrap file is problematic for downstream packaging because it means the x-d-p release can't be compiled on its own (packagers usually don't allow network access during the build, to enforce that they really do have all of the source code available for auditing and LGPL compliance). This is a `git subtree`, which is `git submodule` done right.

Flatpak also uses libglnx. It switched from `git submodule` to `git subtree` a while ago (it was my idea, if I remember correctly) and it has been much more straightforward since then.

tytan652: When it comes to packaging, I consider this is not upstream job if they properly use the buildsystem or VCS. Packager can easily deal with meson subprojects and git submodules and provide the clone themself.

I just found it messier since all of libglnx files are part of the xdp repo.

swick: Personally I'm also not a fan of git subtree because it introduces merge commits which makes the rebase workflow more painful, and I would prefer a meson wrap file, but we have the subtree setup in flatpak and flatpak-builder already and I really don't want to diverge from that.

swick: Well, I changed my opinion after having to deal with git subtree a bit more. We're now using a meson subproject for it, and this PR here needs to be updated accordingly.

swick:
Not sure who to ping for a review. @AdrianVovk, @sophie-h?

sophie-h:
All makes sense to me at first glance! I had looked into always using `statx` directly, and my conclusion was also that this should work well for 'usual' file systems. So I totally agree with the solution :)

What I wanted to look into is how GVfs fuse or fuse in general behaves with this and if there could be a proper solution for that. But that would now be possible by just extending `xdp_file_handle_for_fd` for special cases. And no idea if we have any chance getting something useful out of the backends.

Another mental note I had is checking how call heavy `xdp_file_handle_for_fd` will be and if caching could make sense. Especially if we would do GFfs checks there in the future.

I can do a proper review at a later point.

This is semi part of this PR https://gitlab.gnome.org/swick/libglnx/-/commit/e9cfb9c252513e57d940e9f79ff0e1f98f73d321

subprojects/libglnx.wrap
```diff
@@ -1,4 +1,4 @@
 [wrap-git]
-url = https://gitlab.gnome.org/GNOME/libglnx.git
-revision = ccea836b799256420788c463a638ded0636b1632
-depth = 1
\ No newline at end of file
+url = https://gitlab.gnome.org/swick/libglnx.git
+revision = wip/name-to-handle-at
```
AdrianVovk: Flagging so this doesn't land with a WIP branch on a fork

document-portal/permission-db.c
```diff
@@ -526,39 +526,42 @@ permission_db_lookup (PermissionDb  *self,
...
+{
+  g_auto(GStrv) ids = permission_db_list_ids (self);
+  g_autoptr(GPtrArray) res = NULL;
 
   g_return_val_if_fail (PERMISSION_IS_DB (self), NULL);
-  g_return_val_if_fail (data != NULL, NULL);
+  g_return_val_if_fail (func != NULL, NULL);
 
-  res = g_ptr_array_new ();
+  res = g_ptr_array_new_null_terminated (0, NULL, TRUE);
```
AdrianVovk: If we're touching this anyway, might as well convert it to `GStrvBuilder`?

document-portal/permission-db.c
```diff
@@ -526,39 +526,42 @@ permission_db_lookup (PermissionDb  *self,
...
-            }
-        }
-      g_free (id);
+      entry = permission_db_lookup (self, ids[i]);
+      if (!entry)
+        continue;
+
+      entry_data = permission_db_entry_get_data (entry);
+      if (func (entry_data, user_data))
+        g_ptr_array_add (res, g_steal_pointer (&ids[i]));
```
AdrianVovk: You've introduced a memory leak here. The `ids` array is NULL terminated. So if the function returns true and you steal a pointer out of the array, you've left a NULL in the middle of the array. Thus, you've terminated the array and the rest will never get free'd

document-portal/document-portal.c
```diff
@@ -331,13 +331,175 @@ xdp_file_handle_for_fd (int fd)
...
+  };
+
+  ids = permission_db_filter_ids (db, matches_handle, &find_data);
+  if (ids[0] != NULL)
+    return g_strdup (ids[0]);
+  g_clear_pointer (&ids, g_strfreev);
+
+  find_data.flags |= DOCUMENT_ENTRY_FLAG_TRANSIENT;
+
+  ids = permission_db_filter_ids (db, matches_handle, &find_data);
```
AdrianVovk: `find_id` could take an additional argument: `gboolean ignore_transient`. That can be stored in the FindIdData. Next, `matches_handle` and `matches_dev_ino` can simply mask out `DOCUMENT_ENTRY_FLAG_TRANSIENT` if `match->ignore_transient` is true

Then you can drop `find_id_transient` entirely, along with its kinda awkward repetition of these calls

document-portal/document-portal.c
```diff
@@ -357,10 +519,14 @@ do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_exis
 
   if (reuse_existing)
     {
-      ids = permission_db_list_ids_by_value (db, data);
```
AdrianVovk, line 330: I checked, and it appears that this commit drops the last use of `permission_db_list_ids_by_value` so it should be dropped

document-portal/document-portal.c
```diff
@@ -331,13 +331,175 @@ xdp_file_handle_for_fd (int fd)
...
+  uint64_t st_ino;
+  uint32_t flags;
+
+  /* Only match if entry has no handle (old format or zero-sized byte array) */
+  if (g_variant_n_children (data) >= 5)
+    {
+      g_autoptr(GVariant) handle_variant = NULL;
+      g_autoptr(GBytes) handle = NULL;
+
+      g_variant_get (data, "(^&ayttu@ay)", &path, &st_dev, &st_ino, &flags, &handle_variant);
```
AdrianVovk: Didn't you just introduce a nice helper function for this?

The callback of `permission_db_filter_ids` should just take a `PermissionDbEntry` rather than the raw variant underneath. Then you can use the helper function you already wrote

document-portal/document-store.c
```diff
@@ -156,3 +156,17 @@ document_entry_get_flags (PermissionDbEntry *entry)
...
+
+GBytes *
+document_entry_dup_handle (PermissionDbEntry *entry)
+{
+  g_autoptr(GVariant) v = permission_db_entry_get_data (entry);
+  g_autoptr(GVariant) handle_variant = NULL;
+
+  /* Old format without handle field - return empty GBytes */
+  if (g_variant_n_children (v) < 5)
+    return g_bytes_new (NULL, 0);
```
AdrianVovk: Returning an empty `GBytes` seems like a foot gun. Now you can't just check for non-nullness, but you _also_ have to check for the "isn't null but is empty" edge case. Will forgetting to do that lead to a security vulnerability?

Thus, I suggest we just return NULL if the variant doesn't have the field at all, but also return NULL if it has an empty array stored there. Thus `document_entry_dup_handle` returns either NULL or a valid real handle.

document-portal/document-portal.c
```diff
@@ -331,13 +331,175 @@ xdp_file_handle_for_fd (int fd)
...
+  g_variant_get (data, "(^&ayttu@ay)",
+                 &path,
+                 NULL,
+                 NULL,
+                 &flags,
+                 &handle_variant);
+
+  handle = g_variant_get_data_as_bytes (handle_variant);
+
+  return g_strcmp0 (path, match->path) == 0 &&
```
AdrianVovk: If you do my other suggestions, it would probably make sense to combine `matches_handle` and `matches_dev_ino` into one function to reduce code duplication. You can decide if you should matched based on handle or based on dev+ino by checking that `match->handle` is NULL

In other words:

```c
static char
find_id_matches (PermissionsDbEntry *entry, void *user_data)
{
  FindIdData *match = user_data;
  g_autoptr(GBytes) handle = NULL;
  uint32_t flags;

  if (g_strcmp0 (document_entry_get_path (entry), match->path) != 0)
    return FALSE;

  flags = document_entry_get_flags (entry);
  if (match->ignore_transient)
    flags &= ~DOCUMENT_ENTRY_FLAG_TRANSIENT;
  if (match->flags != flags)
    return FALSE;

  if (!match->handle)
    return match->st_dev == document_entry_get_device (entry) &&
           match->st_ino == document_entry_get_inode (entry);

  handle = document_entry_dup_handle (entry);
  return handle && g_bytes_equal (handle, match->handle);
}

static char *
find_id (...)
{
  ...

  find_data = (FindIdData) {
    .path = path,
    .handle = handle,
    .st_dev = st_dev,
    .st_ino = st_ino,
    .flags = flags,
    .ignore_transient = ignore_transient,
  };

  ids = permission_db_filter_ids (db, find_id_matches, &find_data);
  ...

  /* We didn't have a handle, so we already checked by dev+ino */
  if (!handle)
    return NULL;

  /* We didn't find a match via handle, so fall back to checking by dev+ino */
  find_data.handle = NULL;
  ids = permission_db_filter_ids (db, find_id_matches, &find_data);
  ...

  return NULL;
}
```

document-portal/document-portal.c
```diff
@@ -331,13 +331,175 @@ xdp_file_handle_for_fd (int fd)
...
+         ino_t       st_ino,
+         GBytes     *handle,
+         uint32_t    flags)
+{
+  g_auto(GStrv) ids = NULL;
+  FindIdData find_data;
+
+  find_data = (FindIdData) {
+    .path = path,
+    .handle = handle,
```
AdrianVovk: `handle` is NULL sometimes (actually, it's always NULL as of this commit!), which means `match->handle` is NULL inside of `matches_handle`, and you don't check for this. Thus you call `g_bytes_get_size (match->handle)`, which derefs NULL.

document-portal/document-portal.c
```diff
@@ -510,12 +510,16 @@ do_create_doc (struct stat *parent_st_buf,
...
+    g_variant_ref_sink (g_variant_new ("(^ayttu@ay)",
                                        path,
-                                       (guint64) parent_st_buf->st_dev,
-                                       (guint64) parent_st_buf->st_ino,
-                                       flags));
+                                       (uint64_t) parent_st_buf->st_dev,
+                                       (uint64_t) parent_st_buf->st_ino,
+                                       flags,
+                                       g_variant_new_from_bytes (G_VARIANT_TYPE ("ay"),
+                                                                 handle,
```
AdrianVovk: As of this commit, `handle` is always NULL so you're dereferencing a NULL pointer here

document-portal/document-portal.c
```diff
@@ -525,8 +529,22 @@ do_create_doc (struct stat *parent_st_buf,
...
 
-      /* Reuse pre-existing entry with same path */
-      return g_steal_pointer (&id);
+      if (id)
+        {
+          g_autoptr(PermissionDbEntry) entry_old = permission_db_lookup (db, id);
+          g_autoptr(GBytes) old_handle = document_entry_dup_handle (entry_old);
+
+          /* If the old entry does not contain a handle, we update it */
+          if (g_bytes_get_size (handle) > 0 && g_bytes_get_size (old_handle) == 0)
```
AdrianVovk: Hmm so do we actually need to bother looking up the old entry here? Since we already did the match we can just unconditionally store the new entry back into the store, correct?

document-portal/document-portal.c
```diff
@@ -525,8 +529,22 @@ do_create_doc (struct stat *parent_st_buf,
...
+            {
+              g_debug ("Updating existing entry with a file handle");
+              entry = permission_db_entry_new (data);
+              permission_db_set_entry (db, id, entry);
```
AdrianVovk: I think this just changes variables in memory, and doesn't actually notify the permission store service. Is that correct? Was that the intent?

document-portal/document-portal.c
```diff
@@ -357,10 +519,14 @@ do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_exis
...
-      if (ids[0] != NULL)
-        return g_strdup (ids[0]);  /* Reuse pre-existing entry with same path */
+      id = find_id (path,
+                    parent_st_buf->st_dev,
+                    parent_st_buf->st_ino,
+                    handle,
+                    flags);
+
+      /* Reuse pre-existing entry with same path */
+      return g_steal_pointer (&id);
```
AdrianVovk: Unintended change in behavior: the returned ID can be null (i.e. you didn't find an existing ID to reuse). Before that would fall through to allocating a new ID, but now we just return NULL

document-portal/document-portal.c
AdrianVovk: Looks like you forgot to pass in the handle here!

document-portal/document-portal-fuse.c
```diff
@@ -975,9 +978,20 @@ xdp_domain_get_doc_dir (XdpDomain   *doc_domain,
   if (fstat (dirfd, buf) != 0)
     return -errno;
 
-  if (buf->st_ino != doc_domain->doc_dir_inode ||
-      buf->st_dev != doc_domain->doc_dir_device)
-    return -ENOENT;
+  if (doc_domain->doc_dir_handle != NULL)
```
AdrianVovk, line 981: Here's an example of that "non-null but is empty" footgun I mentioned before

doc_dir_handle isn't ever going to be null! If there's no handle we'll have an empty handle instead. But that means that instead of doing the dev+ino verification that we should do in the case that we don't have a handle, we'll do this other code path instead.

Does that lead to a security issue? Potentially! We compare this empty handle to a handle we obtain from the filesystem. But what if the filesystem doesn't support handles or the lookup fails for any other reason?

Well, then `xdp_file_handle_for_fd` _also_ returns an empty `GBytes`. Uh oh! The condition in the if statement doesn't save us: it checks for NULL (which `xdp_file_handle_for_fd` never returns). Thus `g_bytes_equal` returns TRUE and the check is completely bypassed

document-portal/document-portal.c
```diff
@@ -310,6 +311,26 @@ portal_delete (GDBusMethodInvocation *invocation,
...
+  if (!(fcntl (fd, F_GETFL) & O_PATH))
+    fd = fd_owned = glnx_fd_reopen (fd, O_PATH, NULL);
+
+  if (fd < 0 ||
+      !glnx_name_to_handle_at (fd, "",
+                               AT_EMPTY_PATH | AT_HANDLE_FID,
+                               &handle,
+                               NULL,
+                               NULL))
+    return g_bytes_new (NULL, 0);
```
AdrianVovk: My feedback on `document_entry_dup_handle` regarding the foot gun of returning an empty `GBytes` applies here too

AdrianVovk (changes requested):

swick:
I think I managed to address all the raised issues.

document-portal/document-portal.c
```diff
@@ -331,13 +331,108 @@ xdp_file_handle_for_fd (int fd)
   return g_bytes_new (handle->f_handle, handle->handle_bytes);
 }
 
+typedef struct _FindIdData {
```
AdrianVovk, line 325: nit: I think this commit message should prioritize explaining that you've extended the document ID search codepaths to also search via a handle. The fact that it's now also factored into a reusable function is kinda secondary, IMO

document-portal/document-portal.c
```diff
@@ -357,10 +452,15 @@ do_create_doc (struct stat *parent_st_buf, const char *path, gboolean reuse_exis
...
-        return g_strdup (ids[0]);  /* Reuse pre-existing entry with same path */
+      id = find_id (path,
+                    parent_st_buf->st_dev,
+                    parent_st_buf->st_ino,
+                    handle,
+                    flags,
+                    FALSE /* ignore_transient */);
+
+      /* Reuse pre-existing entry with same path */
+      return g_steal_pointer (&id);
```
AdrianVovk: Looks like the issue I noted in https://github.com/flatpak/xdg-desktop-portal/pull/1950#discussion_r3114775323 is still here:

> Unintended change in behavior: the returned ID can be null (i.e. you didn't find an existing ID to reuse). Before that would fall through to allocating a new ID, but now we just return NULL

I know that a subsequent commit makes this moot but in this commit it is a bug

document-portal/document-portal.c
```diff
@@ -458,25 +453,30 @@ do_create_doc (struct stat *parent_st_buf,
...
-        break;
+      g_debug ("reuse_doc %s", id);
     }
+  else
+    {
+      while (TRUE)
+        {
+          g_autoptr(PermissionDbEntry) existing = NULL;
+
+          if (id != NULL)
```
AdrianVovk: `while (id == NULL)` should do the trick right?

document-portal/document-portal.c
```diff
@@ -331,13 +331,108 @@ xdp_file_handle_for_fd (int fd)
...
+      g_autoptr(GBytes) handle = NULL;
+
+      handle = document_entry_dup_handle (entry);
+      if (!handle || !g_bytes_equal (handle, match->handle))
+        return FALSE;
+    }
+  else
+    {
+      if (match->st_dev != document_entry_get_device (entry) ||
+          match->st_ino == document_entry_get_inode (entry))
```
AdrianVovk: Did you mean `!=`?

swick: urgh, good catch

swick:
Took a while to find the last few refcounting issues but this should be ready now, ignoring the libglnx changes.

subprojects/libglnx.wrap
swick: for reference, https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/70

document-portal/document-store.c
```diff
@@ -157,3 +157,25 @@ document_entry_get_flags (PermissionDbEntry *entry)
...
 }
+
+GBytes *
+document_entry_dup_handle (PermissionDbEntry *entry)
+{
+  g_autoptr(GVariant) v = permission_db_entry_get_data (entry);
+  g_autoptr(GVariant) handle_variant = NULL;
+  g_autoptr(GBytes) handle = NULL;
+
+  /* Old format without handle field - return empty GBytes */
```
AdrianVovk: Nit: "return empty GBytes" is an out-of-date comment

AdrianVovk (approved):
LGTM, aside from glnx stuff not landing yet

subprojects/libglnx.wrap
swick: Merged and updated here to point at the right repo and commit.

swick:
Did an oopsie and pushed a previous version. All good again.

document-portal/document-store.c
```diff
@@ -157,3 +157,49 @@
...
+
+  /* We treat 0-sized handles as no handle. The variant needs to be serializable
+   * into a dbus message, so we don't use an optional here. */
+  if (g_bytes_get_size (handle) == 0)
+    return NULL;
+
+  return g_steal_pointer (&handle);
+}
+
+PermissionDbEntry *
```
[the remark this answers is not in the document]
swick: goblint being useful already

PR #1958, opened 2026-04-09
Misc cleanups

A6GibKm:
Found using github.com/bilelmoussaoui/gobject-lint.


document-portal/gvdb/gvdb-builder.c
```diff
@@ -515,7 +515,7 @@ gvdb_table_get_content (GHashTable     *table,
   str = file_builder_serialise (fb, root);
 
   len = str->len;
-  res = g_bytes_new_take (g_string_free (str, FALSE), len);
+  res = g_bytes_new_take (g_string_free_and_steal (str), len);
```
A6GibKm: Actually, this should be g_bytes_new_take.

PR #1963, opened 2026-04-10
doc: Add no-LLM statement

tytan652:
Based on GNOME Calendar, Libadwaita guidelines.

Portals are an important piece of software in Linux-based platform's ecosystem and I don't think LLMs will help at all even worse it might make potential contributors full-stop considering or trying to contribute.

------

Closes #1933


jadahl:
Not sure this is the right policy to make. See mesa (https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40617) and gtk (https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/9694) for alternatives to take inspiration from.

tytan652:
I might checks those ~~what GTK did, but in no way I will for Mesa's () due to the big difference in maintainer-ship and backing of the project~~.

We already are low on people for Portals, so we really should avoid cutting away volunteers that do not want to deal with LLMs (you could say the opposite, but those if new are likely to just give what the model just spilled out).

swick:
Sorry, but an outright ban is not going to happen. I do like the mesa one though.

tytan652:
Then can we ban any addition meant to help LLMs to interpret/understand the repo/project (e.g. AGENTS.md) ?

Edit: I understand I need to compromise, but encouraging its use is also not good.

pedropaulosuzuki:
I fully agree! The Mesa decision to allow tools built on stolen code from millions of FOSS contributors (and otherwise) is not a good decision. Many people will not want to contribute to repositories where LLMs are accepted in any way, and the [Calendar](https://gitlab.gnome.org/GNOME/gnome-calendar/-/blob/main/CONTRIBUTING.md#use-of-generative-ai) and [LibAdwaita positions](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/CONTRIBUTING.md#use-of-generative-ai) are the only ethical positions to hold until ethical models exist and, even if they exist, they are unpredictable by nature and thus, unreliable. We should follow their example, and not Mesa's. Also, they still allow using LLMs for translating issues and comments, and for understanding concepts when the docs aren't clear on some subject, which is a good compromise. Anything more than that is way too problematic. Those are my two cents.

swick:
Please don't add comments here if you're not contributing.

swick:
If anyone want to add a LLM policy, it needs consensus from committers.

PR #1964, opened 2026-04-10
RFC: Improve interfaces versioning

tytan652:
This is tentative to improve xdg-desktop-portal D-Bus interfaces versioning (unrelated to xdp own versioning but could help).

[RST preview](https://github.com/tytan652/xdg-desktop-portal/blob/improve-versioning/doc/versioning.rst)

We have some portals that have deprecated functionality without a proper way to "API break", some portal in a weird middle-ground (e.g. Notification) and `version` property that do not have the same API expectation across interfaces.

- Major version is on the main interface name e.g. `org.freedesktop.portal.Camera2`, the lack or number suffix means 1
- The `version` in the interface documentation is turned into `revision` and behave similarly to a minor version
- The `version` interface is replaced with `active-revision` and act as a reference of the available feature-set.
  - `version` is kept deprecated on existing interfaces, and new interfaces are expected to only provide `active-revision`
  - If the property is not set by the implementation, assume 1.
  - `active-revision` only shows the active feature-set not the documented revision.
  
It allows to keep two major version of an interface in parallel to allow a deprecation period that could be align more with Desktop environments releases since they provide implementations.

This could also detach the "API breakage" from xdg-desktop-portal own versioning.

PS: This is also meant to enable grounds forward enabling staging portals (in a specific object path) with lower stability expectation.

TODO (maybe in a separated PR ?):
- [ ] Extend tests to check more impl version/revision scenario (mainly portals with impl revision greater than 1)

TODO before merge (if it happens):
- [ ] Check every `active-revision` is properly documented as being introduced after the last version/revision of the interface last release of XDP
- [x] Check GlobalShortcuts version behavior (e.g. did #1968 get merged or not)
- [ ] Check Print (impl) revision assumption

### DIFF
--- a/doc/versioning.rst
+++ b/doc/versioning.rst
@@ -1,0 +1,46 @@
+Versionning
+===========
+
+D−Bus Interfaces
+----------------
+
+D-Bus interfaces are versionned to:
+
+- Enable non-backward compatible changes (e.g. remove deprecated functions,
+  interface overhaul) with major bump.
+- Enable backward compatible changes (e.g. method addition, new property for
+  the optional argument) with minor bump.
+
+Two major version of an interface can coexist to allow a deprecation period for
+the former version.
+
+The major version is determined by the main interface name (to not mix up with
+the deprecated ``version`` property in interfaces documentation).
+
+.. csv-table:: Major Version Example
+    :header: "Interface Name", "Version", "Reason"
+
+    "org.freedesktop.portal.Settings", "1", "Lack of number suffix, default to 1"
+    "org.freedesktop.portal.Camera2", "2", "Suffix number is 2"
+
+Interfaces have a minor version documented as "revision" (previously
+"version" but renamed to avoid confusion), assume 1 if not documented.
+
+Each of those interfaces must have a ``active-revision`` property which is not
+meant to match the documented "revision" but the revision matching the
+feature-set enabled, if missing or not set (returns 0) check the interface
+documentation of the property for the recommended fallback or assumption to do.
+
+The ``version`` property has been deprecated and is not guaranted to be identical
+to the new ``active-revision`` for every interface.
+
+Implementation interfaces
+-------------------------
+
+Portal implementation interfaces are also versionned (and can have revisions) but
+do not require to match their non-implementation counterpart.
+
+When an implementation provides an interfaces, it is expected that the feature-set
+behind the set ``active-revision`` is implemented, if missing or not set (returns
+0\) check the interface documentation of the property for the matching fallback
+behaviour that will be used.
--- a/data/org.freedesktop.background.Monitor.xml
+++ b/data/org.freedesktop.background.Monitor.xml
@@ -28,7 +28,7 @@
     This interface provides APIs related to applications
     that are running in the background.
 
-    This documentation describes version 1 of this interface.
+    This documentation describes revision 1 of this interface.
   -->
   <interface name="org.freedesktop.background.Monitor">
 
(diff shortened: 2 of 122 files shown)


doc/versioning.rst
```diff
@@ -0,0 +1,33 @@
...
+the deprecated ``version`` property).
+
+.. csv-table:: Example
+    :header: "Interface Name", "Version", "Reason"
+
+    "org.freedesktop.portal.Settings", "1", "Lack of number suffix, default to 1"
+    "org.freedesktop.portal.Camera2", "2", "Suffix number is 2"
+
+Interfaces have as minor version a property called ``revision`` (previously
+``version`` but renamed to avoid confusion).
```
tytan652: - Does the property should report the actual revision as a feature-set or the one documented in the interface definition ?
- Edit: If the interface was does the latter, should the deprecated `version` keep the behavior and `revision` applies the former ?
  - Edit 2: Note,  that Settings is a "special" exception.
- Edit 3: Should `revision` be renamed `active-revision` ?

swick:
I struggle to see the point of this. It is already how things work: the version describes the version of the interface. If we need to do a backwards incompatible changes we add a new one (there is a precedence with `CreateSession2` for a method, but the same is true for interfaces).

tytan652:
> It is already how things work: the version describes the version of the interface.

It isn't, some portals have `version` set to the latest whatever the impl version is, other adapt to the impl version.

Interfaces and portal are not aligned on that.

> If we need to do a backwards incompatible changes we add a new one (there is a precedence with `CreateSession2` for a method, but the same is true for interfaces).

You really need to consider that we can't everlastingly add new API and deprecate others endlessly. At some we need to cleanup et introduce a breakage, which can be represented by putting a version in the interface name.
Edit 3: Re-read the sentence, this is not defined and it should be explicitly (implicit is not good).

Edit: It also prepare the possibility to keep two interface major version in parallel for the deprecation period since nothing is defined.

-----

This PR tries to set the same versioning behavior on all interfaces even if some might never change (e.g. Session, Request).

And I want to rely on that to give a try to staging portals, that are meant to be in another object path and less stable meaning by allowing breakage without backward compat by bumping the major version.

Edit: `active-revision` and `revision` are meant to go away from the mess that `version` is, due to the misalignment. And ensure that the `active-revision` represent the actual capabilities or the called portal, unlike e.g. ScreenCast with its `version`.

Edit 2: Interfaces versioning (and API stability) is also not properly documented

swick:
> It isn't, some portals have version set to the latest whatever the impl version is, other adapt to the impl version.

What do you even mean with this? The version tells the client which features and behavior it can expect. The frontend always must deal with impls of any version, but that might or might not affect the API it can provide to the client.

tytan652:
> What do you even mean with this?

Like I said not all portal are like this. Without #1968, GlobalShortcuts was going to behave like ScreenCast, RemoteDesktop or Print. Those 3 by precedence are already not indicating which feature will work, but matches the documented version.

I'm not pulling this PR just for the fun, we do have interfaces where `version` does not behave the same as others.

swick:
The GlobalShortcuts issue was a bug, nothing more. If other portals also have a similar issue, we should look into it. Creating a completely new versioning system in addition to the already existing broken mess is just making everything more complicated.

What would actually be helpful is coming up with a way to test things, so those issues don't repeat, and figuring out how to reduce the mess we currently have.

tytan652:
> The GlobalShortcuts issue was a bug, nothing more.

It wasn't a bug if the the expectation behind its `version` was the same as ScreenCast and co.

My PR just proposed to switch to the other expectation since it was still version 2 only in unstable.

> If other portals also have a similar issue, we should look into it.

Technically `version` being misaligned is not a bug. Why ? Because it was not properly documented so we don't know if it's a bug or because not everybody was having the same vision for this property.

> Creating a completely new versioning system in addition to the already existing broken mess is just making everything more complicated.
> 
> What would actually be helpful is coming up with a way to test things, so those issues don't repeat, and figuring out how to reduce the mess we currently have.

Adding testing was planned if the RFC was considered going the right way.
I would not feel okay, if I got a rejection and realized that I wasted time implementing new tests for nothing (which is more time-consuming to implement since you have to deal with all value that impl version can be and future-proofing).

PR #1966, opened 2026-04-11
Ensure that version is never set to an undocumented version

tytan652:
While working on #1964, I came across those two portals that could end up with version 0 if the implementation was doing something wrong.

Found that notification (and screenshot) portal allows setting a version greater than documented.

While it is the implementation that is in the wrong, portal consumer should not deal with version 0 or one that does not exist within the host xdg-desktop-portal capabilities.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -1298,7 +1298,7 @@ notification_new (XdpContext              *context,
   notification->impl_version =
     MAX (xdp_dbus_impl_notification_get_version (notification->impl), 1);
   xdp_dbus_notification_set_version (XDP_DBUS_NOTIFICATION (notification),
-                                     notification->impl_version);
+                                     MIN (notification->impl_version, 2));
 
   g_object_bind_property (G_OBJECT (notification->impl), "supported-options",
                           G_OBJECT (notification), "supported-options",
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -528,7 +528,6 @@ screenshot_new (XdpDbusImplScreenshot *impl,
                 XdpDbusImplAccess     *access_impl)
 {
   Screenshot *screenshot;
-  g_autoptr(GVariant) version = NULL;
 
   screenshot = g_object_new (screenshot_get_type (), NULL);
   screenshot->access_impl = g_object_ref (access_impl);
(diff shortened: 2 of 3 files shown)


src/notification.c
```diff
@@ -1298,7 +1298,7 @@ notification_new (XdpContext              *context,
   notification->impl_version =
     MAX (xdp_dbus_impl_notification_get_version (notification->impl), 1);
   xdp_dbus_notification_set_version (XDP_DBUS_NOTIFICATION (notification),
-                                     notification->impl_version);
```
whot, line 1301: comment on the commit message: imo this should be "Ensure that the version set is correct"

whot (approved):
LGTM

src/notification.c
tytan652: Commit message updated.

src/screenshot.c
```diff
@@ -539,10 +538,8 @@ screenshot_new (XdpDbusImplScreenshot *impl,
...
   /* Before there was a version property, the version was hardcoded to 2, so
    * make sure we retain that behaviour */
-  version = g_dbus_proxy_get_cached_property (G_DBUS_PROXY (screenshot->impl),
-                                              "version");
   screenshot->impl_version =
-    (version != NULL) ? g_variant_get_uint32 (version) : 2;
+    MAX (xdp_dbus_impl_screenshot_get_version (screenshot->impl), 2);
 
   xdp_dbus_screenshot_set_version (XDP_DBUS_SCREENSHOT (screenshot),
                                    screenshot->impl_version);
```
swick, line 545: Shouldn't this than also be a MIN(impl, max_frontend)?

tytan652: The first commit is for dealing with scenario where the impl returns a lower version than 1 so 0 but for screenshot it's 2 (check already existing comment).

tytan652: And also check the second commit too.

swick: Mh, but then why is `MIN (input_capture->impl_version, 2));` added to the input capture in this commit?

tytan652: It was already existing before the PR, I just moved it to put the MAX first to avoid 0 (more aligned with already existing MAX checks like notification).

PR #1967, opened 2026-04-11
input-capture: Add CreateSession missing deprecation annotation

tytan652:
CreateSession was deprecated but the portal interface did not get the annotation.

I don't think there is a point to add one in the impl.


Sodivad:
> I don't think there is a point to add one in the impl.

Agree.

Sodivad (approved):

PR #1973, opened 2026-04-16
tests: Make XdpAppInfo types dataclass subclasses

swick:
```
This has a few benefits:

* We can construct them at pytest test collection time
* All the fields are clearly spelled out
* We can do a single initialize call to create all the runtime things the app info needs
```

Something I wanted to do for some time now, and will be useful for the Entitlement tests for example.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,7 @@
 
 import tests.xdp_utils as xdp
 
-from typing import Any, Dict, Iterator, Optional, assert_never
+from typing import Any, Dict, Iterator, Optional
 from types import ModuleType
 
 import pytest
@@ -37,6 +37,12 @@ def pytest_sessionfinish(session, exitstatus):
     # tests were skipped
     if exitstatus == pytest.ExitCode.NO_TESTS_COLLECTED:
         session.exitstatus = 77
+
+
+def pytest_make_parametrize_id(config, val):
+    if isinstance(val, xdp.AppInfo):
+        return val.__class__.__name__
+    return None
 
 
 def ensure_environment_set() -> None:
@@ -117,11 +123,11 @@ def create_test_dirs(umockdev: Optional[UMockdev.Testbed]) -> Iterator[None]:
 
 
 @pytest.fixture
-def xdp_mocked_executables(xdp_app_info: xdp.AppInfo) -> list[xdp.ExecutableMock]:
+def xdp_mocked_executables(xdp_app_info_init: xdp.AppInfo) -> list[xdp.ExecutableMock]:
     exe = None
-    if xdp_app_info.kind == xdp.AppInfoKind.FLATPAK:
+    if isinstance(xdp_app_info_init, xdp.AppInfoFlatpak):
         exe = "flatpak"
-    elif xdp_app_info.kind == xdp.AppInfoKind.SNAP:
+    elif isinstance(xdp_app_info_init, xdp.AppInfoSnap):
         exe = "snap"
     else:
         return []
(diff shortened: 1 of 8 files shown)


swick:
ping @whot

tests/conftest.py
```diff
@@ -117,11 +123,11 @@ def create_test_dirs(umockdev: Optional[UMockdev.Testbed]) -> Iterator[None]:
...
 
 @pytest.fixture
-def xdp_mocked_executables(xdp_app_info: xdp.AppInfo) -> list[xdp.ExecutableMock]:
+def xdp_mocked_executables(xdp_app_info_init: xdp.AppInfo) -> list[xdp.ExecutableMock]:
     exe = None
-    if xdp_app_info.kind == xdp.AppInfoKind.FLATPAK:
+    if type(xdp_app_info_init) is xdp.AppInfoFlatpak:
         exe = "flatpak"
-    elif xdp_app_info.kind == xdp.AppInfoKind.SNAP:
+    elif type(xdp_app_info_init) is xdp.AppInfoSnap:
```
whot, line 130: this feels like it should be an `instanceof()` but if there was a reason for the explicit `type()`, just ignore it :)

this applies to a few more explicit `type()` checks below too

tests/test_background.py
```diff
@@ -89,7 +89,7 @@ def test_autostart_desktopfile(self, portals, dbus_con, xdp_app_info):
         assert response.results["background"]
 
         # Unsupported on snap and linyaps
-        if xdp_app_info.kind in {xdp.AppInfoKind.SNAP, xdp.AppInfoKind.LINYAPS}:
+        if type(xdp_app_info) in {xdp.AppInfoSnap, xdp.AppInfoLinyaps}:
```
whot, line 92: same as the `type()` comment above and applies to a few cases below too
```suggestion
        if any(isinstance(xdp_app_info, t) for t in {xdp.AppInfoSnap, xdp.AppInfoLinyaps}):
```

tests/xdp_utils.py
```diff
@@ -243,258 +243,243 @@ def desktop_files_path() -> Path:
...
 @dataclass
 class AppInfo:
     """
-    Interacts with conftest.py via ensure_files and extend_env to make the
-    portal frontend discover the requested XdpAppInfo for incoming connections.
-
     Testing code can use this class to construct a specific XdpAppInfo for the
     xdp_app_info fixture.
+
+    The make it possible to collect AppInfo instances in pytest, construction
```
whot, line 252:
```suggestion
    To make it possible to collect AppInfo instances in pytest, construction
```

tests/conftest.py
```diff
@@ -503,8 +487,8 @@ def xdp_env(
     env["XDG_CURRENT_DESKTOP"] = "test"
     env["PATH"] = xdp_bin_path.as_posix()
 
-    if xdp_app_info:
-        xdp_app_info.extend_env(env)
+    for key, val in xdp_app_info_init._env.items():
```
whot, line 490: if you use `_foo` outside self it probably shouldn't be an underscored "private" property/method

tests/xdp_utils.py
```diff
@@ -243,258 +243,243 @@ def desktop_files_path() -> Path:
...
-        for path, content in self.files.items():
-            path.parent.mkdir(parents=True, exist_ok=True)
-            path.write_bytes(content)
+    desktop_file: str | None
 
-    def extend_env(self, env: dict[str, str]) -> None:
-        for key, val in self.env.items():
-            env[key] = val
+    def __init__(self):
+        self._env = {}
```
whot, line 262: could just move this into the dataclass properties? for what we're doing here it's simple enough

whot (approved):

tests/test_background.py
swick: `isinstance` actually takes a tuple of types to check against

tests/xdp_utils.py
swick: Mh, so I see the dataclass props as things that a user of the class can use to construct it, not for things that the class generates. I added an accessor method to XdpAppInfo instead, but open to suggestions.

tests/test_background.py
whot: TIL!

PR #1974, opened 2026-04-17
goblint integration

swick:
Draft because I want to get this down to zero violations, which requires some work in goblint (specifically an option to ignore a rule, and a feature to declare inline to ignore a specific violation).

Also misses CI integration.


document-portal/document-portal-fuse.c
```diff
@@ -24,24 +24,27 @@
...
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
-#include <errno.h>
+
 #include <fcntl.h>
-#include <stdlib.h>
-#include <assert.h>
-#include <glib/gprintf.h>
+#define FUSE_USE_VERSION 35
```
bilelmoussaoui, line 36: weird that this was moved here... needs a fixture test upstream for sure.

swick: I did this manually and on purpose. this needs to be defined for the fuse lowlevel include

bilelmoussaoui: ah alright then. Good to know

swick:
I updated to the latest image and then fixed the image in CI. I think this should be ready, but it also needs someone else to take a quick look over the changes.

src/xdp-documents.c
```diff
@@ -23,20 +23,20 @@
...
+#include "xdp-documents.h"
+
 #include <errno.h>
 
-#include <sys/types.h>
-#include <sys/stat.h>
 #include <fcntl.h>
-
 #include <gio/gio.h>
 #include <gio/gunixfdlist.h>
```
swick, line 33: @bilelmoussaoui why does this create those two blocks here? AFAIU this should be one alphabetically ordered block here, no?

bilelmoussaoui: it is supposed to be grouped into the following categories separate by a new line
- config.h
- associated header (xdp-documents.h)
- standard c headers
- system libs headers
- current project header

which is what mutter is doing, would you have expected a different result? 

The problem here is that posix / standard c headers are split which probably they shouldn't?

src/xdp-session.c
```diff
@@ -52,7 +52,7 @@ static void xdp_session_skeleton_iface_init (XdpDbusSessionIface *iface);
 
 G_DEFINE_TYPE_WITH_CODE (XdpSession, xdp_session, XDP_DBUS_TYPE_SESSION_SKELETON,
                          G_IMPLEMENT_INTERFACE (G_TYPE_INITABLE,
-                                                g_initable_iface_init)
+                                                g_initable_iface_init);
```
bilelmoussaoui, line 55: this is unexpected, does the rule actually produces this?? if so, an issue should be reported and fixed before merging this.

document-portal/document-portal-fuse.c
```diff
@@ -1805,7 +1805,7 @@ invalidate_dentry_cb (gpointer user_data)
 
   g_list_free (to_invalidate);
 
-  return FALSE;
+  return G_SOURCE_REMOVE;
```
bilelmoussaoui, line 1808: is this not a candidate for _once variants or because of the glib requirement they are not suggested, i guess

src/xdp-session.c
swick: Oh, wow, I didn't notice. And yes, this was completely done by goblint.

src/xdp-documents.c
bilelmoussaoui: re-running it again should produce better results now

src/xdp-session.c
swick: https://github.com/bilelmoussaoui/goblint/issues/56

document-portal/document-portal-fuse.c
swick: Didn't know this should be trying to use g_timeout_add_once. It's available since 2.74, and we require 2.76.

bilelmoussaoui: I am assuming it should be detected, could you try to run `goblint --summary` and see?

src/xdp-documents.c
swick: Mh, I didn't immediately recognize the standard c headers as such. I don't have a strong opinion on it, but differentiating between standard c headers and system libs headers doesn't seem useful to me.

bilelmoussaoui: I guess the other option here would be to make the sorting configurable...

document-portal/document-portal-fuse.c
swick: Pulled again, still the same:
```
podman run --rm -v "$PWD:/workspace:Z" ghcr.io/bilelmoussaoui/goblint:latest . --only use_g_source_constants
Found 1 violation(s):

/workspace/document-portal/document-portal-fuse.c:1808:10
          fuse_lowlevel_notify_inval_entry (session, data->parent_ino, data->name,
                                            strlen (data->name));
        g_free (data);
  ...
    g_list_free (to_invalidate);
  ...
  >  return FALSE;
  ...
  /* Queue an inval_dentry, thereby freeing unused inodes in the dcache
  error: Use G_SOURCE_REMOVE instead of FALSE in GSourceFunc callback
  rule: use_g_source_constants
```

bilelmoussaoui: weird, it does get detected here 
```
user@toolbx ~/P/gobject-format (main)> cargo run --bin goblint -- ~/document-portal-fuse.c --only use_g_source_once
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/goblint /home/user/document-portal-fuse.c --only use_g_source_once`
Found 1 violation(s):

/home/user/document-portal-fuse.c:1832:5
  ...
    XdpInvalidateData *data = g_malloc0 (sizeof (XdpInvalidateData) + strlen (name) + 1);
    data->parent_ino = parent->ino;
    strcpy (data->name, name);
  ...
    if (invalidate_list == NULL)
  >    g_timeout_add (10, invalidate_dentry_cb, NULL);
  ...
    invalidate_list = g_list_append (invalidate_list, data);
  ...
  error: Callback 'invalidate_dentry_cb' always returns G_SOURCE_REMOVE. Use g_timeout_add_once instead of g_timeout_add
  rule: use_g_source_once
```

goblint.toml
```diff
@@ -0,0 +1,4 @@
+min_glib_version = "2.20"
```
AdrianVovk: Maybe I misunderstand what this is about, but don't we depend on glib >= 2.76? Why's this 2.20?

src/xdp-app-info.c
```diff
@@ -79,7 +79,7 @@ typedef struct _XdpAppInfoPrivate
 static void g_initable_init_iface (GInitableIface *iface);
 
 G_DEFINE_ABSTRACT_TYPE_WITH_CODE (XdpAppInfo, xdp_app_info, G_TYPE_OBJECT,
-                                  G_ADD_PRIVATE (XdpAppInfo)
+                                  G_ADD_PRIVATE (XdpAppInfo);
```
AdrianVovk: this seems wrong?

document-portal/document-portal-fuse.c
```diff
@@ -1820,12 +1820,14 @@ queue_invalidate_dentry (XdpInode   *parent,
...
       XdpInvalidateData *data = l->data;
-      if (data->parent_ino == parent->ino && strcmp (name, data->name) == 0)
+      if (data->parent_ino == parent->ino && g_strcmp0 (name, data->name) == 0)
         return;
     }
 
   XdpInvalidateData *data = g_malloc0 (sizeof (XdpInvalidateData) + strlen (name) + 1);
   data->parent_ino = parent->ino;
+  /* We allocated enough memory in data->name for name.
+   * goblint-ignore-next-line: use_g_strlcpy */
```
AdrianVovk: Nit: we replace every instance of `strcmp` with `g_strcmp0` without bothering to define ignores for it when it's not really necessary. Should we bother to try and define an ignore here?

```
size_t name_buf_size = strlen (name) + 1;
XdpInvalidateData *data = g_malloc0 (sizeof (XdpInvalidateData) + name_buf_size)
g_strlcpy (data->name, name, name_buf_size)
``` 

Hell, since we already know the `name_buf_size` we can just call `memcpy`

src/gamemode.c
```diff
@@ -40,8 +42,7 @@
 #include "xdp-permissions.h"
 #include "xdp-utils.h"
 
-#include "gamemode.h"
-
+#define _GNU_SOURCE 1
```
AdrianVovk: This define doesn't do anything if you put it after your  `#include`s, so this seems wrong?

src/xdp-app-info-flatpak.c
```diff
@@ -23,17 +23,19 @@
...
-#include <sys/vfs.h>
-#endif
 
+#include <fcntl.h>
 #include <json-glib/json-glib.h>
 
-#include "xdp-app-info-flatpak-private.h"
 #include "xdp-usb-query.h"
 
+#if HAVE_SYS_VFS_H
```
AdrianVovk, line 35: what even is the rule here being used to order this? These are OS-level headers, why isn't this grouped up with errno.h?

document-portal/gvdb/gvdb-builder.h
```diff
@@ -17,8 +17,7 @@
  * Author: Ryan Lortie <desrt@desrt.ca>
  */
 
-#ifndef __gvdb_builder_h__
-#define __gvdb_builder_h__
+#pragma once
```
AdrianVovk: probably shouldn't be touching gvdb

document-portal/permission-db.c
```diff
@@ -70,12 +70,10 @@ static void initable_iface_init (GInitableIface *initable_iface);
 G_DEFINE_TYPE_WITH_CODE (PermissionDb, permission_db, G_TYPE_OBJECT,
                          G_IMPLEMENT_INTERFACE (G_TYPE_INITABLE, initable_iface_init));
 
-enum {
-  PROP_0,
-  PROP_PATH,
+typedef enum {
+  PROP_PATH = 1,
   PROP_FAIL_IF_NOT_FOUND,
-  N_PROPS,
```
AdrianVovk: churn: a few commits ago you had an identically named commit that added this `N_PROPS`

src/xdp-app-info-snap.c
```diff
@@ -43,14 +43,12 @@ struct _XdpAppInfoSnap
...
 {
-  PROP_0,
-  PROP_DESKTOP_FILE,
-  N_PROPS,
-};
+  PROP_DESKTOP_FILE = 1,
+} XdpAppInfoSnapProps;
 
-static GParamSpec *properties [N_PROPS];
+static GParamSpec *properties [PROP_DESKTOP_FILE + 1];
```
AdrianVovk, line 52: This isn't really the convention right? If you add another property you'd have to magically remember to go and change this array length bound here?!?

document-portal/file-transfer.c
```diff
@@ -236,6 +236,8 @@ file_transfer_stop (FileTransfer *transfer)
   g_hash_table_steal (transfers, transfer->key);
   G_UNLOCK (transfers);
 
+  /* We have a ref and unref in the callback, so transfer is always alive.
+   * goblint-ignore-next-line: g_source_id_not_stored */
```
AdrianVovk: This comment isn't super clear to me, perhaps it needs a reword? It sounds like we're reffing and unreffing in the callback and that doesn't make much sense

Also, it's not entirely clear why we're even scheduling the unref in an idle at all. I assume it's to make sure the destructor runs on the main thread?

Also also, we can just pass `g_object_unref` as the function to execute, no?

So, all in all, maybe we should do something like this?

```c
/* We need to unref the transfer from the main thread
 * goblint-ignore-next-line: g_source_id_not_stored */
g_idle_add_once (g_object_unref, g_steal_pointer (&transfer))
```

Perhaps the explicit ownership transfer into the idle callback would make the warning go away? Worth testing to make sure the warning is still happening. If it is, probably a point of improvement for goblint?

AdrianVovk (changes requested):

src/xdp-app-info-snap.c
bilelmoussaoui: see https://github.com/bilelmoussaoui/goblint/issues/52

src/xdp-app-info-flatpak.c
bilelmoussaoui: The grouping is done for the global includes and the ones inside ifdef separately. First all global includes sorted and then the ones inside ifdefs.

goblint.toml
swick: Yikes, I think I tested things and forgot to change it back.

src/gamemode.c
swick: Urgh, another one of those. @bilelmoussaoui this is not great tbh. Some defines need to be set before an include and goblint will happily move it after them. Not sure if there is anything that can be done really.

src/xdp-app-info.c
swick: There was a thread about it and it's fixed now; just need to run it again.

document-portal/document-portal-fuse.c
swick: Seems to work now. Probably because I had the wrong version in the config.

document-portal/file-transfer.c
swick: /cc @bilelmoussaoui yeah, not sure if there is a way to detect this kind of usage.

src/gamemode.c
swick: The reordering also often exposes missing includes etc., so that one needs manual fixups anyway.

swick:
/cc @bilelmoussaoui 
```c
static void
xdp_request_on_signal_response (XdpDbusRequest *object,
                                guint           arg_response,
                                GVariant       *arg_results)
{
  XdpRequest *request = XDP_REQUEST (object);
  XdpDbusRequestSkeleton *skeleton = XDP_DBUS_REQUEST_SKELETON (object);
  GList      *connections, *l;
  GVariant   *signal_variant;

  connections = g_dbus_interface_skeleton_get_connections (G_DBUS_INTERFACE_SKELETON (skeleton));

  signal_variant = g_variant_ref_sink (g_variant_new ("(u@a{sv})",
                                                      arg_response,
                                                      arg_results));
  for (l = connections; l != NULL; l = l->next)
    {
      GDBusConnection *connection = l->data;
      g_dbus_connection_emit_signal (connection,
                                     request->sender,
                                     g_dbus_interface_skeleton_get_object_path (G_DBUS_INTERFACE_SKELETON (skeleton)),
                                     DESKTOP_DBUS_IFACE ".Request",
                                     "Response",
                                     signal_variant,
                                     NULL);
    }
  g_variant_unref (signal_variant);
  g_list_free_full (connections, g_object_unref);
}
```
goblint didn't find the g_list_free_full, and I think it is supposed to?

src/gamemode.c
bilelmoussaoui: Well, the rule shouldn't move the defines, only the includes. I will add a test case for this; meanwhile, feel free to mark the rule as ignore if you don't like the output/feel like it won't produce any meaningful value

document-portal/file-transfer.c
bilelmoussaoui: Tracking ownership is something we can't easily do without making it produce a bunch of false positives so I would say ignoring it here would be fine :)

bilelmoussaoui:
> goblint didn't find the g_list_free_full, and I think it is supposed to?

It only reports the cases where it would use g_clear_list. What would be the suggestion you would expect for that snippet?

swick:
This should be: `g_autolist(GDBusConnection) connections = NULL;`. I know that the type of the list items is not easy to figure out, so this won't be auto fixable, but just warning that this should be g_autolist would be helpful.

bilelmoussaoui:
I just learned about g_autolist :-) i will add a rule for that

bilelmoussaoui:
> I just learned about g_autolist :-) i will add a rule for that

There is a rule for that now

document-portal/file-transfer.c
pwithnall: Tracking ownership is best left to [a static analyser plugin](https://cgit.freedesktop.org/tartan)

swick:
I think that regressed the use_g_autofree rule:
```
/workspace/document-portal/document-portal.c:827:11
                         XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
                         "Not enough permissions");
            return NULL;
  ...
        if (st_buf.st_dev == fuse_dev)
  ...
  >          g_autofree char *real_path = NULL;
            g_autofree char *id = NULL;
  ...
            /* The passed in fd is on the fuse filesystem itself */
  warning: Consider using g_autofree real_path to avoid manual g_free
  rule: use_g_autofree
```

swick:
I will not be adding more things to this PR. Let's get it in first, and then do a follow up when we have enough new violations on the latest goblint.

This should be good to merge, but I'd still like someone to take another quick look at it.

bilelmoussaoui:
> I think that regressed the use_g_autofree rule:
> 
> ```
> /workspace/document-portal/document-portal.c:827:11
>                          XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_NOT_ALLOWED,
>                          "Not enough permissions");
>             return NULL;
>   ...
>         if (st_buf.st_dev == fuse_dev)
>   ...
>   >          g_autofree char *real_path = NULL;
>             g_autofree char *id = NULL;
>   ...
>             /* The passed in fd is on the fuse filesystem itself */
>   warning: Consider using g_autofree real_path to avoid manual g_free
>   rule: use_g_autofree
> ```

Fixed

.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,29 @@ on:
...
+
+      - name: Run goblint
+        run: goblint --format sarif > goblint.sarif
+
+      - name: Upload SARIF results
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: goblint.sarif
+          category: goblint
+        if: always()
```
bilelmoussaoui: i think the if can be removed

.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,29 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/bilelmoussaoui/goblint:main-a7e87d6
+    permissions:
+      # Required for uploading SARIF results
+      security-events: write
+
+    steps:
+      - name: Check out xdg-desktop-portal
+        uses: actions/checkout@v4
```
bilelmoussaoui: should be v6 as that is the latest

.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,29 @@ on:
         type: string
 
 jobs:
+  goblint:
+    name: Goblint
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/bilelmoussaoui/goblint:main-a7e87d6
```
bilelmoussaoui, line 19: can you point it to main to avoid having to bump it every now and then?

swick: There is value in bumping this only when we intent to bump it. If you add new lints, a PR suddenly has to take them into account as well otherwise.

I think it makes sense to do this eventually when changes slow down.

src/xdp-documents.c
swick: If you understand the logic it becomes clear. I think this is fine. If it turns out to be a problem in practice, I will yell.

document-portal/permission-db.c
```diff
@@ -71,14 +71,12 @@ static void initable_iface_init (GInitableIface *initable_iface);
 G_DEFINE_TYPE_WITH_CODE (PermissionDb, permission_db, G_TYPE_OBJECT,
                          G_IMPLEMENT_INTERFACE (G_TYPE_INITABLE, initable_iface_init));
 
-enum {
-  PROP_0,
-  PROP_PATH,
+typedef enum {
+  PROP_PATH = 1,
   PROP_FAIL_IF_NOT_FOUND,
-  LAST_PROP
```
AdrianVovk, line 76: Since part of the point of doing this is to turn on a compiler warning flag, I presume it's a non-default flag and so we should turn it on in meson?

AdrianVovk (approved):

document-portal/permission-db.c
bilelmoussaoui: Right, given that xdp uses too little properties it might make sense to do it manually but i plan to make the linter automatically fix your set_property for read-only properties or get_property for write-only properties and add the missing props there with assert_not_reached as that is what you would expect missing in the coverage and automatically remove the default case.

swick: > I presume it's a non-default flag and so we should turn it on in meson?

Added one more commit which does that.

PR #1977, opened 2026-04-20
ci: Split out pre-commit and various checks into their own jobs

swick:


.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,41 @@ on:
...
+
+    steps:
+      - name: Check out xdg-desktop-portal
+        uses: actions/checkout@v4
+
+      - name: Check POTFILES.in
+        run: .github/workflows/check-potfiles.sh
+
+      - name: Check Qt annotations
+        run: find -name "*.xml" | xargs -n1 /usr/lib/qt6/bin/qdbusxml2cpp
```
AdrianVovk, line 48: I would assume the reason to split this out is so all three checks run, even if the first check fails. But then here you still combine the POTFILES.in check with the Qt annotations check. Shouldn't this check be split out into its own job as well? Or am I misunderstanding the purpose of this patch in the first place?

.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,41 @@ on:
...
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ inputs.image }}
+
+    steps:
+      - name: Configure environment
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
```
AdrianVovk, line 23: An explanation of what this is about would be helpful too, at least in the commit message, since it's new

.github/workflows/build-and-test.yml
```diff
@@ -12,6 +12,41 @@ on:
...
+          echo XDG_DATA_DIRS=$GITHUB_WORKSPACE/tests/share:/usr/local/share:/usr/share | tee -a $GITHUB_ENV
+
+      - name: Check out xdg-desktop-portal
+        uses: actions/checkout@v4
+
+      - name: Run pre-commit hooks
+        run: |
+          export PYTHONPATH="/root/.local/lib/python$(python3 -c 'import sys; print("{}.{}".format(*sys.version_info))')/site-packages:$PYTHONPATH"
+          export PATH="/root/.local/bin:$PATH"
```
whot, line 32: isn't this superfluous, I don't see anything being installed into those paths? I wonder if this is some leftover from an earlier incarnation.

swick: We use `pip install --user` in the Containerfile, so this is unfortunately still needed.

.github/workflows/build-and-test.yml
swick: This already existed before. Added a comment.

.github/workflows/build-and-test.yml
swick: The reason to split this up was just separation of concerns. Adding more steps to the one job is making it harder to follow. Good point about wanting to run all checks even if one fails. Made it one job where the checking steps have `if: always()`.

AdrianVovk (approved):
LGTM

PR #1978, opened 2026-04-20
build: Move gvdb to a meson subproject

swick:
We already started using meson subprojects for libglnx, another copylib we depend on, so there isn't a good reason to continue using git subtree for another copylib dependency.


src/meson.build
```diff
@@ -153,6 +153,7 @@ common_deps = [
   gio_unix_dep,
   json_glib_dep,
   libglnx_dep,
+  gvdb_dep,
```
AdrianVovk, line 156: you added this as a common dependency, so now we're linking in gvdb into main poral binary where previously we only linked it into the document portal

subprojects/gvdb.wrap
```diff
@@ -0,0 +1,8 @@
+[wrap-git]
+directory = gvdb
+url = https://gitlab.gnome.org/GNOME/gvdb.git
+revision = c6f2359cc1d00f16e0a0e2527fa0bc1882b8b5ab
+depth = 1
+
+[provide]
+dependency_names = gvdb
```
AdrianVovk, line 8: I'm pretty sure you don't need this declaration here:

1. You're already going to provide a `gvdb` dependency because the file is named `gvdb.wrap`

> A wrap file named foo.wrap implicitly provides the dependency name foo even when the [provide] section is missing.

2. You're pulling in the dependency via `subproject()` and `get_variable()` anyway. Using the dependency name here is only for automatically including a fallback on `dependency()` calls.

My vote is to use `dependency()`, just because fundamentally there's no reason to treat glnx and gvdb any differently from any other dependency. They both _could_ work as a shared object and by using `dependency()` we'd seamlessly transition to using them as shared objects if the time ever comes to do so.

meson.build
```diff
@@ -132,6 +140,7 @@ libsystemd_dep = dependency('libsystemd', required: get_option('systemd'))
 gudev_dep = dependency('gudev-1.0', required: get_option('gudev'))
 umockdev_dep = dependency('umockdev-1.0', required: get_option('tests'))
 libglnx_dep = libglnx.get_variable('libglnx_dep')
+gvdb_dep = gvdb.get_variable('gvdb_dep')
```
AdrianVovk, line 143: It's not super obvious for me that gvdb is going to be a static library (or even just a set of source files). Ditto with libglnx... Though I'm not sure there's anything we should do about it. Theoretically we could community our intent here by passing in `default_library=static` as a default option? IDK up to you, feel free to just resolve this.

AdrianVovk (changes requested):

subprojects/gvdb.wrap
swick: glnx doesn't override the dependency. I'll try to change that upstream. Changed gvdb to use `dependency()` though.

meson.build
swick: Added `default_library=static`. It does make the intent more clear.

subprojects/gvdb.wrap
swick: https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/77

AdrianVovk: Is this MR actually necessary though? Seems like if you use a .wrap file the dependency override happens automatically

swick: Yes, it is necessary. How else would it know what overrides what?

A project isn't a dependency. There can be arbitrarily many dependencies and other objects in a project.

AdrianVovk: Because in the wrap file you can have a `[provide]` section that tells meson "any time I ask for these dependencies, pull in this wrap file instead". [Docs](https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section)

And then you read the docs a little further and it says:

> A wrap file named foo.wrap implicitly provides the dependency name foo even when the [provide] section is missing.

Thus: if you pull in a subproject via `foobar.wrap` then you can just say `dependency('foobar')` and it'll work

swick: Right, but what is `foobar`? That's not a meson object. The docs also say this:

> This mechanism assumes the subproject calls meson.override_dependency('foo-1.0', foo_dep) so Meson knows which dependency object should be used as fallback.

The part that I missed is that you can specify which dependency object overrides a `dependency()` invocation in the parent project:

```
[provide]
libglnx=libglnx_dep
```

swick: And that snippet does indeed work. Added another commit.

bilelmoussaoui (approved):

subprojects/gvdb.wrap
AdrianVovk: > This mechanism assumes the subproject calls meson.override_dependency('foo-1.0', foo_dep) so Meson knows which dependency object should be used as fallback

Totally missed that :open_mouth:

PR #1981, opened 2026-04-27
Screenshot: Add target selection API

bhack:
## Summary

This picks up the Screenshot target-selection API work from the closed PR #1415 and addresses the review feedback from that thread.

- add Screenshot portal version 3 with a `target` option
- add public and implementation `AvailableTargets` properties
- validate requested targets before forwarding them to v3 backends
- preserve v2 backend compatibility by not forwarding `target` to v2 implementations
- add integration coverage for valid, invalid, unavailable, and v2-compatible target handling

## API Notes

`target` is a single requested target value, not a bitmask. `AvailableTargets` is the advertised bitmask of backend-supported targets.

Supported values are:

- `1`: screen
- `2`: window
- `4`: area

If `target` is omitted, the portal preserves previous Screenshot behavior. Public Screenshot version is capped at version 3, and v2 backends continue to expose public version 2.

This is intended as the API plumbing for the target-selection part of #1065. It does not fully resolve broader screenshot-tool parity requests such as delay, pointer visibility, decorations, or deterministic coordinate-region capture.

## Testing

- `git diff --check`
- `xmllint --noout data/org.freedesktop.impl.portal.Screenshot.xml data/org.freedesktop.portal.Screenshot.xml`
- `pre-commit run --show-diff-on-failure --color=always --all-files` in the project CI container
- `meson test -C /tmp/build-xdp-review-final integration/screenshot --print-errorlogs` in the project CI container
- full suite previously run in the project CI container: 26 passed, 1 skipped, 0 failed

### DIFF
--- a/data/org.freedesktop.impl.portal.Screenshot.xml
+++ b/data/org.freedesktop.impl.portal.Screenshot.xml
@@ -27,7 +27,7 @@
 
       This screenshot portal lets sandboxed applications request a screenshot.
 
-      This documentation describes version 2 of this interface.
+      This documentation describes version 3 of this interface.
   -->
   <interface name="org.freedesktop.impl.portal.Screenshot">
     <!--
@@ -51,6 +51,20 @@
 
           Hint whether the dialog should offer customization before taking a screenshot.
           Defaults to no.
+
+        * ``target`` (``u``)
+
+          The screenshot target. It is a single target value, not a bitmask, and
+          must be one of the targets advertised in AvailableTargets. If this option
+          is omitted, the backend should preserve the previous Screenshot
+          behavior. Available targets are:
+
+          - ``1``: Screen: The entire screen.
+          - ``2``: Window: A window selected by the user.
+          - ``4``: Area: An area selected by the user.
+          - ``8``: Active Window: The currently active window.
+
+          This option was added in version 3 of this interface.
 
         * ``permission_store_checked`` (``b``)
 
@@ -104,6 +118,19 @@
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
 
+    <!--
+        AvailableTargets:
+
+        A bitmask of available screenshot targets. Available targets are:
+
+        - ``1``: Screen: The entire screen.
+        - ``2``: Window: A window selected by the user.
+        - ``4``: Area: An area selected by the user.
+        - ``8``: Active Window: The currently active window.
+
+        This property was added in version 3 of this interface.
+    -->
+    <property name="AvailableTargets" type="u" access="read"/>
     <property name="version" type="u" access="read"/>
   </interface>
 </node>
--- a/src/screenshot.c
+++ b/src/screenshot.c
@@ -26,6 +26,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
@@ -43,6 +44,14 @@
 #include "xdp-portal-config.h"
 #include "xdp-request.h"
 #include "xdp-utils.h"
+
+typedef enum
+{
+  SCREENSHOT_TARGET_SCREEN = 1u << 0,
+  SCREENSHOT_TARGET_WINDOW = 1u << 1,
+  SCREENSHOT_TARGET_AREA = 1u << 2,
+  SCREENSHOT_TARGET_ACTIVE_WINDOW = 1u << 3,
+} ScreenshotTarget;
 
 typedef struct _Screenshot Screenshot;
 typedef struct _ScreenshotClass ScreenshotClass;
@@ -185,9 +194,52 @@ screenshot_done (GObject *source,
   g_task_run_in_thread (task, send_response_in_thread_func);
 }
 
-static XdpOptionKey screenshot_options[] = {
+static gboolean
+validate_screenshot_target (Screenshot *screenshot,
+                            uint32_t    target,
+                            GError    **error)
+{
+  uint32_t available_targets;
+
+  switch (target)
+    {
+    case SCREENSHOT_TARGET_SCREEN:
+    case SCREENSHOT_TARGET_WINDOW:
+    case SCREENSHOT_TARGET_AREA:
+    case SCREENSHOT_TARGET_ACTIVE_WINDOW:
+      break;
+    default:
+      g_set_error (error, XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   "Invalid screenshot target %x", target);
+      return FALSE;
+    }
+
+  available_targets =
+    xdp_dbus_screenshot_get_available_targets (XDP_DBUS_SCREENSHOT (screenshot));
+  if (!(available_targets & target))
+    {
+      g_set_error (error, XDG_DESKTOP_PORTAL_ERROR, XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
+                   "Unavailable screenshot target %x", target);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+validate_screenshot_target_filter (const char  *key,
+                                   GVariant    *value,
+                                   GVariant    *options,
+                                   gpointer     user_data,
+                                   GError     **error)
+{
+  return validate_screenshot_target (user_data, g_variant_get_uint32 (value), error);
+}
+
+static XdpOptionKey screenshot_options_v3[] = {
   { "modal", G_VARIANT_TYPE_BOOLEAN, NULL },
-  { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL }
+  { "interactive", G_VARIANT_TYPE_BOOLEAN, NULL },
+  { "target", G_VARIANT_TYPE_UINT32, validate_screenshot_target_filter }
 };
 
 static gboolean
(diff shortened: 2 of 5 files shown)


data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -104,6 +117,18 @@
...
+
+        A bitmask of available screenshot modes. Available modes are:
+
+        - ``1``: Screen: The entire screen.
+        - ``2``: Window: A window selected by the user.
+        - ``4``: Area: An area selected by the user.
+
+        This property was added in version 3 of this interface.
+    -->
+    <property name="AvailableModes" type="u" access="read"/>
```
swick: Can we call this something else? Mode is not really descriptive of what this is about. Maybe something like "target"?

src/screenshot.c
```diff
@@ -190,6 +198,72 @@ static XdpOptionKey screenshot_options[] = {
...
 
+static gboolean
+validate_screenshot_mode (Screenshot *screenshot,
+                          uint32_t    mode,
+                          GError    **error)
+{
+  uint32_t available_modes;
+
+  if (__builtin_popcount (mode) != 1 ||
+      (mode & ~SCREENSHOT_MODE_ALL) != 0)
```
swick: Not sure about this. If we changed the SCREENSHOT_MODE definitions into an enum and just switched on it, then we won't need the builtin and the SCREENSHOT_MODE_ALL.

src/screenshot.c
```diff
@@ -390,11 +478,20 @@ handle_screenshot (XdpDbusScreenshot *object,
...
 {
+  Screenshot *screenshot = (Screenshot *) object;
   XdpRequest *request = xdp_request_from_invocation (invocation);
   g_autoptr(GTask) task = NULL;
+  g_autoptr(GError) error = NULL;
 
   g_debug ("Handle Screenshot");
 
+  if (screenshot->impl_version >= 3 &&
+      !validate_screenshot_mode_option (screenshot, arg_options, &error))
```
swick: Why is this not using `xdp_filter_options` like above?

tests/test_screenshot.py
```diff
@@ -19,6 +19,15 @@
     signature="sv",
 )
 
+SCREENSHOT_MODE_SCREEN = 1
```
swick: from enum import Flag

tests/test_screenshot.py
```diff
@@ -46,8 +55,132 @@ def set_permission(self, dbus_con, appid, permission):
...
+        properties_intf = dbus.Interface(
+            xdp.get_xdp_dbus_object(dbus_con), "org.freedesktop.DBus.Properties"
+        )
+        available_modes = properties_intf.Get(
+            "org.freedesktop.portal.Screenshot",
+            "AvailableModes",
+        )
+        assert int(available_modes) == (
+            SCREENSHOT_MODE_SCREEN | SCREENSHOT_MODE_WINDOW | SCREENSHOT_MODE_AREA
+        )
```
swick, line 80: Put this in its own test case. `test_version` is really only about the interface version.

tests/test_screenshot.py
```diff
@@ -19,6 +19,15 @@
     signature="sv",
 )
 
+SCREENSHOT_MODE_SCREEN = 1
+SCREENSHOT_MODE_WINDOW = 2
+SCREENSHOT_MODE_AREA = 4
+SCREENSHOT_MODES = (
```
swick: I usually define good and bad cases, so I would call this SCREENSHOT_MODES_GOOD and add `SCREENSHOT_MODES_BAD = (0, 3, 8)`

src/screenshot.c
```diff
@@ -363,9 +418,23 @@ handle_screenshot_in_thread_func (GTask *task,
 
   xdp_request_set_impl_request (request, impl_request);
 
-  xdp_filter_options (options, &opt_builder,
```
swick, line 366: So, the filtering here is just weird. All of this should be done as early as possible, making it possible to return a proper error to the caller. `handle_screenshot_in_thread_func` only gets called from `handle_screenshot` so that's where you should do all the input validation.

src/screenshot.c
```diff
@@ -363,9 +418,23 @@ handle_screenshot_in_thread_func (GTask *task,
...
+                               screenshot, &error))
+        {
+          g_warning ("Invalid screenshot options: %s", error->message);
+          send_response (request, 2, g_variant_builder_end (&opt_builder));
+          return;
+        }
+    }
+  else
+    {
+      xdp_filter_options (options, &opt_builder,
```
swick: I think always calling `xdp_filter_options` with `screenshot_options_v3` should be fine. Well behaving clients with version < 3 didn't put `target` in there.

swick (approved):
Needs squashing down and a bit more explanation in the commit message, but otherwise this LGTM.

swick:
Required acks from KDE and GNOME. I'm going to ack from the GNOME side if no one else has objections (@jadahl, @GeorgesStavracas). @Sodivad can you ack for the KDE side?

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -52,6 +52,19 @@
...
 
+        * ``target`` (``u``)
+
+          The screenshot target. It is a single target value, not a bitmask, and
+          must be one of the targets advertised in AvailableTargets. If this option
+          is omitted, the backend should preserve the previous Screenshot
+          behavior. Available targets are:
+
+          - ``1``: Screen: The entire screen.
+          - ``2``: Window: A window selected by the user.
```
JakobDev, line 63: I think we should also add a target for the current active window. This is the option I mostly use when taking Screenshots and I also have a few scripts that do this.

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -52,6 +52,19 @@
...
+          - ``4``: Area: An area selected by the user.
```
Sodivad, line 64: If it's not a bitmask, why is 3 omitted?

swick: Because it *is* a bitmask in the property. Maybe there is a better way to word this.

Sodivad: Ah right for `AvailableTargets` we need it as a bitmask.

Sodivad:
Acked by KDE

data/org.freedesktop.impl.portal.Screenshot.xml
bhack: Added this as `Active Window` target value `8`, and updated both public/backend XML, frontend validation, the mock backend `AvailableTargets`, and integration coverage. It remains advertised through `AvailableTargets`, so backends can opt in when they support focused-window capture.

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -27,7 +27,7 @@
```
swick, line 27: The commit message needs to be wrapped at 80 chars.

Photon89:
Hi guys,

I'm involved in Shutter's development and I really have to restrain myself from writing in caps, because this PR would really be a game changer for us and various other screenshot capturing tools after so many years of struggling with Wayland support! 

As far as I can see, this PR already enables capturing any open window, the currently active window, a rectangular selection and of course the whole screen.

This would already massively improve what is available to date, so HUGE thanks for working on this!

Some more thoughts (which shouldn't affect merging this PR and may or may not be addressed at some point in the future):

1. Shutter (and most probably most other screenshot tools) has an option to capture the mouse pointer (or not to capture it). It is some kind of modifier to all targets and would require to read the cursor coordinates at capture time.
2. For selection mode, Shutter has the option to provide a default selection geometry (position of the top left corner and height/width), for example to caption a given portion of the screen from a CLI call.
3. Shutter has the option to add a delay option of n seconds. The delay starts after the user gives all necessary input (clicks on the desired window, confirms the desired selection etc.). So Shutter would either need to access the API twice (for getting the screen state to get user input and for the actual capture after the delay), or maybe it could supply a delay modifier to do everything in one go.

Again, please don't let yourself slow down by my comment, I just thought that it might be useful to already think about these things while defining the API.

bhack:
@swick Do you think we could include also these points?

bhack:
@Photon89 Thanks, these are useful requirements from screenshot-tool use cases. I agree they are worth exploring, but I would prefer not to block or expand this PR, since the current target-selection API is already focused and reviewed.

To make the possible follow-up shape concrete, I opened a draft prototype against this branch in my fork:

https://github.com/bhack/xdg-desktop-portal/pull/1

It experiments with a possible v4 API for:

- `include_cursor`
- `capture_delay`, with the delay owned by the backend after any user selection
- `selection_geometry`, as a suggested default area geometry

This is not meant as a required change for this PR, just a reviewable hypothesis for the follow-up API design if maintainers want to discuss these modifiers.

Photon89:
Of course, absolutely no hurry, it will take a lot of time for us to adapt the v3 functionality anyway! Thanks for having a look at it!

Photon89:
@swick Are there any news from Gnome regarding this? Maybe you could link to the corresponding issue? Thanks! :smiley:

bhack:
> @swick Are there any news from Gnome regarding this? Maybe you could link to the corresponding issue? Thanks! 😃

I have a draft gnome PR let me know if it is required but I suppose we need to wait an official portal release tag or not?

Photon89:
I don't know what the status at Gnome end is, thus asking. I tried to use the new v3 API in a Fedora VM (which ships xdg-desktop-portal 1.21.2) but neither Gnome nor KDE supported the newly defined targets so far.

PR #1982, opened 2026-04-28
trash: Fix up trashing files, part 2

swick:
/cc @smcv

### DIFF
--- a/src/trash.c
+++ b/src/trash.c
@@ -186,14 +186,26 @@ ignore_trash_mount_fd (XdpAppInfo *app_info,
         return FALSE;
 
       if (strstr (mount_options, "x-gvfs-notrash") != NULL)
-        return TRUE;
+        {
+          g_debug ("Ignoring the trash dir, because mount options include x-gvfs-notrash");
+          return TRUE;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/trash.c
```diff
@@ -561,20 +561,60 @@ open_parent (int          fd,
 }
```
jadahl, line 573: > We need to mount id of the host
namespace to find the appropriate trash location.

Something wrong with the grammar here.

jadahl (approved):

peteruithoven:
What I understand is that the security fix was backported, but this fix was not. Which causes issues on Ubuntu 24.04.4. So I've opened: https://bugs.launchpad.net/ubuntu/+source/xdg-desktop-portal/+bug/2166546

peteruithoven:
A fix has been proposed and is available through: ppa:ubuntu-security-proposed/ppa
On Ubuntu Noble (elementary OS 8.1 Built on Ubuntu 24.04.5 LTS) it fixes the issue. 
More info: https://bugs.launchpad.net/ubuntu/+source/xdg-desktop-portal/+bug/2166546

PR #1988, opened 2026-04-30
ci: Update to Ubuntu 26.04 as image base

tytan652:
Now that Ubuntu 26.04 is released, ubuntu:latest is (to be) swapped to use it.
To avoid breakage the tag is changed to explicitly use it.

- Dropped RST generator glib2 workaround
- Make Qt annotations check only the data directory, otherwise gresource will also be parsed.

### DIFF
--- a/.github/workflows/Containerfile
+++ b/.github/workflows/Containerfile
@@ -2,7 +2,7 @@
 # When this file is changed, or one needs to rebuild the image for another
 # reason, bump the `IMAGE_TAG` in the container.yml workflow.
 
-FROM ubuntu:latest
+FROM ubuntu:26.04
 
 RUN apt update
 RUN apt upgrade -y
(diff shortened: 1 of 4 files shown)


jadahl (approved):

swick:
@tytan652 you have to enable the OCI registry in your fork and make sure the action run there to get the CI to pass (we can't allow PRs to build images into the flatpak org registry).

tytan652:
It is enabled but I don't think @jadahl has permission to touch my registry.

Le 5 mai 2026 22:16:41 GMT+02:00, Sebastian Wick ***@***.***> a écrit :

swick:
Mind force pushing to your repo then?

tytan652:
Okay, here what happened: The workflow uses the actor's registry and on a push or pr the actor is the one who pushed.

So if the container is not on the pusher registry, it will fail on the PR.

swick:
Updated to image tag to today

swick:
@tytan652 well, it tries to push to `flatpak/xdg-desktop-portal` so you'll have to do it instead :(

PR #1989, opened 2026-05-01
Repo-wide changes

swick:
I am planning to do a stable release soon, after which it's the perfect time to do some larger, repo-wide changes:

* Make it REUSE compliant
* Move shared code to a new `shared/` dir
* Rename `src/` to `desktop-portal`

@jadahl @AdrianVovk @Sodivad


.github/workflows/build-and-test.yml
```diff
@@ -54,7 +54,7 @@ jobs:
     container:
       image: ${{ inputs.image }}
       env:
-        CFLAGS: -Wp,-D_FORTIFY_SOURCE=2
+        CFLAGS: -Wp,-D_FORTIFY_SOURCE=2,-Werror
```
jadahl: `meson setup .. -Dwerror=true`

document-portal/document-portal-fuse.c
```diff
@@ -24,7 +24,7 @@
 #include <glib/gprintf.h>
```
jadahl, line 24: Should add this commit to `.git-blame-ignore-revs`.

data/org.freedesktop.background.Monitor.xml
```diff
@@ -1,23 +1,7 @@
 <?xml version="1.0"?>
 <!--
- Copyright (C) 2022 Endless OS Foundation, LLC
```
jadahl, line 3: While this is AFAIK technically fine to remove, I do think it has more value than just being about copyright from a legal point of view. I'm not objecting, but my preference is to leave these as is, but I'm interested in others opinion.

swick:
(Removed a bunch of other commits again to keep it focused)

data/org.freedesktop.background.Monitor.xml
swick: I'm going ahead and merge this as is because no one seems to care. It's still in the history if we want to revisit this. Feel free to ping people on matrix or something.

Issue #304, opened 2019-02-22
Add a portal to see currently open windows


johan-bjareholt:
I am a maintainer of the ActivityWatch project and we are polling the currently focused windows name and title. This is not an issue under Xorg but on Wayland this is a problem as there is no common API between compositors. I have discussed this shortly with both a wlroots and Gnome developer and they both seem to agree that exposing such data would be best solved by adding a xdg-desktop-portal API for this.

Wlroots and KWin already have APIs for this (gnome-shell too but it's disabled by default) but they are all different, so a xdg-desktop-portal API would significantly simplify things.

Suggestions on properties for windows, methods and signals that would be good to have:

Window properties:
- app_id (string)
- title (string)
- focused (bool)

Signals:
- FocusedApplicationChanged
- RunningApplicationsChanged

Methods:
- GetRunningApplications (array of windows)

Links to prior discussions with Gnome and wlroots developers:
- https://gitlab.gnome.org/jadahl/gnome-shell/commit/a31e33d5813dfe7c77c1a3a58c9f107d5242b1cb#note_339241
- https://gitlab.gnome.org/GNOME/gnome-shell/issues/983
- https://www.reddit.com/r/linux/comments/as1dd0/we_are_the_sway_wlroots_developers_ask_us_anything/egskqq9/

TingPing:
It sounds very niche. What other applications would use this portal?

johan-bjareholt:
Discord is another example, but another use-case would be for panels which can support multiple compositors (maybe not a good use-case for flatpak, but a use-case non the less). There are also competitors to ActivityWatch which are not open source (most notably RescueTime) which would have the same use-case as us.

I guess it's kind of niche, but not extremely niche. Sooner or later some API like this would be needed for wayland and to me it seems like xdg-desktop-portal is the best fit.

matthiasclasen:
This is not the kind of information we would normally want to leak into sandboxes, I think.

johan-bjareholt:
@matthiasclasen Neither should screenshots by default as that contains the same information and more, but you need to explicitly allow permissions for that right?

matthiasclasen:
A screenshot is an explicit operation. You take a screenshot of an individual window, present the result to the user, and ask him: "share this with "SpyApp deluxe" ?" 

What you are asking for is hard to present to the user in a meaningful form.

johan-bjareholt:
[...]

PR #1992, opened 2026-05-04
Screenshot: Prototype capture modifiers

bhack:
This prototypes a v4 extension for the Screenshot portal API that forwards additional capture modifiers to backends that advertise Screenshot implementation version 4.

Changes included:

- document public and implementation Screenshot v4 options for include_cursor, capture_delay, and selection_geometry
- forward the new options only to v4 backends while keeping v3 and older backend behavior unchanged
- validate capture_delay bounds and require selection_geometry to be used with the Area target and nonzero size
- extend the screenshot mock template and integration coverage for forwarding, filtering, and invalid values

Compatibility notes:
Backends below version 4 keep seeing only the existing pre-v4 option set. The public portal version is capped to the backend-supported API version, now up to 4.


Photon89:
@bhack Thanks for the work that went into v3 and now also v4!

@DarthGandalf, @vadi2: Please have a look, maybe there are more things which I missed!

Play-InTheWay:
Another option useful alongside enabling or disabling the cursor or having a delay, is including window borders with or without shadow; those options are especially used by Spectacle but are also an option for the interactive screenshot portal.
image
image

bhack:
@Play-InTheWay Thanks, that makes sense as another screenshot-tool modifier, especially for window and active-window targets.

I think this probably needs a bit of backend/desktop agreement before adding it here, because decorations and shadows can mean different things depending on client-side/server-side decorations and compositor effects.

Play-InTheWay:
I was checking up on options around screenshotting apps, and for the most part, superficially and without the technical knowledge, I would say that with this MR most of the things are there; however, I have some questions or bring light to some considerations about the targets:

1. With the “screen” target, I guess you mean the current active screen, but it may be worth it to also allow a target to select a monitor?

2. I have some doubts about the area target, because with the current non-interactive portal, it is possible to do an area screenshot; the app just asks for a full-screen screenshot, and then it is cropped by the user with the app's own UI. This is useful because it allows putting overlays, like magnifying glass, rulers, moving the selection, resizing, etc., like Flameshot does today. So I wonder if that would be possible with the specific area target or if it has some benefits over the current approach; for me, it seems to be a distinction more from the app side.

3. With the “Window” target, it will allow selecting the current open windows and screenshotting them? If not, would it need an additional [portal](https://github.com/flatpak/xdg-desktop-portal/issues/304) or a [Wayland protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/187)?

4. I have noticed that Shutter has even more targets, like by workspace, menus, and tooltips. Those are needed, or another approach could be taken to take those kinds of screenshots? @Photon89

Photon89:
Very good points, I knew, I'd miss something!

1. When using Compiz, Shutter shows a dropdown with available viewports (virtual desktops) and lets the user choose any of them for capturing. As far as I understand, it just switches to the specified viewport and does a full screen capture. But when using Xfwm (and probably other WMs), this functionality doesn't work. I think, the viewports dropdown has been developed in a time when Compiz was popular, and doesn't support other WMs so far. But of course, if xdg-desktop-portal can allow to access a specific virtual desktop in a desktop-agnostic way, this would revive this functionality and make it universally available, which would be more than is currently possible in X11.
2. I think, Flameshot has been built around the limitation that only full screen capture is possible. So, most probably, in its current state it won't make use of the now available new targets and still perfectly function by accessing full screen capture only. But on the long run it could make use of the new targets and implement new functionality. So I'd consider it a new chance, not an issue.
3. As far as I can see, it is already possible in v3 with target number 8: https://github.com/flatpak/xdg-desktop-portal/pull/1981/changes
4. Regarding workspaces, see 1. Regarding menus and tooltips: Indeed, that's what I forgot. But I'm not sure, how it is currently implemented, whether it is some X11 feature that Shutter is using or some magic that Shutter does internally. I'll try to find some time to look into Shutter's code and report back!

edit: Regarding window borders and shadows: That's actually another good catch! Maybe it would be possible to discriminate between "window content only" and "window content + decoration (+ shadow if available)", so, to have only two instead of three options, if shadows are the part causing trouble?

Play-InTheWay:
1. I think I may explain it somewhat badly; what I basically wanted to propose there was a target for different monitors. It's not that needed, as you can always change the focus fast and take the screenshot, but nonetheless it is a step less. About the workspaces, I would respond in the 4 point.

2. I wasn't concerned that Flameshot would stop working with the new modes; but that I do not see the new area target as really needed, because the screenshot tools like Flameshot can already do area selection with the current non-interactive approach, and I suspect that depending on the implementation, the area target could be worse than what we have today, as it may offer fewer features (no overlays).

3. Yes, the active window selection is perfectly feasible, and with that you can do a lot. What I mean is about the window target, the 2. For example, for Shutter, it is about the function where you can screenshot a window in the drop menu directly from Shutter; my doubt is that in order to do that, you would need to offer through the portal the information of all the active windows, their names, and icons, but it seems somewhat out of the scope of just one portal.

4. Actually with Shutter, when I tried it today, it worked perfectly fine on Plasma; it even detected the name of the workspace, and yes, what it does is that it changes the focus to that workspace and makes the screenshot. That exact behavior wouldn't be possible on Wayland, but as you mention, with the portals it could be better and more reliable.
About the menus and tooltips, it is kind of a hit or miss; it depends a lot on the toolkit and app, but as with the workspaces, maybe with the portals it will be possible to make them work more reliably.

5. I don't know a lot about the portals, and others could answer that question better, but as I understand it, it would be something to be solved by the portal backend devs.

Play-InTheWay:
Going back to the main point of this MR, another possible modifier that I have thought about is offering the option to not freeze the screen at the moment of doing the screenshot.

Other would be about the possibility of remembering and using the last used area, but it probably would be possible with the default selection geometry already in the MR.

bhack:
I think some of the points raised here are valid follow-ups, but probably need backend/desktop agreement before being added to the portal contract.

A few notes on the specific cases:

- Remembering the last area should already be covered by `selection_geometry`: the app can store the previous rectangle and pass it back as the suggested default for the area selector.
- The `area` target is not meant to replace tools like Flameshot doing full-screen capture plus their own rich overlay. Those apps can keep using the existing/full-screen path. The area target gives apps a backend/compositor-managed selection path when they do not want to implement that UI themselves.
- `window` is intended as user-mediated window selection, not as exposing a list of windows, names, or icons to the app. Window enumeration feels like a separate privacy-sensitive API/protocol question.
- Monitor capture, workspaces, menus/tooltips, decorations/shadows, and “do not freeze the screen” all sound like plausible future extensions, but I think they need more design input from backend maintainers before we bake names/semantics into this PR.

I can also clarify the XML docs a bit so `screen` does not read as “active monitor” and `window` does not imply app-visible window enumeration.

Play-InTheWay:
Thank you for the clarification; I understand it better.

- About the area target, maybe it will be worth it to clarify in the docs that it is meant for a quick backend-managed selection, and if it is wanted to include overlays like magnifying glass, text, rulers, crosshairs, tools, etc., the fullscreen target will be a better option.

- However, I still don't understand how the window selection can be done without exposing that information. Maybe it would be a selection where you have to manually click the window you want to capture (like the “select window” option in Spectacle), or it will be more like the screencast portal, where a backend-managed list of windows will appear?

For the others, well, I guess then we will have to wait until we receive some feedback, or those would be more for a v5?

Photon89:
> About the area target, maybe it will be worth it to clarify in the docs that it is meant for a quick backend-managed selection, and if it is wanted to include overlays like magnifying glass, text, rulers, crosshairs, tools, etc., the fullscreen target will be a better option.

I think that app devs will understand that. 

> However, I still don't understand how the window selection can be done without exposing that information. Maybe it would be a selection where you have to manually click the window you want to capture (like the “select window” option in Spectacle), or it will be more like the screencast portal, where a backend-managed list of windows will appear?

As far as I understand, the user will be presented with the possibility to click a window on his screen and the window he clicked on will be captured, all of which is managed by the backend. Is this correct, @bhack ?

bhack:
> As far as I understand, the user will be presented with the possibility to click a window on his screen and the window he clicked on will be captured, all of which is managed by the backend. Is this correct, @bhack ?

Yes, that is the intended model.
The app requests `target = Window`, but the actual selection is backend-managed and user-mediated. The app should not receive a list of windows, titles, icons, or window identifiers through this option. The backend may implement the user interaction however fits the desktop: for example clicking a window directly, showing a trusted backend-owned picker, or another desktop-specific UI.

The important part for the portal contract is that the app only gets the final screenshot result, not the window inventory or the selected window identity.

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -66,6 +66,32 @@
...
+
+          This option was added in version 4 of this interface.
+
+        * ``selection_geometry`` (``(iiuu)``)
+
+          Suggested default geometry for the area selector as x, y, width,
+          height in the compositor's logical coordinate space. Width and height
+          must be greater than zero. This option is only valid when ``target``
+          is set to ``4`` (Area), and implementations may ignore or adjust the
+          geometry.
```
jadahl: The "compositor logical coordinate space" isn't defined anywhere.

data/org.freedesktop.impl.portal.Screenshot.xml
```diff
@@ -66,6 +66,32 @@
...
+
+        * ``capture_delay`` (``u``)
+
+          Delay in seconds before the screenshot is captured. The delay starts
+          after any required user input, such as selecting a window or area,
+          has completed. The value must not be greater than 3600.
+
+          This option was added in version 4 of this interface.
+
+        * ``selection_geometry`` (``(iiuu)``)
```
jadahl: Where would this information come from? There is no "select area" portal functionality, nor should there be, I'd say. Perhaps this could be replaced with a "interactive-select-area" kind of flag?

Photon89:
@bhack I see, thanks for clarifying! Shutter indeed has both the user selection and also a dropdown list of open windows, but the latter one always seemed like a gimmick which nobody really uses to me, so not many will cry if we drop this feature.

data/org.freedesktop.impl.portal.Screenshot.xml
bhack: I dropped `selection_geometry` from this iteration, so we no longer need to define a coordinate space in the Screenshot API here.

data/org.freedesktop.impl.portal.Screenshot.xml
bhack: This was underspecified. The intent was to let clients suggest a remembered/default area, but without a portal-defined source for that geometry or coordinate space it is not a good contract. I dropped `selection_geometry`; if we want remembered/default area selection later, it should be designed separately with clearer semantics.

Photon89:
Another thought which I'm not sure whether it is possible:

In X Shutter can access the window name or the viewport number that it is capturing. This data is then used to to generate a suiting file name for the screenshot. Is this possible to implement, or is this information restricted?

Photon89:
Hi, just wanted to ask about the status of this PR, since it looks like it has accumulated some merge conflicts over the time. Would be a real pity if it won't make it and gets closed...

bhack:
I can rebase it eventually but I had no major review

swick:
For this to move forward, you need at least one backend implementation. I think the addition is fine, and I can't imagine KDE having a problem either. The only thing that might need a bit of work in the protocol is maybe making capture_delay optional somehow.

Photon89:
Sadly, the v3 API version which was introduced by #1981 has been implemented in `xdg-desktop-portal-xapp` only so far, as far as I know. But I am figuring, the earlier the API gets updated, the earlier the implementations will follow. So maybe the `xdg-desktop-portal-xapp` implementation can be used as basis to implement the capture modifiers.

bhack:
In any case I've resolved the conflict

bhack:
I’ve opened a draft GNOME backend implementation:
https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/269

It includes the existing v3 target-selection work and adds support for the v4 `include_cursor` and `capture_delay` options. It preserves the previous behavior when either option is omitted and applies the delay after area selection.

One backend limitation surfaced: GNOME Shell’s `ScreenshotArea` method has no cursor parameter, so `include_cursor=true` cannot currently be honored for area captures. The draft MR returns an explicit error for that combination rather than silently ignoring it. This may need some agreement on the portal semantics while the API is still under review.

I tested the backend against both the released `xdg-desktop-portal 1.21.1` interface and the v4 implementation XML from this PR.

Photon89:
Thanks for moving this forward! Some thoughts:

1. Is it maybe possible to query which modifiers the backend supports (similarly to how supported targets can be queried)? Then 3rd party tools can query the backend's capabilities and adjust their GUI accordingly (disable unsupported options). There even could be a return value for each sensible target/modifier combination, which would "solve" the Gnome include_cursor + area capture issue.
2.  Why was it necessary to drop the selection geometry modifier?
3. Another idea (not sure how realistic): An option for the area capture to enable/disable confirmation (confirmation enabled: selection stays there, user can modify it and then needs to confirm it for example by pressing the Enter key; confirmation disabled: screenshot is taken as soon as user release mouse button while drawing the selection).
4. Since Gnome doesn't seem to have intentions to move forward their backend implementation, it would be great if the draft PR at Gnome's Gitlab could lose its draft status at some point. :smiley:

bhack:
Being able to query modifier support per target makes sense, especially given the GNOME area/cursor limitation. A global list wouldn’t capture that distinction. I’d like to hear the maintainers thoughts on whether to add that discovery mechanism here.

I dropped `selection_geometry` following review as the coordinate space wasn’t defined, and it wasn’t clear where an application would obtain meaningful coordinates. 
Remembering an area could still be useful, but it needs a clearer contract.

The confirmation option sounds worth discussing separately. I’d prefer to keep this PR focused on cursor inclusion and capture delay.
The GNOME MR is a draft while we validate runtime behavior and discuss these limitations the intention is to move it forward once those points are addressed.

PR #1993, opened 2026-05-04
ScreenCast: Add audio stream support

### LINKED ISSUE #957: Screencast: send an audio stream to capture alongside video
Currently, afaik, the screencast portal only supports video capture. This means that when an application wants to capture audio as well, like Discord or a browser, it needs to have a separate dialog (afaict) and permissions to access all of PipeWire.

Allowing the screencast portal to send some audio from the captured application can make the process of screen sharing easier, as users don't need to select audio capture separately. This means that an end user can capture a game with OBS, for example, and it'll automatically give OBS the appropriate audio sources.

Nuhiat-Arefin:
Closes #957 
## Summary

  This adds frontend API support for requesting audio alongside video in the ScreenCast portal.

  The ScreenCast interface is bumped to version 6 and gains a new `audio` option in `SelectSources()`. When supported by the backend, `Start()` may now return multiple PipeWire streams for a selected source, with each stream carrying a `media_type` property of either `"video"` or `"audio"`.
## Details

  - Add an `audio` boolean option to `SelectSources()`.
  - Add `media_type` stream metadata returned by `Start()`.
  - Default streams without `media_type` to video for compatibility with existing backends.
  - Strip the `audio` option before forwarding requests to ScreenCast backends older than v6.
  - Track stream media type internally with `ScreenCastStreamType`.

### DIFF
--- a/data/org.freedesktop.portal.ScreenCast.xml
+++ b/data/org.freedesktop.portal.ScreenCast.xml
@@ -25,7 +25,7 @@
 
       The Screen cast portal allows to create screen cast sessions.
 
-      This documentation describes version 6 of this interface.
+      This documentation describes version 7 of this interface.
   -->
   <interface name="org.freedesktop.portal.ScreenCast">
     <!--
@@ -97,6 +97,20 @@
         * ``multiple`` (``b``)
 
           Whether to allow selecting multiple sources. Default is no.
+
+        * ``audio`` (``b``)
+
+          Whether to request audio streams associated with the selected sources.
+          Default is no.
+
+          If set to true, the application is capable of receiving audio streams
+          returned by the :ref:`org.freedesktop.portal.ScreenCast.Start` method.
+          The portal implementation may still return no audio streams if none are
+          available, if it cannot determine a suitable audio source, or if the
+          active portal backend implements an older interface version that does
+          not support audio.
+
+          This option was added in version 7 of this interface.
 
         * ``cursor_mode`` (``u``)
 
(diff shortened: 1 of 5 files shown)


data/org.freedesktop.portal.ScreenCast.xml
```diff
@@ -195,14 +209,27 @@
...
+          The array will contain streams for a single source if 'multiple' (see
           org.freedesktop.portal.ScreenCast.SelectSources())
-          was set to 'false', or at least one stream if
+          was set to 'false', or at least one source's streams if
           'multiple' was set to 'true' as part of the
           org.freedesktop.portal.ScreenCast.SelectSources() method.
 
+          Since version 6, a single selected source may be represented by
+          multiple media streams, such as one video stream and one or more audio
+          streams.
```
tytan652: I think it would be saner for API consumer to have only one audio stream/node to deal with per video stream.

tytan652 (commented):
Adding this without Audio portal to help find expected audio nodes will not be easy (I didn't said impossible).

Sodivad:
If I see correctly this pushes the burden of selecting the correct audio stream on to the backends? How involved is this?  I expect this to have corner cases that one needs to be aware of. Could this be shared somehow?

Nuhiat-Arefin:
> If I see correctly this pushes the burden of selecting the correct audio stream on to the backends? How involved is this? I expect this to have corner cases that one needs to be aware of. Could this be shared somehow?

Yes, the backend needs to select the correct audio stream. As discussed in the Matrix channel, cadubentzen is working on the backend implementation.

Sodivad:
> Yes, the backend needs to select the correct audio stream. As discussed in the Matrix channel, cadubentzen is working on the backend implementation.

But there are multiple backends? So we will duplicate the code everywhere?

cadubentzen:
Hi @Sodivad! I've been prototyping how this could look like on the -gnome and -kde backends but haven't settled on the approach to share just yet, hoping to do that soon with an RFC and draft MRs.

In any case, I agree with the sentiment that if we can make the backend implementations less involved and share the common logic to handle corner cases well and reduce duplication, we should.

PR #1995, opened 2026-05-04
Add more pre-commit checks, commit message check, update goblint linter, and move goblint to pre-commit

swick:
Draft because I'm investigating switching from pre-commit to prek.


document-portal/permission-db.c
```diff
@@ -68,6 +68,8 @@ typedef struct
 
 static void initable_iface_init (GInitableIface *initable_iface);
 
+/* cleanup defined in header
+ * goblint-ignore-next-line: missing_autoptr_cleanup */
 G_DEFINE_TYPE_WITH_CODE (PermissionDb, permission_db, G_TYPE_OBJECT,
```
bilelmoussaoui, line 71: this shouldn't trigger at all, please open an issue

bilelmoussaoui:
While at it, i would recommend running the dead_code rule, testing it locally here shows some dead code that can be removed.

document-portal/permission-db.c
bilelmoussaoui: Fixed in https://github.com/bilelmoussaoui/goblint/commit/77115366130a33731f637356c98b9d7a6e4c65c3

swick:
Weird, for me it doesn't show anything.

bilelmoussaoui:
> Weird, for me it doesn't show anything.

It requires a meson builddir. It will complain if you run with --verbose

swick:
Right, I get "Meson introspection not available - proceeding without public/private distinction" now, but I have a builddir called `build` which should get picked up, and even if I add `build_dir = "build"` it still shows the same message.

bilelmoussaoui:
Ah I assume you are using the container, which means we have to install meson inside of it as well given that we run `meson introspect --all`

bilelmoussaoui:
> Ah I assume you are using the container, which means we have to install meson inside of it as well given that we run `meson introspect --all`

if you pull again, it should work now

swick:
Not happy where the pre-commit stuff is going, so I pulled out the goblint things to #1996.

PR #1997, opened 2026-05-05
Move all checking to pre-commit in a container for CI and local

swick:
Probably for some time after the stable release. This gives us a single way to run checks in the CI and locally, while not having to depend on the host.

It also adds a bunch of linters and a commit message check, which should help with reducing repetitive review comments, and allows us to build more scripts to check things easily.


swick:
(note to myself: check out prek)

Issue #1862, opened 2025-11-12
RemoteDesktop - Request clarification of "persist_mode" behaviors


jordansissel:
### Operating System

N/A

### XDG Desktop Portal version

1.20

### Desktop Environment

GNOME

### Expected Behavior

(this is more a spec issue than an implementation/bug, but, I'll do my best to fill this form out)

A call to `SelectDevices` with `persist_mode` = `2` should persist until explicitly revoked. I do not explicitly revoke permission, so I should not be prompted a second time when using a previous `restore_token`.

### Current Behavior

A `restore_token` works until i log out and log back in, after which that same `restore_token` does not work correctly - instead, it prompts me again.

### Steps to Reproduce

Summary: Using a `restore_token` fails after logging out and logging back in.

1. call CreateSession
2. call SelectDevices with `persist_mode = 2`
3. call Start
4. save the `restore_token`
5. start again, call CreateSession
6. call SelectDevices with `restore_token` set to the previous value, and `persist_mode = 2`
7. call Start, it works.
8. Log out
9. log back in as the same user
10. repeat previous CreateSession + SelectDevices + Start with the same `restore_token` value
11. observe that the `restore_token` is ignored and that a prompt is shown instead of using the `restore_token`

### Anything else we should know?

The spec states:

> 2: Permissions persist until explicitly revoked

I did not explicitly revoke permission, but permission was revoked anyway.

I seek clarity on the specification here on what constitutes "explicitly revoked" or on which scenarios one should expect implicit revocation. Once I've got that information, I can decide what to do next - file a bug with gnome, change which interface my software is using, submit a PR to this spec to update the wording, etc.

Issue #1371, opened 2024-06-07
Screencast, persist mode does not keep the same monitor selection


fufesou:
### Operating System

Ubuntu 23.10

### XDG Desktop Portal version

1.18

### Desktop Environment

GNOME

### Expected Behavior

I'm using ScreenCast to get screenshot.

I have two monitors.

When I use the restore token in a persist mode, the monitor to share is not the one I've choosed. It should keep the same monitor to share.

### Current Behavior

https://github.com/flatpak/xdg-desktop-portal/assets/13586388/3598dd7d-c94b-4e2e-bd4f-abbac70324cb

You can see:
1. I've choosed the second monitor, and the screenshot is from what I choose.
2. But when I run the second time with the restore token, the screenshot is another monitor.

### Steps to Reproduce

Just test the following python script, it is mainly from https://gitlab.gnome.org/-/snippets/19

```python
#!/usr/bin/python3

import re
import signal
import dbus
import sys
from gi.repository import GLib
from dbus.mainloop.glib import DBusGMainLoop

import gi
gi.require_version('Gst', '1.0')
from gi.repository import GObject, Gst

DBusGMainLoop(set_as_default=True)
Gst.init(None)

loop = GLib.MainLoop()

bus = dbus.SessionBus()
request_iface = 'org.freedesktop.portal.Request'
screen_cast_iface = 'org.freedesktop.portal.ScreenCast'

restore_token = ''
if len(sys.argv) > 1:
    restore_token = sys.argv[1]

pipeline = None

def terminate():
    if pipeline is not None:
        self.player.set_state(Gst.State.NULL)
    loop.quit()

request_token_counter = 0
session_token_counter = 0
sender_name = re.sub(r'\.', r'_', bus.get_unique_name()[1:])

def new_request_path():
    global request_token_counter
    request_token_counter = request_token_counter + 1
    token = 'u%d'%request_token_counter
    path = '/org/freedesktop/portal/desktop/request/%s/%s'%(sender_name, token)
    return (path, token)

def new_session_path():
    global session_token_counter
    session_token_counter = session_token_counter + 1
    token = 'u%d'%session_token_counter
    path = '/org/freedesktop/portal/desktop/session/%s/%s'%(sender_name, token)
    return (path, token)

def screen_cast_call(method, callback, *args, options={}):
    (request_path, request_token) = new_request_path()
    bus.add_signal_receiver(callback,
                            'Response',
                            request_iface,
[...]

PR #2001, opened 2026-05-12
fix: remove premature delete_persistent_permissions before backend validation

herrschmidt:
## Problem

`restore_token` with `persist_mode=2` (PERSISTENT) fails after logout/login — the portal dialog reappears on every session restart. The token works within a session, but not across sessions.

Reported in:
- #1862 "restore_token works until i log out and log back in"
- rustdesk/rustdesk#9435 "persist_mode not working"

## Root Cause

In `xdp_session_persistence_replace_restore_token_with_data()`, the `restore_token` is deleted from the PermissionStore **before** the backend (e.g. gnome-screencast) validates the restore data. The critical code is in `src/xdp-session-persistence.c:232-237`.

## Fix

Remove the premature `delete_persistent_permissions()` call (3 lines). The delete is unnecessary because:
1. On successful restore, `Start()` generates a fresh token
2. `generate_and_save_restore_token()` already handles cleanup
3. The PermissionStore entry is overwritten on next `Start()` anyway

**Change:** 3 lines removed from `src/xdp-session-persistence.c`.

## Tested

- ✅ Ubuntu 26.04, GNOME/Wayland, RustDesk 1.4.6
- ✅ Within session: no dialog on reconnect
- ✅ Across session (logout → login): no dialog, instant connection
- ✅ Build: 138/138 successful

## Related

- #1371 "persist mode does not keep monitor selection"
- #638 Original screencast restore PR

### DIFF
--- a/src/xdp-session-persistence.c
+++ b/src/xdp-session-persistence.c
@@ -229,12 +229,6 @@ xdp_session_persistence_replace_restore_token_with_data (XdpSession *session,
                 xdp_session_persistence_get_persistent_permissions (session,
                                                                     table,
                                                                     restore_token);
-              if (restore_data)
-                {
-                  xdp_session_persistence_delete_persistent_permissions (session,
-                                                                         table,
-                                                                         restore_token);
-                }
             }
 
           if (restore_data &&
@@ -381,6 +375,16 @@ xdp_session_persistence_replace_restore_data_with_token (XdpSession *session,
   else
     {
       *in_out_persist_mode = XDP_SESSION_PERSISTENCE_MODE_NONE;
+
+      /* Clean up persistent permissions when the backend did not
+       * return restore_data. The token was kept in the store
+       * (no longer deleted early in replace_restore_token_with_data)
+       * — if the backend doesn't return restore_data, clean up here. */
+      if (*in_out_restore_token)
+        xdp_session_persistence_delete_persistent_permissions (session,
+                                                               table,
+                                                               *in_out_restore_token);
+
       g_clear_pointer (in_out_restore_token, g_free);
     }
 


Sodivad:
Can you expand more on why things fail?

> In xdp_session_persistence_replace_restore_token_with_data(), the restore_token is deleted from the PermissionStore before the backend (e.g. gnome-screencast) validates the restore data.

How does this cause restoration to fail?

herrschmidt:
Hi, this is the failure mechanism:

The bug is a two-stage **consume-before-validate** race:

**Stage 1 — Token consumed prematurely:**
When a client calls `SelectSources(persist_mode=2, restore_token="<uuid>")`, the function `replace_restore_token_with_data()` in `src/xdp-session-persistence.c` does three things:
1. Looks up the token in the PermissionStore (GVDB database)
2. **Immediately deletes it** (the `delete_persistent_permissions()` call on line 234)
3. Passes the `restore_data` to the backend for validation

**Stage 2 — Backend validation fails (silently):**
The backend (e.g. `xdg-desktop-portal-gnome`'s screencast implementation) calls `restore_stream_from_data()` → `find_monitor_by_string()`. This function matches a monitor identifier string against currently connected monitors. After a logout/login cycle:
- The GPU re-initializes
- Display connectors are re-enumerated
- Monitor identifier strings can change (different connector ID assignment, different EDID path)

So `find_monitor_by_string("SAM:Q95A:0x01000e00")` fails because the monitor is now `"SAM:Q95A:0x01001100"` or similar. The restore fails, the backend shows the "Select screen" dialog, and the user has to manually confirm.

**Why this breaks persistence across sessions:**
- Token was deleted in Stage 1 ✓
- Restore failed in Stage 2 ✗
- A **new** token is only generated when the session is successfully started AND closed (`generate_and_save_restore_token()` runs on `session.close()`)
- Since the session wasn't successfully restored (dialog appeared instead), no new token is written
- `persist_mode=2` silently degrades to `NONE`

Within a single session, this works because the monitor is still connected with the same identifier — `find_monitor_by_string()` succeeds, the session is properly restored, a fresh token is generated, and the cycle continues.

**The fix:** The delete is moved to after the backend response. If the backend successfully restores, the old token entry is simply overwritten. If it fails (or `persist_mode=NONE` is used), the cleanup happens in `replace_restore_data_with_token()`'s else-branch. I've updated the PR with this approach, it now properly handles both paths.

swick:
I think this change makes sense, /ping @whot @jadahl

src/xdp-session-persistence.c
```diff
@@ -381,6 +375,16 @@ xdp_session_persistence_replace_restore_data_with_token (XdpSession *session,
...
+
+      /* Clean up persistent permissions when the backend did not
+       * return restore_data. The token was kept in the store
+       * (no longer deleted early in replace_restore_token_with_data)
+       * — if the backend doesn't return restore_data, clean up here. */
+      if (*in_out_restore_token)
+        xdp_session_persistence_delete_persistent_permissions (session,
+                                                               table,
+                                                               *in_out_restore_token);
+
```
whot, line 387: ok, probably too niche to worry about but: if the restore_data comes from the transient permissions we now pass the token into persistent permissions to delete associated data. Assuming the token is unique enough this shouldn't be an issue but feels off anyway.

Also not immediately sure why the same issue doesn't affect transient permissions but I've paged much of this out by now.

whot (commented):
Note on the commit message: xdp doesn't use conventional commits, better teach the agent that.

swick:
Re-reading this, it also seems problematic that portal-gnome doesn't seem to handle the restore token correctly in find_monitor_by_string. Can you please report this as well?

Anyway, the commit message needs updating (right prefix, line length) and the comments are too verbose, the typical AI slop.

Issue #1983, opened 2026-04-29
Regression: D-Bus service does not start in MATE (1.21.1)


mcrha:
### Operating System

Fedora Linux 44

### XDG Desktop Portal version

1.21

### XDG Desktop Portal version (Other)

xdg-desktop-portal-1.21.1-1.fc44.x86_64

### Desktop Environment

Cinnamon / MATE / Xfce

### Expected Behavior

Service starts, I'm logged in MATE

### Current Behavior

Service fails to start, indicated also [in gtk4 apps](https://gitlab.gnome.org/GNOME/gtk/-/work_items/8181).
Looking into the journalctl from the time of an attempt to start the service I see this:
```
Apr 29 11:15:26 xxx systemd[1900]: graphical-session.target - Current graphical user session is inactive.
Apr 29 11:15:26 xxx systemd[1900]: Dependency failed for xdg-desktop-portal.service - Portal service.
Apr 29 11:15:26 xxx systemd[1900]: xdg-desktop-portal.service: Job xdg-desktop-portal.service/start failed with result 'dependency'.
```
which is kinda nonsense, because there are exported these:
```
XDG_CURRENT_DESKTOP=MATE
DISPLAY=:0
```

Packages:
```
xdg-desktop-portal-1.21.1-1.fc44.x86_64
systemd-259.5-1.fc44.x86_64
```

It was not the case before upgrade from Fedora 43 to Fedora 44, which used:
```
xdg-desktop-portal-0:1.20.4-1.fc43.x86_64
systemd-0:258.7-1.fc43.x86_64
```

### Steps to Reproduce

see above, the service does not start. I also used "double-click" on Session Bus->`org.freedesktop.portal.Desktop` in D-Feet, which errors as "NameHasNoOwner" and "startup job failed (3)" error message in the UI (D-Feel does not let me copy the error message, but it's kinda long to re-type here letter by letter).

mcrha:
I edited `/usr/lib/systemd/user/xdg-desktop-portal.service` and commented out these lines:
```
PartOf=graphical-session.target
Requisite=graphical-session.target
After=graphical-session.target
```
which feels like circular dependency to me (you cannot require, be part of and run after something, but I do not know systemd, thus whatever), which made it work...  until the next update. ;)

silentsonata:
I also have this same problem when running Hyprland on Fedora 44. Commenting out the same three lines resolves the problem for me as well.

KatsuroDev:
[...]

PR #2003, opened 2026-05-13
Revert Service Dependency Regression

KatsuroDev:
Revert dependency to `dbus.service` rather than `graphical-session.target`.

This fixes the issue when running on Hyprland for example, the dependency failed job causing the service to not run. #1983

Reverting commit 4d284de29d1d0740c9b80b634631a8f253287680

### DIFF
--- a/src/xdg-desktop-portal.service.in
+++ b/src/xdg-desktop-portal.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=Portal service
 PartOf=graphical-session.target
-Requisite=graphical-session.target
-After=graphical-session.target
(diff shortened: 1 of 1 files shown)


Erick555:
Similar problem was reported [before](https://github.com/flatpak/xdg-desktop-portal/pull/1941) and it was considered as [config issue](https://github.com/flatpak/xdg-desktop-portal/pull/1941#issuecomment-4109911265) in specific DE.

KatsuroDev:
I would assume #1983 will be closed for the same reason. Thanks for the clarity! I was hoping someone would help since reverting seemed weird.

PR #2004, opened 2026-05-16
xdp: resolve directories properly

mcatanzaro:
Since sometime last year, I have generally been presented with an error message whenever opening a file upload file chooser in Epiphany (for example, when trying to attach any file on GitHub). It looks like this:

Oops! Something went wrong.

Unable to find “/home/mcatanzaro/example.txt”. Please check the spelling and try again.

The problem lies in how the file chooser portal resolves document portal paths to host system paths. Say the user uses the document portal to open the aforementioned /home/mcatanzaro/example.txt. That is a host system path. The corresponding document portal path will look different, say /run/user/1000/doc/99bfa7fa/example.txt. Epiphany then trims off the final component and saves /run/user/1000/doc/99bfa7fa as the last opened directory. Later, the user tries to open a different file. Epiphany remembers the last upload directory and passes it to GtkFileDialog, which provides it to xdg-desktop-portal via the current_folder option. Finally, in handle_open_file(), xdg-desktop-portal looks at the current_folder option and resolves it from a document portal path to a host path. The result is, perhaps surprisingly, /home/mcatanzaro/example.txt. That's right: we provided a directory, but got back a regular file instead! Definitely not what the application wanted, and yet it's arguably the most reasonable behavior, because of course the path /run/user/1000/doc/99bfa7fa definitely corresponds to the regular file /home/mcatanzaro/example.txt, and it would be a little weird for it to resolve to just /home/mcatanzaro, right?

Perhaps we really should resolve the parent directory of a full document portal path to the parent on the host filesystem. I'm not sure. I'm definitely not smart enough to reason about what might possibly break if we were to try changing that. Fortunately, there's no need, because xdg-desktop-portal always knows whether it wants a folder or a file, so it can simply indicate what it wants.

This refines: d037b5c3f91b68ca208a9a41b6e18e6a3a659e05

### DIFF
--- a/src/file-chooser.c
+++ b/src/file-chooser.c
@@ -515,7 +515,8 @@ handle_open_file (XdpDbusFileChooser *object,
     if (current_folder)
       {
         const char *path_from_app = g_variant_get_bytestring (current_folder);
-        g_autofree char *host_path = xdp_resolve_document_portal_path (path_from_app);
+        g_autofree char *host_path = xdp_resolve_document_portal_path (path_from_app,
+                                                                       XDP_RESOLVE_DOCUMENT_TO_DIRECTORY);
 
         g_variant_builder_add (&options, "{sv}", "current_folder",
                               g_variant_new_bytestring (host_path));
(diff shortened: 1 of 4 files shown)


mcatanzaro:
I'm suspicious of the preexisting code in open-uri.c. It's resolving the same path twice, and I don't know why. Presumably that is a mistake?

swick:
> I'm suspicious of the preexisting code in open-uri.c. It's resolving the same path twice, and I don't know why. Presumably that is a mistake?

Maybe? Either way, this is a good improvement, so merging.

PR #2008, opened 2026-05-25
network-monitor: Connect to property notifications

jameswestman:
Most GNetworkMonitor implementations emit network-changed only when the set of available networks changes, not when the properties change. Add signal connections to ensure `changed` is emitted when one of the properties changes.

Also, use g_idle_add_full() to coalesce signal emissions, since one added/removed network could also trigger multiple property changes at the same time.

Related to [GLib #3963](https://gitlab.gnome.org/GNOME/glib/-/work_items/3963).

### DIFF
--- a/desktop-portal/network-monitor.c
+++ b/desktop-portal/network-monitor.c
@@ -23,6 +23,8 @@ struct _NetworkMonitor
   XdpDbusNetworkMonitorSkeleton parent_instance;
 
   GNetworkMonitor *monitor;
+
+  guint emit_source_id;
 };
 
 struct _NetworkMonitorClass
@@ -183,7 +185,7 @@ handle_can_reach (XdpDbusNetworkMonitor *object,
       g_autoptr(GSocketConnectable) address = NULL;
 
       address = g_network_address_new (hostname, port);
-      g_network_monitor_can_reach_async (nm->monitor, address, NULL, can_reach_done, g_object_ref (invocation)); 
+      g_network_monitor_can_reach_async (nm->monitor, address, NULL, can_reach_done, g_object_ref (invocation));
     }
 
   return G_DBUS_METHOD_INVOCATION_HANDLED;
@@ -200,13 +202,22 @@ network_monitor_iface_init (XdpDbusNetworkMonitorIface *iface)
 }
 
 static void
-on_network_changed (GObject  *object,
-                    gboolean  network_available,
-                    gpointer  user_data)
+emit_changed (gpointer user_data)
 {
   NetworkMonitor *nm = user_data;
 
   xdp_dbus_network_monitor_emit_changed (XDP_DBUS_NETWORK_MONITOR (nm));
+
+  nm->emit_source_id = 0;
+}
+
+static void
+schedule_emit_changed (NetworkMonitor *nm)
+{
+  if (nm->emit_source_id > 0)
+    return;
+
+  nm->emit_source_id = g_idle_add_once (emit_changed, nm);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


jameswestman:
Looking at the documentation for the `changed` signal, I wonder if this is correct, or if there should be separate signals (or properties?) for available/metered/connectivity, like GNetworkMonitor has (see the linked issue). My goal is to be able to show in a UI when background downloads are paused, and be able to automatically restart them when the network is no longer metered.

src/network-monitor.c
```diff
@@ -264,6 +297,21 @@ network_monitor_new (void)
                            network_monitor,
                            G_CONNECT_DEFAULT);
 
+  g_signal_connect_object (network_monitor->monitor, "notify::connectivity",
+                           G_CALLBACK (on_network_property_notify),
+                           network_monitor,
+                           G_CONNECT_DEFAULT);
```
swick:
```suggestion
                           G_CALLBACK (on_network_property_notify),
                           schedule_emit_changed,
                           G_CONNECT_SWAPPED);
```

We can skip on_network_property_notify entirely.

src/network-monitor.c
```diff
@@ -217,14 +219,45 @@ network_monitor_iface_init (XdpDbusNetworkMonitorIface *iface)
...
+  return G_SOURCE_REMOVE;
+}
+
+static void
+schedule_emit_changed (NetworkMonitor *nm)
+{
+  if (nm->emit_source_id > 0)
+    return;
+
+  nm->emit_source_id = g_idle_add_full (G_PRIORITY_DEFAULT_IDLE, emit_changed, g_object_ref (nm), g_object_unref);
```
swick: This keeps the object alive just to send a changed signal. Would be better to not ref it and clear the idle source in dispose instead. Also `g_idle_add_once` exists.

swick: Do we know that property changes and the `network-changed` signal will all arrive until the idle fires?

swick:
ping @jameswestman

src/network-monitor.c
jameswestman: Good point, and same with on_network_changed.

src/network-monitor.c
jameswestman: Fixed.

jameswestman: It seems like it combines some property changes, but still sometimes emits multiple changed signals. Here's what happens when I turn my wired connection off:

```
notify::network-available (true)
notify::connectivity (full)
network-changed (network_available=false)
** Emitting changed signal
notify::network-available (false)
notify::connectivity (local)
** Emitting changed signal
network-changed (network_available=false)
** Emitting changed signal
```

And when I turn it back on:
```
network-changed (network_available=false)
** Emitting changed signal
network-changed (network_available=false)
** Emitting changed signal
notify::network-available (false)
notify::connectivity (local)
network-changed (network_available=true)
** Emitting changed signal
notify::network-available (true)
notify::connectivity (limited)
** Emitting changed signal
notify::connectivity (full)
notify::network-available (true)
notify::connectivity (full)
network-changed (network_available=false)
** Emitting changed signal
notify::network-available (true)
notify::connectivity (full)
network-changed (network_available=true)
** Emitting changed signal
```

I'm by no means an expert in networking or NetworkMonitor, but it seems that turning the connection on/off starts a cascade of low-level changes that can all get reported separately.

swick: Urgh, there isn't a "done" signal or something like that? We could hold for some timeout before applying the changes or something like that, but that would all be racy.

jameswestman: It doesn't look like there is. Those are all the signals and properties on GNetworkManager. I wonder if some of the noise could be reduced on the GLib side, though.

Issue #2009, opened 2026-05-30
[Feature Request] Keyboard Layout params in RemoteDesktop


Decodetalkers:
### Operating System

Archlinux

### XDG Desktop Portal version

1.21

### Desktop Environment

wlroots

### Expected Behavior

Now the client cannot control the keyboard layout of the remote. For example, my computer is using the us layout, but the controlled side is using JP layout, and if the control side follow the controlled machine layout, and create a virtual keyboard, and then I will using us keyboard with a jp keyboard layout

### Current Behavior

No way to set the keyboard layout when starting a remote control

Decodetalkers:
NOT PLAN

PR #2010, opened 2026-05-30
RFC: add keyboard_layout key in RemoteDesktop

Decodetalkers:
I think it is good to let controlling side can use the keyboard layout they want. We will always meet the case that the controlled side is using a different keyboard layout, and xdg-desktop-portal can only know the keyboard_layout of controlled side, it will never know what keyboard_layout is the controlling side using. So I submitted the rcf

solved: https://github.com/flatpak/xdg-desktop-portal/issues/2009

### DIFF
--- a/data/org.freedesktop.impl.portal.RemoteDesktop.xml
+++ b/data/org.freedesktop.impl.portal.RemoteDesktop.xml
@@ -106,6 +106,14 @@
 
           This option was added in version 2 of this interface.
 
+        * ``keyboard_layout`` (``s``)
+          Optional key. What the keyboard layout is used after start a remoting task
+
+          If it is not set, the keyboard layout will follow the controlled side. For
+          example, if the controlled side is using us keyboard, the connection will also
+          use us keyboard. If the key is set, when the controlled side is using jp keyboard,
+          controlling side can still use us keyboard to control the desktop.
+
         For available device types, see the AvailableDeviceTypes property.
     -->
     <method name="SelectDevices">


jadahl:
Keyboard layouts are more complicated than this, and it's not really possible to describe with a string or set of rules, as they are described with non-trivial files describing the whole layout. In practice it means that an application would need to send a complex file describing the layout via fd passing that would then have to be parsed and processed in the display server, and forwarded to all the Wayland clients / Xwayland.

Sending complicated files from a sandboxed application that needs to be parsed by complex parsers by the desktop environment and unrelated applications is not a good approach from a security perspective, thus I don't think it is something we should do for portals.

We do have complicated files passed to the desktop environment in other portals, e.g. images, but it's much simpler to handle such files securely.

What you can do instead is in your application parse the current keyboard layout, and find keycodes that result in the same keysym. With the new libei version just released, there is also a ei_text interface that lets you reliably send text as UTF-8, or keysyms directly, letting the compositor handle things itself.

Decodetalkers:
Emm.. Sorry , I can not follow.. How can the xdp backend know the keyboard layout that is not on the machine? If without xdg-desktop-portal telling it. What can I do with libei? But I think even in that way, we still need to define a rules to pass such information..

jadahl:
> Emm.. Sorry , I can not follow.. How can the xdp backend know the keyboard layout that is not on the machine? If without xdg-desktop-portal telling it. What can I do with libei? But I think even in that way, we still need to define a rules to pass such information..

It doesn't necessarily need to. If you press the keycode C with layout L it translates to the keysym S, which translates tot the unicode character U in e.g. a text input.

libei lets you send the S directly, which then can be used in the display server to come up with a layout L' and a keycode C' that in existing clients results in the same S, without forwarding untrusted layout files to everything consuming keyboard events.

Decodetalkers:
ok, then I close this pr. Thanks for your explanation!

PR #2011, opened 2026-05-30
remote-desktop: Handle D-Bus proxy reconnection when backend restarts

SnotacusNexus:
## Summary

- Monitor `g-name-owner` on the impl D-Bus proxy for `RemoteDesktop` and `ScreenCast`
- Recreate the proxy when the backend reconnects after a restart
- Without this, if a portal backend (e.g. `xdg-desktop-portal-hyprland`) restarts, async D-Bus calls through the stale proxy silently never complete

## Background

`xdg-desktop-portal` creates D-Bus proxies to backend implementations once at startup with `G_DBUS_PROXY_FLAGS_NONE`. If the backend restarts (e.g., crashes and gets re-spawned by D-Bus activation), the proxy holds a stale connection to the old process. Async D-Bus calls through a stale proxy never complete — the callback silently never fires because `g_dbus_proxy_set_default_timeout()` is set to `G_MAXINT`.

## Implementation

Both `remote-desktop.c` and `screen-cast.c` get the same treatment:

1. **Signal handler** `on_impl_g_name_owner_notified()` — watches `notify::g-name-owner` on the proxy:
   - On disconnect (name_owner → NULL): sets a `"impl-was-connected"` flag
   - On reconnection (NULL → name_owner with flag set): triggers proxy recreation
   - On first connection: just records the flag (initial startup)

2. **Reconnect helper** — creates a fresh proxy with the same connection/bus-name/path, re-binds properties, and wires the signal on the new proxy. The old proxy is safely held via `g_autoptr` during the swap so in-flight signal emission completes cleanly.

3. Signal is connected in `init_remote_desktop()` / `init_screen_cast()` right after the portal object is created.

## Test plan

- [x] Build passes with zero warnings (verified: `meson setup build -Dtests=disabled && ninja`)
- [x] Restart portal backend and verify new D-Bus calls succeed instead of hanging

### DIFF
--- a/src/remote-desktop.c
+++ b/src/remote-desktop.c
@@ -1639,13 +1639,104 @@ remote_desktop_iface_init (XdpDbusRemoteDesktopIface *iface)
 }
 
 static void
+on_impl_g_name_owner_notified (GObject    *object,
+                                GParamSpec *pspec,
+                                gpointer    user_data);
+
+static void
+remote_desktop_reconnect_impl (RemoteDesktop *remote_desktop);
+
+static void
 remote_desktop_dispose (GObject *object)
 {
   RemoteDesktop *remote_desktop = (RemoteDesktop *) object;
 
+  /* Signal handler on impl is auto-disconnected when impl is finalized */
   g_clear_object (&remote_desktop->impl);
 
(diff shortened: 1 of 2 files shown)


swick:
This is not how GDBus proxy objects work and we always create them for well-known names.

PR #2012, opened 2026-05-31
pipewire: Disable loading module-rt

ford-prefect:
If the PipeWire daemon restarts, the Camera implementation restarts its connection to PipeWire. As part of that process, it creates a pw_context which causes module-rt to be loaded, which in makes a blocking call out to the Realtime portal. Since that blocking call occurs on the main loop, the process is then deadlocked until the blocking DBus call times out.

We don't need the ability to run anything in realtime, so it is okay to just disable loading module-rt. In parallel, we'll make PipeWire have a more reasonable timeout for its DBus calls as well.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -2,9 +2,9 @@
 # SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
 
 [general]
-ignore=title-must-not-contain-word, body-min-length, body-is-missing, author-valid-email
+ignore=title-must-not-contain-word, body-max-line-length, body-min-length, body-is-missing, author-valid-email
 ignore-merge-commits=true
-extra-path=.gitlint.conf/commit-msg-rules.py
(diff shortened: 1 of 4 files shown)


swick:
> In parallel, we'll make PipeWire have a more reasonable timeout for its DBus calls as well.

Where is that happening?

ford-prefect:
> > In parallel, we'll make PipeWire have a more reasonable timeout for its DBus calls as well.
> 
> Where is that happening?

Pushed to PipeWire [here](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/aef3d02173d162b7dfce495da9f04d8722d43811).

swick:
Please clarify this in the commit message, I expected this to be something we in xdp and not in the pipewire source code itself.

ford-prefect:
Done.

swick:
Thanks!

PR #2018, opened 2026-06-08
Release 1.22.0 (fixup)

swick:
Went through the checklist again and noticed I screwed up. Fortunately we can still fix this before the release.


GeorgesStavracas (approved):
LGTM. Don't forget to bump the release version after the 1.22.0 release

PR #2019, opened 2026-06-09
[1.20] secret: Init asynchronously

jadahl:
When run in a nested D-Bus session, the secret portal backend tends to fail to launch due to another secret service running within the same $XDG_RUNTIME_DIR. Handle this without blocking by initializing the secret portal asynchronously.

Based on an earlier version of
https://github.com/flatpak/xdg-desktop-portal/pull/1814.

### DIFF
--- a/src/secret.c
+++ b/src/secret.c
@@ -192,26 +192,50 @@ secret_class_init (SecretClass *klass)
 }
 
 GDBusInterfaceSkeleton *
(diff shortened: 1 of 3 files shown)


swick:
I've noticed that we officially support 1.20 still. It has not received any of the security fixes though, so I'm not sure if we should pretend that this is still supported.

PR #2021, opened 2026-06-12
document-portal: fix open-file for atypical XDG_RUNTIME_DIR

stacyharper:
When the XDG_RUNTIME_DIR is not the typical /run/user/UID, the whole file picker then portals fails to give an accessible path to the app:

https://github.com/flatpak/flatpak/issues/6494
https://github.com/flatpak/flatpak/issues/5573

The document portal correctly create the mount point, and handle the request as it should. The problem is that portals are not mapping the paths back to the internal path /run/flatpak/doc.

documents_mountpoint is the host path. When using the typical path /run/user/UID/doc, it works because this runtime dir is also mounted in the bwrap, but this is not the documented /run/flatpak/doc/ path:

doc/documents-and-fuse.rst:

> FUSE
> ----
> 
> Inside the Flatpak sandbox, files for which the app has permission to access are
> available at ``/run/flatpak/doc``.

This xdp_register_document method is only used by portals, to call the desktop portal. The portals then return those paths directly. I think we can just enforce the expected path prefix here.

### DIFF
--- a/desktop-portal/open-uri.c
+++ b/desktop-portal/open-uri.c
@@ -232,7 +232,7 @@ launch_application_with_uri (const char *choice_id,
       if (writable)
         flags |= XDP_DOCUMENT_FLAG_WRITABLE;
 
-      ruri = xdp_register_document (uri, app_id, flags, &local_error);
+      ruri = xdp_register_document (uri, app_id, info, flags, &local_error);
       if (ruri == NULL)
         {
           g_warning ("Error registering %s for %s: %s", uri, app_id, local_error->message);
--- a/desktop-portal/xdp-documents.c
+++ b/desktop-portal/xdp-documents.c
@@ -54,6 +54,7 @@ xdp_init_document_proxy (GDBusConnection  *connection,
 char *
 xdp_register_document (const char        *uri,
                        const char        *app_id,
+                       GDesktopAppInfo   *app_info,
                        XdpDocumentFlags   flags,
                        GError           **error)
 {
@@ -72,6 +73,7 @@ xdp_register_document (const char        *uri,
   int version;
   gboolean handled_permissions = FALSE;
   DocumentAddFullFlags full_flags;
+  g_autofree char *fuse_path = NULL;
 
   g_return_val_if_fail (app_id != NULL && *app_id != '\0', NULL);
 
@@ -201,7 +203,9 @@ xdp_register_document (const char        *uri,
       return g_filename_to_uri (doc_path, NULL, error);
     }
 
-  doc_path = g_build_filename (documents_mountpoint, doc_id, basename, NULL);
+  fuse_path = xdp_desktop_app_info_get_doc_mountpoint (app_info);
+  doc_path = g_build_filename (fuse_path, doc_id, basename, NULL);
+
   return g_filename_to_uri (doc_path, NULL, error);
 }
 
--- a/shared/xdp-utils.h
+++ b/shared/xdp-utils.h
@@ -205,5 +205,8 @@ gboolean xdp_desktop_app_info_is_sandboxed (GDesktopAppInfo *info);
 /* Returns the sandboxed app ID, or the desktop file basename as fallback */
 char * xdp_desktop_app_info_get_app_id (GDesktopAppInfo *info);
 
+/* Returns the document fuse mountpoint */
+char * xdp_desktop_app_info_get_doc_mountpoint (GDesktopAppInfo *info);
+
 #define XDP_EXPORT_TEST XDP_EXPORT
 #define XDP_EXPORT __attribute__((visibility("default"))) extern
--- a/tests/xdp_utils.py
+++ b/tests/xdp_utils.py
@@ -318,13 +318,15 @@ class AppInfoFlatpak(AppInfo):
             self.instance_id = "1234567890"
 
         if not self.desktop_entry:
-            self.desktop_entry = b"""
+            desktop_entry_str = f"""
 [Desktop Entry]
 Version=1.0
 Name=Example App
 Exec=true %u
 Type=Application
+X-Flatpak={self.app_id}
 """
+            self.desktop_entry = desktop_entry_str.encode("UTF-8")
 
         if not self.metadata:
             metadata_str = f"""
--- a/desktop-portal/account.c
+++ b/desktop-portal/account.c
@@ -85,7 +85,11 @@ send_response_in_thread_func (GTask        *task,
       if (xdp_app_info_is_host (request->app_info))
         ruri = g_strdup (image);
       else
-        ruri = xdp_register_document (image, xdp_app_info_get_id (request->app_info), XDP_DOCUMENT_FLAG_NONE, &error);
+        ruri = xdp_register_document (image,
+                                      xdp_app_info_get_id (request->app_info),
+                                      xdp_app_info_get_gappinfo (request->app_info),
+                                      XDP_DOCUMENT_FLAG_NONE,
+                                      &error);
 
       if (ruri == NULL)
         {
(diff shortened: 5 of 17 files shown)


stacyharper:
The tests checks that the returned path is available. The /run/flatpak/doc/ is not available from the host, so obviously this not works. I'd say we should probably remap the file paths later in the workflow.

swick:
This won't work. For a host app the current code is correct. If this is really XdpAppInfo specific, then we'll need to add the appropriate API to the XdpAppInfo.

stacyharper:
Thanks for the clue. I have dedicated a new API for this. There is an existing "remap_path", but it seems to be only used internally, and include the opposite mapping from "/run/flatpak/doc" to the runtime dir. That seems better to build the path correctly, instead of remapping it afterward, so this new methods feels better to me.

swick:
The only thing that's really XdpAppInfo specific is where the mount point is.

stacyharper:
> The only thing that's really XdpAppInfo specific is where the mount point is.

You meant I should not pass this as argument?

swick:
The handling of doc_id and basename is always the same, so there is no need to put them in the XdpAppInfo. It should only abstract away the crucial difference: the mount point of the document fuse.

stacyharper:
Fair. My current patch abstract the whole document path build, hiding the prefix away. It works great because building the runtime + UID + "doc" path require an additional string memory management, for those specific prefixes. Do you think this API is okay?

swick:
I would like to see the XdpAppInfo API changed to be only about the mount point of the document portal fuse.

stacyharper:
done!

desktop-portal/open-uri.c
```diff
@@ -248,7 +249,7 @@ launch_application_with_uri (const char *choice_id,
       if (writable)
         flags |= XDP_DOCUMENT_FLAG_WRITABLE;
 
-      ruri = xdp_register_document (uri, choice_id, flags, &local_error);
+      ruri = xdp_register_document (uri, choice_id, app_info, flags, &local_error);
```
swick: I just noticed this. We're launching an app, register the document for the app we're launching and then hand the app the URI for the document as the caller would see it. This is wrong, it needs to be the path as the spawned app will see it.

Not entirely sure how to fix this...

stacyharper: I'd say we should have access to both app infos at this opint, and pass the spawned app one here

edit: well I'm adding the appinfo as argument to launch_application_with_uri, so I should be able to pass the correct one

swick: The XdpAppInfo is a thing that only exists (and only *should exist) for active connections from a specific launched app instance to xdp. You need another concept of a typed (flatpak, host, ...), un-launched app.

swick: Something similar to https://github.com/flatpak/xdg-desktop-portal/pull/2046/changes/f3a2e7c7fe77dc5886d99306614fa5b1c33d5089 probably will work: `xdp_desktop_app_info_get_doc_mountpoint`

stacyharper: If I'm right, you mean to handle this specific case of open-uri using this GDesktopAppInfo instead. So I have to hybrid the two approaches, or invert this dependency as in: https://github.com/flatpak/xdg-desktop-portal/pull/2021/changes/c950e571c99b8f68bda61318d4aa386fca8d6d47?

swick: I think you can drop `xdp_app_info_document_path` entirely, use `xdp_app_info_get_gappinfo` and feed that to `xdp_desktop_app_info_get_doc_mountpoint`.

stacyharper: `xdp_app_info_get_gappinfo` returns a `GAppInfo *`, not a `GDesktopAppInfo *`. I'm not sure on how to close the gap, but I tried to re-use the available `app_id`, rebuild the `desktop_id`, and grab the `GDesktopAppInfo` with it: https://github.com/flatpak/xdg-desktop-portal/commit/15d3318ead35d29c452177d16e21daa861b1b1a6#diff-bd6406e677bd10f04a18fcdac8749e6a7f4d277920dbeb3de646f26068ac26b0R80

swick: Might be able to change all those functions in utils.c to use `GAppInfo` (it's an interface, `GDesktopAppInfo` is an object implementing it).

stacyharper: `GAppInfo` does not have a `get_string` method to look at a `.desktop` entry. It seems to be meaningfull for other kind of "system installed application". For the XDP context, what are the use cases for `GAppInfo` alone? Shouldn't we just use `GDesktopAppInfo` everywhere instead?

swick: Yeah, you can look into using `GDesktopAppInfo` in XdpAppInfo then.

stacyharper: Done!

desktop-portal/open-uri.c
```diff
@@ -199,7 +199,8 @@ is_file_uri (const char *uri)
 }
 
 static gboolean
-launch_application_with_uri (const char *choice_id,
+launch_application_with_uri (XdpAppInfo *app_info,
+                             const char *choice_id,
```
swick: This is still wrong. We don't launch the `XdpAppInfo *app_info`, we launch `const char *choice_id`. What we have to pass down to `xdp_register_document` is a `GDesktopAppInfo` that you get from the `choice_id`.

stacyharper: ah right, that make sense. Sorry for the noise. Done

shared/xdp-utils.h
```diff
@@ -205,5 +205,8 @@ gboolean xdp_desktop_app_info_is_sandboxed (GDesktopAppInfo *info);
 /* Returns the sandboxed app ID, or the desktop file basename as fallback */
 char * xdp_desktop_app_info_get_app_id (GDesktopAppInfo *info);
 
+/* Returns the document fuse mountpoint */
+char * xdp_desktop_app_info_get_doc_mountpoint (GDesktopAppInfo *info);
```
swick, line 209: The new function deserves its own commit.

tests/xdp_utils.py
```diff
@@ -318,13 +319,15 @@ def __post_init__(self):
...
 
         if not self.desktop_entry:
-            self.desktop_entry = b"""
+            desktop_entry_str = f"""
 [Desktop Entry]
 Version=1.0
 Name=Example App
 Exec=true %u
 Type=Application
+X-Flatpak={self.common_id}
```
swick: This is always the app_id.

swick: Can also be its own commit

desktop-portal/xdp-documents.c
```diff
@@ -201,7 +203,9 @@ xdp_register_document (const char        *uri,
       return g_filename_to_uri (doc_path, NULL, error);
     }
 
-  doc_path = g_build_filename (documents_mountpoint, doc_id, basename, NULL);
+  fuse_path = xdp_desktop_app_info_get_doc_mountpoint(app_info);
```
swick:
```suggestion
  fuse_path = xdp_desktop_app_info_get_doc_mountpoint (app_info);
```

swick (commented):
Getting there...

swick:
Oh, and the commit message should not be prefixed with `chore` but what it touches, so either `app-info` or `xdp` if it touches kind of everything.

stacyharper:
done!

swick (approved):

swick:
Thanks! Now we just need to get https://github.com/flatpak/xdg-desktop-portal/pull/2046 merged.

stacyharper:
I have pushed a reworded commit, replacing tabs by whitespaces

PR #2023, opened 2026-06-12
document-portal: Ensure we always deal with valid App IDs

swick:
In some code paths, we failed to validate the app id argument which gets passed by an attacker. We now validate the app id as soon as we get it from the dbus call, and also at the place where we would fail catastrophically if it is not.

In particular, a well-crafted app id could cause parsing of arbitrary files on the host.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -733,7 +733,7 @@ app_has_file_access (const char *target_app_id,
   g_autofree char *res = NULL;
   g_autofree char *arg = NULL;
 
-  if (target_app_id == NULL || target_app_id[0] == '\0')
+  if (target_app_id == NULL || !xdp_is_valid_app_id (target_app_id))
     return FALSE;
 
   if (g_str_has_prefix (target_app_id, "snap."))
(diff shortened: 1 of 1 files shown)


document-portal/document-portal.c
```diff
@@ -733,7 +733,9 @@ app_has_file_access (const char *target_app_id,
   g_autofree char *res = NULL;
   g_autofree char *arg = NULL;
 
-  if (target_app_id == NULL || target_app_id[0] == '\0')
+  if (target_app_id == NULL ||
+      target_app_id[0] == '\0' ||
+      !xdp_is_valid_app_id (target_app_id))
```
ebassi: `xdp_is_valid_app_id()` already returns false for an empty string; it adds a critical for `NULL`, sadly, which I would actually replace since this is not a library, and precondition checks are not a thing

```c
if (target_app_id == NULL || !xdp_is_valid_app_id (target_app_id))
  return FALSE;
```

document-portal/document-portal.c
```diff
@@ -792,6 +794,15 @@ portal_add_full (GDBusMethodInvocation *invocation,
   g_variant_get (parameters, "(@ahu&s^a&s)",
                  &array, &flags, &target_app_id, &permissions);
 
+  if (target_app_id[0] != '\0' &&
+      !xdp_is_valid_app_id (target_app_id))
```
ebassi, line 796:
```suggestion
  if (!xdp_is_valid_app_id (target_app_id))
```

document-portal/document-portal.c
```diff
@@ -1085,6 +1096,15 @@ portal_add_named_full (GDBusMethodInvocation *invocation,
       return;
     }
 
+  if (target_app_id[0] != '\0' &&
+      !xdp_is_valid_app_id (target_app_id))
```
ebassi, line 1098:
```suggestion
  if (!xdp_is_valid_app_id (target_app_id))
```

ebassi (commented):
Looks good, though we can remove a couple of additional checks in favour of just having `xdp_is_valid_app_id()` everywhere.

document-portal/document-portal.c
swick: This check here (and below) specifically *allows* an empty target_app_id.

jadahl (approved):

PR #2034, opened 2026-06-18
tests: add ScreenCast portal integration test

glamberson:
The ScreenCast portal had no integration test. This adds one, with a python-dbusmock template for `org.freedesktop.impl.portal.ScreenCast` that drives the CreateSession/SelectSources/Start flow through the frontend.

The test covers the `pipewire-serial` stream property added in #1942: a version 6 backend advertises it and the frontend forwards it to the application, while a version 5 backend does not, and the portal reports version 5 accordingly.

Suggested in #1942.


swick:
Thank you, that's a good start!

PR #2036, opened 2026-06-20
Make it possible to implement portals with libdex

swick:
This is the last round of changes we need to finally port over some portals to libdex.

Includes https://github.com/flatpak/xdg-desktop-portal/pull/2035


swick:
The CI is failing because libdex ues a ctor to call `dex_init()` which spawns threads. We have to either work around that or fix libdex.

swick:
Worked around by compiling libdex against liburing, the next release will turn off the shared object constructor so we should switch to that (and require it in meson) once it's available.

ref: https://gitlab.gnome.org/GNOME/libdex/-/work_items/36

swick:
Please review @jadahl @AdrianVovk

swick:
CI is failing now because of https://gitlab.gnome.org/GNOME/libdex/-/work_items/37, but everything else passes so reviews would be appreciated.

swick:
Workaround for that is in place as well, so CI should be good.

desktop-portal/xdp-context.c
```diff
@@ -203,6 +203,36 @@ method_needs_request (GDBusMethodInvocation *invocation)
   return method_info ?  method_info->uses_request : TRUE;
 }
 
+static gboolean
+authorize_callback_fiber (GDBusInterfaceSkeleton *interface,
```
AdrianVovk, line 231: I'm not entirely following why this is a separate function. I get that we want to change `xdg_app_info_registery_ensure_for_invocation_sync` -> something async in the fiber codepath. Is that the only reason?

Because I _think_ we could detect this at runtime: We can check if `DEX_IS_DBUS_INTERFACE_SKELETON` on the interface, and if it is then we can get its flags, see if it has the handle_in_fiber flag set, and redirect execution?

This way we can minimize some of the code duplication?

desktop-portal/xdp-permissions.h
```diff
@@ -20,20 +22,38 @@ char **xdp_get_permissions_sync (XdpAppInfo *app_info,
                                  const char *table,
                                  const char *id);
 
+DexFuture * xdp_permissions_get_future (XdpAppInfo *app_info,
```
AdrianVovk, line 25: Is it intentional that the naming convention is backwards from the old methods?

`xdp_<verb>_permission[s]_<variant>` became `xdp_permission[s]_<verb>_<variant>`. I think the new name is probably better.

Did you just decide it's not worth doing the refactor to make the names consistent since we'll eventually just lose the old variants?

desktop-portal/xdp-request-future.c
```diff
@@ -0,0 +1,306 @@
...
+  XdpAppInfo *app_info;
+  XdpDbusImplRequest *impl_request;
+  GDBusInterfaceSkeleton *skeleton;
+  char *id;
+  gboolean exported;
+} XdpRequestFuture;
+
+static void xdp_request_skeleton_iface_init (XdpDbusRequestIface *iface);
+
+G_DEFINE_FINAL_TYPE_WITH_CODE (XdpRequestFuture,
```
AdrianVovk: This name makes it sound like it's a type of Future, even though it's not. How about `XdpFiberRequest` or even `XdpDexRequest`?

When all the portals a ported to libdex, I presume we'll lose the old `XdpRequest` and can rename this to `XdpRequest` instead?

desktop-portal/xdp-request-dex.c
```diff
@@ -0,0 +1,306 @@
...
+                    G_CALLBACK (request_authorize_callback),
+                    request);
+
+  if (!g_dbus_interface_skeleton_export (G_DBUS_INTERFACE_SKELETON (request),
+                                         g_dbus_interface_skeleton_get_connection (request->skeleton),
+                                         request->id,
+                                         &error))
+      return dex_future_new_for_error (g_steal_pointer (&error));
+
+  return dex_future_new_for_object (request);
```
AdrianVovk, line 222: Missing a g_steal_pointer?

desktop-portal/xdp-permissions.c
```diff
@@ -48,6 +48,42 @@ xdp_get_permissions_sync (XdpAppInfo *app_info,
...
+  g_autoptr(XdpDbusImplPermissionStoreLookupResult) result = NULL;
+  g_autofree char **permissions = NULL;
+
+  result = dex_await_boxed (dex_ref (future), NULL);
+
+  if (!result ||
+      !g_variant_lookup (result->permissions, app_id, "^a&s", &permissions))
+    return dex_future_new_for_pointer (NULL);
+
+  return dex_future_new_for_pointer (g_strdupv (permissions));
```
AdrianVovk: why strdup and not steal?

desktop-portal/xdp-session-future.c
```diff
@@ -0,0 +1,448 @@
...
+  XdpAppInfo *app_info;
+  XdpDbusImplSession *impl_session;
+  GDBusInterfaceSkeleton *skeleton;
+  char *id;
+  gboolean exported;
+} XdpSessionFuture;
+
+static void xdp_session_skeleton_iface_init (XdpDbusSessionIface *iface);
+
+G_DEFINE_FINAL_TYPE_WITH_CODE (XdpSessionFuture,
```
AdrianVovk: Similar comments re: naming as `XdpRequestFuture`

desktop-portal/xdp-session-dex.c
```diff
@@ -0,0 +1,448 @@
...
+
+static void
+xdp_session_future_class_init (XdpSessionFutureClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->dispose = xdp_session_future_dispose;
+
+  signals[SESSION_CLOSED] =
+    g_signal_new ("session-closed",
```
AdrianVovk, line 142: Should we have some better naming here? I'm pretty sure we have a `closed` signal, which means "we're notifying the app that this session was closed by x-d-p". But `session-closed`, here, is the opposite: "we're notifying the rest of x-d-p that the app closed the session".

Perhaps we should call the session `closed-by-peer` or something? So it's a little more descriptive what `Session::closed` vs `Session::session-closed` mean

desktop-portal/xdp-session-dex.c
```diff
@@ -0,0 +1,448 @@
...
+                    G_CALLBACK (session_authorize_callback),
+                    session);
+
+  if (!g_dbus_interface_skeleton_export (G_DBUS_INTERFACE_SKELETON (session),
+                                         g_dbus_interface_skeleton_get_connection (session->skeleton),
+                                         session->id,
+                                         &error))
+      return dex_future_new_for_error (g_steal_pointer (&error));
+
+  return dex_future_new_for_object (session);
```
AdrianVovk, line 269: Are we missing a `g_steal_pointer` here too?

shared/xdp-app-info-registry.c
```diff
@@ -55,101 +96,217 @@ xdp_app_info_registry_new (void)
...
+          break;
+        default:
+          g_assert_not_reached ();
+        }
+    }
 
-  app_info = g_hash_table_lookup (registry->app_infos, sender);
-  if (!app_info)
-    return NULL;
+  g_hash_table_remove (registry->channels, sender);
```
AdrianVovk, line 223: We update registry->channels here without holding the mutex

shared/xdp-app-info-registry.c
```diff
@@ -7,29 +7,70 @@
...
 #include "xdp-app-info.h"
+#include "xdp-dex.h"
 
 struct _XdpAppInfoRegistry
 {
   GObject parent_instance;
 
-  GHashTable *app_infos; /* unique dbus name -> app info */
-  GMutex app_infos_mutex;
+  GMutex mutex;
```
AdrianVovk: We should document what the mutex is protecting (as far as I can tell, it's the `channels` field?)

desktop-portal/xdp-context.c
swick: > We can check if DEX_IS_DBUS_INTERFACE_SKELETON on the interface

No, all skeletons are derive from there now, we'd have to check the flags. But honestly the methods are very different and the single method becomes harder to follow:

```c
static gboolean
authorize_callback (GDBusInterfaceSkeleton *interface,
                    GDBusMethodInvocation  *invocation,
                    gpointer                user_data)
{
  XdpContext *context = XDP_CONTEXT (user_data);
  g_autoptr(DexFuture) future = NULL;
  g_autoptr(XdpAppInfo) app_info = NULL;
  DexDBusInterfaceSkeletonFlags flags;
  gboolean is_fiber;

  flags = dex_dbus_interface_skeleton_get_flags (DEX_DBUS_INTERFACE_SKELETON (interface));
  is_fiber = (flags & DEX_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_FIBER);

  future = xdp_app_info_registry_ensure_future (context->app_info_registry,
                                                invocation);

  if (!is_fiber)
    dex_thread_wait_for (dex_ref (future), NULL);

  app_info = dex_await_object (g_steal_pointer (&future), &error);
  if (app_info == NULL)
    {
      g_dbus_method_invocation_return_error (invocation,
                                             G_DBUS_ERROR,
                                             G_DBUS_ERROR_ACCESS_DENIED,
                                             "Portal operation not allowed: %s", error->message);
      return FALSE;
    }

  g_object_set_data (G_OBJECT (invocation), "xdp-app-info", app_info);

  if (!is_fiber && method_needs_request (invocation))
    {
      if (!xdp_request_init_invocation (invocation, context, app_info, &error))
        {
          g_dbus_method_invocation_return_gerror (invocation, error);
          return FALSE;
        }
    }

  return TRUE;
}
```

I'd like to keep it separate.

desktop-portal/xdp-permissions.h
swick: The entire thing is on my "bad code that needs to be cleaned up" list but I resisted the urge to change it all, but also didn't want to follow the really backwards naming convention.

desktop-portal/xdp-request-future.c
swick: Naming with libdex is a bit hard. Usually for futures there is no need to write it in the function name because it returns a DexFuture. For fibers, things get tricky because there isn't a clear type hint that this must run as part of a fiber, so we usually add a `_fiber` postfix. The `XdpRequestFuture` API surface does not require a fiber! So I would object to names involving fiber. Eventually I would like this to become the only `XdpRequest` indeed. I'm also not sure if I like `XdpDexRequest` because it sounds like a `DexRequest` specifically for xdp.

It all sucks, not sure renaming this will get us a better result.

desktop-portal/xdp-request-dex.c
swick: dex_future_new_for_object docs: The data is owned by the caller of the function

desktop-portal/xdp-session-dex.c
swick: It is actually a signal we emit when the request is closed for any reason: app calling close, impl calling close, peer disconnect. We could differentiate between them, but usually you're just interested in the fact that it got closed, for whatever reason.

desktop-portal/xdp-permissions.c
swick: Stuck to the existing `_sync` version, and it creates a shallow copy in permissions, so it can't be stolen. We can however just get a deep copy from the GVariant.

desktop-portal/xdp-context.c
AdrianVovk: Ah, the other difference is requests being different. Alright, fine

desktop-portal/xdp-request-future.c
AdrianVovk: How about `XdpRequestDex`?

desktop-portal/xdp-session-dex.c
AdrianVovk: I see, could we then just emit the `closed` signal? Rather than defining a second signal?

I understand that emitting `closed` emits the signal to the bus too, but if we order things right then an app calling `Close()` on us can un-export us from the bus before we emit the `close` signal. We can skip dbus signal emission in that case, if we're not exported onto the bus anymore. This way, though, there's always one signal that's triggered in all cases

desktop-portal/xdp-permissions.c
```diff
@@ -54,15 +54,15 @@ xdp_permissions_get_future_finally_cb (DexFuture *future,
...
 
   result = dex_await_boxed (dex_ref (future), NULL);
 
   if (!result ||
-      !g_variant_lookup (result->permissions, app_id, "^a&s", &permissions))
+      !g_variant_lookup (result->permissions, app_id, "^as", &permissions))
     return dex_future_new_for_pointer (NULL);
 
-  return dex_future_new_for_pointer (g_strdupv (permissions));
+  return dex_future_new_for_pointer ((char **) g_steal_pointer (&permissions));
```
AdrianVovk: isn't this causing a leak? What happens if the DexFuture is free'd w/o being awaited?

desktop-portal/registry.c
```diff
@@ -39,88 +39,110 @@ G_DEFINE_TYPE_WITH_CODE (Registry, registry,
...
+    xdp_app_info_registry_insert_future (registry->app_info_registry,
+                                         invocation,
+                                         DEX_FUTURE (dex_ref (app_info_promise)));
+
+  /* Then we check if we actually should allow the caller to update the
+   * app id */
+  {
+    g_autoptr(XdpAppInfo) detected_app_info = NULL;
+
+    detected_app_info = xdp_app_info_new_for_invocation_sync (invocation,
```
AdrianVovk, line 70: We probably need a FIXME here too? For sync invocation on a fiber? And also the other _sync method invocations in this commit?

shared/xdp-app-info-registry.c
```diff
@@ -55,101 +96,224 @@ xdp_app_info_registry_new (void)
...
+static DexFuture *
+work_fiber (XdpAppInfoRegistry *registry,
+            const char         *sender)
 {
-  XdpAppInfo *app_info = NULL;
+  DexChannel *channel;
 
-  G_MUTEX_AUTO_LOCK (&registry->app_infos_mutex, locker);
+  g_mutex_lock (&registry->channels_lock);
+  channel = g_hash_table_lookup (registry->channels, sender);
```
AdrianVovk: Do you need to ref the channel while you hold it? What happens if another thread tries to remove the channel, but you're not holding the mutex? Smells like a use-after-free but I didn't think it through _too_ thoroughly

AdrianVovk: Looking a little closer, seems that this is fine because we only ever remove a channel from the hash table in this function. On the other hand, if that ever changes we'd be introducing a use-after-free. Perhaps it's best to just ref the thing and hold an owned reference to avoid any possible issues?

desktop-portal/xdp-permissions.c
swick: Got rid of all the `dex_future_new_for_pointer` calls by replacing them with `dex_future_new_take_boxed.

shared/xdp-app-info-registry.c
swick: Reffing now to make it more obviously correct.

desktop-portal/xdp-request-future.c
swick: Not bad. I'm kind of resisting though because I have all those portals ported in https://github.com/flatpak/xdg-desktop-portal/pull/1855 which makes it easy to actually test the code here. If I do it, it will be shortly before merging.

swick:
Fixed issues related to shutdown and cancellation.

desktop-portal/xdp-session-dex.c
```diff
@@ -0,0 +1,453 @@
...
+  object_class->dispose = xdp_session_future_store_dispose;
+}
+
+XdpSessionFutureStore *
+xdp_session_future_store_new (size_t session_offset)
+{
+  XdpSessionFutureStore *store;
+
+  store = g_object_new (XDP_TYPE_SESSION_FUTURE_STORE, NULL);
+  store->session_offset = session_offset;
```
AdrianVovk, line 452: What's this about? I looked at your PR that ports a few portals to this API and you always pass 0

AdrianVovk (commented):
Everything looks good, modulo my previous unresolved feedback:
- Renaming `XdpRequestFuture`/`XdpSessionFuture`
- Swapping around `session-closed` signal to `closed`

I don't feel particularly strongly about these so feel free to merge w/o it if you don't agree with those changes

desktop-portal/xdp-session-dex.c
swick: The `XdpSessionFuture` class is final. Subclassing the session has made things more complex for no good reason, so the pattern I want to see now is that if you need to store more data with your session, you embed the session in your own struct/object. You then construct the `XdpSessionFutureStore` with `xdp_session_future_store_new (G_STRUCT_OFFSET (MySessionWrapper, session))`.

But honestly, maybe it makes sense to split this into two constructors, and maybe even a bit more macro magic?

`xdp_session_future_store_new (void)`
`xdp_session_future_store_new_wrapped (MySessionWrapper, session)`

desktop-portal/xdp-session-dex.c
swick: I'm actually confused myself now what's going on there and what the semantics were supposed to be...

swick: The XdpSessionFuture is subclassing the xdp.Session dbus skeleton. That one has client and server "things" on it, one being the closed signal which gets emitted on the client side and can be used on the service-side to emit the dbus signal (xdp is the service-side). I'd rather not overload the dbus signal in one direction (frontend->app) with the general "the session got closed somehow" gobject signal.

kolunmi:
it looks like libdex isn't always built with gdbus support; would it make sense to add a libdex.wrap and change to
```meson
libdex_dep = dependency('libdex-1', version: '>=1.1', default_options: ['-Dgdbus=enabled'])
```
?

swick:
Not sure if that's a good idea. We use meson subprojects for things we compile into our binaries, not things we dynamically link to. If distros don't enable gdbus support in libdex, they will have problems with shipping xdp anyway.

desktop-portal/xdp-context.c
```diff
@@ -90,14 +90,37 @@ xdp_context_dispose (GObject *object)
...
+          if (DEX_IS_DBUS_INTERFACE_SKELETON (skeleton))
+            dex_dbus_interface_skeleton_cancel (DEX_DBUS_INTERFACE_SKELETON (skeleton));
+          g_dbus_interface_skeleton_unexport (skeleton);
+        }
+
+      g_clear_pointer (&context->exported_portals, g_hash_table_unref);
+    }
+
+  while (g_main_context_iteration (NULL, FALSE))
+    ;
```
tytan652, line 118: Wouldn't be better to cancel the cancellable before draining ?

desktop-portal/xdp-request-future.c
swick: Renamed

desktop-portal/xdp-session-dex.c
swick: I'll leave it as is

desktop-portal/xdp-session-dex.c
swick: done that

desktop-portal/xdp-context.c
swick: Good point, done

Issue #1314, opened 2024-03-23
Optional information actually required in GlobalShortcuts::CreateSession


dcz-self:
### Operating System

Fedora 38

### XDG Desktop Portal version

1.18

### Desktop Environment

KDE

### Expected Behavior

Contents of `options` parameter described as "optional further information" in https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html#org-freedesktop-portal-globalshortcuts-createsession are not required for a successful call.

### Current Behavior

Call errors out unless both `session_handle_token` and `handle_token` are present: `org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code0: Missing token`.

dbus-monitor log:
```
method call time=1711192202.821524 sender=:1.1659 -> destination=:1.64 serial=8 p
ath=/org/freedesktop/portal/desktop; interface=org.freedesktop.portal.GlobalShort
cuts; member=CreateSession
   array [
   ]
error time=1711192202.824365 sender=:1.64 -> destination=:1.1659 error_name=org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code0 reply_serial=8
   string "Missing token"
```

### Steps to Reproduce

1. Download https://gitlab.gnome.org/dcz/portal-poker/-/commit/e8e92aaab902a1c80885db2d6498669051973751
2. Run `./portal-poker.py portal GlobalShortcuts CreateSession`

matthiasclasen:
Yeah, that documentation is misleading. It was probably copied from the screencast portal, where sessions made a first appearance.

swick:
The docs are no longer calling those options optional

PR #2038, opened 2026-06-21
xdp-session: Remove g_assert(session->token != NULL) — dead code, crashes portal instead of returning error

### LINKED ISSUE #2037: xdp_session_initable_init g_assert aborts process when session_handle_token is NULL (dead code after assertion)
## Summary

`xdp_session_initable_init` uses `g_assert(session->token != NULL)` which aborts the entire portal process when a session is created without `session_handle_token`.

The code already has a graceful error check for this case at lines 290-294, but it is dead code because the `g_assert` on line 279 fires first, causing SIGABRT.

## Impact

A call to `org.freedesktop.portal.RemoteDesktop.CreateSession` (or any portal that creates XdpSession) without `session_handle_token` in the options dict will abort `xdg-desktop-portal` instead of returning a proper error. This also affects any backend that creates internal sessions without a token during portal initialization.

## Affected Code

`desktop-portal/xdp-session.c` in `xdp_session_initable_init`:

https://github.com/flatpak/xdg-desktop-portal/blob/main/desktop-portal/xdp-session.c#L277-L294
""
  g_assert (session->sender != NULL);
  g_assert (session->token != NULL);    // <-- ABORTS when token is NULL
  g_assert (session->context != NULL);

  ...

  if (!session->token)                  // <-- dead code, never reached
    {
      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Missing token");
      return FALSE;
    }
""

## Fix

Remove the `g_assert(session->token != NULL)` line. The existing graceful check at line 290 already returns a proper GError for this case.

> Note: This is a different manifestation of the same root cause as #1314 (GlobalShortcuts::CreateSession fails with 'Missing token'). In that issue the error is returned gracefully; here it crashes the entire process.

## Environment

- xdg-desktop-portal 1.22.1-1.1 (CachyOS/Arch Linux)
- Compositor: MangoWM (wlroots)
- Triggered by hypr-kdeconnect-fix portal backend during startup when portals.conf explicitly routes RemoteDesktop to hypr-kdeconnect

SK-DEV-AI:
## Summary

Removes `g_assert(session->token != NULL)` in `xdp_session_initable_init` — the assertion aborts the portal process when a session is created without `session_handle_token`, making the graceful error code below it unreachable.

## Details

The `g_assert` at `desktop-portal/xdp-session.c:279` fires when a session is created without `session_handle_token` in the GObject construction properties. The same code immediately after (lines 290-294) already handles this case gracefully with `g_set_error(error, ..., "Missing token"); return FALSE;`.

By removing the assertion, the graceful fallback becomes reachable and the portal returns a proper D-Bus error instead of aborting.

## Related

Fixes #2037 — this is the same root cause as #1314 (GlobalShortcuts 'Missing token') but manifests as a crash instead of a graceful error.

## Testing

- Triggered by creating any portal session without `session_handle_token`
- After fix: returns `Missing token` error via D-Bus
- Before fix: `SIGABRT`, portal process crashes

### DIFF
--- a/desktop-portal/xdp-session.c
+++ b/desktop-portal/xdp-session.c
@@ -277,7 +277,6 @@ xdp_session_initable_init (GInitable     *initable,
   int i;
 
   g_assert (session->sender != NULL);
-  g_assert (session->token != NULL);
   g_assert (session->context != NULL);
 
   sender_escaped = g_strdup (session->sender + 1);


swick:
Thanks! The commit message needs to be re-formatted (80 char lines), and the tag `Fixes: 54a08384 ("request, session: Claim requests and sessions object paths")` should be added to indicate where the issue was introduced.

swick:
Did the changes, thanks again for tracking this down.

PR #2042, opened 2026-06-23
xdp: Switch to a new versioning scheme

swick:
We currently have something which resembles semantic versioning, but the minor releases are stable when the number if even, and unstable when the number is odd.

This also means the first number is fixed at 1 because we never break backwards compatibility.

A lot of people get confused by the odd/even scheme. Distros like Fedora have been shipping unstable releases. Distros like Arch have not been testing any of the unstable releases. Systems like GNOME OS have started shipping xdp from git main in their nightly where we will notice when something went wrong immediately.

Releases are work. Doing them needs to have a benefit. It does not seem like there is any benefit for us to do the unstable releases.

Given that unstable releases are a waste of time and that the major version is fixed, this commit changes the project over to a scheme where...

* all releases are stable releases
* the version number is major.minor
* minor versions get released to deal with security issues and bugs
* only the latest major release is supported

The next release will be version 23.0 because we would have releases 1.23.0 in the old versioning scheme.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -21,7 +21,7 @@ body:
     label: "XDG Desktop Portal version"
     description: "What version of XDG Desktop Portal are you using?"
     options:
-    - "1.23"
+    - "23"
     - "1.22"
     - "1.20"
     - Git
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@ SPDX-License-Identifier: LGPL-2.1-or-later
 SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
 -->
 
-Changes in 1.23.0
+Changes in 23.0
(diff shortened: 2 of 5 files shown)


tytan652:
> Given that unstable releases are a waste of time and that the major version is fixed

Just for clarity or being addressed explicitly here, the new scheme will not have things alike to `.dev`, `.alpha`, `.beta` or `.rc` even when it's a version in development ?

swick:
Correct. Testing can be done by building from main, just like GNOME OS does.

swick:
@jadahl @GeorgesStavracas @smcv @Sodivad @AdrianVovk

Sodivad:
I agree the current scheme is just too confusing.

AdrianVovk (approved):
I've been in favor of something like this since #1779. So, LGTM!

(Also, we should note that this PR supersedes #1779)

GeorgesStavracas (approved):
No objections

PR #2043, opened 2026-06-24
Make USB Portal companion documentation available

swick:
The D-Bus interface is usually pretty terse on how things work which makes sense because it is about the API. We could use some higher level documentation about how it all fits together though. We in fact already have such documentation about the USB Portal, but never hooked it up to anything.


doc/org.freedesktop.portal.Usb.rst
```diff
@@ -0,0 +1,86 @@
...
+Device acquisition and release
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once an app has determined which devices it wants to access, the app
+can call the ``AcquireDevices()`` method. This method may prompt a dialog
+for the user to allow or deny the app from accessing specific devices.
+
+If permission is granted, xdg-desktop-portal tries to open the device
+file on the behalf of the requesting app, and pass the file
+descriptor down. [#usb-logind]_
```
AdrianVovk, line 55: I wouldn't say "tries to" here. Let's just say it opens the device. We can keep the note about logind though, but fact of the matter is that there's a good chance this will change soon. We've already discussed the necessary changes in systemd to make this feasible for us here in x-d-p. IIRC, we decided that logind will allow you to `TakeDevice()` if you're not the session controller, and that will come with a Polkit prompt

doc/org.freedesktop.portal.Usb.rst
```diff
@@ -0,0 +1,86 @@
...
+
+- Apps can call the ``EnumerateDevices()`` method, which gives a snapshot
+  of the current devices to the app.
+
+- Apps can create a device monitoring session with ``CreateSession()``
+  which sends the list of available devices on creation, and also
+  notifies the app about connected and disconnected devices.
+
+Only devices that the app is allowed to see are reported in both
+cases.
```
AdrianVovk, line 41: This is mediated by a static permission (entitlement?) right? We should probably document that somehow?

AdrianVovk (commented):
Minor notes, otherwise LGTM

doc/org.freedesktop.portal.Usb.rst
swick: It is not yet an entitlement, but that will change, and I have a version around which changes the entire section accordingly. One thing after the other.

doc/org.freedesktop.portal.Usb.rst
swick: All fair points, but this is just changing the format from md to rst. I'll leave this as is for now.

PR #2044, opened 2026-06-25
Make the docs generated from the dbus xml interface files more flexible

swick:
* Allows adding toplevel headings in the interface description comment instead of always being under "Description"
* Allows including the properties, methods, and signals sections which makes it possible to put content before and after

This also inlines the USB companion doc.

The flexibility will become specifically useful when we're going to add docs about entitlements to portals.


swick:
@AdrianVovk

swick:
Going to merge in a few days if no one takes a look at it.

AdrianVovk (approved):

PR #2046, opened 2026-06-26
Fixes and cleanups around desktop id -> app id mapping

swick:


Sodivad (approved):

duude21:
Thanks a lot for fixing this!!

There could be made a point, that last choice should not be used at all. But rather the default for that mime type, as it does create confusion if the default is not used. see this discussion https://github.com/flatpak/xdg-desktop-portal/discussions/1874

PR #2047, opened 2026-06-26
Fix PipeWire and WirePlumber issues

tytan652:
* pipewire: Fix drop-in configuration

   For some reason `{}` was working when I tested at the time of my old testing
   Was an array with `[]` since a real while, so there no need to worry about compat.
   https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/90b0410280da65f5b8f879da7f204735b6d27cb1
* pipewire: Fix WirePlumber module properties enum

   I blame gobject-linter on that one, since CI forced me to swap approach but didn't caught that I missed to add `= 1`.

### DIFF
--- a/pipewire/pipewire.conf
+++ b/pipewire/pipewire.conf
@@ -7,10 +7,9 @@ context.properties = {
     module.portal = false
 }
 
-context.modules = {
+context.modules = [
     { name = libxdg-desktop-portal
         flags = [ ifexists ]
         condition = [ { module.xdp-desktop-portal = !false } ]
     }
-
-}
+]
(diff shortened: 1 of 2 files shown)


swick:
Thanks for keeping on top of this. Out of interest, how did you notice?

tytan652:
Manual testing on GNOME OS with self-made sysext before and while making https://gitlab.gnome.org/GNOME/gnome-build-meta/-/merge_requests/4808

- `pipewire.access` was still set to `portal` rather than `xdg-desktop-portal` and PipeWire logged the config issue

- WirePlumber was segfaulting due to missing core assertion and I found critical log about property 0 while running `wireplumber` in a terminal

PR #2048, opened 2026-06-26
Fix Settings with multiple impls

swick:


MyChaOS87:
Checked the new behavior (only in regard to my reported issue) and can confirm this solves it.

swick:
I'm going to merge this in a few days if there are no reviews until then. /ping @ebassi @jadahl @AdrianVovk

AdrianVovk (changes requested):
So right now we have an approach that will slowly use more and more RAM as settings change. Whenever a setting changes we'll add it to the hash table and never remove it. It's probably not the biggest deal (how often are settings changing?) but still is avoidable.

Really the problem is when a lower-priority Settings impl emits `Changed()` on a setting that's actually present in a higher-priority Settings impl. We don't need to compare values or anything like that: we just need to ensure that the impl emitting `Changed()` is the highest priority impl that provides that setting.

So: rather than keeping around values of emitted signals in a hash table, what we could do is loop through the impls and check if any one of them have the setting. If an impl with higher priority than the one that emitted `Changed()` has the setting, then we simply don't emit anything. If we find the impl that emitted `Changed()` before finding any other, then we pass the `Changed()` signal through

This way, we don't need to keep track of `last_emitted` and our memory usage doesn't need to go up

swick:
Then you still need to track the highest priority impl per settings key, no? I'm also unsure if that actually works in all situations. If there was nothing tracked, all impls race and they can return in any order.

We have libdex now so it should be possible to just walk the impls when we get a change, until we find the first impl which has a result for the key. That does not require any bookkeeping between changes.

swick:
@AdrianVovk updated the PR. This now ports over to libdex and then implements this:

> We have libdex now so it should be possible to just walk the impls when we get a change, until we find the first impl which has a result for the key. That does not require any bookkeeping between changes.

swick:
also /ping @TheEvilSkeleton @kolunmi for reviews

desktop-portal/settings.c
```diff
@@ -186,154 +190,239 @@ settings_handle_read_one (XdpDbusSettings       *object,
...
+
+      result = dex_await_boxed (
+        xdp_dbus_impl_settings_call_read_future (g_ptr_array_index (self->impls, i),
+                                                 data->ns,
+                                                 data->key),
+        &error);
+
+      if (result != NULL)
+        return dex_future_new_for_boolean (FALSE);
+    }
```
kolunmi: Currently, you are awaiting the can these calls of `xdp_dbus_impl_settings_call_read_future` in sequence. Can they be run in parallel with `dex_future_allv` after collecting these futures into an array? You do this elsewhere

here is a basic example:
```c
futures = g_ptr_array_new_with_free_func (dex_unref);
for (guint i = 0; i < n; i++)
  g_ptr_array_add (futures, function_that_returns_future (array[i]));
dex_await (dex_future_allv ((DexFuture *const *) futures->pdata, futures->len), NULL);
for (guint i = 0; i < n; i++)
  {
    DexFuture *future = g_ptr_array_index (futures, i);
    const GValue *value = dex_future_get_value (future, &local_error);
    if (value != NULL)
      do_something_with_value (value);
    else
      {
        do_something_with_error (local_error);
        g_clear_error (&local_error);
      }
  }
```

desktop-portal/settings.c
```diff
@@ -186,154 +190,239 @@ settings_handle_read_one (XdpDbusSettings       *object,
...
 
   return G_DBUS_METHOD_INVOCATION_HANDLED;
 }
 
+DEX_DEFINE_CLOSURE_TYPE (SettingsChanged, settings_changed,
+                         DEX_DEFINE_CLOSURE_OBJECT (XdpSettings, self),
+                         DEX_DEFINE_CLOSURE_VALUE (size_t, impl_idx),
+                         DEX_DEFINE_CLOSURE_POINTER (char *, ns, g_free),
+                         DEX_DEFINE_CLOSURE_POINTER (char *, key, g_free),
+                         DEX_DEFINE_CLOSURE_POINTER (GVariant *, value, g_variant_unref))
```
kolunmi: What is your rationale for using this over the new `dex_scheduler_spawnv`?

TheEvilSkeleton (changes requested):
I notice that https://github.com/flatpak/xdg-desktop-portal/commit/4b114eabf13ccb070e3bfdbfc36716870aa0c903 does more than just porting to libdex fiber dispatch, which makes it harder to review. As the commit message states: "Also modernize the GObject boilerplate" — could you modernize the boilerplate in one commit and port to libdex fiber dispatch in another?

desktop-portal/settings.c
swick: Yeah, it's a bit cleaner than what I had in create_impl_proxies, so I stole it. However, what the function here does, is check if any of the higher priority impls shadow the impl the signal is from, and I think it makes sense to go through them in-order to avoid issuing unnecessary calls.

desktop-portal/settings.c
swick: Because I forgot that it exists already (:

swick:
> could you modernize the boilerplate in one commit and port to libdex fiber dispatch in another?

I don't think there is much to gain. Those ports usually are just a bit messy. What I found helps in those cases is the split-view diff viewer. In the upper right there is a diff-view settings cog thingy and you can switch it to split-view there. Then you can see the old and new version side-by-side and it's actually not too bad to review it then.

swick:
Also added cancellation to the fiber which I somehow forgot before.

swick:
Arrgh, cancellation is hard. I think I got it right now.

desktop-portal/settings.c
kolunmi: Ah yes, makes sense

TheEvilSkeleton:
> > could you modernize the boilerplate in one commit and port to libdex fiber dispatch in another?
> 
> I don't think there is much to gain. Those ports usually are just a bit messy. What I found helps in those cases is the split-view diff viewer. In the upper right there is a diff-view settings cog thingy and you can switch it to split-view there. Then you can see the old and new version side-by-side and it's actually not too bad to review it then.

Thanks, that helps a lot

swick:
This now depends on https://github.com/flatpak/xdg-desktop-portal/pull/2109 to fix the race that makes the test fail sometimes.

desktop-portal/xdp-context.c
```diff
@@ -384,7 +385,27 @@ init_portal_in_fiber (XdpContext   *context,
...
+  g_autoptr(DexFuture) all = NULL;
+
+  if (pending_inits == NULL)
+    return;
+
+  all = dex_future_allv ((DexFuture *const *) pending_inits->pdata,
+                         pending_inits->len);
+
+  while (dex_future_is_pending (all))
+    g_main_context_iteration (NULL, TRUE);
```
pwithnall: I’m not super familiar with the threading/main context model in xdg-desktop-portal, but iterating the global default main context here seems a little suspicious. It’s probably fine at the moment, but it’s generally more robust to iterate the thread-default main context instead, then the function (e.g. `xdp_context_register()`) is idempotent to main contexts being pushed higher up in its call stack.

desktop-portal/settings.c
```diff
@@ -25,6 +25,7 @@ struct _XdpSettings
   XdpDbusSettingsSkeleton parent_instance;
 
   GPtrArray *impls; /* XdpDbusImplSettings */
+  GCancellable *cancellable;
```
pwithnall:
```
GCancellable *cancellable;  /* (owned) (not nullable) */
```

I find it’s always handy to annotate class members with their ownership date and nullability. Helps make future maintenance clearer.

desktop-portal/settings.c
```diff
@@ -226,17 +227,77 @@ settings_handle_read_one (XdpDbusSettings       *object,
...
-  xdp_dbus_settings_emit_setting_changed (XDP_DBUS_SETTINGS (self),
-                                          arg_namespace, arg_key,
-                                          arg_value);
+  size_t impl_idx;
+
+  for (impl_idx = 0; impl_idx < self->impls->len; impl_idx++)
+    {
+      if (g_ptr_array_index (self->impls, impl_idx) == impl)
+        break;
+    }
```
pwithnall: I think this can be simplified to `g_ptr_array_find (self->impls, impl, &impl_idx)`.

desktop-portal/settings.c
swick: Do you do this for all fields?

pwithnall: Where I think it will clarify things, yes. Not religiously. I wouldn’t add `(owned)` for a `char *` field though, for example, as the presence/absence of `const` makes that clear. `(nullable)`/`(not nullable)` are useful on basically any pointer member though.

pwithnall:
This LGTM now

PR #2049, opened 2026-06-28
pipewire,xdp: Refactor property keys under XDP namespace

tytan652:
Centralize common keys in one header file and move away from PipeWire namespace for xdp custom properties.

PS: Tested under GNOME OS with a self-made sysext (Camera, ScreenCast)

### DIFF
--- a/desktop-portal/camera.c
+++ b/desktop-portal/camera.c
@@ -18,6 +18,7 @@
 #include "xdp-impl-dbus.h"
 #include "xdp-permissions.h"
 #include "xdp-portal-config.h"
+#include "xdp-pw-keys.h"
 #include "xdp-request.h"
 #include "xdp-utils.h"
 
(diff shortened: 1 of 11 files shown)


desktop-portal/camera.c
```diff
@@ -416,8 +417,8 @@ create_pipewire_remote (Camera *camera,
       return FALSE;
     }
 
-  pipewire_properties = pw_properties_new ("pipewire.access.xdg-desktop-portal.is_portal", "true",
-                                           "xdg-desktop-portal.monitor", "Camera",
+  pipewire_properties = pw_properties_new (XDP_PW_NAMESPACE ".is_portal", "true",
```
swick, line 420: Any reason why those props don't get their own define?

desktop-portal/meson.build
```diff
@@ -145,6 +145,7 @@ incs_xdg_desktop_portal = [
   xdp_utils_includes,
   include_directories('.'),
   include_directories('../document-portal'),
+  include_directories('../pipewire'),
```
swick: Can we rather define xdp_pipewire_includes and so on?

desktop-portal/camera.c
tytan652: They are only used there (and I plan to replace the camera monitor actual implementation).

desktop-portal/meson.build
tytan652: Maybe, I need to see if re-ordering folders in the main `meson.build` is not an issue.

tytan652: Re-ordered to add `xdp_pipewire_includes` replacing all include_directories with this variable.

PR #2050, opened 2026-07-01
Rename to meson.options and bump version

TheEvilSkeleton:


swick:
Not ideal that gitlint only seems to check the latest commit here but all commits from the merge queue...

I would suggest to run the githook locally anyway, but I'll try to fix this.

TheEvilSkeleton:
This is the following error I get when running gitlint:

```
gitlint..................................................................Failed
- hook id: gitlint
- exit code: 2

5: UC2 Line exceeds max length (233>80): "../../../../../xdg-desktop-portal/meson.build:244: WARNING: Project targets '>= 1.1' but uses feature deprecated since '0.62.0': pkgconfig.generate variable for builtin directories. They will be automatically included when referenced"
8: UC2 Line exceeds max length (142>80): "See: https://mesonbuild.com/Release-notes-for-0-62-0.html#pkgconfiggenerate-will-now-include-variables-for-builtin-directories-when-referenced"
```

I believe this is because the commit message

> This addresses the following warning:
> ```
> ../../../../../xdg-desktop-portal/meson.build:244: WARNING: Project targets '>= 1.1' but uses feature deprecated since '0.62.0': pkgconfig.generate variable for builtin directories. They will be automatically included when referenced
> ```
> 
> See: https://mesonbuild.com/Release-notes-for-0-62-0.html#pkgconfiggenerate-will-now-include-variables-for-builtin-directories-when-referenced

is too long. I prefer to keep the commit message descriptive (thus, long), but this means CI fails.

swick:
There are escape hatches. using 4 spaces at the beginning of the line for example, and footnote style (`[2]: long link`)

So something like this should work:

```
build: Don't explicitly include builtin directories

This addresses the following deprecation[1] warning:

    ../../../../../xdg-desktop-portal/meson.build:244: WARNING: Project targets '>= 1.1' but uses feature deprecated since '0.62.0': pkgconfig.generate variable for builtin directories. They will be automatically included when referenced

[1]: https://mesonbuild.com/Release-notes-for-0-62-0.html#pkgconfiggenerate-will-now-include-variables-for-builtin-directories-when-referenced
```

TheEvilSkeleton:
> \[1\]: https://mesonbuild.com/Release-notes-for-0-62-0.html#pkgconfiggenerate-will-now-include-variables-for-builtin-directories-when-referenced

Apparently, adding a colon right after the reference fails the linter, but "[1] https://mesonbuild.com/Release-notes-for-0-62-0.html#pkgconfiggenerate-will-now-include-variables-for-builtin-directories-when-referenced" works

swick:
You can adjust the python script to add a few more exceptions that make sense (and we could also use better docs about the commit message requirements). I mostly don't want to keep repeating the same thing in every review.

swick:
Thanks!

PR #2051, opened 2026-07-02
Stop using G_GNUC_CONST in _get_type functions

A6GibKm:
As per g_type_ensure's documentation it is technically incorrect to mark _get_type fns with G_GNUC_CONST since they have side-effects on their first run.

See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5223 and https://blogs.gnome.org/mcatanzaro/2026/06/29/your-_get_type-function-is-not-g_gnuc_const-part-two/ for more details.

### DIFF
--- a/desktop-portal/account.c
+++ b/desktop-portal/account.c
@@ -39,7 +39,7 @@ struct _AccountClass
   XdpDbusAccountSkeletonClass parent_class;
 };
 
-GType account_get_type (void) G_GNUC_CONST;
+GType account_get_type (void);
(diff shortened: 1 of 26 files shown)


jadahl (approved):

swick:
Please add a prefix to the commit message. I really have to fix the gitlint thing.

A6GibKm:
what would be the prefix for something that applies globally?

A6GibKm:
Went with xdp for now.

PR #2053, opened 2026-07-03
Update sl.po (Slovenian l10n)

filmsi:
This is translation update for Slovenian (sl).

### DIFF
--- a/po/sl.po
+++ b/po/sl.po
@@ -1,13 +1,15 @@
 # Slovenian translation for xdg-desktop-portal.
-# SPDX-License-Identifier: LGPL-2.1-or-later
-# SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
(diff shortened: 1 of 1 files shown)


po/sl.po
```diff
@@ -1,13 +1,15 @@
 # Slovenian translation for xdg-desktop-portal.
-# SPDX-License-Identifier: LGPL-2.1-or-later
-# SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
+# Copyright (C) 2024 xdg-desktop-portal's COPYRIGHT HOLDER
+# This file is distributed under the same license as the xdg-desktop-portal package.
+#
+# Martin Srebotnjak <miles@filmsi.net>,  2024, 2025, 2026.
 #
```
swick, line 6: This makes CI fail because it removes the SPDX license information

filmsi: So, what do you want me to do? It is a comment and it is true.

PR #2054, opened 2026-07-03
background: Find apps when monitoring starts

scarab-systems:
## Summary

- Discover existing background applications when the background monitor starts, instead of running a special portal-initialization check.
- Start the monitor cycle after the backend `RunningApplicationsChanged` signal and Flatpak instance directory monitor are connected.
- Add a regression that creates a pre-existing Flatpak instance and verifies `BackgroundApps` is populated on startup.

Fixes #1947.

## Testing

- `git diff --check`
- `PYTHONPYCACHEPREFIX=<tmp> python3 -m py_compile tests/test_background.py tests/templates/background.py`
- Not run locally: `meson test -C _build integration/background --print-errorlogs` (local environment does not have pytest or a configured Meson build dir)

### DIFF
--- a/desktop-portal/background.c
+++ b/desktop-portal/background.c
@@ -696,6 +696,14 @@ monitor_background_in_thread (GTask        *task,
                               GCancellable *cancellable)
 {
   Background *background = source_object;
+  gboolean initial_check = GPOINTER_TO_INT (task_data);
+
+  if (initial_check)
+    {
+      check_background_apps (background);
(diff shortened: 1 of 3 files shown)


swick:
That doesn't sound like the right approach to fixing the issue. Instead of monitoring always, starting the monitoring should ensure that finds all the apps, no?

swick:
Okay, I have no interest with talking to an LLM agent.

PR #2055, opened 2026-07-05
document-portal: clean up xdg_unparse_permissions

kolunmi:
The previous implementation used GPtrArray instead of GStrvBuilder to make the strv, but neither are necessary since the max possible length is known ahead of time. Thus, we can use a stack approach

### DIFF
--- a/document-portal/document-store.c
+++ b/document-portal/document-store.c
@@ -14,19 +14,20 @@
 const char **
 xdg_unparse_permissions (DocumentPermissionFlags permissions)
 {
-  g_autoptr(GPtrArray) array = g_ptr_array_new ();
+  g_autoptr(GStrvBuilder) builder = NULL;
+
+  builder = g_strv_builder_new ();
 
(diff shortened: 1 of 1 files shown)


swick:
This makes the code even more brittle than before: any change requires updating multiple places, the memdup completely erases types and requires manual size calculations. The performance argument doesn't convince me either because this hasn't been shown to be a performance problem, and it is extremely unlikely to be a performance problem because it gets called around once per portal call.

Using GStrvBuilder here would at least make the code more robust and more obviously correct.

kolunmi:
@swick understandable, I switched this pr to use GStrvBuilder instead, otherwise feel free to close

swick:
Thanks!

kolunmi:
@swick just close this pr, I had forgotten that strvbuilder calls strdup and thus this will leak, sorry

swick:
oh, yeah, it returns a new array of `const char *`...

PR #2057, opened 2026-07-07
githooks: Allow exceptions in gitlint's max-line-length rule

TheEvilSkeleton:
Currently, the rule contains separate exceptions for references
as well as tags, the latter of which are defined inside a fixed
list of keywords. As a consequence, it requires us to manually
update the list, should we add more tags in the future.
For references, they strictly disallow colons (:) after the
closing bracket. For example "\[1\]:" won't work.[1]

This patch consolidates both use cases into one and requires
following with a colon. Anything starting with a capital
letter followed by a string of characters, then followed by a colon
are treated as a tag. Additionally, number inside a bracket,
e.g. "\[1\]" are also treated as a tag.

[1]: https://github.com/flatpak/xdg-desktop-portal/pull/2050#issuecomment-4904784798

### DIFF
--- a/.gitlint.conf/custom-body-max-line-length.py
+++ b/.gitlint.conf/custom-body-max-line-length.py
@@ -4,13 +4,6 @@
 import re
 from gitlint.rules import BodyMaxLineLength
 
-TAGS = [
-    "Signed-off-by",
-    "Co-authored-by",
-    "Fixes",
-    "Closes",
-]
-
 
 class CustomBodyMaxLineLength(BodyMaxLineLength):
     name = "custom-body-max-line-length"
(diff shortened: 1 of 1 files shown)


swick:
A more generic prefix like `githooks` would be more appropriate. Check the git log on what already exists. The point here is that it should be easy to recognize what part is being changed.

Change looks good though.

swick:
while you're at it: do you mind making the exception generic for all things that look like tags and not just the fixed list?

TheEvilSkeleton:
I don't mind. I'll work on it later

TheEvilSkeleton:
Should colons be optional or mandatory when making references, e.g. "[1] ..."?

TheEvilSkeleton:
I made the use of a colon mandatory under any circumstances, but if you would like to have it be optional for references, I can change the regex a bit.

.gitlint.conf/custom-body-max-line-length.py
```diff
@@ -21,16 +25,8 @@ def validate(self, line, commit):
         if line.startswith(" " * 4):
             return None
 
-        # Ignore reference lines (e.g. [2]: https://example.org/foobar)
-        ret = re.match(r"^\[\d+\] ", line)
-        if ret is not None:
+        if re.match(TAGS, line):
```
swick: Let's move this inline here so everything is at one place, and adjust the commit message tag to githooks, and we're good here.

TheEvilSkeleton: > Let's move this inline here so everything is at one place

I forced pushed the suggestions and addressed this.

> and adjust the commit message tag to githooks

I haven't done this yet as I have a hard time parsing this. What do you mean by "adjust[ing] the commit message tag"?

swick: The commit message currently reads: "max-line-length: Make tags generic". "max-line-length" is the tag. It's too specific. Call it "githooks: Allow exceptions for generic tags in gitlint's max-line-length rule" or something like that, and optimally add a justification in the body.

TheEvilSkeleton: Thanks a lot for explaining!

swick:
Please write a comment or something when you're done addressing a reivew. github doesn't send out anything when you just push changes.

swick:
Thanks!

PR #2059, opened 2026-07-07
Improve git commit message linting and docs 

swick:


swick:
Tested locally, in CI without a PR and now here with a bad test commit. Finally works as expected.

PR #2062, opened 2026-07-08
Update Ukrainian translation

yurchor:
Thanks in advance for reviewing.


swick:
Hey, we're now strict about our commit messages, which is why CI is failing. Can you change the commit message to start with: `po: `?

yurchor:
> Hey, we're now strict about our commit messages, which is why CI is failing. Can you change the commit message to start with: `po: `?

Seems to be done. Thanks.

PR #2063, opened 2026-07-11
po: Update sl.po

filmsi:
Updated Slovenian translation.

### DIFF
--- a/po/sl.po
+++ b/po/sl.po
@@ -2,12 +2,15 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 # SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
 #
+# Translators:
+# Martin Srebotnjak <miles@filmsi.net>, 2024, 2025, 2026.
+#
 msgid ""
 msgstr ""
 "Project-Id-Version: xdg-desktop-portal main\n"
-"Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-06-07 13:57+0200\n"
-"PO-Revision-Date: 2025-10-26 09:08+0100\n"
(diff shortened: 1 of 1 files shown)


filmsi:
@swick This is now updated sl.po with the right copyright text.

swick:
Thanks. We introduced a commit message check in CI. Do you get to choose the commit message in the tools you're using or is this auto generated?

Either way, to pass the current rules you'd have to prefix it, so something like: "po: Update sl.po".

filmsi:
No, I did all manually. It would me easier for me just to send you a po file.

filmsi:
I do not use any tools, just po file editor.

swick:
well, all you have to do is change the commit message (git commit --amend) and then force push.

filmsi:
I do not use git. I only edit the po file. I am not a developer. I translate UI.

swick:
Right, that's why I'm asking how the commit message got created. Does the tool allow you to change it easily? If not, I can adjust the commit message rules.

filmsi:
No, I created the issue/merge request via web interface (https://github.com/flatpak/xdg-desktop-portal/) so I cannot rename it.

filmsi:
So, can you please fix this?

swick:
Thanks for the translations! Fixed the commit message. Would be great if you can commit with a commit message which conforms to our rules yourself to make my life a bit easier.

filmsi:
Well...
It would make the life of translators (and yours) even much easier by allowing automatic pushing from GNOME Damned Lies or automatic creation of issues from Damned Lies (like glib uses from Damned Lies to gitlab.gnome.org) if you do not trust translators for automatic pushing of translations to master.
Here is an example - I added updated translation to Damned Lies and itself it created an issue:
https://l10n.gnome.org/vertimus/glib/main/po/sl/

swick:
@filmsi do you have any pointers on how we could integrate things properly?

PR #2064, opened 2026-07-14
githooks: Run when podman and docker are both installed

iinuwa:
When both `podman` and `docker` are available, `command -v podman docker` produces

```
/usr/bin/podman
/usr/bin/docker
```

which causes the container run execution to expand to `/usr/bin/podman /usr/bin/docker inspect xdg-desktop-portal:latest ...`.

Selecting the first in the output fixes the issue


swick:
Thank you!

PR #2068, opened 2026-07-20
doc: Shorten interface member names in HTML

iinuwa:
While working on the docs for the credential portal (which are almost complete!), I noticed that the interface name is repeated throughout the method, signal and property members, but the actual interface name by itself never appears on the page (since we replace the title with the shortened name). The lack of the interface name was confusing to me when I began working with portals, and the length of the member names is especially noticeable in the right-side table-of-contents.

I'm wondering if it would be desirable to have a shortened name for the interface members too. 

Removing this would require having some sort of convention/automation to include the interface name somewhere.

I tried it out, and it looks decent in the main page and the table of contents, and the anchors all stay the same, so no broken links.

Here's a visual diff:
Left: After, Right: Before

image

### DIFF
--- a/doc/fix-rst-dbus.py
+++ b/doc/fix-rst-dbus.py
@@ -27,7 +27,7 @@ def is_section_label(line):
     return False
 
 
-def split_sections(lines, output_prefix):
+def split_sections(lines, output_prefix, interface_name):
     """Split Properties, Methods, Signals into separate fragment files.
 
     Returns (desc_lines, found_sections)."""
@@ -58,7 +58,11 @@ def split_sections(lines, output_prefix):
         with open(fragment_path, "w") as f:
             heading_written = False
             skip_transition = False
+            skip_next = False
             for line in section_lines:
+                if skip_next:
+                    skip_next = False
+                    continue
                 if is_section_label(line):
                     continue
                 # Convert ---- overline+underline to ~~~~ underline-only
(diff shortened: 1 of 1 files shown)


swick:
Yup, seems like a good idea.

swick:
@iinuwa any reason why this is a draft?

iinuwa:
@swick I had planned to double-check the HTML diff to clean up anything weird (e.g. if a prefix being removed made the reference ambiguous). But if you don't think that's necessary, I'm pretty happy with it as is.

swick:
Actually checked it out now. There are a few issues. The anchors can now collide between e.g. methods and signals (so I think it only really should change what we show in the title), and we have to show the interface name somewhere. It's entirely missing right now which is super confusing.

iinuwa:
> The anchors can now collide between e.g. methods and signals (so I think it only really should change what we show in the title)

Oh, that's surprising; I thought I had verified that the anchors don't change, just the text of the link. I'll look into that.

> and we have to show the interface name somewhere.

Would it be ok if I manually added the interface name to each portal (as in the picture screenshot above), or would you want that to be automated? It's been a while, but I think extracting that was pretty tricky.

swick:
It would be nicer if the interface was coming from the actual XML file because it reduces the risk of something going wrong.

iinuwa:
OK, looked closer into the anchor issues and addressed the problem. Now anchors are created for the long name (`org-freedesktop-portal-account-getuserinformation`) and the short one (`getuserinformation`), each pointing to the same place in the document. This means that old links will still work. As a caveat, since links on the page will point to the new anchor format, we'd break links if we reverted this.

I also added `()` as a suffix to methods to more easily distinguish them in the prose as methods instead of properties or signals.

Finally, I added the interface name to the top of the document, as extracted from the existing title.

@swick, this is ready for review now, sorry for taking so long to address.

<details>
<summary>Sample screenshot</summary>
image
</details>

---
There are a few places where are just using ` ``MemberName`` ` or `#MemberName` directly in the XML instead of `` :ref:`org.freedesktop.portal.Portal.MemberName` ``, which means they are not linked in the former case, or still use the full name of the member in the latter case. It might be nice to update those in a follow-up, but I have not addressed those here.

swick:
Looking better. There are still links which use the fully qualified name, in Clipboard::SelectionTransfer for example. RemoteDesktop::CreateSession on the other hand links to other interfaces, which is now indistinguishable.

PR #2078, opened 2026-07-29
remote-desktop: Fix check_position stream lookup and unknown stream sizes

### LINKED ISSUE #2077: NotifyPointerMotionAbsolute/NotifyTouchDown always fail with "Invalid position" for Virtual-source RemoteDesktop streams
## Bug description

Any absolute pointer/touch positioning call (`NotifyPointerMotionAbsolute`,
`NotifyTouchDown`, `NotifyTouchMotion`) against a RemoteDesktop session
backed by a `Virtual`-source ScreenCast stream (i.e. an Extend-mode /
headless session, as opposed to mirroring a real monitor) fails
unconditionally with:

    org.freedesktop.DBus.Error.Failed: Invalid position

This makes it impossible for any RemoteDesktop client to inject absolute
pointer or touch input into a virtual monitor stream at all - only
relative pointer motion works.

## Root cause

`check_position()` in `desktop-portal/remote-desktop.c` has two bugs:

1. Its `stream` parameter (the target stream id) is shadowed by a
   loop-local variable of the same name, so it's never actually used to
   select which known stream to validate the position against.
2. A `Virtual` stream's size isn't known yet when `Start()` returns (unlike
   a `Monitor` stream, whose size is known immediately from the real
   output's current mode) - it only gets fixed once the PipeWire consumer
   negotiates a format. Until then it's recorded as `(0, 0)` in
   `RemoteDesktopSession->streams`, and the bounds check `x < width` can
   never pass for `width == 0`, so every position is rejected.

## Reproduction

Request a `Virtual` source via `ScreenCast.SelectSources`, start a
combined RemoteDesktop+ScreenCast session, then call
`NotifyPointerMotionAbsolute` or `NotifyTouchDown` with the returned
stream's node id and any in-range (x, y). It always fails with "Invalid
position", with debug logging showing the matched stream's recorded size
as `(0, 0)`.

Reproduced and debugged against the 1.21.1 release.

## Fix

PR incoming: fixes both issues in `check_position()` - select the stream
by its actual pipewire node id, and treat an unnegotiated `(0, 0)` size as
"unknown, don't reject" instead of "zero-sized, reject everything".

vicajilau:
Fixes #2077

`check_position()` has two independent bugs, one commit each.

## 1. The `stream` parameter was never used

The loop variable shadowed the parameter, so the check accepted a position
that fell inside *any* of the session's streams instead of the one the
caller named.

This makes the check **stricter**: a position that is only in bounds for a
different stream of the session, or a node id that doesn't belong to the
session at all, is now rejected.

## 2. An absent `size` was treated as a zero-sized stream

`size` is Optional in the ScreenCast interface, and a backend leaves it out
for streams whose size it doesn't know at `Start()`, which is the case for
virtual streams. `collect_screen_cast_stream_data()` then left the recorded
size at `(0, 0)`, and `x < width` never holds for `width == 0`, so every
absolute position was rejected. `NotifyPointerMotionAbsolute`,
`NotifyTouchDown` and `NotifyTouchMotion` failed with "Invalid position" on
every call, leaving relative pointer motion as the only usable input.

The fix records whether the backend actually reported a size, mirroring the
existing `has_pipewire_serial` handling, and skips the bounds check when it
didn't, rather than treating `(0, 0)` as a sentinel.

## Testing

There was no coverage of absolute positioning at all. Added four cases:
in-bounds, out-of-bounds, unknown node id, and a stream without a `size`
property. The two that pin the bugs above fail on `main` and pass here.

## Open question

Should the portal be doing this bounds check at all, or is the compositor
the authority? It can only validate against what the backend chose to
report, and rejecting by default breaks the legitimate case. If you'd
rather the backend always reported a size, that's a mutter-side fix and
commit 2 could be dropped; commit 1 stands either way.

### DIFF
--- a/desktop-portal/remote-desktop.c
+++ b/desktop-portal/remote-desktop.c
@@ -863,6 +863,15 @@ check_position (XdpSession *session,
       if (screen_cast_stream_get_pipewire_node_id (stream) != node_id)
         continue;
 
+      /* The "size" stream property is optional, and backends omit it for
+       * streams whose size isn't known when the session is started, such as
+       * virtual streams. Without it we have no bounds to validate against,
+       * so accept the position and leave it to the backend; rejecting would
+       * make absolute positioning unusable on those streams.
+       */
+      if (!screen_cast_stream_has_size (stream))
+        return TRUE;
+
       screen_cast_stream_get_size (stream, &width, &height);
 
       return x >= 0.0 && x < width && y >= 0.0 && y < height;
--- a/desktop-portal/remote-desktop.c
+++ b/desktop-portal/remote-desktop.c
@@ -848,7 +848,7 @@ check_notify (XdpSession *session,
 
 static gboolean
 check_position (XdpSession *session,
-                uint32_t stream,
+                uint32_t node_id,
                 double x,
                 double y)
 {
@@ -860,11 +860,12 @@ check_position (XdpSession *session,
       ScreenCastStream *stream = l->data;
       int32_t width, height;
 
+      if (screen_cast_stream_get_pipewire_node_id (stream) != node_id)
+        continue;
+
       screen_cast_stream_get_size (stream, &width, &height);
 
-      if (x >= 0.0 && x < width &&
-          y >= 0.0 && y < height)
-        return TRUE;
+      return x >= 0.0 && x < width && y >= 0.0 && y < height;
     }
 
   return FALSE;
--- a/desktop-portal/screen-cast.c
+++ b/desktop-portal/screen-cast.c
@@ -56,6 +56,7 @@ struct _ScreenCastStream
   uint32_t id;
   int32_t width;
   int32_t height;
+  gboolean has_size;
   uint64_t pipewire_serial;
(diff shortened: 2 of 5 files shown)


vicajilau:
Heads-up since review was already requested: I've pushed a reworked version, so the earlier one is stale. Sorry for the churn.

What changed:

- **Split into two commits.** The shadowing fix and the unknown-size fix are independent problems, and the first one stands on its own.
- **Corrected the root cause.** My first version claimed the size wasn't known "until the PipeWire consumer negotiates a format". That was wrong: the portal never learns the size from PipeWire at all, it just copies whatever the backend put in the `Streams` vardict at `Start()`. The actual cause is that `size` is optional and the portal was treating "absent" as "zero".
- **Dropped the `(0, 0)` sentinel** in favour of tracking whether the property was present, mirroring the existing `has_pipewire_serial` handling.
- **Added tests.** There was no coverage of absolute positioning at all.

Details, the behaviour change this implies, and an open design question are in the updated description.

PR #2079, opened 2026-07-29
document-portal: Track permission store changes

### LINKED ISSUE #197: single-writer assumption between document portal and permission store
The document portal keeps a cache of its permission store table, and assumes that it is the only writer to it. but the permission store offers a generally available dbus api for writing, so there is really no guarantee of that. There should be some cache-invalidation signal in the PermissionStore api so the document portal can know when to reload its cache from disk.

north-echo:
Closes #197.

The document portal loads `$XDG_DATA_HOME/flatpak/db/documents` once in `main()` and never observes writes made by the permission store, which is a separate process holding the same table. `flatpak permission-remove` and `flatpak permission-reset` therefore return success and empty the canonical entry, while the running portal keeps authorizing FUSE reads *and* writes for the removed app until it is restarted.

Analysis against current `main` is in https://github.com/flatpak/xdg-desktop-portal/issues/197#issuecomment-5121068375.

Rewritten since the last review. The previous version applied the contents of the `Changed` signal and counted the portal's own writes so it could skip their echoes. @swick asked what happens when the permission store restarts and when a change happens concurrently in both processes. The second question was the real one: the counting scheme identified echoes positionally, so an external change landing inside that window was consumed as though it were ours. That is gone.

### What this does

A `Changed` for the `documents` table marks the entry stale and the portal re-reads it with `Lookup`, rather than applying the payload. The payload carries the state the store held when it applied the change, so applying it directly means telling our own echoes apart from external ones, which means pairing signals to writes. Re-reading needs neither: whoever made the change, the answer is whatever the store holds now.

That only holds while the store is not behind us, and it can be. Our writes to it are asynchronous, so between applying a change locally and the store acknowledging it, a re-read would undo what we just applied. So the portal counts the writes it has outstanding for an entry, and a `Changed` arriving while any of them is unacknowledged is recorded rather than acted on; the read happens once they are all acknowledged.

That is not the echo pairing again. Nothing is matched to anything: signals arriving in any order still produce exactly one read, taken once the store has caught up. It is closer to your dirty-on-`Changed` suggestion, with per-id granularity rather than marking the whole table.

Other details:

- Only permissions are taken. The document metadata is ours and the store keeps a mirror of it.
- The db lock is dropped before invalidating the affected FUSE nodes, as `portal_delete()` already requires.
- Every entry is re-read when the store's name owner reappears. A restart is a gap in the `Changed` stream: whatever happened while it was away was never announced, and it comes back from whatever reached disk, which can be behind what it had already told us.
- A deleted entry is recognised by the wire name of the store's error rather than with `g_error_matches()`. The D-Bus name mapping for `XDG_DESKTOP_PORTAL_ERROR` is registered lazily, on first use of the quark, so a reply decoded before anything in the process has used that domain lands in `G_IO_ERROR` and the match silently fails.

### Limitations

Eventual, not synchronous, consistency. `handle_delete_permission()` emits `Changed` before `ensure_writeout()`, so the portal catches up within a main-loop iteration, but a window remains in which the stale grant is still honoured. Staleness goes from unbounded, until the portal restarts, to bounded by main-loop and bus latency. Closing it completely would mean making one component authoritative for these decisions, which is a much larger change than this.

A restart issues one `Lookup` per held document rather than a single `List`. That is a burst at large table sizes, and `List` is an easy swap if you would prefer it.

### Testing

Three new tests, all of which fail on unpatched `main`:

- `test_external_revocation` revokes through the permission store the way the flatpak CLI does.
- `test_external_revocation_after_store_restart` does the same through an instance that replaced a killed one.
- `test_external_deletion` removes the whole entry rather than one app's permissions. The store answers a later lookup with `NotFound`, which is a different path through the portal.

`tests/conftest.py` grows a `restart_xdg_permission_store` fixture, factored out of the existing one so a restart goes through the same startup path, including the ASAN preload handling.

Results, built as CI builds it (`-Dinstalled-tests=true -Dtests=enabled -Db_sanitize=address -Db_lundef=false -Dwerror=true`):

- No compiler warnings.
- `integration/documents`: 11 passed, repeated 4 times, no flakes. With the same tests against unpatched `main`, 3 failed and 8 passed.
- The installed-tests run, `gnome-desktop-testing-runner xdg-desktop-portal/integration-documents.test`, repeated 10 times: 10 passed.

Only persistent documents are affected: `Add(fd, reuse_existing, persistent)` mirrors an entry into the permission store only when `persistent` is true, so the tests cannot use `export_file()` from the test helpers, which passes `persistent=False`.

Rebased onto current `main`.

### Disclosure

I reported the security impact of this to `flatpak-security@lists.freedesktop.org` on 2026-07-19. The maintainers elected to handle it publicly upstream rather than as an embargoed issue.

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -46,6 +46,121 @@ XdpAppInfoRegistry *app_info_registry;
 
 G_LOCK_DEFINE (db);
 
+/* The documents table is not ours alone. The permission store is a separate
+ * process writing the same table, so `flatpak permission-remove`,
+ * `flatpak permission-reset` and any other PermissionStore client can change
+ * an entry behind our back. We keep a private PermissionDb because the FUSE
+ * layer needs a synchronous answer, which means a change we do not observe
+ * leaves us authorizing access the user has already revoked.
+ *
+ * Rather than apply the contents of the Changed signal, treat the entry as
+ * stale and re-read it. The signal carries the state the store held when it
+ * applied the change, so applying it directly means telling our own echoes
+ * apart from external ones, which means pairing signals to writes. Re-reading
+ * needs neither: whoever made the change, the answer is whatever the store
+ * holds now.
+ *
+ * That only holds while the store is not behind us. Our writes to it are
+ * asynchronous, so between applying a change locally and the store
+ * acknowledging it, a re-read would undo what we just applied. A Changed
+ * arriving in that window is therefore recorded, not acted on, and the read
+ * happens once our writes are acknowledged.
+ *
+ * https://github.com/flatpak/xdg-desktop-portal/issues/197
+ */
+typedef struct
+{
+  guint    writes;   /* our writes to the store not yet acknowledged */
+  gboolean reading;  /* a Lookup is in flight */
+  gboolean queued;   /* a Changed arrived while we could not read */
+} EntrySync;
+
+/* id -> EntrySync. Guarded by the db lock. */
+static GHashTable *entry_sync;
+
+static void start_refresh (const char *id, EntrySync *sync);
+
+static EntrySync *
+entry_sync_get (const char *id)
+{
+  EntrySync *sync = g_hash_table_lookup (entry_sync, id);
+
+  if (sync == NULL)
+    {
+      sync = g_new0 (EntrySync, 1);
+      g_hash_table_insert (entry_sync, g_strdup (id), sync);
+    }
+
+  return sync;
+}
+
+static void
+entry_sync_release (const char *id,
+                    EntrySync  *sync)
+{
+  if (sync->writes == 0 && !sync->reading && !sync->queued)
+    g_hash_table_remove (entry_sync, id);
+}
+
+/* Re-read the entry, unless the store is still catching up with us or a read
+ * is already under way. Call with the db lock held. */
+static void
+refresh_entry (const char *id,
+               EntrySync  *sync)
+{
+  if (sync->writes > 0 || sync->reading)
+    {
+      sync->queued = TRUE;
+      return;
+    }
+
+  sync->queued = FALSE;
+  start_refresh (id, sync);
+}
+
+/* One of our writes to the store has been acknowledged, or failed, in which
+ * case it emits no Changed and there is nothing further to wait for. */
+static void
+store_write_cb (GObject      *source_object,
+                GAsyncResult *res,
+                gpointer      user_data)
+{
+  g_autofree char *id = user_data;
+
+  g_autoptr(GVariant) ret = NULL;
+  g_autoptr(GError) error = NULL;
+  EntrySync *sync;
+
+  ret = g_dbus_proxy_call_finish (G_DBUS_PROXY (source_object), res, &error);
+  if (ret == NULL)
+    {
+      g_dbus_error_strip_remote_error (error);
+      g_warning ("Failed to update the permission store for %s: %s", id, error->message);
+    }
+
+  XDP_AUTOLOCK (db);
+
+  sync = g_hash_table_lookup (entry_sync, id);
+  if (sync == NULL)
+    return;
+
+  if (sync->writes > 0)
+    sync->writes--;
+
+  if (sync->writes == 0 && sync->queued && !sync->reading)
+    refresh_entry (id, sync);
+  else
+    entry_sync_release (id, sync);
+}
+
+/* Call with the db lock held. */
+static void
+note_store_write (const char *id)
+{
+  entry_sync_get (id)->writes++;
+}
+
+
 char **
 xdp_list_apps (void)
 {
@@ -92,6 +207,7 @@ do_set_permissions (PermissionDbEntry    *entry,
 
   if (persist_entry (new_entry))
     {
+      note_store_write (doc_id);
       xdg_permission_store_call_set_permission (permission_store,
                                                 TABLE_NAME,
                                                 FALSE,
@@ -99,7 +215,7 @@ do_set_permissions (PermissionDbEntry    *entry,
                                                 app_id,
                                                 perms_s,
                                                 NULL,
-                                                NULL, NULL);
+                                                store_write_cb, g_strdup (doc_id));
     }
 }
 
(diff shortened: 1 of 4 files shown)


north-echo:
A note on the first commit, since a `permission-db` change is not obviously part of this.

`remove_permissions()` never unrefs the child returned by `g_variant_iter_next_value()`. That is a new reference, and `g_variant_builder_add_value()` takes its own, so the loop leaked every child it walked. `add_permissions()` already does exactly that unref, which is what convinced me it was an oversight rather than deliberate.

It was latent because nothing in the document portal called `permission_db_entry_remove_app_permissions()` — only `xdg-permission-store` reached it, from `handle_delete_permission()`. The new handler here is the first caller in a binary whose leaks CI checks, which is how it surfaced: the ASan job failed on the first push of this branch and bisecting the failure led here rather than to my own code.

I kept it as its own commit so it can be judged separately, and it is a fix worth having regardless of what happens to the rest of this PR.

For completeness, my own contribution to that failure was in the second commit and is fixed there: `store_write_cb()` treated `g_dbus_proxy_call_finish()` as returning a boolean when it returns a transfer-full `GVariant *`, leaking a reply per permission-store write.

Both ASan jobs are green now, and the full integration suite reports no leak summaries.

swick:
What happens if

* the permission store gets restarted?
* the same change happens concurrently in the permission store and the document portal?

north-echo:
Both good questions. The first is fine; the second I think you have found the weak point of the design, and I would rather change it than defend it.

### Permission store restart

I tested this: grant a persistent document to an app, confirm FUSE access, `SIGKILL` the permission store, wait for the name to drop, start a fresh instance, then revoke through the new one. The revocation takes effect on the running document portal. `GDBusProxy` follows the well-known name across owner changes, so the subscription survives and signals from the new instance arrive normally. In-flight writes at the moment of the kill fail with `NoReply`, and `store_write_cb` drops the pending count for them, so nothing is left stuck.

What the patch does *not* do is resync on restart. If the store dies with a writeout still queued, and `emit_changed()` runs before `ensure_writeout()` so a change can be announced and then lost, the new instance comes back from disk in an older state than the portal's copy and nothing announces the difference. For revocation that direction fails safe: the portal is the more restrictive of the two. The reverse case leaves the portal holding a grant the store no longer records, and a later revocation of it returns `NotFound` without emitting `Changed`, so the portal keeps it. That is a durability gap rather than something this patch introduces, but the patch would be a reasonable place to close it by reloading the table when the name owner changes.

### Concurrent change

You are right that this is where the counting scheme is weakest. The counter identifies our own echoes positionally: with N writes outstanding for an id, the next N `Changed` signals for that id are assumed to be ours. If an external change lands inside that window, it is consumed as though it were an echo, and our echo is later applied as though it were external.

Working through the orderings, the mis-pairing does not appear to leave the portal inconsistent with the store, because the payload is a full permissions snapshot rather than a delta and the count always drains, so the last signal for an id is always applied and it carries the store's current state. I ran 25 rounds of portal-side grant immediately followed by an external `DeletePermission` and lost no revocations, but I could not force the inverted ordering from a test: the portal's write is always on the bus first, so the store sees it first. So treat that as weak evidence, not a demonstration that the race is benign.

What does concern me is that the correctness of it rests on invariants that are not stated anywhere and are easy to break later: one `Changed` per write, full snapshots rather than deltas, and the count always draining. If a `Changed` is ever missed, dropped in transit or lost across a store restart at the wrong moment, the count stays high and the next external change for that id is swallowed. One event, self-healing afterwards, but that one event could be the revocation that matters.

### What I would rather do

Two changes, if you agree with the direction:

1. Match echoes by content instead of by count. Keep the last permissions map we sent to the store for each id; if an incoming `Changed` matches it, it is ours and we ignore it. Order-independent, and a missed signal cannot leave behind state that swallows a later change.

2. Reload the documents table when the permission store's name owner changes, which covers both the restart case above and any divergence we did not observe.

That keeps the shape of the patch. If you would rather go further, having the portal stop keeping an independently mutable `PermissionDb` and treat the store as authoritative for these decisions, I am happy to do that instead; it is the better end state and it is your architecture, so I would rather ask than guess.

Either way I will hold off pushing until you say which you prefer.

north-echo:
Rebased onto current main, CI green. The only conflict was the import block in tests/test_documents.py, where your reorganization collided with the import time my test needed.

swick:
> That keeps the shape of the patch. If you would rather go further, having the portal stop keeping an independently mutable PermissionDb and treat the store as authoritative for these decisions, I am happy to do that instead; it is the better end state and it is your architecture, so I would rather ask than guess.

I don't know. The document portal needs fast access to the permissions which necessitates some degree of cached state, so I don't think we will get around some sort of cache management.

The question is just how to make this really robust. The current code just seems a bit fragile to me. Maybe we should dirty all our state as dirty whenever we receive any Changed signal? There are lots of different ideas which could be explored.

north-echo:
Makes sense, your dirty-on-Changed idea is better than the delta-applying I had.

I don't know the caching side as well as you do, so treat this as how I'm thinking about it rather than a firm design: keep the PermissionDb but stop mutating it in place. On a documents Changed, mark it stale and reload on next lookup, store as source of truth.

Restart's the part I'm least sure of. When I tested a store restart the proxy followed the new owner and things kept working, but that wouldn't cover changes made while it was down, since those signals just don't arrive. So reloading on NameOwnerChanged when it reappears would close that gap. Does that line up with how you see it?

Concurrency looks like it falls out of the same thing: we write to the store too, so our own writes echo back, but stale-and-reload doesn't need to know whose change it was, it just re-reads.

I can put a version together for you to look at if the direction's right.

swick:
That might work. I also didn't dig too deep into the issue yet, so I'm not sure of how much help I am, but this might be worth exploring.

north-echo:
Pushed the rewrite, rebased onto current `main`.

The echo counting is gone. A `Changed` now marks the entry stale and the portal re-reads it from the store, so there is nothing to pair signals to and no invariant about one signal per write.

Re-reading alone was not enough, and it took two red CI runs to see why: our writes to the store are asynchronous, so between applying a change locally and the store acknowledging it the store is behind us, and a re-read in that window destroys a grant we just made. That is the same failure the counting was there to prevent. So the portal now counts its outstanding writes per entry and defers the read until they are acknowledged. Nothing is matched to anything, and signals in any order still produce exactly one read once the store has caught up.

That is your dirty-on-`Changed` idea with per-id granularity rather than the whole table. If you would rather it be whole-table, that is a smaller change and I will make it.

Restart is handled by re-reading every entry we hold when the store's name owner reappears, which also closes the durability gap I flagged as unfixed last time.

Three new tests, all failing on unpatched `main`: revocation through the store, the same through an instance that replaced a killed one, and deleting the whole entry, which takes a different path since the store answers the follow-up lookup with `NotFound`. That last one turned up a separate bug worth knowing about: `g_error_matches()` against `XDG_DESKTOP_PORTAL_ERROR` misses, because the D-Bus name mapping is registered lazily and a reply decoded before anything in the process has used that domain lands in `G_IO_ERROR`. It now compares the wire name.

Known weak points are in the description rather than buried: the consistency window is unchanged, and a restart issues one `Lookup` per held document instead of a single `List`.

PR #2080, opened 2026-07-29
Minor improvements to type checking and pre-commit configuration

TheEvilSkeleton:
See commits.

In future, I would like to use MyPy with [`--strict`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict) to force type checking throughout Python.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
     permissions:
       packages: write
     with:
-      image: xdg-desktop-portal-checks:2026-07-07.1
+      image: xdg-desktop-portal-checks:2026-07-30.1
       containerfile: containerfiles/check.containerfile
 
   build-and-test:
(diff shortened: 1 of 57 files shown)


swick:
We need to re-build the image in CI, so you have to bump the image tag here https://github.com/flatpak/xdg-desktop-portal/blob/da6c30ea6685630b7c5b3988a3702cd51e1d3a42/.github/workflows/main.yml#L24

TheEvilSkeleton:
Thanks. I pushed the above commit to see if running `pre-commit run --all-files --config containerfiles/pre-commit-config.yaml` locally yields the same results as CI, as I had errors locally and would like to see if CI will have the same errors.

TheEvilSkeleton:
Can confirm, I have the same errors as these locally: https://github.com/TheEvilSkeleton/xdg-desktop-portal/actions/runs/30488608585/job/90701677956

PR #2082, opened 2026-07-31
tests: Enforce typing

TheEvilSkeleton:
Depends on https://github.com/flatpak/xdg-desktop-portal/pull/2080.

See commits.

Since this enables strict mode in MyPy, this also introduces 891 errors.


tests/templates/geoclue2.py
```diff
@@ -21,7 +21,7 @@
 
 
 class GeoClueClient(mockobject.DBusMockObject):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
```
zeldaAhmed: I am not a huge fan of having the `None` return annotations, especially for built in method overrides. It just seems like more noise when all methods without a return type will return `None` implicitly. Maybe the XDP maintainers disagree, and im not sure if its possible to set mypy to not complain in this scenario.

tests/test_clipboard.py
```diff
@@ -105,20 +109,25 @@ def cb_activated(session_handle, options):
...
         if type == SessionType.INPUT_CAPTURE:
             return self.start_input_capture_session(dbus_con)
         raise ClipboardTypeError(dbus_con, type)
 
-    def test_request_clipboard_and_start_session(self, portals, dbus_con, type):
+    def test_request_clipboard_and_start_session(
+        self, portals, dbus_con: dbus.Bus, type: SessionType
+    ) -> None:
         _, clipboard_enabled = self.start_session(dbus_con, type)
+        assert portals == ":3"
```
zeldaAhmed: :3
(Dont forget to remove this)

tests/test_inputcapture.py
```diff
@@ -16,8 +16,7 @@
 counter = count()
 
 
-def default_zones():
-    return [(1024, 768, 0, 0), (640, 480, 1024, 0)]
+DEFAULT_ZONES = ((1024, 768, 0, 0), (640, 480, 1024, 0))
 
 
 @pytest.fixture
```
zeldaAhmed, line 25: pytest fixture methods also need return type annotation, I think

zeldaAhmed (approved):

tests/templates/geoclue2.py
TheEvilSkeleton: While it is more noise, it is more explicit. If I don't see an explicit `None`, I look into the code and see what it returns otherwise, as I just assume it was never annotated.

tests/test_inputcapture.py
TheEvilSkeleton: If it's needed, I'm sure one of the [891 errors](https://github.com/flatpak/xdg-desktop-portal/actions/runs/30664359353/job/91267763559?pr=2082#step:5:3609) will mention.

TheEvilSkeleton:
From 891 errors to 797 :face_with_spiral_eyes:

PR #2084, opened 2026-08-01
tests: Bump timeout

TheEvilSkeleton:
The "xdg-desktop-portal:integration/notification" test usually succeeds at just below the 200 seconds, which is the current timeout.

When working on #2080, the aforementioned test reached the [timeout](https://github.com/TheEvilSkeleton/xdg-desktop-portal/actions/runs/30593873504/job/91041890326#step:6:190), forcing the entire job to fail. However, rerunning it 'fixed' it. To avoid this from happening again, the timeout is being bumped to 225 seconds.

This PR has two commits: The first adds the capabilities to manually set the time for individual tests. The second bumps the timeout for the `notifications` test.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -112,6 +112,35 @@ pytest_files = [
   'test_wallpaper.py',
 ]
 
+tests = {
+  'account': {},
+  'background': {},
+  'camera': {},
+  'clipboard': {},
+  'documents': {},
+  'document_fuse': {},
+  'dynamiclauncher': {},
+  'email': {},
+  'filechooser': {},
+  'globalshortcuts': {},
+  'inhibit': {},
+  'inputcapture': {},
+  'location': {},
+  'notification': {},
+  'openuri': {},
+  'permission_store': {},
+  'print': {},
+  'registry': {},
+  'remotedesktop': {},
+  'settings': {},
+  'screencast': {},
+  'screenshot': {},
+  'shutdown': {},
+  'trash': {},
+  'usb': {},
+  'wallpaper': {},
+}
+
 template_files = [
   'templates/access.py',
   'templates/account.py',
@@ -137,12 +166,11 @@ template_files = [
   'templates/xdp_utils.py',
 ]
 
-foreach pytest_file : pytest_files
-  testname = pytest_file.replace('.py', '').replace('test_', '')
+foreach testname, params : tests
   test(
     'integration/@0@'.format(testname),
     run_test,
-    args: [meson.current_source_dir() / pytest_file] + pytest_args,
+    args: [meson.current_source_dir() / 'test_@0@.py'.format(testname)] + pytest_args,
     depends: [
       xdg_desktop_portal,
       xdp_validate_icon,
@@ -152,7 +180,7 @@ foreach pytest_file : pytest_files
     ],
     env: pytest_env,
     suite: ['integration'],
-    timeout: 200,
+    timeout: params.get('timeout', 200),
   )
 endforeach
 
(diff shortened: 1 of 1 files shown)


TheEvilSkeleton:
I'm marking this as draft as I would like to run the job 15 times to check if "xdg-desktop-portal:integration/notification" exceeds the timeout when it goes over 200 seconds.

TheEvilSkeleton:
Looks like most, if not all the jobs are running under 150 seconds, so I'll just mark it as ready for review

swick:
How about we make it possible to override the timeout of specific tests? The problem with bumping all to a high timeout is that when something starts hanging, you have to wait a long time before it becomes visible as a failure.

TheEvilSkeleton:
Pushed. What are your thoughts on this implementation?

tests/meson.build
```diff
@@ -83,34 +83,35 @@ if pymod.find_installation('python3', modules: ['xdist'], required: false).found
...
-  'test_screencast.py',
-  'test_screenshot.py',
-  'test_shutdown.py',
-  'test_trash.py',
-  'test_usb.py',
-  'test_wallpaper.py',
-]
+default_timeout = 200
+pytest_files = {
+  'test_account.py': default_timeout,
```
swick: I think you could do better with adding something like this:

```
pytests = {
  'account': {},
  'background': {'timeout': 250},
}

test(
  timeout: test.get('timeout', 200),
)
```

and then use pytest_files/template_files only for installing.

TheEvilSkeleton: Indeed, that's much cleaner

TheEvilSkeleton:
> tests/meson.build:188:24: ERROR: Foreach expects exactly 2 variables for iterating over objects of type dict

Do you happen to know what this error means?

Edit: never mind, I think I figured it out

tests/meson.build
```diff
@@ -137,7 +137,9 @@ template_files = [
   'templates/xdp_utils.py',
 ]
 
-foreach pytest_file : pytest_files
+foreach i : range(pytest_files.keys().length())
```
swick: `foreach file, params : pytest_files` should work

tests/meson.build
```diff
@@ -83,34 +83,34 @@ if pymod.find_installation('python3', modules: ['xdist'], required: false).found
...
-  'test_remotedesktop.py',
-  'test_settings.py',
-  'test_screencast.py',
-  'test_screenshot.py',
-  'test_shutdown.py',
-  'test_trash.py',
-  'test_usb.py',
-  'test_wallpaper.py',
-]
+pytest_files = {
```
swick: I think it would be cleaner to separate the file tracking we need to install the tests from the test definitions.

So, `pytest_files` should track the actual files we need to install (including `__init__.py` and so on), and `tests` should be a map of `test name` -> `options`.

Yes, more code, but cleaner.

swick (commented):
getting there

tests/meson.build
TheEvilSkeleton: > and `tests` should be a map of `test name` -> `options`.

What is "`options`"? Did you mean timeout?

swick: the dict which defines the test:
```
tests = {
  'account': {'timeout': 225},
}


foreach testname, params : tests
  test(
    'integration/@0@'.format(testname),
    args: [meson.current_source_dir() / 'test_@0@.py'.format(testname)] + pytest_args,
    timeout: params.get('timeout', 200),
  )
endforeach
```

testname = pytest_file.replace('.py', '').replace('test_', '')

TheEvilSkeleton: Thanks

TheEvilSkeleton:
Done

swick:
Thank you!

PR #2087, opened 2026-08-03
Begin porting utils to Dex API

kolunmi:
continuation: see discussion on https://github.com/swick/xdg-desktop-portal/pull/2


shared/xdp-utils.c
```diff
@@ -149,11 +149,38 @@ xdp_pid_fd_result_new (uint32_t pid,
   return result;
 }
 
+static XdpPidFdResult *
+fiber_helper_get_pidfd_legacy (GDBusConnection *connection,
```
swick: Not a huge fan of the name. Probably should be something along the lines of `connection_get_pidfd_legacy` but maybe with fiber in the name?

swick:
Please squash the fixups into the appropriate commits (I guess there is only one here).

shared/xdp-utils.c
kolunmi: I usually like to prefix functions that must be run in a fiber with `fiber_` so it is absolutely clear they cannot be run in another context

swick: Sure, but the other one has a `_fiber` postfix. We should at least be consistent. I think there are other places already which have a postfix?

kolunmi: That's the "root" fiber function, but now that the trampolining is implemented, there is nothing really distinguishing them. So `connection_get_pidfd_legacy_fiber` it is

kolunmi:
@swick done

swick:
I still see a bunch of fixup commits.

kolunmi:
ah I see, I misunderstood you

kolunmi:
should be good now

swick:
The commit message is not exactly great now.

shared/xdp-app-info.h
```diff
@@ -25,6 +26,11 @@ G_DECLARE_DERIVABLE_TYPE (XdpAppInfo,
                           XDP, APP_INFO,
                           GObject);
 
+DexFuture * xdp_app_info_new_for_invocation (GDBusMethodInvocation *invocation);
+
+DexFuture * xdp_app_info_new_for_registered (GDBusMethodInvocation *invocation,
```
swick, line 31: Just noticed this is unused. The Registry portal is still using the sync variant.

Both the sync app info constructors would then become dead code and could be removed in the next commit.

kolunmi:
updated

shared/xdp-app-info.h
kolunmi: Yes, I included that in anticipation for the registry portal to use it later. What action would you like me to take here?

swick: Should be trivial to change the registry portal just like you changed the app info registry. Maybe make this PR 3 commits: first add the future variants of the app info constructors, then in the second replace all the users of the sync variant, and then remove the dead code in the third.

shared/xdp-app-info.c
```diff
@@ -370,6 +370,78 @@ xdp_app_info_new (const char  *sender,
...
+xdp_app_info_new_for_invocation (GDBusMethodInvocation *invocation)
+{
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  const char *sender = g_dbus_method_invocation_get_sender (invocation);
+
+  return dex_future_then (xdp_connection_get_pidfd (connection, sender),
+                          app_info_new_for_invocation_then,
+                          g_object_ref (invocation),
+                          g_object_unref);
+}
```
zeldaAhmed, line 402: This is more of a question then a comment, but would public methods like these require documentation? For my sake as someone whos understanding of Libdex isn't the best, it would definitely help 😄

zoeyAhmed (approved):

shared/xdp-app-info.c
swick: If you have suggestion, feel free to add them. I would be against adding documentation on libdex concepts but not against higher level descriptions and documentation on the expected types of the future.

shared/xdp-app-info.h
kolunmi: @swick done, let me know if it needs further adjustments

kolunmi:
ah CI failed, give me a moment

swick:
Thank you!

PR #2088, opened 2026-08-03
trash: Allow trash directories to have any mode

swick:
The code was over-eager with what it considers a valid trash directory. While we should make sure that we only create trash dirs which are private to the user (so 0700), there appear to be enough systems where the trash directories have other permissions that we should allow them to be used.

We still emit warnings about this, because it might be unexpected that trashed files are readable or writeable by other users.

Fixes: 37aaaeac ("trash: Implement trashing securely")


GeorgesStavracas (approved):

swick:
Thanks!

Issue #2033, opened 2026-06-17
Duplicate Settings Backend Resolution Breaks Runtime Theme-Switching


MyChaOS87:
### Operating System

ArchLinux

### XDG Desktop Portal version

1.22

### Desktop Environment

Other

### Desktop Environment (Other)

niri

### Expected Behavior

- If `org.freedesktop.impl.portal.Settings` is explicitly mapped/resolved to `gtk` in a high-priority user portals config (e.g. `niri-portals.conf`), only `xdg-desktop-portal-gtk` should be loaded as the settings backend. Dynamic runtime theme changes (e.g., toggling light/dark mode via GSettings) should propagate smoothly to GTK and Electron/Chromium applications.
- **This is the correct behavior observed in version `1.20.4`**, where only `gtk.portal` is loaded for the `Settings` interface and theme switching works perfectly. 

### Current Behavior

- Both `gtk` and `gnome` settings backends are loaded simultaneously. This is because the configuration merging loop in `xdg-desktop-portal` version `1.22.0` falls back to merging default backends from lower-priority configuration files (like the system-wide fallback `/usr/share/xdg-desktop-portal/portals.conf`) even when an explicit preference was already resolved.
- Running both settings backends concurrently under a non-GNOME environment (like the `niri` compositor) leads to duplicate and conflicting `SettingChanged` D-Bus signals, which breaks runtime theme-switching in GTK (Libadwaita) and Electron applications.

### Steps to Reproduce

1. Set up a Wayland session using a non-GNOME compositor (e.g., `niri`) with the following `~/.config/xdg-desktop-portal/niri-portals.conf`:
   ```ini
   [preferred]
   default=gnome;gtk;
   org.freedesktop.impl.portal.Access=gtk;
   org.freedesktop.impl.portal.Notification=gtk;
   org.freedesktop.impl.portal.Secret=gnome-keyring;
   org.freedesktop.impl.portal.Settings=gtk;
   ```
2. Launch `xdg-desktop-portal` version `1.22.0` with `G_MESSAGES_DEBUG=all` and observe that both backends are loaded for settings:
   ```
   Using gtk.portal for org.freedesktop.impl.portal.Settings (config)
   Using gnome.portal for org.freedesktop.impl.portal.Settings (config)
   ```
3. Subscribe to the D-Bus settings signal:
   ```bash
   dbus-monitor "type='signal',interface='org.freedesktop.portal.Settings',member='SettingChanged'"
   ```
[...]

PR #2097, opened 2026-08-10
portal-config: Read multiple config files from each directory

maxcabrajac:
Hi!
This is a small follow up to #1867.
On #1867, we started taking into account multiple config files instead of just the first one found, but I feel like we made an oversight when we kept reading a single config file per directory.

Current behavior:
A person uses gnome and kde with a DE-agnostic user-level portal config.
After a while, they want to try out cosmik and want to override the portal used for a single interface on it and not on the other DEs.
They write a cosmik-portals.conf changing it and now all DE-agnostic portal configs they have are ignored on cosmik.

Expected behavior:
cosmik specifig configuration are overlayed over the user's DE-agnostic configs.

### DIFF
--- a/desktop-portal/xdp-portal-config.c
+++ b/desktop-portal/xdp-portal-config.c
@@ -374,9 +374,9 @@ out:
 }
 
 static PortalConfig *
-load_portal_configuration_for_dir (gboolean    opt_verbose,
-                                   const char *base_directory,
-                                   const char *portal_file)
+load_config_file (gboolean    opt_verbose,
+                  const char *base_directory,
+                  const char *portal_file)
 {
   g_autoptr(GKeyFile) key_file = NULL;
   g_autofree char *path = NULL;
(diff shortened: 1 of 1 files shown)


swick:
Have you seen https://github.com/flatpak/xdg-desktop-portal/pull/2048 ?

maxcabrajac:
Hey @swick !

I took a look at #2033 and #2048 and got that the issue was partially caused by a mistake on #1867 (sorry for that =/), but I don't really see how they correlate to this PR, could you elaborate?

This PR just adds more files to the set of files consumed by the xdp config resolution code.

swick:
> They write a cosmik-portals.conf changing it and now all DE-agnostic portal configs they have are ignored on cosmik.

> Expected behavior:
> cosmik specifig configuration are overlayed over the user's DE-agnostic configs.

Yeah, no, I don't think this is reasonable. There isn't a "DE-agnostic config". If you use a DE, you write a portal config for it.

maxcabrajac:
https://github.com/flatpak/xdg-desktop-portal/blob/a68391a7eaaada041bed30157ed391534b5b462a/desktop-portal/xdp-portal-config.c#L472-L486

What I meant by "DE-agnostic config" was `portals.conf` (unprefixed). It is used if no DE specific (with a `<DE>-` prefix) exists.

swick:
I understood, but there never was an attempt at making those config files overlays. They are supposed to be per-DE configs, and `portals.conf` is a fallback for when there isn't a DE-specific config.

PR #2100, opened 2026-08-12
Changes to XdpRequestDex API

swick:
See commits


pwithnall (approved):
It would be useful to explain _why_ you’re making the change in “request-dex: Make exporting the frontend request explicit”, in the commit message.

Otherwise, this LGTM

swick:
Good point, added this:

```
Often it's required to provide the backend with a request id/handle
which RequestDex provides, but to avoid races, we have to wait until
some time after issuing the call to the backend, before exporting the
request to apps.
```

PR #2101, opened 2026-08-12
Fix a race between apps Request.Close() and the impl creating its ImplRequest

### LINKED ISSUE #854: Closing a request can fail due to backend not having created a request object yet.
An application quickly closing a request (e.g. to cancel a file chooser dialog) can result in a race condition where this close event is not handled in the backend.

For example:
1. An application calls `org.freedesktop.portal.FileChooser.OpenFile`.
2. `xdg-desktop-portal` creates an `org.freedesktop.portal.Request` object to handle this request. It listens to the `Close` method call on this object.
3. `xdg-desktop-portal` calls `org.freedesktop.impl.portal.FileChooser.OpenFile` and passes the object path of the request create in step 2.
4. `xdg-desktop-portal` returns the path of the request object to the application.
5. Once `xdg-desktop-portal-gnome` receives the `org.freedesktop.impl.portal.FileChooser.OpenFile` request it creates an `org.freedesktop.impl.portal.Request` with the provided object path. It listens for the `Close` method call on this object to close the dialog.

If the application calls `org.freedesktop.portal.Request.Close` before step 5 above, `xdg-desktop-portal` will receive this and then call `org.freedesktop.impl.portal.Request.Close` and fail due to the backend not having yet created this object:
```
GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path “/org/freedesktop/portal/desktop/request/1_6909/dart2665461754”
```

swick:
Fixes https://github.com/flatpak/xdg-desktop-portal/issues/854.

The big issue is that this requires a signal from the impl Request for the frontend to know that it's safe to export the Request to the app. We also don't have versioning on the impl Request interface, so we need the portal to provide the Request version.

Unfortunately all of this means that the frontend must rewrite the portal with libdex, then add a property to the portal, and then impls need to support impl Request v2.

Depends on: https://github.com/flatpak/xdg-desktop-portal/pull/2100

/cc @jadahl @AdrianVovk or anyone else who wants to dig into this mess

### DIFF
--- a/data/org.freedesktop.impl.portal.Account.xml
+++ b/data/org.freedesktop.impl.portal.Account.xml
@@ -59,5 +59,16 @@
       <annotation name="org.qtproject.QtDBus.QtTypeName.Out1" value="QVariantMap"/>
       <arg type="a{sv}" name="results" direction="out"/>
     </method>
+
+    <!--
+        request-version:
+
+        The version of the :doc:`org.freedesktop.impl.portal.Request`
+        interface supported by this backend. See that interface's
+        documentation for details.
+
+        If this property is not present, version 1 is assumed.
+    -->
+    <property name="request-version" type="u" access="read"/>
   </interface>
 </node>
--- a/data/org.freedesktop.impl.portal.Request.xml
+++ b/data/org.freedesktop.impl.portal.Request.xml
@@ -17,6 +17,13 @@
 
       The portal can abort the interaction by calling
       :ref:`org.freedesktop.impl.portal.Request.Close` on the Request object.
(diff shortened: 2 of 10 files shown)


swick:
Same issue exists in the Session. If we want to got down this route, we should do the same for Session.

AdrianVovk:
I wonder: should we bother going through and fixing this comprehensively like this in the portals v1 codebase? Or could we take on a little tech debt, and do the much easier solution of just keeping track of the failed Close() request and trying again a few seconds later?

PR #2099, opened 2026-08-11
Documentation cleanup

swick:
This is directly out of an LLM. Someone should probably look over the functional changes.

PR #2102, opened 2026-08-12
data: Improve Clipboard, InputCapture, RemoteDesktop, ScreenCast docs

swick:
(only the last commit is relevant, sits on top of https://github.com/flatpak/xdg-desktop-portal/pull/2099)

This is straight out of an LLM, but I'm not the right person to review this properly. @whot @jadahl, please review, completely rewrite, or do whatever else with this, but I think that the idea of expanding on the docs for those portals is useful.

### DIFF
--- a/data/org.freedesktop.portal.RemoteDesktop.xml
+++ b/data/org.freedesktop.portal.RemoteDesktop.xml
@@ -12,6 +12,62 @@
       The Remote desktop portal allows to create remote desktop sessions.
 
       This documentation describes version 2 of this interface.
+
+      Session lifecycle
+      ~~~~~~~~~~~~~~~~~
+
+      A remote desktop session is created with ``CreateSession()``. The
+      application then calls ``SelectDevices()`` to choose which device
+      types to remote control (keyboard, pointer, touchscreen). Calling
+      ``Start()`` presents a dialog to the user and, upon success, returns
+      the set of devices granted.
+
+      Cross-portal integration
+      ~~~~~~~~~~~~~~~~~~~~~~~~
+
+      A remote desktop session can be used with other portals by passing
+      the session created here to their methods:
+
+      - :ref:`org.freedesktop.portal.ScreenCast`: Call ``SelectSources()``
+        and ``OpenPipeWireRemote()`` with the remote desktop session to
+        capture screen content alongside input control. The session must
+        be started and stopped through this portal.
+
+      - :ref:`org.freedesktop.portal.Clipboard`: Call
+        ``RequestClipboard()`` before starting the session to enable
+        clipboard sharing.
+
+      Input methods
+      ~~~~~~~~~~~~~
+
+      There are two ways to send input events to the remote desktop
+      session:
+
+      - **EIS (recommended):** Call ``ConnectToEIS()`` after starting the
+        session to obtain a file descriptor for a libei sender context.
+        Input events are sent via the EI protocol. Once an EIS connection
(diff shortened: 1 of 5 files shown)


TheEvilSkeleton:
> @whot @jadahl, please review, completely rewrite, or do whatever else with this, but I think that the idea of expanding on the docs for those portals is useful.

Have you asked them if they're comfortable reviewing LLM-generated commits, and have they confirmed that they're comfortable with them?

swick:
I'm not going to ask for permission to ask for something.

TheEvilSkeleton:
I don't like telling people what they should or should not do, but when it comes to something that is inherently impersonal like prompting an LLM and pinging maintainers to untangle the potential mess, I would highly suggest asking for consent beforehand.

I hope I'm not crossing anyone's boundaries by commenting this, and I am not speaking on the behalf of @whot or @jadahl. If I were to be a maintainer, I'd be really uncomfortable reviewing this especially if it was coming from another maintainer, because 1. it is inherently an impersonal process: what's being shoved isn't code written by a human being but by a model, so there was no thought process, and therefore the submitter cannot give a genuine answer if they were to be asked why it was written this way; and 2. because of the previous point, I would feel stranded trying to untangle the potential mess created by the LLM without a proper point of contact.

Additionally, it would be hard for me to refuse to review if I were a maintainer, because this would potentially create a conflict between at least two maintainers which is something every maintainer would prefer to avoid.

In any case, if @whot and @jadahl are not comfortable reviewing this (understandably), then I'll review it.

swick:
> so there was no thought process

I certainly had a thought process when I prompted the LLM to create this. The result won't be much worse (if any) than I would have produced with my limited understanding of the specific portals. I'm explicitly not forcing anyone to do anything here either.

> Additionally, it would be hard for me to refuse to review if I were a maintainer, because this would potentially create a conflict between at least two maintainers which is something every maintainer would prefer to avoid.

I have enough trust in both of them to tell me when they are unhappy with something I've done.

jadahl:
I don't think unfiltered LLM output should be expected be reviewed by anyone other than the one submitted the change, before it was submitted. Looking at the commit message it's also hard to know what was improved, and in what way, and why.

swick:
Like I said in the PR description: do whatever you want with the generated output. I don't care if you use it or not. What I care about is that those portals get proper documentation. They are clearly missing a high level description on how everything fits together. 

Those portals are not my area of expertise, but yours and Peters, so I expect you to improve the docs one way or the other.

data/org.freedesktop.portal.RemoteDesktop.xml
```diff
@@ -12,6 +12,62 @@
...
+      - **D-Bus Notify methods:** Use the various ``Notify*`` methods
+        (``NotifyPointerMotion()``, ``NotifyKeyboardKeycode()``, etc.) to
+        send input events over D-Bus.
+
+      Session persistence
+      ~~~~~~~~~~~~~~~~~~~
+
+      Remote desktop sessions support persistence using the
+      ``persist_mode`` and ``restore_token`` options in
+      ``SelectDevices()``. Persistence for combined remote desktop and
```
whot, line 60: fwiw, the persistence bits are missing from the inputcapture documentation

whot (approved):
Last commit only but this is good enough IMO. Doesn't read like an LLM and while it's not perfect it's much better than what we had before. Thanks.

swick:
Thanks Peter! Rebased this on the rest of the doc changes and fixed the doc generate script to deal with empty methods/signals/properties sections.

PR #2106, opened 2026-08-13
file-transfer: Don't truncate parent inode numbers

### LINKED ISSUE #2105: FileTransfer rejects files when the parent inode does not fit in a signed int
### Operating System

Arch Linux

Other

1.22.1

### Desktop Environment

KDE Plasma

### Expected Behavior

Dragging a local file from Dolphin into a sandboxed Flatpak application should export the file through the FileTransfer/Documents portal and provide the application with an accessible `/run/flatpak/doc/...` path, regardless of the parent directory's inode number.

### Current Behavior

Drag-and-drop works from some directories but fails consistently from others on XFS.

The error is:

`kf.coreaddons: Failed to retrieve files from portal: QDBusError("org.freedesktop.portal.Error.NotAllowed", "Invalid parent directory")`

The Flatpak then receives the inaccessible host path instead of a document-portal path.

The behavior correlates with the parent directory inode:
```
/home/admin/Downloads
inode=537159281
works

/home/admin/Downloads/test1
inode=537213101
works

/home/admin/Downloads/test2
inode=2952790157
fails

/home/admin/Downloads/test3
inode=5641209410
fails
```

### Steps to Reproduce

1. On an XFS filesystem, create/find a directory whose inode is greater than INT_MAX.
2. Put an image in that directory.
3. Open Dolphin (or presumably any file manager under any DE/WM)
4. Open a sandboxed Flatpak application which accepts files through drag-and-drop with no broad host filesystem permission. (io.github.gimletlove.imagecompare in my case) 
5. Drag the image from Dolphin into the Flatpak application.
6. Drag and drop fails with `org.freedesktop.portal.Error.NotAllowed: Invalid parent directory` 
7. Repeat from a directory whose inode fits in a signed 32-bit integer; the drop succeeds.

The apparent cause is xdg-document-portal storing st_ino and st_dev in a signed int instead of ino_t and dev_t and passing them to document_add_full() as int *, instead of using ino_t and dev_t throughout.

Although reproduced on XFS, ext4 inode numbers are unsigned 32-bit and may also exceed INT_MAX.

gimletlove:
FileTransfer stores st_dev and st_ino in int before passing them to document_add_full(). This truncates inode numbers above INT_MAX and causes parent directory validation to fail.

Use dev_t and ino_t throughout.

Closes #2105

### DIFF
--- a/document-portal/document-portal.c
+++ b/document-portal/document-portal.c
@@ -868,8 +868,8 @@ portal_add_full (GDBusMethodInvocation *invocation,
  */
 char **
 document_add_full (int                      *fd,
-                   int                      *parent_dev,
-                   int                      *parent_ino,
+                   dev_t                    *parent_dev,
+                   ino_t                    *parent_ino,
                    DocumentAddFullFlags     *documents_flags,
                    int                       n_args,
                    XdpAppInfo               *app_info,
--- a/document-portal/document-portal.h
+++ b/document-portal/document-portal.h
@@ -28,8 +28,8 @@ gboolean validate_fd (int           fd,
                       GError      **error);
 
 char ** document_add_full (int                      *fd,
-                           int                      *parent_dev,
-                           int                      *parent_ino,
+                           dev_t                    *parent_dev,
+                           ino_t                    *parent_ino,
                            DocumentAddFullFlags     *documents_flags,
                            int                       n_args,
                            XdpAppInfo               *app_info,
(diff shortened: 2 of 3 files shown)


GeorgesStavracas (approved):
LGTM. I don't know how this can be made into a test though...

gimletlove:
> LGTM. I don't know how this can be made into a test though...

I skipped an automated test because i don't know of a way to allocate an inode number to a directory either, they are allocated by the fs. 
The fix was verified on my personal system where i ran into this issue in the first place.

Issue #2091, opened 2026-08-05
ScreenCast: OpenPipeWireRemote handler deadlocks the portal after a monitor reconfiguration


tonymitchell:
### Operating System

Ubuntu 26.04 (GNOME Shell 50.1-0ubuntu1.1, mutter 50.1-0ubuntu2.2), Wayland, libpipewire 1.6.2, 4x 2560x1440

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.21.1+ds-1ubuntu3 (backend: xdg-desktop-portal-gnome 50.0-0ubuntu1)

### Desktop Environment

GNOME

### Expected Behavior

When a monitor is switched off or on while a client holds a ScreenCast session,
the compositor closes the session, the client creates a new one, and the portal
carries on serving D-Bus as normal.

### Current Behavior

An OpenPipeWireRemote handler blocks forever tearing down its PipeWire context,
while holding a session lock the main loop needs. Every portal method then hangs
for every client in the session -- file choosers, screenshots, screen sharing --
not just the capture client.

Worker thread, stuck joining a PipeWire thread that never exits:

    #5  __pthread_clockjoin_ex
    #6  pw_thread_loop_stop            libpipewire-0.3
    #7  pw_thread_loop_destroy         libpipewire-0.3
    #8  libpipewire-module-rt.so
    #9  pw_impl_module_destroy
    #10 pw_context_destroy
    #11 pipewire_remote_destroy        src/pipewire.c:232
    #12 handle_open_pipewire_remote    src/screen-cast.c:1064
    #18 _xdp_dbus_screen_cast_skeleton_handle_method_call
    #19 dispatch_in_thread_func        gio/gdbusinterfaceskeleton.c:548

Main loop, blocked on that session's mutex as soon as any peer disconnects:

    #1  g_mutex_lock_slowpath (mutex=0x7795d0005cb8)
    #3  g_mutex_locker_new   (mutex=0x7795d0005cb8)
    #4  on_peer_disconnect   (peer=":1.45320", user_data=0x7795d0005c90)
                                          src/xdp-session.c:280
    #10 on_peer_disconnect                src/xdp-context.c:319
    #11 name_owner_changed                src/xdp-utils.c:167
    #12 emit_signal_instance_in_idle_cb   gio/gdbusconnection.c:4207
    #16 g_main_loop_run                   src/xdp-main.c:244

[...]

PR #2107, opened 2026-08-14
open-uri: Don't hold the request lock across ShowItems

### LINKED ISSUE #2093: OpenDirectory deadlocks for 25s when the file manager must be D-Bus activated (portal blocks on its own mutex in on_peer_disconnect)
## Summary

`OpenURI.OpenDirectory` takes **exactly 25 seconds** whenever the file manager has to be D-Bus activated (i.e. Nautilus is not already running). If Nautilus is already running the same call completes in ~0.2 s.

The portal's main thread is blocked on **its own mutex** in `on_peer_disconnect()`, reached from `name_owner_changed` — which fires precisely because the Nautilus it just activated appeared on the bus. Meanwhile that new Nautilus is blocked in a *synchronous* D-Bus call back into the portal from `gtk_init_check()`. Neither side proceeds until the 25 s GDBus default timeout expires.

This looks like the same deadlock class as #636 ("Slow startup due to deadlock if o.fd.FileManager1 is implemented by GtkApplication", fixed via #639), resurfacing on the `OpenDirectory` path.

## Versions

- `xdg-desktop-portal` **1.22.1**
- `xdg-desktop-portal-gnome` 50.0
- `nautilus` 50.2.2
- `gtk4` 4.22.4
- GNOME Shell 50.3, Wayland
- Fedora 44 (Nobara), stock portal configuration (`gnome-portals.conf`, gnome + gtk backends only)

## Reproducer

```python
import gi, os, sys, time
gi.require_version("Gio", "2.0")
from gi.repository import Gio, GLib

path = sys.argv[1]
fd = os.open(path, os.O_RDONLY | os.O_CLOEXEC)
fdlist = Gio.UnixFDList.new()
idx = fdlist.append(fd)
bus = Gio.bus_get_sync(Gio.BusType.SESSION, None)
loop = GLib.MainLoop()
t0 = time.time()

def on_resp(conn, sender, obj, iface, signal, params):
    print("RESPONSE after %.2f s  code=%s" % (time.time() - t0, params.unpack()[0]))
    loop.quit()

bus.signal_subscribe("org.freedesktop.portal.Desktop",
    "org.freedesktop.portal.Request", "Response", None, None,
    Gio.DBusSignalFlags.NONE, on_resp)

bus.call_with_unix_fd_list_sync(
    "org.freedesktop.portal.Desktop", "/org/freedesktop/portal/desktop",
    "org.freedesktop.portal.OpenURI", "OpenDirectory",
    GLib.Variant("(sha{sv})", ("", idx, {})),
    GLib.VariantType("(o)"), Gio.DBusCallFlags.NONE, 120000, fdlist, None)
loop.run()
```

```bash
# cold -- Nautilus not running
pkill -x nautilus; sleep 3
python3 repro.py ~/Downloads
#   RESPONSE after 25.03 s  code=0

# warm -- Nautilus already running
nautilus --gapplication-service & sleep 5
python3 repro.py ~/Downloads
#   RESPONSE after 0.23 s  code=0
```

Note the async shape: `OpenDirectory` returns its `Request` handle in ~0 s in **both** cases. The stall is only visible by subscribing to `org.freedesktop.portal.Request.Response`, which is easy to miss when measuring.

A direct `org.freedesktop.FileManager1.ShowItems` call that bypasses the portal is **0.02 s** cold or warm, so Nautilus itself is never slow.

Journal, one line per occurrence, always exactly 25.0 s after the call:

```
xdg-desktop-portal[4336]: Failed to call ShowItems: Timeout was reached
```

After the timeout the portal falls back to launching the file manager again, so the user gets **two** windows.

## Backtraces during a live stall

`xdg-desktop-portal` main thread (`eu-stack -p <pid>`). Frames at `0x5605…` are the portal binary itself:

```
TID 4336:
#0  0x00007f31075d637d syscall
#1  0x00007f3107c151e2 g_mutex_lock_slowpath
#2  0x0000560591720462 on_peer_disconnect.lto_priv.2
#3  0x00007f3107954e7c g_closure_invoke
#4  0x00007f3107973849 signal_emit_unlocked_R.isra.0
#5  0x00007f3107975a28 signal_emit_valist_unlocked
#6  0x00007f3107975cb8 g_signal_emit_valist
#7  0x00007f3107975d73 g_signal_emit
#8  0x000056059171719e on_peer_disconnect.lto_priv.1
#9  0x000056059172138b name_owner_changed.lto_priv.0
#10 0x00007f3107a93d30 emit_signal_instance_in_idle_cb
#11 0x00007f3107be2524 g_idle_dispatch
#12 0x00007f3107be0f24 g_main_context_dispatch_unlocked.lto_priv.0
#13 0x00007f3107be5038 g_main_context_iterate_unlocked.isra.0
#14 0x00007f3107be52e7 g_main_loop_run
#15 0x00005605916af093 main
```

The freshly activated `nautilus` main thread at the same moment (`Threads: 1` on the first sample — still in the dynamic linker):

```
TID 2214949:
#0  0x00007fce52b45312 __syscall_cancel_arch
#3  0x00007fce52bb3426 ppoll
#4  0x00007fce54c42125 g_main_context_iterate_unlocked.isra.0
#5  0x00007fce54c422e7 g_main_loop_run
#6  0x00007fce54ea1b7f g_dbus_connection_send_message_with_reply_sync
#7  0x00007fce54eb2c68 g_dbus_connection_call_sync_internal
#8  0x00007fce54eb2fd6 g_dbus_connection_call_sync
#9  0x00007fce53e8cf1b check_portal_interface
#10 0x00007fce53e467e9 _gdk_wayland_display_open
#11 0x00007fce53eabdc6 gdk_display_manager_open_display
#12 0x00007fce53b2eb8b gtk_init_check
#13 0x000056443289fb5f nautilus_application_startup
#17 0x00007fce54e7fe1a g_application_register
#19 0x00007fce54e80325 g_application_run
#20 0x0000564432898e73 main
```

Sampling the portal every 3 s across a stall shows its main thread deep in a nested `g_main_loop_run` (frame #14) from t≈6 s through t≈21 s, returning to a normal `ppoll` at t≈24 s when the timeout fires.

## Analysis

1. `OpenDirectory` → portal calls `org.freedesktop.FileManager1.ShowItems`.
2. Nobody owns that name, so D-Bus activates `nautilus --gapplication-service`.
3. Nautilus appears on the bus → `NameOwnerChanged` → the portal's `name_owner_changed` → `on_peer_disconnect` → blocks acquiring a lock.
4. Nautilus, still in `gtk_init_check()`, makes a **synchronous** call to the portal via GTK's `check_portal_interface()`.
5. Deadlock until the 25 s GDBus timeout.

Any GTK 4 application owning `org.freedesktop.FileManager1` should hit this, since GTK 4 on Wayland unconditionally queries the portal during display open and there is no longer a `GTK_USE_PORTAL` escape hatch.

Practically this is reached from Chrome's "Show in folder", which routes through `OpenDirectory`. Launching the file manager from the dock is a direct activation that never touches the portal and is always fast — which is likely why this is under-reported.

## Workaround

Keep the file manager warm so the portal never has to cold-start it — a systemd user service that pings the bus name every 15 s, re-activating Nautilus outside any portal call:

```ini
[Service]
ExecStart=/bin/sh -c 'while :; do gdbus call --session \
  --dest org.freedesktop.FileManager1 \
  --object-path /org/freedesktop/FileManager1 \
  --method org.freedesktop.DBus.Peer.Ping >/dev/null 2>&1; sleep 15; done'
Restart=always
```

It must *ping* rather than run `nautilus --gapplication-service` directly, otherwise it competes with D-Bus activation for the bus name and restart-loops.

Happy to gather more data or test a patch.

ravi-arnan:
Closes #2093.

`OpenDirectory` was reported as taking exactly 25 seconds whenever the file manager has to be
D-Bus activated. The mechanism the reporter identified is right, but it needs much less to trigger
than the report assumes, and the damage is wider: while it happens, **every portal is unresponsive
for every client**, not just the caller.

## What actually happens

- `open-uri.c:591` takes `REQUEST_AUTOLOCK (request)` and holds it for the whole of
  `handle_open_in_thread_func()`, which includes
- `open-uri.c:715`, a synchronous `ShowItems` call with `timeout = -1`, i.e. the GDBus default of
  25 seconds. That is where the reported 25.03s comes from: the timeout is the only thing that
  ends it.
- `shared/xdp-utils.c:67` runs on the main loop and fires for every unique name that leaves the
  bus, and
- `xdp-request.c:183` takes that same request mutex there.

So the main loop parks in `on_peer_disconnect` until the blocking call gives up.

## Measurements

A file manager stub that delays its `ShowItems` reply by 8s, and an unrelated
`OpenURI.SchemeSupported` call timed from a third connection:

| in flight | peer disconnect | probe |
|---|---|---|
| nothing | yes | 0.01s |
| OpenDirectory | no | 0.01s |
| OpenDirectory | yes | **6.70s** |

No Nautilus, no GTK and no call back into the portal are needed. Any unrelated client dropping off
the session bus is enough. `RUN_IN_THREAD` does not protect the other portals either: the probe
method is exported with it (`open-uri.c:1171`) and still stalls, because GDBus dispatches the
invocation on the main context before handing it to the thread pool.

## The fix

Drop the request lock around the `ShowItems` call and take it again afterwards. Nothing touched in
between belongs to the request: the fd is stolen out of it at `open-uri.c:605`, and the uri is
built from locals.

The request can be closed while the lock is not held, so the worker re-checks `request->exported`
before emitting the response. That check is load-bearing rather than defensive:
`xdp_request_unexport()` (`xdp-request.c:296`) does an unconditional `g_object_unref()`, so without
it a caller that disconnects at the wrong moment gets the request unexported twice.

An earlier revision of this PR moved the `PEER_DISCONNECT` emit off the main loop instead. That was
the wrong layer, and it is gone: `xdp-context.c` is untouched here.

## Testing

- New regression test in `tests/test_openuri.py`. It **fails on unpatched main at 3.70s** on the
  probe after the disconnect, while the probe before it still passes, and passes with this change.
- Full `meson test`: 28 ok, 0 failed, 1 skipped.
- The re-check path was exercised directly under an ASAN build matching CI
  (`-Db_sanitize=address -Db_lundef=false`): 15 callers killed while their own requests were parked
  inside `ShowItems`, so `on_peer_disconnect` unexports underneath the worker in exactly the window
  where the lock is not held. No use-after-free, no double unref, no crash, and the portal keeps
  serving afterwards. That stress is not in the diff, it is slow and process-heavy; happy to add it
  if you want it as an installed test.

## Note on #2091

`ScreenCast: OpenPipeWireRemote` holds the **session** lock across a PipeWire teardown, which is
the same shape one layer over. This PR does not address it, since the fix here is specific to the
open-uri call site.

### DIFF
--- a/desktop-portal/xdp-context.c
+++ b/desktop-portal/xdp-context.c
@@ -69,6 +69,7 @@ struct _XdpContext
   GHashTable *exported_portals; /* iface name -> GDBusInterfaceSkeleton */
   GHashTable *registered_object_paths; /* char *object_path set */
   GMutex registered_object_paths_lock;
+  GThreadPool *peer_disconnect_pool;
 
   GCancellable *cancellable;
 };
@@ -122,6 +123,12 @@ xdp_context_dispose (GObject *object)
   g_clear_object (&context->lockdown_impl);
   g_clear_object (&context->access_impl);
   g_clear_object (&context->app_info_registry);
+
+  if (context->peer_disconnect_pool)
+    {
+      g_thread_pool_free (context->peer_disconnect_pool, FALSE, TRUE);
+      context->peer_disconnect_pool = NULL;
+    }
 
   if (context->registered_object_paths)
     {
@@ -360,13 +367,35 @@ xdp_context_get_portal (XdpContext *context,
   return g_hash_table_lookup (context->exported_portals, interface);
 }
 
+/* Emitted from a worker thread, never from the main loop.
+ *
+ * The handlers take the per-request and per-session locks, and those locks are
+ * held across blocking work: open-uri holds the request lock for the whole of
+ * handle_open_in_thread_func(), which includes a synchronous ShowItems call to
+ * the file manager. Emitting on the main loop parks the whole daemon behind
+ * that for as long as the blocking work lasts.
+ *
+ * The pool is a dedicated one rather than the shared pool behind
+ * g_task_run_in_thread(), because handlers that block would otherwise starve
+ * every other GTask user in the process, GDBus method dispatch included.
+ */
+static void
+emit_peer_disconnect_in_thread (gpointer data,
+                                gpointer user_data)
+{
+  XdpContext *context = XDP_CONTEXT (user_data);
+  g_autofree char *peer = data;
+
+  g_signal_emit (context, signals[PEER_DISCONNECT], 0, peer);
+}
+
 static void
 on_peer_disconnect (const char *name,
                     gpointer    user_data)
 {
   XdpContext *context = XDP_CONTEXT (user_data);
 
-  g_signal_emit (context, signals[PEER_DISCONNECT], 0, name);
+  g_thread_pool_push (context->peer_disconnect_pool, g_strdup (name), NULL);
 
   dex_future_disown (xdp_app_info_registry_delete_future (context->app_info_registry,
                                                           name));
@@ -401,6 +430,9 @@ xdp_context_register (XdpContext       *context,
   portal_errors = XDG_DESKTOP_PORTAL_ERROR;
 
   g_set_object (&context->connection, connection);
+
+  context->peer_disconnect_pool =
+    g_thread_pool_new (emit_peer_disconnect_in_thread, context, -1, FALSE, NULL);
 
   context->peer_disconnect_handle_id =
     xdp_connection_track_peer_disconnect (connection,
--- a/tests/test_openuri.py
+++ b/tests/test_openuri.py
@@ -4,11 +4,15 @@
 # This file is formatted with Python Black
 
 import os
+import subprocess
+import sys
+import time
 from pathlib import Path
 from typing import Any
 
 import dbus
 import pytest
+from gi.repository import GLib
 
 import tests.xdp_doc_utils as xdp_doc
 import tests.xdp_utils as xdp
@@ -77,6 +81,34 @@ def required_templates():
         "appchooser": {},
         "lockdown": {},
     }
+
+
+# Answers ShowItems after a delay, without blocking its own main loop, so a
+# request stays pending in the portal for a known amount of time.
+SLOW_FILE_MANAGER = """
+import dbus
+import dbus.service
+import dbus.mainloop.glib
+from gi.repository import GLib
+
+
+class FileManager1(dbus.service.Object):
+    @dbus.service.method("org.freedesktop.FileManager1",
+                         in_signature="ass", out_signature="",
+                         async_callbacks=("ok", "err"))
+    def ShowItems(self, uris, startup_id, ok, err):
+        GLib.timeout_add(SHOW_ITEMS_DELAY_MS, lambda: ok() or False)
+
+
+dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+bus = dbus.SessionBus()
+name = dbus.service.BusName("org.freedesktop.FileManager1", bus)
+FileManager1(bus, "/org/freedesktop/FileManager1")
+print("ready", flush=True)
+GLib.MainLoop().run()
+"""
+
+SHOW_ITEMS_DELAY_MS = 5000
 
 
 class TestOpenURI:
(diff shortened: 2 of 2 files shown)


swick:
No way. Get rid of the locking or make it more fine-grained.

ravi-arnan:
Fair enough, dropping this approach.

I will narrow the locking in open-uri instead, so the request lock is not held across the
synchronous `ShowItems` call.

One constraint that shapes the patch, in case you have a preference on how to handle it:
`xdp_request_unexport()` does an unconditional `g_object_unref()`, so once the lock no longer spans
the call, the worker has to re-check `request->exported` under the lock before emitting the response
and unexporting, or it races `on_peer_disconnect`. The fd is not a problem, `open-uri.c:605` already
steals it out of the request.

Do you want this kept to the open-uri path, or the request and session locking reworked more
broadly? #2091 is the same shape on the session side, with the session lock held across a PipeWire
teardown.

Assisted-by: Claude:opus-5

swick:
Try to keep it in the open-uri path. In general this locking should all go away when we port the portals to libdex.

PR #2108, opened 2026-08-14
Don't block the main loop on peer disconnect

### LINKED ISSUE #2093: OpenDirectory deadlocks for 25s when the file manager must be D-Bus activated (portal blocks on its own mutex in on_peer_disconnect)

ravi-arnan:
Closes #2093. Replaces #2107, which took the wrong layer (it moved the `PEER_DISCONNECT` emit off the main loop in `xdp-context.c`); this PR instead fixes the disconnect handler itself.

`OpenDirectory` was reported as taking exactly 25 seconds whenever the file manager has to be D-Bus activated. The mechanism the reporter identified is right, but it needs much less to trigger than the report assumes, and the damage is wider: while it happens, **every portal is unresponsive for every client**, not just the caller.

## What actually happens

- A portal handler (e.g. `open-uri.c` `handle_open_in_thread_func`) takes `REQUEST_AUTOLOCK(request)` and holds it across a blocking call. For `ShowItems` that call has `timeout = -1`, i.e. the GDBus default of 25 seconds.
- `xdp-context.c` emits `peer-disconnect` on the **main loop** for every unique name that leaves the bus.
- `xdp-request.c` `on_peer_disconnect` (and the session equivalent) take the *same* request/session lock.

So the main loop parks in `on_peer_disconnect` until the blocking call gives up, and every other portal stalls with it.

## Measurements

A file manager stub that delays its `ShowItems` reply, and an unrelated `OpenURI.SchemeSupported` call timed from a third connection:

| in flight | peer disconnect | probe |
|---|---|---|
| nothing | yes | 0.01s |
| OpenDirectory | no | 0.01s |
| OpenDirectory | yes | **6.70s** |

No Nautilus, no GTK and no call back into the portal are needed. Any unrelated client dropping off the session bus is enough.

## The fix

Move the disconnect work off the main loop. `on_peer_disconnect` for requests and sessions now:

- does the peer check **lock-free** (the sender is fixed for the object's lifetime, so it never changes while the request/session exists), which lets the main loop skip the lock for every disconnect event that is not this object's owner, and
- runs the lock-taking disconnect sequence in a **worker thread** (`g_task_run_in_thread`), so a handler holding the lock no longer parks the main loop. The sequence is still serialized by the lock, so there is no double close/unexport.

This covers both the reported `ShowItems` stall and the same-shape session case (`ScreenCast.OpenPipeWireRemote` joining a PipeWire thread under the session lock, #2091). `xdp-session-persistence.c` is intentionally left alone: its handler uses its own `transient_permissions_lock`, not the request/session lock.

## Testing

A regression test was written and used to validate this, then dropped from the diff at swick's request (too specific for this issue; the libdex port will make the remaining lock sites moot anyway).

### DIFF
--- a/desktop-portal/xdp-request.c
+++ b/desktop-portal/xdp-request.c
@@ -180,16 +180,14 @@ get_token (GDBusMethodInvocation *invocation)
 }
 
 static void
-on_peer_disconnect (XdpContext *context,
-                    const char *peer,
-                    gpointer    user_data)
+on_peer_disconnect_in_thread_func (GTask        *task,
+                                   gpointer      source_object,
+                                   gpointer      task_data,
+                                   GCancellable *cancellable)
 {
-  XdpRequest *request = XDP_REQUEST (user_data);
+  XdpRequest *request = XDP_REQUEST (source_object);
 
   REQUEST_AUTOLOCK (request);
-
-  if (g_strcmp0 (request->sender, peer) != 0)
-    return;
 
   if (!request->exported)
     return;
@@ -199,6 +197,30 @@ on_peer_disconnect (XdpContext *context,
 
   xdp_request_unexport (request);
   xdp_context_unclaim_object_path (request->context, request->id);
+}
+
+static void
+on_peer_disconnect (XdpContext *context,
+                    const char *peer,
+                    gpointer    user_data)
+{
+  XdpRequest *request = XDP_REQUEST (user_data);
+  g_autoptr(GTask) task = NULL;
+
+  /* The sender is fixed for the lifetime of the request, so the peer
+   * check is safe without the lock. This lets the main loop skip the
+   * lock for every disconnect event that is not this request's owner,
+   * which would otherwise stall every portal while a handler holds it. */
+  if (g_strcmp0 (request->sender, peer) != 0)
+    return;
+
+  /* The disconnect sequence needs the lock, but taking it here would
+   * block the main loop if a portal handler is currently holding it
+   * (e.g. open-uri blocking on ShowItems for the full D-Bus timeout).
+   * Run it in a worker thread instead. */
+  task = g_task_new (request, NULL, NULL, NULL);
+  g_task_set_source_tag (task, on_peer_disconnect_in_thread_func);
+  g_task_run_in_thread (task, on_peer_disconnect_in_thread_func);
 }
 
 gboolean
--- a/desktop-portal/xdp-session.c
+++ b/desktop-portal/xdp-session.c
@@ -244,18 +244,39 @@ xdp_session_authorize_callback (GDBusInterfaceSkeleton *interface,
 }
 
 static void
+on_peer_disconnect_in_thread_func (GTask        *task,
+                                   gpointer      source_object,
+                                   gpointer      task_data,
+                                   GCancellable *cancellable)
+{
+  XdpSession *session = XDP_SESSION (source_object);
+
+  SESSION_AUTOLOCK_UNREF (g_object_ref (session));
+
+  xdp_session_close (session, FALSE);
+}
+
+static void
 on_peer_disconnect (XdpContext *context,
                     const char *peer,
                     gpointer    user_data)
 {
   XdpSession *session = XDP_SESSION (user_data);
+  g_autoptr(GTask) task = NULL;
 
-  SESSION_AUTOLOCK (session);
-
+  /* The sender is fixed after session init, so the peer check is safe
+   * without the lock. This keeps the main loop from taking the session
+   * lock on every disconnect event, which would stall portals while a
+   * handler holds it (e.g. screen-cast joining a PipeWire thread). */
   if (g_strcmp0 (session->sender, peer) != 0)
     return;
 
-  xdp_session_close (session, FALSE);
+  /* The disconnect sequence needs the lock, but taking it here would
+   * block the main loop if a portal handler is currently holding it.
+   * Run it in a worker thread instead. */
+  task = g_task_new (session, NULL, NULL, NULL);
+  g_task_set_source_tag (task, on_peer_disconnect_in_thread_func);
+  g_task_run_in_thread (task, on_peer_disconnect_in_thread_func);
 }
 
 static gboolean


swick:
Can you find all the instances where we hold a lock over a potentially long running call which can interfere with other peers?

I also appreciate the test but I think we should just drop it. It's too specific for this issue, and we should rewrite all portals with libdex anyway, at which point we don't need any of this.

ravi-arnan:
Test dropped, the PR is just the open-uri change now.

For the audit I scripted it rather than eyeballing: index every function in `desktop-portal/` and
`shared/`, mark a function as blocking if it contains a blocking primitive or calls something
already marked blocking (iterated to a fixed point), then report functions that take a lock and
then reach one. **32 sites**, and they are almost all one pattern.

## 31 of them: `xdp_dbus_impl_request_proxy_new_sync()` under the request lock

Every portal that creates an impl Request does this, e.g. `account.c:185`:

```c
REQUEST_AUTOLOCK (request);

impl_request = xdp_dbus_impl_request_proxy_new_sync (
  g_dbus_proxy_get_connection (G_DBUS_PROXY (account->impl)),
  G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
  ...
```

The flags skip properties but not auto-start, so construction is still a synchronous round trip
that can activate the backend. Nine of these hold `SESSION_AUTOLOCK_UNREF` at the same time.

| file | sites | file | sites |
|---|---|---|---|
| `input-capture.c` | 4 | `inhibit.c` | 2 |
| `screen-cast.c` | 4 | `print.c` | 2 |
| `file-chooser.c` | 3 | `screenshot.c` | 2 |
| `global-shortcuts.c` | 3 | `account.c` `background.c` `dynamic-launcher.c` | 1 each |
| `remote-desktop.c` | 3 | `email.c` `location.c` `secret.c` `usb.c` `wallpaper.c` | 1 each |

Full list with line numbers, if useful:

```
account.c:170 handle_get_user_information          location.c:487 handle_start_in_thread_func
background.c:1081 handle_request_background        print.c:129 handle_print
dynamic-launcher.c:645 handle_prepare_install      print.c:247 handle_prepare_print
email.c:195 handle_compose_email                   remote-desktop.c:341 handle_create_session
file-chooser.c:468 handle_open_file                remote-desktop.c:529 handle_select_devices
file-chooser.c:591 handle_save_file                remote-desktop.c:735 handle_start
file-chooser.c:757 handle_save_files               screen-cast.c:263 handle_create_session
global-shortcuts.c:220 handle_create_session       screen-cast.c:519 handle_select_sources
global-shortcuts.c:386 handle_bind_shortcuts       screen-cast.c:884 handle_start
global-shortcuts.c:519 handle_list_shortcuts       screenshot.c:345 handle_screenshot_in_thread_func
inhibit.c:185 handle_inhibit                       screenshot.c:510 handle_pick_color
inhibit.c:401 handle_create_monitor                secret.c:111 handle_retrieve_secret
input-capture.c:240 handle_create_session          usb.c:1126 handle_acquire_devices
input-capture.c:604 handle_start                   wallpaper.c:108 handle_set_wallpaper_in_thread_func
input-capture.c:780 handle_get_zones
input-capture.c:939 handle_set_pointer_barriers
```

These are milder than the `ShowItems` case in this PR: the backend is normally already running, so
the round trip is short, and it is not activating a GTK app inside the critical section. But it is
the same shape, it has no timeout, and a slow or restarting backend blocks the main loop for
everyone.

## The one that is not that pattern

`screen-cast.c:984 handle_open_pipewire_remote()` holds `SESSION_AUTOLOCK_UNREF` and reaches
`pw_context_destroy` through `pipewire_remote_destroy()` (`pipewire.c:217`), which joins a PipeWire
thread. That is #2091, and the audit found it independently rather than because I went looking.

## Caveats on the method

The script parses braces, does not expand macros, resolves callees by name, and cannot see calls
through function pointers. That last one is not hypothetical here: `pw_context_destroy` is passed
to `g_clear_pointer()` rather than called directly, so an earlier version of the script missed
`handle_open_pipewire_remote` entirely. Anything reached only through a callback or a vfunc is
still invisible, so treat 32 as a floor.

Happy to take any subset of these in follow-ups, or to leave them if the libdex port is close
enough that churning them now is wasted work. The `_proxy_new_sync` ones look mechanical enough
that a single pass could do them, since the proxy does not need the lock at all.

swick:
Mh, I see, this is a structural issue. Maybe the better idea then is to handle XdpRequests on_peer_disconnect peer check without a lock (it should not change over the lifetime of the request) and then run the disconnect sequence in a thread with the lock held. Same for session.

ravi-arnan:
That is the right direction for the structural fix. Pulling the peer check out of the lock and running the disconnect sequence in a thread would let every portal drop the same lock without any per-call unlock/re-lock dance, and it would fix the session case too (`handle_open_pipewire_remote` in screen-cast.c).

This PR is scoped to the one reported stall: `ShowItems` in open-uri holding the lock while it can block for the full D-Bus timeout. The diff is just the unlock/re-lock around that single call and the `exported` recheck after. It does not try to fix the other 31 sites (those are milder, short backend round trips, not a GTK activation), and they will all become irrelevant once the libdex port lands.

Happy to file the structural follow-up (peer check without lock + threaded disconnect for request and session) as a separate issue or PR if you want it tracked separately from this one.

swick:
lets just fix the structural issue here

ravi-arnan:
Reworked to the structural fix you described, dropped the scoped open-uri unlock/re-lock.

- `on_peer_disconnect` for requests and sessions now does the peer check lock-free (the sender is fixed for the object's lifetime), then runs the disconnect sequence in a worker thread with the lock held. So a handler holding the lock (open-uri on `ShowItems`, screen-cast joining a PipeWire thread) no longer parks the main loop and stalls every other portal.
- `xdp-session-persistence.c` is left alone: its handler uses its own `transient_permissions_lock`, not the request/session lock, so it is not part of this stall.

The 32-site audit still stands as background, but with this change the remaining `_proxy_new_sync`-under-lock sites become moot for the reported symptom, so I have not touched them here.

desktop-portal/xdp-request.c
```diff
@@ -201,6 +199,28 @@ on_peer_disconnect (XdpContext *context,
...
+   * lock for every disconnect event that is not this request's owner,
+   * which would otherwise stall every portal while a handler holds it. */
+  if (g_strcmp0 (request->sender, peer) != 0)
+    return;
+
+  /* The disconnect sequence needs the lock, but taking it here would
+   * block the main loop if a portal handler is currently holding it
+   * (e.g. open-uri blocking on ShowItems for the full D-Bus timeout).
+   * Run it in a worker thread instead. */
+  g_autoptr(GTask) task = g_task_new (request, NULL, NULL, NULL);
```
swick: variable declarations only at the beginning of a scope. Also CI is failing because:

```
  error: GTask 'task' created without g_task_set_source_tag
  rule: g_task_source_tag
```

ravi-arnan: Fixed both: declarations moved to the start of the scope, and added `g_task_set_source_tag` in both `on_peer_disconnect` handlers (request and session). Force-pushed.

swick:
The commit message should not reference any previous revision which doesn't exist anymore. We also don't use fix/feat/... It should be two commits, one with `request: ` and one with `session: `.

ravi-arnan:
Split into the two commits you asked for:

- `request: run disconnect off the main loop` (xdp-request.c only)
- `session: run disconnect off the main loop` (xdp-session.c only)

Neither message references the old revision anymore, and there is no fix/feat/ prefix. CI is green on the new head (6 ok, 1 skipped), merge state clean.

swick:
Thanks!

PR #2109, opened 2026-08-14
context: Wait on async init to finish before owning the D-Bus name

swick:
If we do not wait for the initialization to finish, some portals might not be exported on the bus before we own the public dbus name, which can result in apps trying to use a portal and get unexpected errors.

Async is still useful because all portals can initialize in parallel.

Fixes: 58d75ebb ("secret: Init asynchronously")
Fixes: 8558c706 ("secret: Init async in fiber")

### DIFF
--- a/desktop-portal/xdp-context.c
+++ b/desktop-portal/xdp-context.c
@@ -71,6 +71,7 @@ struct _XdpContext
   GMutex registered_object_paths_lock;
 
   GCancellable *cancellable;
+  GPtrArray *pending_inits; /* DexFuture */
 };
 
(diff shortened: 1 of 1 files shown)


pwithnall:
One minor review comment left on this at https://github.com/flatpak/xdg-desktop-portal/pull/2048#pullrequestreview-4950940937 (since the commit here is included in that PR too).

swick:
Iterating the thread-default main context now. AFAICS this isn't an issue currently, but you're right.

swick:
Oops, merged the other one first. Same commit, but github doesn't pick it up.

PR #2111, opened 2026-08-15
screencast: Fix options double-unref on invalid persistence

Nuhiat-Arefin:
`handle_select_sources()` keeps the filtered options in a `g_autoptr(GVariant)` and passes that pointer to `replace_screen_cast_restore_token_with_data()`.

The helper takes ownership of the same reference through another autoptr. For a RemoteDesktop session with persistence enabled, it returns `InvalidArgument` without giving that reference back to the caller. As a result, both autoptrs unref the same `GVariant`, which triggers a GLib critical and can lead to heap corruption or a crash.

Fix this by stealing the local pointer back into `*in_out_options` before returning the error. This keeps ownership with the caller on the failure path. The existing ScreenCast success path is unchanged: there, the local autoptr still cleans up the options variant that was replaced.

Add a regression test for a request containing both a nonzero `persist_mode` and a `restore_token`. The test checks that:

* the existing `InvalidArgument` error and message are unchanged;
* the invalid request is not sent to the backend; and
* the portal is still alive and responsive after the request.

The existing test harness makes GLib criticals fatal, so the regression test detects the invalid cleanup by verifying that the portal remains alive and responsive after rejecting the request.

### DIFF
--- a/desktop-portal/screen-cast.c
+++ b/desktop-portal/screen-cast.c
@@ -493,6 +493,7 @@ replace_screen_cast_restore_token_with_data (XdpSession *session,
                        XDG_DESKTOP_PORTAL_ERROR,
                        XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
                        "Remote desktop sessions cannot persist");
+          *in_out_options = g_steal_pointer (&options);
           return FALSE;
         }
     }
--- a/tests/test_screencast.py
+++ b/tests/test_screencast.py
@@ -11,7 +11,7 @@ import tests.xdp_utils as xdp
 
 @pytest.fixture
 def required_templates():
-    return {"screencast": {}}
+    return {"remotedesktop": {}, "screencast": {}}
 
 
 def start_stream(dbus_con, screencast_intf):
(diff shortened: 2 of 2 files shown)


swick:
Good catch! Isn't the more proper fix to steal the `in_out_options`?

```
  options = g_steal_pointer (in_out_options);
```

swick:
> the portal is still alive and responsive after the request.

I  think that deserves a comment on the `xdp.check_version`. It's very subtle and easy to miss.

Nuhiat-Arefin:
> Good catch! Isn't the more proper fix to steal the `in_out_options`?
> 
> ```
>   options = g_steal_pointer (in_out_options);
> ```

Updated

Nuhiat-Arefin:
> > the portal is still alive and responsive after the request.
> 
> I think that deserves a comment on the `xdp.check_version`. It's very subtle and easy to miss.

Done

desktop-portal/screen-cast.c
```diff
@@ -493,6 +493,7 @@ replace_screen_cast_restore_token_with_data (XdpSession *session,
                        XDG_DESKTOP_PORTAL_ERROR,
                        XDG_DESKTOP_PORTAL_ERROR_INVALID_ARGUMENT,
                        "Remote desktop sessions cannot persist");
+          *in_out_options = g_steal_pointer (&options);
```
swick, line 496: That doesn't look right. The in-out gvariant is owned, then ownership is transferred to the local options (leaving `*in_out_poptions == NULL`). So I think we should simply return here without any further changes.

desktop-portal/screen-cast.c
swick: And here we want to pass `&options` to `xdp_session_persistence_replace_restore_token_with_data`. Then transfer ownership back to `in_out_options` unconditionally below.

desktop-portal/screen-cast.c
Nuhiat-Arefin: Changed it.

desktop-portal/screen-cast.c
Nuhiat-Arefin: Changed it.

swick:
Sorry to send you an a detour. `xdp_session_persistence_replace_restore_token_with_data` is awful to use, you got it right the first time around.

I changed it back (but kept the comment).

swick:
Also created https://github.com/flatpak/xdg-desktop-portal/pull/2114 to clean this mess up a bit more.

PR #2113, opened 2026-08-16
registry: Reject the insert promise when propagating an error

swick:
The main commit is a bug fix that was noticed in GNOME OS OpenQA.

Then there is another fix for a race that probably never actually happens and a bit of improvements for the logging.

### DIFF
--- a/shared/xdp-app-info-registry.c
+++ b/shared/xdp-app-info-registry.c
@@ -14,6 +14,7 @@ struct _XdpAppInfoRegistry
   GObject parent_instance;
 
   GHashTable *app_infos; /* unique dbus name -> XdpAppInfo* */
+  GMutex app_infos_lock; /* protects app_infos */
   GHashTable *channels; /* unique dbus name -> DexChannel* */
   GMutex channels_lock; /* protects channels */
 };
@@ -83,6 +84,9 @@ xdp_app_info_registry_dispose (GObject *object)
   if (registry->channels)
     g_mutex_clear (&registry->channels_lock);
 
+  if (registry->app_infos)
+    g_mutex_clear (&registry->app_infos_lock);
+
   g_clear_pointer (&registry->app_infos, g_hash_table_unref);
   g_clear_pointer (&registry->channels, g_hash_table_unref);
 
@@ -111,6 +115,7 @@ xdp_app_info_registry_new (void)
   registry->app_infos = g_hash_table_new_full (g_str_hash, g_str_equal,
                                                g_free,
                                                g_object_unref);
+  g_mutex_init (&registry->app_infos_lock);
 
   registry->channels = g_hash_table_new_full (g_str_hash, g_str_equal,
                                               g_free,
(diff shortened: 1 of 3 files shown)


shared/xdp-app-info-registry.c
```diff
@@ -316,10 +316,12 @@ xdp_app_info_registry_ensure_future (XdpAppInfoRegistry    *registry,
 {
   g_autoptr(QueueData) data = NULL;
   const char *sender = g_dbus_method_invocation_get_sender (invocation);
-  XdpAppInfo *existing = g_hash_table_lookup (registry->app_infos, sender);
+  XdpAppInfo *existing;
 
   G_MUTEX_AUTO_LOCK (&registry->channels_lock, locker);
 
+  existing = g_hash_table_lookup (registry->app_infos, sender);
```
pwithnall: If you’re putting this access to `registry->app_infos` under `channels_lock`, then:
 - The code comment by the declaration of `channels_lock` needs to be amended to say it also protects `app_infos`
 - Other references of `registry->app_infos` also need to be locked similarly

swick: I thought I could get away with not locking app_infos in the fiber because the check for a channel would make it impossible for the fiber to modify it, but app_infos is global to all senders, so you're right.

All access to `registry->app_infos` is now under a new lock.

shared/xdp-app-info-registry.c
```diff
@@ -153,23 +158,34 @@ work_fiber (XdpAppInfoRegistry *registry,
           {
             XdpAppInfo *app_info;
 
+            g_debug ("XdpAppInfoRegistry: Ensure for sender %s", sender);
+
+            g_mutex_lock (&registry->app_infos_lock);
             app_info = g_hash_table_lookup (registry->app_infos, sender);
+            g_mutex_unlock (&registry->app_infos_lock);
```
pwithnall, line 167: This critical section could race with another one which removes the `app_info` from `registry->app_infos`, and the final reference on `app_info` could be dropped after the `g_mutex_unlock()` here, leaving it dangling.

tl;dr: You need to take a strong reference on `app_info` before unlocking the mutex. (This is a common pattern which typically should be applied everywhere a mutex is used to query a reference to a `GObject`.)

shared/xdp-app-info-registry.c
```diff
@@ -153,23 +158,34 @@ work_fiber (XdpAppInfoRegistry *registry,
...
+                             sender, error->message);
                     dex_promise_reject (data->promise, g_steal_pointer (&error));
                     break;
                   }
 
+                g_mutex_lock (&registry->app_infos_lock);
                 g_hash_table_insert (registry->app_infos,
                                      g_strdup (sender),
                                      g_steal_pointer (&app_info_owned));
+                g_mutex_unlock (&registry->app_infos_lock);
```
pwithnall, line 194: What happens if this has raced with another thread which has added an app info for this `sender` already? The code will currently overwrite that hash table entry (and handle the memory allocation/frees correctly). Is that the behaviour you want? From my limited understanding of the app info registry, it probably is, but I thought I should check.

shared/xdp-app-info-registry.c
```diff
@@ -190,25 +214,33 @@ work_fiber (XdpAppInfoRegistry *registry,
...
               }
 
             g_debug ("Inserted app info %s for sender %s",
                      xdp_app_info_get_id (app_info), sender);
+
+            g_mutex_lock (&registry->app_infos_lock);
             g_hash_table_insert (registry->app_infos,
                                  g_strdup (sender),
                                  g_steal_pointer (&app_info));
+            g_mutex_unlock (&registry->app_infos_lock);
```
pwithnall, line 241: Same comment as above

swick:
I'm considering to drop the fast path. One of the ideas with libdex was that we can drop threads and not have to care about locking anymore...

shared/xdp-app-info-registry.c
swick: We only ever use the delete shortcut if there is no fiber running, which means nothing can delete it concurrently.

But fair, let's just do the safe thing...

shared/xdp-app-info-registry.c
swick: > What happens if this has raced with another thread which has added an app info for this sender already

Can't happen because other threads only ever modify the hash table if the fiber is not running.

swick:
Anyway, this should be ready now.

shared/xdp-app-info-registry.c
pwithnall: You could add a `gboolean key_did_not_exist = g_hash_table_insert (…); g_assert (key_did_not_exist);` (and an accompanying comment) to assert that’s true every time.

swick: Did that now.

pwithnall:
LGTM :)

pwithnall (approved):

swick:
Thanks!

PR #2114, opened 2026-08-19
session-persistence: Fix replace_restore_token_with_data onwership

swick:
The xdp_session_persistence_replace_restore_token_with_data function
would takek an in-out GVariant, and then set a new value, but not free
the old value. This API is super confusing and requires the caller to
potentially free the in-out param. This has caused double-free and
memory leak bugs.

This change ensures that the in-out options is always a valid, owned
GVariant, which simplifies the callers.

### DIFF
--- a/desktop-portal/xdp-session-persistence.c
+++ b/desktop-portal/xdp-session-persistence.c
@@ -246,6 +246,7 @@ xdp_session_persistence_replace_restore_token_with_data (XdpSession *session,
       g_clear_pointer (&value, g_variant_unref);
     }
 
+  g_clear_pointer (in_out_options, g_variant_unref);
   *in_out_options = g_variant_ref_sink (g_variant_builder_end (&options_builder));
 }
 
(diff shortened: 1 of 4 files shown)


desktop-portal/xdp-session-persistence.c
```diff
@@ -246,6 +246,7 @@ xdp_session_persistence_replace_restore_token_with_data (XdpSession *session,
       g_clear_pointer (&value, g_variant_unref);
     }
 
+  g_clear_pointer (in_out_options, g_variant_unref);
   *in_out_options = g_variant_ref_sink (g_variant_builder_end (&options_builder));
```
whot, line 250: Is this some stable API? Could make the options a normal readonly argument only and return the new variant.

whot (approved):
Typo in second commit message `takek`

desktop-portal/xdp-session-persistence.c
swick: Tried that. Gets worse.

whot: story of my life :)

PR #2117, opened 2026-08-21
tests: Defer request closing until export

razzeee:
Request close timers could fire before asynchronous portal methods exported their request paths under load. This caused Close() to fail with UnknownMethod while the backend waited until timeout.

Arm the timer after the method reply, propagate close errors, and remove pending sources when the call finishes.

### DIFF
--- a/tests/xdp_utils.py
+++ b/tests/xdp_utils.py
@@ -519,6 +519,7 @@ class Closable:
         self._mainloop: GLib.MainLoop | None = None
         self._impl_closed = False
         self._bus = bus
+        self._close_source_id: int | None = None
 
         self._closable = type(self).__name__
         assert self._closable in ("Request", "Session")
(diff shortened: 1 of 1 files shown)


swick:
Not entirely sure what you're trying to fix here but we recently fixed exporting the async init portal exporting (https://github.com/flatpak/xdg-desktop-portal/pull/2109). This looks a bit more like you found the race is Request.Close calls which we need to systematically get rid of, for which I have https://github.com/flatpak/xdg-desktop-portal/pull/2101.

razzeee:
Let me rebase my other work and see if I can safely drop this then.

razzeee:
Had another look on a rebased branch. My  tests were still failing, even with this.

This fixes them https://github.com/razzeee/xdg-desktop-portal/commit/b28fe0fbcda5ec885430b2f129c109b70852a060

swick:
That by itself can't be a complete fix, but it does look like there might be something to it, so a new PR would be nice.

Issue #1953, opened 2026-04-07
O_RDONLY|O_NOFOLLOW on /proc/<pid>/root always returns ELOOP — unsandboxed process registration silently fails, breaking KDE screen sharing


OrcVole:
### Operating System

Bazzite 43.20260406 (Fedora 43 immutable / OSTree / composefs)

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.20.3-2.fc43.x86_64

### Desktop Environment

Other

### Desktop Environment (Other)

KDE Plasma 6.6.3 (Wayland session)

### Expected Behavior

The portal successfully registers unsandboxed processes (such as KWin) as host applications. KWin's screencast plugin creates the `zkde_screencast_unstable_v1` Wayland global. Screen sharing and PipeWire screen capture work normally.

### Current Behavior

Every registration attempt from an unsandboxed process fails with:

`Could not register app ID: Unable to open /proc/<pid>/root
`
On KDE Plasma, this prevents KWin's screencast plugin from creating `zkde_screencast_unstable_v1`. Screen sharing silently fails for the entire session — no error is shown to the user.

The failure is caused by the portal opening `/proc/<pid>/root` with `O_RDONLY | O_NOFOLLOW`. That path is a kernel magic symlink; POSIX requires `ELOOP` when `O_NOFOLLOW` is used on any symlink with `O_RDONLY`. This is not a race, not SELinux, not seccomp — it is a flag combination that is unconditionally invalid for a symlink on any Linux system.

Confirmed via Python on the affected system:
```python
os.open('/proc/<pid>/root', os.O_PATH | os.O_NOFOLLOW)
# → OK

os.open('/proc/<pid>/root', os.O_RDONLY | os.O_NOFOLLOW)
# → FAIL: [Errno 40] Too many levels of symbolic links
```

The string `O_PATH fd was opened O_NOFOLLOW` appears in the portal binary, confirming the code path. The error repeats on every portal restart across the entire session lifetime — it is not a startup timing issue.

### Steps to Reproduce

1. Log in to a KDE Plasma 6 Wayland session on any immutable Fedora derivative (Bazzite, Kinoite, Silverblue).
2. Attempt to share your screen — via OBS PipeWire capture, browser WebRTC, or any application that calls `org.freedesktop.portal.ScreenCast.CreateSession`.
3. Nothing happens. The chooser dialog never appears.

To confirm the failure directly:
```bash
[...]

Issue #1490, opened 2024-10-27
KeepassXC attempts to use a privileged file when using portals and can't open a File Chooser


txtsd:
### Operating System

Arch Linux

### XDG Desktop Portal version

1.18

### Desktop Environment

wlroots

### Desktop Environment (Other)

Sway

### Expected Behavior

KeepassXC should open a File Chooser

### Current Behavior

KeepassXC does nothing

### Steps to Reproduce

1. Create a file ~/.config/xdg-desktop-portal/portals.conf
2. Fill it with
  ```
  [preferred]
  default=wlr;gtk
  org.freedesktop.impl.portal.FileChooser=lxqt

  ```
3. Log out and log back in
4. Run KeepassXC and hit the `Open Database` button.
5. Nothing happens
...

### Anything else we should know?

Here's a dbus-monitor log:

```
method call time=1730043932.721391 sender=:1.65 -> destination=org.freedesktop.portal.Desktop serial=84 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.portal.FileChooser; member=OpenFile
   string "x11:1"
   string "Open database"
   array [
      dict entry(
         string "directory"
         variant             boolean false
      )
      dict entry(
         string "filters"
         variant             array [
               struct {
                  string "KeePass 2 Database "
                  array [
                     struct {
                        uint32 0
                        string "*.kdbx"
                     }
                  ]
               }
               struct {
                  string "All files "
                  array [
                     struct {
                        uint32 0
                        string "*"
                     }
                  ]
               }
            ]
      )
      dict entry(
         string "handle_token"
         variant             string "qt2944820299"
      )
      dict entry(
         string "modal"
         variant             boolean true
      )
      dict entry(
         string "multiple"
         variant             boolean false
      )
   ]
method call time=1730043932.721809 sender=:1.28 -> destination=org.freedesktop.DBus serial=108 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=GetConnectionUnixProcessID
   string ":1.65"
method return time=1730043932.721820 sender=org.freedesktop.DBus -> destination=:1.28 serial=4294967295 reply_serial=108
   uint32 2504
error time=1730043932.722082 sender=:1.28 -> destination=:1.65 error_name=org.freedesktop.DBus.Error.AccessDenied reply_serial=84
   string "Portal operation not allowed: Unable to open /proc/2504/root"

```

[...]

PR #2119, opened 2026-08-26
app-info: treat any EACCES opening /proc/<pid>/root as not sandboxed

TalonGrayson:
## Summary

`open_flatpak_info()` / `open_linyaps_info()` in `shared/xdp-app-info-*.c` decide
whether a calling process is Flatpak/linyaps-confined by opening
`/proc/<pid>/root`. If that `open()` fails with `EACCES`, the code only treats
it as "not actually confined, keep looking" when `statfs()` on that path
reports `FUSE_SUPER_MAGIC` — the known toolbox-container case. Any other
`EACCES` is treated as a hard failure (`G_IO_ERROR_FAILED`), which aborts
`xdp_app_info_new()` entirely and denies the whole portal request.

That FUSE-only check misses a different, unrelated source of the identical
`EACCES`: the kernel marks a process **non-dumpable** whenever it gains
capabilities via file capabilities (`setcap`) at `exec()`. Once non-dumpable,
`/proc/<pid>/root` (like `/proc/<pid>/mem`, `/maps`, etc.) can't be opened by
any other process — including one with the same UID — without
`CAP_SYS_PTRACE`. This is standard kernel behavior, independent of filesystem.

In practice this breaks portal calls (`RemoteDesktop`, `ScreenCast`, ...) from
any ordinary, unsandboxed desktop app that happens to carry a capability for
an unrelated reason. Two concrete examples I hit on a real system:

- `kdeconnectd` ships `cap_net_admin` (for LAN device discovery), which breaks
  KDE Connect's remote-input feature.
- `kwin_wayland` itself ships `cap_sys_nice`, which by the same mechanism would
  break its own screen-sharing/portal registration in the same circumstances.

Neither of these processes is a Flatpak or linyaps sandbox, but a real one of
those is *always* dumpable and same-UID-openable — so an `EACCES` here can
never legitimately come from an actual Flatpak/linyaps process. This PR
simplifies both functions to treat **any** `EACCES` on that open as
`XDP_APP_INFO_ERROR_WRONG_APP_KIND` unconditionally, instead of gating it on
`statfs()` reporting one specific filesystem magic number, so app-info lookup
keeps falling through to snap/host detection exactly the way it already does
for the FUSE case.

## Reproduction

On Fedora Atomic (Bazzite/Kinoite, KDE Plasma 6.7, `xdg-desktop-portal`
1.22.1), every `kdeconnectd` remote-input request failed with:

```
Could not register app ID: Unable to open /proc/<pid>/root
```

Confirmed with a minimal reproducer performing the exact same `openat()` the
portal does, against `kdeconnectd`'s PID from an unrelated same-UID process —
fails with `EACCES`. The same reproducer against `plasmashell` (no file caps)
succeeds; against `kwin_wayland` (`cap_sys_nice`) it fails identically. `getcap`
confirms `kdeconnectd` carries `cap_net_admin=ep`.

With this patch applied (built and layered via `rpm-ostree override replace`,
then rebooted), remote input works correctly.

## Notes

- This is somewhat related to #1953, though that report is specifically about
  an `O_NOFOLLOW`/`ELOOP` failure mode on FUSE rootfs; this is a distinct
  failure mode (non-dumpable target process) that happens to surface through
  the same code path and error message.
- I didn't add an automated test — reliably reproducing a non-dumpable target
  process (or a FUSE-EACCES one) from the existing `umockdev`-based test
  harness isn't straightforward. Happy to take direction on how you'd want
  this covered.


swick:
If I understand the changes here correctly, then making a sandboxed process non-dumpable would result in xdp treating it as unsandboxed. That doesn't sound like a good idea?

TalonGrayson:
Good point, thank you. I've pushed a change to address that. It now only treats EACCES as not sandboxed if the process also has actual capabilities set. If it's EACCES with no capabilities, it falls through to the hard error like before.

The FUSE check is unchanged. Still fixes the original bug, from what I've been able to test.

shared/xdp-utils.c
```diff
@@ -1091,6 +1091,83 @@ xdp_pidfd_get_pidns (int      pidfd,
...
+  gboolean has_caps = FALSE;
+  FILE *f;
+  size_t keylen = 0;
+  size_t vallen = 0;
+  ssize_t n;
+  int fd;
+
+  g_return_val_if_fail (pid > 0, FALSE);
+
+  status_path = g_strdup_printf ("/proc/%u/status", (guint) pid);
```
smcv, line 1131: x-d-p needs to be really careful with this sort of thing. There are two ways that a malicious process can arrange for what appears in `/proc` to be misleading:

* `execve()`: if the creator of the socket hands it off to a subprocess and then execs a setuid/setgid/setcap executable, that executable replaces its pid, so now `/proc/$pid/status` will reflect the elevated privileges
* pid reuse: if the creator of the socket hands it off to a subprocess and then exits, its pid could be reused by an ordinary system process. This is *unlikely* to be conveniently privileged, but there might be ways that a sufficiently creative sandboxed app can arrange for the system to start a lot of privileged processes so that the odds are in its favour.

This is why D-Bus and polkit access control never reads files in `/proc/$pid`, and instead exclusively relies on facts that can be obtained from the kernel in a race-free way (`GetConnectionCredentials()`, which on Linux is implemented in terms of `SO_PEERCRED`, `SO_PEERSEC` and similar getsockopt calls).

Flatpak avoids the setuid `execve()` attack by unconditionally setting `PR_SET_NO_NEW_PRIVS`, and avoids (or at least strongly mitigates) both attacks by arranging that the sandboxed process is not actually talking to the portal directly. Instead, the sandboxed process talks to xdg-dbus-proxy, which talks to the portal but is not under the sandboxed app's full control, so the sandboxed app can't induce the proxy to exec anything or exit at a convenient time.

Is linyaps equally careful? Are we sufficiently confident that it will always be this careful in the future, for it to be safe for this to be security-critical?

TalonGrayson: Oh, that's super interesting. Thanks for flagging... I'll go away and consider these points.

smcv: On sufficiently recent kernels, the other way to be more confident about the provenance of a remote process is to refer to it via a pidfd (obtained from `SO_PEERPIDFD` or D-Bus `GetConnectionCredentials()["ProcessFD"]`), rather than a process ID.

swick:
I'm not super happy about this. Ideally we'd move away from parsing things in /proc/$pid/root at all, and we are currently working towards that goal.

TalonGrayson:
> I'm not super happy about this. Ideally we'd move away from parsing things in /proc/$pid/root at all, and we are currently working towards that goal.

No worries, if I've missed the mark I'm happy for this to be closed. I just can't use Input Mode on my Legion Go running Bazzite at the moment, so I'll raise any issue instead of a PR.

No harm, no foul. Please close if that's what's best for the project! Thank you 🙏

smcv:
Is there an issue report for this? I think it would be valuable to have one, that describes the problem without assuming that any particular solution is right, so that discussion about ways to solve it doesn't have to happen on a PR that encapsulates a particular attempt at a solution. Perhaps "can't identify unsandboxed clients with elevated capabilities as unsandboxed"?

Turning this around, instead of trying to prove that the process is not Flatpak / not Snap / not linyaps / etc., is there a way that a process can prove that it is explicitly part of the TCB? I've never been very happy with the way x-d-p needs to check whether the client is Flatpak or Snap or (etc.), and then if it's none of those, assume/hope that it's in the TCB. This was done as a pragmatic approach, so that existing un-sandboxed apps can use x-d-p (possibly implicitly, via libraries like GTK and Qt) without needing to do anything special - and for that use-case, it's probably a necessary evil.

But, there are a finite number of processes with elevated capabilities, and those processes are already doing something unusual (and are responsible for ensuring that they can't be tricked into exercising those capabilities on behalf of other processes, although I'm sceptical whether that's actually practically feasible). So, could we perhaps require those processes to register with a service that exists on the un-sandboxed session bus before they call into any portal APIs, and document that sandboxing frameworks must not allow sandboxed apps to contact that service? (similar to the way sandboxed apps can't be allowed to contact `o.fd.systemd1` or `o.fd.Flatpak`)

smcv:
> require those processes to register with a service that exists on the un-sandboxed session bus before

Perhaps the `o.fd.host.portal.Registry` interface can become that, but at the moment it's on the same bus name as the rest of x-d-p, and we're gating access to it by relying on the same app-identification that isn't working in this case. For that to work, it'd have to be exposed on a separate D-Bus name.

smcv:
> Is there an issue report for this? I think it would be valuable to have one, that describes the problem without assuming that any particular solution is right, so that discussion about ways to solve it doesn't have to happen on a PR that encapsulates a particular attempt at a solution. Perhaps "can't identify unsandboxed clients with elevated capabilities as unsandboxed"?

We can use #1490 for this.

smcv (changes requested):
I think this approach is too dangerous, and I'd be inclined to close the PR without merging it, but perhaps the comments I left on this PR (which I've now forwarded to https://github.com/flatpak/xdg-desktop-portal/issues/1490#issuecomment-5526699834) could be a basis for solving this in a safer way.

TalonGrayson:
Agreed - I took a stab but I'm in over my head, so I'm more than happy to close this PR. Lessons were learned, best of luck with the fix 🙂

PR #2120, opened 2026-08-27
request-dex: Handle peer disconnect during proxy creation

razzeee:
Raised separately as mentioned in https://github.com/flatpak/xdg-desktop-portal/pull/2117#issuecomment-5442583500

### DIFF
--- a/desktop-portal/xdp-request-dex.c
+++ b/desktop-portal/xdp-request-dex.c
@@ -157,11 +157,31 @@ typedef struct _RequestImplProxyCreateData {
   XdpAppInfo *app_info;
   GDBusInterfaceSkeleton *skeleton;
   char *id;
+  gulong peer_disconnect_handler;
+  gboolean peer_disconnected;
 } RequestImplProxyCreateData;
+
+static void
+on_pending_peer_disconnect (XdpContext *context,
+                            const char *peer,
+                            gpointer    user_data)
+{
+  RequestImplProxyCreateData *data = user_data;
+
+  if (g_strcmp0 (xdp_app_info_get_sender (data->app_info), peer) == 0)
+    data->peer_disconnected = TRUE;
+}
 
 static void
 request_impl_proxy_create_data_free (RequestImplProxyCreateData *data)
 {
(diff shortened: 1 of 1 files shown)


razzeee:
@swick please sanity check this

ravi-arnan (commented):
Read this from the threading side, since I did the #2108 lock audit and the peer-disconnect rework there. Scope honesty first: `xdp-request-dex.c` has no in-tree callers yet and I have no meson/libdex to build with locally, so this is a static review; compile is covered by CI.

Verified and holding up:

- This closes the exact gap #2117 deferred: a peer vanishing between the path claim and proxy-ready no longer produces a half-built request.
- The ownership flow is correct: on success everything is stolen into the request, so the free function degrades to a no-op, and unclaim only runs when `id != NULL`, so no double-unclaim path.
- Side-effect fix worth naming: the base free function never unclaimed the object path (checked against the parent commit), so a *failed* proxy creation leaked the claim. This PR unclaims on every non-success exit.
- `G_IO_ERROR_CANCELLED` is the right error for "caller went away", and the sender-match is consistent with the existing `on_peer_disconnect` handlers.

One question (threading, decides two things): `peer_disconnected` is written in `on_pending_peer_disconnect`, which runs on the `GDBusConnection` dispatch thread via `NameOwnerChanged`, and read in `on_impl_request_proxy_created`, a dex continuation. If those two run on different threads, (1) the plain-`gboolean` flag is a data race with no visibility guarantee — `gint` + `g_atomic_int_set/get` removes the question for one word; (2) there is a clear-then-connect window, a disconnect emitted after the pending handler is cleared but before the request's own handler connects is missed entirely, leaving a request created and later exported for a dead peer with no further event to tear it down. A re-check of the flag after `g_signal_connect_object`, routed into teardown if set, would close it. If the dex side is guaranteed to run on the same thread as the D-Bus dispatch, both points dissolve. Which is it?

Nit: the one-line commit message doesn't mention the unclaim-on-failure side fix. One sentence would help future readers.

No blocking issues from me; leaving approval to the maintainers.

PR #2124, opened 2026-09-02
authorize: don't dereference a NULL GError in the deny path

### LINKED ISSUE #2123: xdg-desktop-portal SIGSEGV in authorize_callback (NULL error deref at xdp-context.c:236) on host GTK app exit
# xdg-desktop-portal SIGSEGV in `authorize_callback` — NULL `error` deref at `xdp-context.c:236`

## Summary

`xdg-desktop-portal` crashes (SIGSEGV) in `authorize_callback` at
`src/xdp-context.c:236` when it tries to build an "Access Denied" error
message from a `GError` that is `NULL`. The `NULL` error originates in
`xdp_connection_get_pidfd_sync` (`src/xdp-utils.c`), where a failed
`GetConnectionCredentials` D-Bus call can return `FALSE` **without setting
`error`** (the immediately preceding `g_propagate_error (error, ...)` is
asserted away because `src == NULL`).

The race is hit when the caller process is exiting/concurrently vanishing,
so the D-Bus name owner is no longer resolvable to a pidfd.

## Environment

- Fedora 44 (`xdg-desktop-portal-1.22.1-1.fc44`)
- GNOME Shell 50.4 / mutter 50.4, Wayland session
- Recurring since at least 2026-09-02; **10 identical cores in one day**

## Crash signature

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x... in authorize_callback (interface=<optimized out>,
        invocation=0x..., user_data=0x...) at ../src/xdp-context.c:236
236	      g_dbus_method_invocation_return_error (invocation,
...
info locals:
  context = 0x...
  app_info = 0x0
  error = 0x0          <-- NULL
```

Register/dump at the faulting instruction:

```
=> 0x... <authorize_callback+1164>:  mov    0x8(%rax),%rbx
rax  0x0
rdi  0xffffffff                     (GDBusError -1)
#0  0x... in authorize_callback () at ../src/xdp-context.c:236
```

The worker thread is a gdbus dispatch thread (`dispatch_in_thread_func`),
i.e. the crash happens while authorizing an incoming portal method call.

### 1. `authorize_callback` dereferences `error->message` when `app_info == NULL`

```c
// src/xdp-context.c:230-241
app_info = xdp_app_info_registry_ensure_for_invocation_sync (context->app_info_registry,
                                                             invocation,
                                                             NULL,
                                                             &error);
if (app_info == NULL)
  {
    g_dbus_method_invocation_return_error (invocation,
                                           G_DBUS_ERROR,
                                           G_DBUS_ERROR_ACCESS_DENIED,
                                           "Portal operation not allowed: %s", error->message);
    return FALSE;
  }
```

`GError *error = NULL;` (line 229). If the helper returns `NULL` without
writing to `error`, line 239 dereferences `error->message` → SIGSEGV.

### 2. `xdp_connection_get_pidfd_sync` can return `FALSE` with `error` still `NULL`

```c
// src/xdp-utils.c:184-211
reply = g_dbus_connection_call_with_unix_fd_list_sync (connection,
                                                       DBUS_DBUS_NAME, ...,
                                                       "GetConnectionCredentials",
                                                       ...,
                                                       &local_error);

if (!reply)
  {
    if (g_error_matches (local_error, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_INTERFACE))
      return xdp_connection_get_pid_legacy (...);   // legacy path

    g_propagate_error (error, g_steal_pointer (&local_error));  // <-- local_error may be NULL
    return FALSE;
  }
```

When the D-Bus call fails but **`local_error` is also `NULL`** — the signature
observed right before every crash:

```
xdg-desktop-portal[...]: g_propagate_error: assertion 'src != NULL' failed
```

`g_propagate_error (error, NULL)` hits the `g_return_if_fail (src != NULL)`
assert, which warns and returns without setting `error`. The function then
returns `FALSE` with `*error` still `NULL`, and the NULL propagates up to
`authorize_callback`.

### 3. Why `local_error` can be `NULL`

`g_dbus_connection_call_with_unix_fd_list_sync` performs a synchronous call to
`org.freedesktop.DBus.GetConnectionCredentials` for the caller's unique bus
name and tries to obtain the caller's **pidfd**. When the caller process
is exiting / its name is being released concurrently, the credential/pidfd
lookup can fail without a populated `GError` (e.g. "UnixProcessIdUnknown /
No such process"). This is corroborated by adjacent session logs:

```
gnome-session: Could not create transient scope ... UnixProcessIdUnknown: ... No such process
```

## A notable common frame: `monitor_background_in_thread`

Every one of the analysed cores also contains thread `monitor_background_in_thread`
(`xdg-desktop-portal + 0x4a22b`) at frame #5, in addition to the crashing
`dispatch_in_thread_func` / `authorize_callback` thread. This strongly suggests
the recurring trigger is the **Background / application-monitoring** area of
the portal (applications registered as background/running-on-demand apps),
rather than an arbitrary interface.

## Triggering pattern (high confidence)

The crash correlates strongly with **host (non-Flatpak) GTK/WebKit apps that
call portals and then exit/restart quickly**. On the affected system the
recurring trigger is a long-running WebKitGTK host app (Ryn shell + WebKit)
that frequently quits/restarts; each crash occurs seconds after one of those
quit/restart transitions, while portal is still authorizing an in-flight
portal method for a process that is concurrently disappearing.

App-launch/quit → portal-crash pairs observed within one day (CST):
- 03:45:18 app quit  → 03:45:55 crash
- 03:49:01 app quit  → 03:50:23 crash
- 04:48:19 app quit  → 04:48:29 crash
- 04:55:09 app quit  → 04:55:19 crash
- 20:49:12 app quit  → 20:49:26 crash
- 20:50:21 app quit  → 20:50:35 crash
- 21:30:12 shell restart → 21:30:40 crash

All 10 cores share the identical faulting frame (`authorize_callback`
at `xdp-context.c:236`, same `+0x7affc` offset, same `g_propagate_error`
assert immediately preceding).

## Why this isn't the (already-fixed) #1718 / #1729

#1718/#1729 fixed the "Flatpak app without `APP_ID.desktop`" authorization
path in 1.20.3. This crash is on a **different path**: the failure is in
resolving the caller's **credentials/pidfd** (`xdp-connection.c` /
`xdp-utils.c`) for a host (non-Flatpak) call, which is still present in
`1.22.1`. The two are distinct.

## Suggested fix

Make `authorize_callback` robust to a `NULL` `error`:

```c
if (app_info == NULL)
  {
    g_dbus_method_invocation_return_error (invocation,
                                           G_DBUS_ERROR,
                                           G_DBUS_ERROR_ACCESS_DENIED,
                                           "Portal operation not allowed");
    return FALSE;
  }
```

And/or ensure `xdp_connection_get_pidfd_sync` (and callers) never return
`FALSE` with `*error` left `NULL` — e.g. treat a `NULL local_error` after a
failed call as its own error and set one via `g_set_error` before returning.

## Attachments

- 10 coredumps (`sig=SEGV`, exec `/usr/libexec/xdg-desktop-portal`), all
  `authorize_callback` at `xdp-context.c:236`
- Journal asserts `g_propagate_error: assertion 'src != NULL' failed` on
  every occurrence, immediately before the core is written
- This report can be reproduced on Fedora 44 / xdg-desktop-portal 1.22.1
  by running a host GTK app that calls a portal and exits during the call.

ZK-Andy:
Fixes #2123

`xdp_app_info_registry_ensure_for_invocation_sync` can return `NULL` `app_info`
without setting `*error`, for example when the caller's credentials/pidfd
cannot be resolved because the caller process is exiting concurrently (see
`g_dbus_connection_call...GetConnectionCredentials` failures reported as
`g_propagate_error: assertion 'src != NULL' failed`).

In that case `error` is `NULL`, and the deny path

```c
"Portal operation not allowed: %s", error->message
```

dereferences a `NULL` pointer, crashing `xdg-desktop-portal` with `SIGSEGV`
in `authorize_callback` (both the async `authorize_callback_fiber` and the
sync `authorize_callback` paths).

This change makes the deny path robust by only using `error->message` when it
is set, falling back to a literal otherwise. The authorization decision is
unchanged — only the error message is now always safe to construct.

We reproduced this on Fedora 44 / xdg-desktop-portal 1.22.1 with a host
(non-Flatpak) GTK/WebKit app that calls a portal and exits while the call is
in flight. Note: the crash also exists on current `main`; please confirm
before merge.

### DIFF
--- a/shared/xdp-utils.c
+++ b/shared/xdp-utils.c
@@ -177,6 +177,32 @@ connection_get_pidfd_legacy_fiber (GDBusConnection *connection,
   return xdp_pid_fd_result_new (pid, -1);
 }
 
+/*
+ * connection_get_pidfd_fiber_reject:
+ *
+ * Build a rejected future for the pidfd path, always with a non-NULL error.
+ *
+ * @local_error may be NULL when the underlying libdex/GDBus call fails without
+ * populating an error (for example when the caller's process is exiting
+ * concurrently, so its credentials/pidfd can no longer be resolved). Passing a
+ * NULL GError to dex_future_new_for_error() would create a rejected future
+ * whose error is NULL, which is then propagated to dex_await_object() and ends
+ * up dereferenced by authorize_callback() (xdp-context.c) as error->message,
+ * crashing xdg-desktop-portal with a SIGSEGV.
+ */
+static DexFuture *
+connection_get_pidfd_fiber_reject (GError      *local_error,
+                                   const char  *sender)
+{
+  if (local_error != NULL)
+    return dex_future_new_for_error (g_steal_pointer (&local_error));
+
+  return dex_future_new_reject (G_IO_ERROR,
+                                G_IO_ERROR_FAILED,
+                                "Failed to get pidfd for sender '%s'",
+                                sender);
+}
+
 static DexFuture *
 connection_get_pidfd_fiber (GDBusConnection *connection,
                             const char      *sender)
@@ -216,7 +242,7 @@ connection_get_pidfd_fiber (GDBusConnection *connection,
             return dex_future_new_take_boxed (XDP_TYPE_PID_FD_RESULT,
                                               g_steal_pointer (&result));
         }
-      return dex_future_new_for_error (g_steal_pointer (&local_error));
+      return connection_get_pidfd_fiber_reject (local_error, sender);
     }
 
   reply = dex_await_variant (dex_ref (dex_future_set_get_future_at (DEX_FUTURE_SET (future), 0)), NULL);
@@ -235,7 +261,7 @@ connection_get_pidfd_fiber (GDBusConnection *connection,
         return dex_future_new_take_boxed (XDP_TYPE_PID_FD_RESULT,
                                           g_steal_pointer (&result));
       else
-        return dex_future_new_for_error (g_steal_pointer (&local_error));
+        return connection_get_pidfd_fiber_reject (local_error, sender);
     }
 
   pid = g_variant_get_uint32 (process_id);
@@ -255,7 +281,7 @@ connection_get_pidfd_fiber (GDBusConnection *connection,
 
   pidfd = g_unix_fd_list_get (fd_list, fd_id, &local_error);
   if (pidfd < 0)
-    return dex_future_new_for_error (g_steal_pointer (&local_error));
+    return connection_get_pidfd_fiber_reject (local_error, sender);
 
   return dex_future_new_take_boxed (XDP_TYPE_PID_FD_RESULT,
                                     xdp_pid_fd_result_new (pid, g_steal_fd (&pidfd)));


smcv (changes requested):
> `xdp_app_info_registry_ensure_for_invocation_sync` can return `NULL` `app_info` without setting `*error`

That would be a bug in `xdp_app_info_registry_ensure_for_invocation_sync`. See https://docs.gtk.org/glib/error-reporting.html for the conventions used.

If the function is failing a precondition check (`g_return_val_if_fail`) as a result of a programming error, then the caller needs to avoid calling it with the preconditions not met.

Otherwise, `xdp_app_info_registry_ensure_for_invocation_sync` needs to always set the `error` before returning `NULL`.

ZK-Andy:
Hi @smcv — thanks for the review. You're right that the caller-side guard was papering over the root cause, not fixing it. I've reworked the PR to address it where the error actually originates, and kept `authorize_callback` unchanged.

### What changed (root-cause fix, not a caller guard)

The PR now fixes [`connection_get_pidfd_fiber`](https://github.com/flatpak/xdg-desktop-portal/blob/0b6e3ce/shared/xdp-utils.c) in `shared/xdp-utils.c`. That function is what becomes the `NULL *error` the callback later dereferences:

- `connection_get_pidfd_fiber` had three failure paths that did
  ```c
  return dex_future_new_for_error (g_steal_pointer (&local_error));
  ```
  When the underlying D-Bus call (`GetConnectionCredentials` / pidfd resolution) fails **without populating `local_error`** — the typical case being the caller process exiting concurrently, so its credentials/pidfd can no longer be resolved — it built a rejected future carrying a `NULL` GError.
- That NULL error rides up `xdp_app_info_registry_ensure_for_invocation_sync` → `xdp_app_info_new_for_invocation` → `dex_await_object(..., &error)`, and `authorize_callback` then does `error->message` → SIGSEGV.

So the fix introduces a helper that always produces a non-NULL error:

```c
static DexFuture *
connection_get_pidfd_fiber_reject (GError      *local_error,
                                   const char  *sender)
{
  if (local_error != NULL)
    return dex_future_new_for_error (g_steal_pointer (&local_error));

  return dex_future_new_reject (G_IO_ERROR,
                                G_IO_ERROR_FAILED,
                                "Failed to get pidfd for sender '%s'",
                                sender);
}
```

and routes all three failure branches through it. `authorize_callback` / `authorize_callback_fiber` in `xdp-context.c` are left exactly as upstream (the `"Portal operation not allowed: %s", error->message` stays).

This matches the convention you linked: a function that fails now always sets the error when it returns a failure result — here the rejected future always carries a real GError instead of letting a NULL one escape.

### Verification

- `gcc -fsyntax-only` on `shared/xdp-utils.c` passes with the project's flags (`-Wall -Wextra -Werror -Wno-unused-parameter ...`), exit 0.
- I could not do a full local `ninja` link because the libdex on this machine (`libdex-1.1.0`) predates `dex_scheduler_spawnv`, so `settings.c` hits an unrelated `implicit declaration` (filed separately in #2125, PR #2126). That's independent of this change — `xdp-utils.c` already includes `xdp-dex.h` and is unaffected.

Happy to adjust if you'd prefer the helper to live elsewhere, or to also harden the `dex_future_new_for_error` callers in `xdp-app-info.c` (I left those untouched here to keep the change focused on the crash path).

swick:
> When the underlying D-Bus call (GetConnectionCredentials / pidfd resolution) fails without populating local_error

If the observations here are true, then *that* is the root issue and the patch papers over it. This should not happen. either something in `dex_dbus_connection_call_with_unix_fd_list` or how the future eventually resolves with `dex_await` would then be the culprit.

ZK-Andy:
I'm closing this PR. After digging into the crash, I found that the fix direction here was wrong, and the problem it targets no longer exists on current `main`. Apologies for the noise.

## What the crash actually was

The bug reproduces on Fedora 44 / xdg-desktop-portal 1.22.1. A host (non-Flatpak) GTK/WebKit app (DeepSeek Harness Desktop) that calls a portal and exits while the call is in flight crashes `xdg-desktop-portal` with SIGSEGV in `authorize_callback`.

From the coredump + journal:
- crashed frame: `authorize_callback` at `src/xdp-context.c:236` (the `"Portal operation not allowed: %s", error->message` deny path)
- locals: `app_info = 0x0`, `error = 0x0`; faulting instruction `mov 0x8(%rax),%rbx` with `rax = 0` (deref of `error->message` through NULL)
- journal immediately before: `g_propagate_error: assertion 'src != NULL' failed`
- core strings: sender is `io.github.ZK-Andy.dotnet-deepseek-harness-desktop`; in-flight interfaces include `org.freedesktop.portal.Notification/Background/FileChooser/Screenshot/OpenURI`; and `org.freedesktop.DBus.Error.UnixProcessIdUnknown` was produced on the legacy pidfd path.

## The real path in 1.22.1

In 1.22.1 the caller path is the **synchronous** one:

```
xdp_app_info_new_for_invocation_sync
  → xdp_connection_get_pidfd_sync  (src/xdp-utils.c)
    → g_dbus_connection_call_with_unix_fd_list_sync fails
    → g_propagate_error (error, g_steal_pointer (&local_error))
```

When the D-Bus call fails but `local_error` is also NULL (the exiting-caller case), `g_propagate_error(error, NULL)` trips the `src != NULL` assertion, leaves `*error == NULL`, and that NULL rides to `authorize_callback`'s `error->message` -> SIGSEGV.

## Why I'm closing it

This synchronous path has **already been removed upstream**. Commits on 2026-08-05:

- `81d24113` "utils: port appinfo constructors to Dex API" — future-ized `xdp_app_info_new_for_invocation_sync` (made it async).
- `aaf47004` "utils: clean up dead sync app info construction code" — removed the remaining sync code for good.

current `main` (and this PR's base) no longer contain `xdp_connection_get_pidfd_sync`; the call is now async via `connection_get_pidfd_fiber`. I verified with a small repro against libdex that the async `dex_await` failure path returns a **non-NULL** error (`"Too many futures failed"`), so the async path does not produce the NULL GError that this crash comes from.

The original PR direction (a `connection_get_pidfd_fiber_reject` helper in `xdp-utils.c`) was also wrong: my repro showed that failure in `connection_get_pidfd_fiber` yields a non-NULL error, so that path wasn't the source of the NULL — fixing it there would have been the "papers over the root cause" pattern rather than a real fix.

So there is no change to make on current `main`; closing this PR as not needed. (If the goal is to backport a fix to 1.22.x, that belongs on the release branch, not here.)

Apologies again for the churn, and thank you @smcv for the review that pointed out the caller-side guard was masking the actual issue.

ZK-Andy:
Closing as not needed — see the reasoning above. The synchronous path this targeted was already removed upstream (2026-08-05), and the async replacement does not produce the NULL GError this crash came from. No change is required on current main.

PR #2126, opened 2026-09-02
settings: include xdp-dex.h for the dex_scheduler_spawnv fallback

### LINKED ISSUE #2125: settings.c fails to build when libdex does not provide dex_scheduler_spawnv
# settings.c fails to build when libdex does not provide dex_scheduler_spawnv

## Summary

`desktop-portal/settings.c` calls `dex_scheduler_spawnv()` but does not
`#include "xdp-dex.h"`. This only fails to compile when
`HAVE_DEX_SCHEDULER_SPAWNV` is 0 — i.e. when the linked `libdex` does not
export `dex_scheduler_spawnv` itself — because in that case the fallback
declaration lives in `xdp-dex.h` (guarded by `#if !HAVE_DEX_SCHEDULER_SPAWNV`),
which `settings.c` never includes.

## Reproduction

Build with a libdex version that lacks `dex_scheduler_spawnv`
(observed on Fedora 44 / `libdex-1.1.0-1.fc44`), with `-Dwerror=true`:

```
meson setup _build -Dwerror=true
ninja -C _build
```

Meson detects the missing function:

```
Checking for function "dex_scheduler_spawnv" with dependency libdex-1: NO
```

and then `ninja` fails in `settings.c`:

```
../desktop-portal/settings.c: In function 'on_impl_settings_changed':
../desktop-portal/settings.c:287:7: error: implicit declaration of function 'dex_scheduler_spawnv'; did you mean 'dex_scheduler_spawn'? [-Wimplicit-function-declaration]
  287 |       dex_scheduler_spawnv (NULL, 0,
      |       ^~~~~~~~~~~~~~~~~~~~
      |       dex_scheduler_spawn
../desktop-portal/settings.c:287:7: error: passing argument 1 of 'dex_future_first' makes pointer from integer without a cast [-Wint-conversion]
```

## Root cause

`dex_scheduler_spawnv` is a portal-internal wrapper defined in
`shared/xdp-dex.c` and declared in `shared/xdp-dex.h`:

```c
// shared/xdp-dex.h
#if !HAVE_DEX_SCHEDULER_SPAWNV
DexFuture *
dex_scheduler_spawnv (DexScheduler *scheduler,
                      gsize         stack_size,
                      GCallback     callback,
                      guint         n_params,
                      ...);
#endif /* !HAVE_DEX_SCHEDULER_SPAWNV */
```

The other callers (`shared/xdp-app-info-registry.c`, `shared/xdp-utils.c`)
include `xdp-dex.h`. `settings.c` does not, so when the fallback declaration
is required it is not visible and `dex_scheduler_spawnv` is treated as an
implicit function.

## Suggested fix

Add `#include "xdp-dex.h"` to `desktop-portal/settings.c`.

(Root-cause alternative that would also cover any future caller: have
`xdp-utils.h` or a common shared header include `xdp-dex.h`. But the minimal,
targeted fix is the `settings.c` include, consistent with the other callers.)

## Note

This is independent from the fixed crash in
[`authorize_callback`](https://github.com/flatpak/xdg-desktop-portal/issues/2123)
and the corresponding PR
[#2124](https://github.com/flatpak/xdg-desktop-portal/pull/2124).
It was found while trying to build that PR locally.

ZK-Andy:
Fixes #2125

`settings.c` calls `dex_scheduler_spawnv()` but never includes `shared/xdp-dex.h`, where the `!HAVE_DEX_SCHEDULER_SPAWNV` fallback declaration lives. On a libdex build without the function, meson reports `dex_scheduler_spawnv ... : NO` and the build fails with an implicit-declaration error at `settings.c:287`.

Adding the include is consistent with the other callers (`xdp-app-info-registry.c`, `xdp-utils.c`) which already include `xdp-dex.h`.

Verified: `gcc -fsyntax-only` on `settings.c` passes after this change.

### DIFF
--- a/desktop-portal/settings.c
+++ b/desktop-portal/settings.c
@@ -16,6 +16,7 @@
 
 #include "xdp-context.h"
 #include "xdp-dbus.h"
+#include "xdp-dex.h"
 #include "xdp-impl-dbus.h"
 #include "xdp-portal-config.h"
 #include "xdp-utils.h"


swick:
Thanks!

PR #2129, opened 2026-09-04
Add support for experimental portals

### LINKED ISSUE #2066: Add support for experimental portals
While working on new portals, it would be beneficial to have a feature flag system to get portals merged into `main` incrementally while they are being worked on to have smaller PRs/reviews and to get easier feedback from the community by having the patches in the central XDP repository and being able to enable the experimental portal at runtime. 

The in-progress credential portal, for example, uses the environment variable `XDG_DESKTOP_PORTAL_ENABLE_EXPERIMENTAL=portal1,portal2,...` as a feature flag.

Some considerations:
- what is the mechanism for providing the feature flag? An environment variable as above seems apt.
- How should we separate experimental portals at the D-Bus level?
  - what path should host experimental portals (`/org/freedesktop/portal/desktop/experimental`?)
  - should we have `experimental` somewhere in the interface name
- This requires work in the code generation and documentation scripts to visually differentiate the experimental portals in the documentation, and if we use a separate interface prefix, to strip the experimental interface prefix from the page title and generated method/class names.

iinuwa:
Fixes #2066.

### DIFF
--- a/desktop-portal/xdp-context.c
+++ b/desktop-portal/xdp-context.c
@@ -44,6 +44,7 @@
 #include "xdp-portal-config.h"
 #include "xdp-request.h"
 #include "xdp-session-persistence.h"
+#include "xdp-types.h"
 #include "xdp-utils.h"
 
 enum
@@ -341,13 +342,27 @@ xdp_context_take_and_export_portal (XdpContext             *context,
         }
     }
 
-  if (g_dbus_interface_skeleton_export (skeleton,
-                                        context->connection,
-                                        DESKTOP_DBUS_PATH,
-                                        &error))
-    g_debug ("Providing portal %s", name);
+  if (flags & XDP_CONTEXT_EXPORT_FLAGS_EXPERIMENTAL)
+    {
+      if (g_dbus_interface_skeleton_export (skeleton,
+                                            context->connection,
+                                            DESKTOP_EXPERIMENTAL_DBUS_PATH,
+                                            &error))
+        g_debug ("Providing experimental portal %s", name);
+      else
+        g_warning ("Exporting portal failed: %s", error->message);
+    }
+
   else
-    g_warning ("Exporting portal failed: %s", error->message);
+    {
+      if (g_dbus_interface_skeleton_export (skeleton,
+                                            context->connection,
+                                            DESKTOP_DBUS_PATH,
+                                            &error))
+        g_debug ("Providing portal %s", name);
+      else
+        g_warning ("Exporting portal failed: %s", error->message);
+    }
 
   g_hash_table_insert (context->exported_portals,
                        g_strdup (name),
--- a/doc/experimental-portals.rst
+++ b/doc/experimental-portals.rst
@@ -1,0 +1,66 @@
+..
+   SPDX-License-Identifier: LGPL-2.1-or-later
+   SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
+
+Experimental Portals
+====================
+
+New portals are often developed in collaboration with both desktop developers
+and app developers. In order to get feedback on the portal, it may be
+beneficial to implement an experimental portal interface. These interfaces are
+intended to be temporary and do not promise any stability guarantees:
+developers should expect that these interfaces may change or become unavailable
+at any time.
+
+Interface Naming Conventions
+----------------------------
+
+Experimental portal frontends must follow the following conventions:
+
+- The frontend object path must be
+  ``/org/freedesktop/portal/desktop/experimental``. The backend object path
+  should be the same.
+- The interface name must follow the pattern
+  ``org.freedesktop.portal.Name.X#``, where ``#`` is a number that increments
+  on major changes to the experimental portal.
+- When the portal is promoted to stable, the ``.X#`` suffix in the interface
+  will be dropped, and the version property will be reset to ``1``.
+- The ``version`` property is a number that increments with minor changes to
+  the portal that are backwards compatible with previous client implementations
+  under the same major interface version. Clients should abort if the major
+  version ``X#`` does not match exactly or if the ``version``
+  property is lower than what they expect.
+
+Experimental Merge Requirements
+-------------------------------
+
+In order to be merged into the project as an experimental portal, experimental
+frontend implementations must meet minimum stability requirements, including:
+
+- The intent and design of the portal must be discussed with and accepted by
+  one of the committers.
+- The code must be reasonably correct so as not to cause security issues for
+  other portals running in the same process.
+- Code and commit message hygiene as documented in :doc:`pull-requests` must be
+  followed.
+
+Exposing Experimental Portals
+-----------------------------
+
+Like all other portals, experimental portals will not start unless a backend is
+configured and discovered at startup, so application developers will have to
+install and configure an appropriate backend in order for the frontend to
+become available. No other mechanism is required for exposing the experimental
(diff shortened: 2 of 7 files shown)


desktop-portal/xdp-context.c
```diff
@@ -341,9 +343,17 @@ xdp_context_take_and_export_portal (XdpContext             *context,
...
     }
 
+  if (flags & XDP_CONTEXT_EXPORT_FLAGS_EXPERIMENTAL)
+    {
+      object_path = DESKTOP_EXPERIMENTAL_DBUS_PATH;
+    }
+  else
+    {
+        object_path = DESKTOP_DBUS_PATH;
+    }
```
TheEvilSkeleton, line 365:
```suggestion
  if (flags & XDP_CONTEXT_EXPORT_FLAGS_EXPERIMENTAL)
    object_path = DESKTOP_EXPERIMENTAL_DBUS_PATH;
  else
    object_path = DESKTOP_DBUS_PATH;
```

TheEvilSkeleton (changes requested):

desktop-portal/xdp-context.c
iinuwa: I went ahead and moved the whole export block to the if statement so that it'd be clearer in the logs when the portal is being exported as experimental.

iinuwa:
@tytan652 or @swick, any comments on this documentation for experimental portals?

sjtrotter:
hi there,  i'm implementing some experimental backends that utilize this in anticipation of the merge.

full disclosure: i'm using claude and codex to assist with my builds.

originally my issue was with freerdp; i connect on a govcloud remote desktop that requires oauth sign in as well as smartcard support.

in exploring the code and playing around with it, i eventually landed on the idea for two separate portals: one for web authentication, and one for a smartcard/client certificate chooser.

the one i think has the most potential for acceptance is the cert chooser: https://github.com/sjtrotter/xdg-desktop-portal-certificate

i have a small fix that changes how the name is generated. this version parses the name as e.g. " Certificate . X 1" and the fix changes it to "Certificate (Experimental X1)". @iinuwa i can submit it to your branch as a PR for inclusion if you'd like to see it?

iinuwa:
@sjtrotter, sure that'd be helpful!

sjtrotter:
okay, just pushed it -- it should be with you for review

doc/experimental-portals.rst
```diff
@@ -0,0 +1,66 @@
...
+----------------------------
+
+Experimental portal frontends must follow the following conventions:
+
+- The frontend object path must be
+  ``/org/freedesktop/portal/desktop/experimental``. The backend object path
+  should be the same.
+- The interface name must follow the pattern
+  ``org.freedesktop.portal.Name.X#``, where ``#`` is a number that increments
+  on major changes to the experimental portal.
```
orowith2os, line 25: Backends usually expose these under names and paths like `org.freedesktop.impl.portal`, don't they?

doc/experimental-portals.rst
```diff
@@ -0,0 +1,66 @@
...
+frontend to application developers.
+
+Stable Promotion
+----------------
+
+In order to be promoted to stable, the experimental portal must:
+
+- Go through final review by the committers, including frontend code and
+  documentation for both the frontend and backend.
+- Be implemented and merged into at least one major backend implementation.
```
orowith2os, line 63: Why not 3? GNOME, KDE, and Cosmic, for example. Those are the major desktop environments I see floating around as of late, and having more than one major backend implementation means potentially exposing different problems that wouldn't show up otherwise.

doc/experimental-portals.rst
```diff
@@ -0,0 +1,66 @@
...
+at any time.
+
+Interface Naming Conventions
+----------------------------
+
+Experimental portal frontends must follow the following conventions:
+
+- The frontend object path must be
+  ``/org/freedesktop/portal/desktop/experimental``. The backend object path
+  should be the same.
```
swick, line 22: What's the reason for the different object path? The way I see it we either change the object path or we change the interface name. The interface name is conceptually the better fit because it describes the interface/API.

doc/experimental-portals.rst
```diff
@@ -0,0 +1,66 @@
...
+  on major changes to the experimental portal.
+- When the portal is promoted to stable, the ``.X#`` suffix in the interface
+  will be dropped, and the version property will be reset to ``1``.
+- The ``version`` property is a number that increments with minor changes to
+  the portal that are backwards compatible with previous client implementations
+  under the same major interface version. Clients should abort if the major
+  version ``X#`` does not match exactly or if the ``version``
+  property is lower than what they expect.
+
+Experimental Merge Requirements
```
swick, line 34: Not sure if this section actually says anything? All of that is trivially the case and follows from our existing guidelines.

doc/experimental-portals.rst
```diff
@@ -0,0 +1,66 @@
...
+  followed.
+
+Exposing Experimental Portals
+-----------------------------
+
+Like all other portals, experimental portals will not start unless a backend is
+configured and discovered at startup, so application developers will have to
+install and configure an appropriate backend in order for the frontend to
+become available. No other mechanism is required for exposing the experimental
+frontend to application developers.
```
swick, line 54: That is not how other portals work. There are some portals which do not require backends and they start automatically. I really think that we do need an explicit toggle for experimental portals.

swick (commented):
Thanks for pushing this.

doc/experimental-portals.rst
tytan652: In this case we will need to introduce staging portals, because not every non-stable needs to be behind an opt-in.

doc/experimental-portals.rst
tytan652: No it's worse because experimental portal should not be under the same path as stable ones.

Edit: The different paths is even the base of the idea discussed in the matrix room.

doc/experimental-portals.rst
iinuwa: no, the object path for backends is also typically `/org/freedesktop/portal/desktop`. `impl` is included in the backend interface though.

doc/experimental-portals.rst
iinuwa: I think that having 3 would slow down adoption too much. I think it would be better to keep it as 1 implementation and do a V2 if issues that require breaking changes come up in response to another implementation.

doc/experimental-portals.rst
iinuwa: I felt it was worth repeating here that these minimum requirements. For example, a portal might not be feature-complete before being merged as an experimental portal, but its security and code commit requirements need to match the standards of the rest of the project. Having it explicitly listed here can reduce arguments about it later on, rather than new contributors attempting to bypass security requirements "because it's just experimental."

Up to you if you prefer not to have the section.

doc/experimental-portals.rst
tytan652: We really need to make explicit to **app developer**  of when they are using non-stable Portal.

And putting them in the same object path as stable portals would be a wrong step.

We must not make stability-level confusing but clearly explicit (in code), we are at app level being too "cryptic" is the last thing we should do.

The object path is heavily visible and human readable so using `/org/freedeskktop/portal/experimental` should be heavily considered.

PS: If app developer don't to want to learn/understand some basics of D-Bus (or any other IPC), it's their problem.

PR #2130, opened 2026-09-05
build: Use libdex 1.2beta as subproject

iinuwa:
Since 81d24113d088685d26902d6d89e7dbe9e18ac2bb, we require more features from unreleased libdex, which now supports being used as a Meson subproject, so this PR uses that instead of our polyfills.

Draft for now because I can't test the container because Ubuntu's repos are being flaky right now.


iinuwa:
Looks like tests pass on gcc, but there's an upstream warning about atomics performance when with clang.

It seems that one way we can get around this by using `-mcx16` to require the `CMPXCHG16B` instruction, which is available on x86_64 processors since 2006. But the easy way to do that (setting `c_args` in the `default_options` of the libdex subproject) doesn't seem to work; maybe libdex doesn't respect c_args?

swick:
I'd rather not add subprojects for dependencies we link to from the system. You're the second person though, so I'm a bit unsure if we should just do it, but I really want to avoid packagers using libdex as a subproject.

iinuwa:
Yeah, that's sticky. If the system does provide a high enough version, then the system dependency is preferred, so this is a fallback for older systems/less bleeding edge, which is a bit better, but it still makes it easy for a packager to use the subproject.

I guess there's a few directions we could go:
- temporarily allow the subproject while libdex settles
  - - Perhaps we could gate it on debug builds only so that developers can continue working on it?
- leave it and wait until distros catch up
- Put a hold on bumping the version of libdex until it's been taken up by a couple of distros.

swick:
1.2 has been released now and we should bump our own version requirement to that, and I also don't think it's too bad to require building a single other dependency to get xdp working.

iinuwa:
So you're saying you don't want libdex as a subprojects? Ok, if we require 1.2, I think it will fail with a clearer error, so that works for me.

I can make that change when I get back to a computer.

iinuwa:
Dropped in favor of #2157

Issue #2133, opened 2026-09-08
Deadlock when selecting hundreds of files in a snap sandboxed application


rolandd:
### Operating System

Ubuntu 26.04

### XDG Desktop Portal version

1.22

### Desktop Environment

GNOME

### Expected Behavior

when selecting many files in a sandboxed application (eg Firefox running as a snap, uploading many photos to a website), the selected files should be made available to the sandboxed app.

### Current Behavior

It seems that there are two overlapping bugs that lead to a deadlock: first, checking whether file access is allowed spawns a helper process for each file ("snap routine file-access <snap> <path>" in the snap case), and document_add_full() holds the DB lock across all such calls; second, the helper for running the subprocess xdp_spawn() calls back into the glib main loop with a default context, which means it dispatches all events, including new D-Bus requests. This seems to create a situation where the portal process gets deadlocked in a recursive main loop call.

### Steps to Reproduce

1. Go to a website that accepts multiple file uploads (eg Google Photos) in a sandboxed browser (eg Firefox running as a snap) and select a few hundred photos to upload.

### Anything else we should know?

Also, this seems somewhat related to #2056, which is not a deadlock, but a symptom of spawning a flatpak helper process individually for 1000s of files in a loop.

aleasto:
Should be closed as fixed by https://github.com/flatpak/xdg-desktop-portal/pull/2134

Issue #2056, opened 2026-07-07
multi-file selection in sandboxed apps cause slowdowns


alexballas:
### Operating System

Ubuntu 26.04

### XDG Desktop Portal version

Other

### XDG Desktop Portal version (Other)

1.21.1

### Desktop Environment

GNOME

### Expected Behavior

When thousands of files are selected in a sandboxed Flatpak app, the app should get them immediately.

### Current Behavior

When thousands of files are selected in a sandboxed Flatpak app, the app does not get them immediately. First, the portal grants access to each file one by one. The slow path is essentially: send_response() → xdp_register_document() → portal_add_full() → app_has_file_access() → xdp_spawn("flatpak", "info", "--file-access=<path>", "<app-id>"). That last command runs once per file, so tiny delays add up into minutes.

I created a proof-of-concept app to demonstrate the issue, but it should be reproducible in any fully sandboxed app.

### Steps to Reproduce

```c
#include <gio/gio.h>
#include <glib.h>
#include <stdio.h>

#define PORTAL_BUS_NAME "org.freedesktop.portal.Desktop"
#define PORTAL_OBJECT_PATH "/org/freedesktop/portal/desktop"
#define FILE_CHOOSER_IFACE "org.freedesktop.portal.FileChooser"
#define REQUEST_IFACE "org.freedesktop.portal.Request"

typedef struct {
	GMainLoop *loop;
	GTimer *timer;
	guint32 response;
	gboolean got_response;
	gboolean timed_out;
} AppState;

static const char *
response_name(guint32 response)
{
	switch (response) {
	case 0:
		return "success";
	case 1:
		return "cancelled";
	case 2:
		return "failed";
	default:
		return "unknown";
	}
}

static void
print_returned_uris(GVariant *results)
{
	GVariant *uris = g_variant_lookup_value(results, "uris", G_VARIANT_TYPE("as"));

	if (uris == NULL) {
		g_print("Returned URI count: 0 (no 'uris' result)\n");
		return;
	}

	gsize count = g_variant_n_children(uris);
	g_print("Returned URI count: %" G_GSIZE_FORMAT "\n", count);

	gsize preview = count < 5 ? count : 5;
	for (gsize i = 0; i < preview; i++) {
		const char *uri = NULL;
		g_variant_get_child(uris, i, "&s", &uri);
		g_print("  %" G_GSIZE_FORMAT ": %s\n", i + 1, uri);
	}

	if (count > preview)
[...]

PR #2134, opened 2026-09-08
Fix deadlock when selecting many files in a sandboxed application

rolandd:
As described in #2133, selecting many (100s) files in a sandboxed application such as snap-confined Firefox can lead to a deadlock in the xdg-desktop-portal due to the locking and glib main loop context around spawning the access-checking helper process. Fix both xdg_spawn to use its own context to limit event sources, and also the lock scope around the document db. Also add tests that pass with the fixes in place but fail without them.

### DIFF
--- a/tests/test-xdp-utils.c
+++ b/tests/test-xdp-utils.c
@@ -194,6 +194,40 @@ test_app_id_via_systemd_unit (void)
 }
 #endif /* HAVE_LIBSYSTEMD */
 
+static gboolean default_context_dispatched = FALSE;
+
+static gboolean
+idle_cb (gpointer data)
+{
+  default_context_dispatched = TRUE;
+  return G_SOURCE_REMOVE;
+}
+
+static void
+test_spawn_isolates_context (void)
+{
+  g_autoptr(GError) error = NULL;
+  g_autofree char *output = NULL;
+  g_autoptr(GSource) source = g_idle_source_new ();
+
+  g_source_set_callback (source, idle_cb, NULL, NULL);
+  g_source_attach (source, NULL); /* Attached to default global GMainContext */
+
+  default_context_dispatched = FALSE;
+
+  output = xdp_spawn (&error, "true", NULL);
+  g_assert_no_error (error);
+
+  /* The nested main loop in xdp_spawn_full() must run on a dedicated
+   * GMainContext and must NOT dispatch events attached to the default
+   * GMainContext. */
+  g_assert_false (default_context_dispatched);
+
+  /* Now iterate the default context and verify the idle source is dispatched */
+  g_main_context_iteration (NULL, FALSE);
+  g_assert_true (default_context_dispatched);
+}
+
 int main (int argc, char **argv)
 {
   g_test_init (&argc, &argv, NULL);
@@ -205,5 +239,6 @@ int main (int argc, char **argv)
 #if HAVE_LIBSYSTEMD
   g_test_add_func ("/app-id-via-systemd-unit", test_app_id_via_systemd_unit);
 #endif
+  g_test_add_func ("/spawn/isolates-context", test_spawn_isolates_context);
   return g_test_run ();
 }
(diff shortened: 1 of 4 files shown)


aleasto (approved):
changes seem correct to me

tests/test-xdp-utils.c
```diff
@@ -194,6 +194,40 @@ test_app_id_via_systemd_unit (void)
...
+
+static gboolean
+idle_cb (gpointer data)
+{
+  default_context_dispatched = TRUE;
+  return G_SOURCE_REMOVE;
+}
+
+static void
+test_spawn_isolates_context (void)
```
swick: This test is hyper specific to the one thing you just fixed, so lets just drop it.

tests/test_documents.py
```diff
@@ -335,6 +337,89 @@ def test_host_paths_xattr(self, xdg_document_portal, dbus_con):
         host_path = xdp_doc.get_host_path_attr(mountpoint / doc_id / "a" / "b" / "c")
         assert host_path == base_path / "b" / "c"
 
+    def test_add_as_needed_by_app_reentrancy(
```
swick: Again seems to be rather specific (typical AI generated test). It also has a lot of magic timeouts, and tries to detect a hang instead of just letting it hang and the let the test timeout take care.

swick (commented):
It's a bit unfortunate that xdp_spawn blocks now, because it means it will block threads in the document portal, and all of the desktop portal when new apps connect. It's obviously better than causing deadlocks, so lets get this merged. In the long run we should figure out how to avoid the blocking though.

rolandd:
> It's a bit unfortunate that xdp_spawn blocks now, because it means it will block threads in the document portal, and all of the desktop portal when new apps connect. It's obviously better than causing deadlocks, so lets get this merged. In the long run we should figure out how to avoid the blocking though.

Yes, although I think all current helpers are relatively short-lived, although even just the overhead of starting many processes leads to issues like #2056 . I think there are two independent paths that could be pursued to improve this, although they're both bigger changes:

- Move calling helpers to a continuation style (async or maybe Dex fiber), so waiting for the helper can use the top-level main loop - just need to make sure we don't hold locks across helper calls (my second patch). I'll investigate how big a change this is.
- Have the snap and flatpak permissions helpers support a peristent daemon mode, so we could spawn them and then just send queries / get responses back on a pipe, which would make individual queries much faster. This is obviously a big cross-project job.

tests/test-xdp-utils.c
rolandd: will drop, was useful while developing this to make sure the fix actually did something but probably is not a regression risk going forward.

tests/test_documents.py
rolandd: again wanted to prove end-to-end that the change actually fixed something, I feel this test has some value but can drop as well if you want. can definitely get rid of the hang detection, the only downside is that in the failing case the overall default timeout is longer, but that should never happen unless things regress.

swick: > I feel this test has some value

Yes, it's not awful but could maybe be generalized a bit more and the hang "detection" should be dropped for sure.

swick:
Also thanks @aleasto for the review, much appreciated

rolandd:
I took a look at moving the spawn helper to return a Dex future that could be used asynchronously, but it doesn't seem to buy much given the current structure of the places that call external helpers. The document portal already has a thread pool, so waiting for the helper on a fiber just unblocks one thread for a few milliseconds; and the snap app info and icon/sound validation are not using Dex fibers yet.

Seems like it would be a pretty invasive rewrite to make the places spawning helpers structured so they do anything useful while waiting for external processes.

I updated my branch to get rid of the specific C unit test, and tried to make the python test cleaner / more generic, updated this PR with the new test commit.

swick:
> Seems like it would be a pretty invasive rewrite to make the places spawning helpers structured so they do anything useful while waiting for external processes.

Yeah, I didn't think so. It's something to keep in mind if we ever change the code in a more fundamental way, but nothing that could be fixed with a small patch.

PR #2135, opened 2026-09-09
shared/utils: use correct index for duplicate PID mappings

kylin-liwn:
In xdp_map_pids_full, the loop that handles duplicate PID entries wrote to res[idx] instead of res[i]. When the input pids array contains duplicate values, only the first matching position was filled while duplicates remained 0, causing the function to incorrectly report "Process ids could not be found" and fail.


swick:
Good catch! Please use an appropriate git commit message tag, like `shared/utils: ` instead if `fix`.

kylin-liwn:
Thanks for the review! Good point — I'll amend the commit message to use shared/utils: as the tag and force-push shortly.

PR #2137, opened 2026-09-09
Experiment: Add a varlink interface for a few portals

jthoward64:
Ths PR adds a varlink interface for the settings, input-capture, clipboard, and secret portals. These portals were chosen as a representative set of most of the behaviors that the different portals demonstrate.

I generally tried to avoid touching dbus code, but two deeper refactors were needed.

1. Add an abstraction over dbus invocation/varlink client called XdpPeer
2. Expanded dex code and ported input capture to it

As varlink itself does not support sending file descriptors, I followed the sd-varlink convention, which is supported by a number of varlink libraries including zlink (Rust) and my fork of libvarlink (C, used here). This does mean any portals that need file descriptors can't be proxied over TCP or what have you, but supporting that felt far out of scope.
I did consider using upgrade to to implement a socket-support mode and a basic mode so that simpler portals (like settings) could be proxied, but that didn't feel reasonable.

I came up with the concept of an instances to fill in some of the stuff dbus normally provides. Whereas in dbus you can pretty easily track where messages come from, with varlink it is a bit trickier in edge cases. Therefore while it is optional for a client to register with `org.freedesktop.host.portal.Registry`, I made the equivalent varlink interface mandatory, as well as adding the concept of an instance on top.

The first call each connection on the socket makes must always be to `org.freedesktop.host.portal.Registry.Register` or to `org.freedesktop.host.portal.Registry.Claim`. Register makes a `XdpAppInfo` (using a hint passed to Register for host apps, and the flatpak/snap/etc id for sandboxed ones) and stores it, returning an instance id. Future connections by that app must `Claim` that instance with the returned id. Host apps can claim any other host app's isntance, but sandboxed apps have to match on engine, app, and instance (the sandbox concept, not this; in hindsight i should have picked a better name). A given instance is lazily cleaned up once all connections that use it are closed, though a client can re-register again for free.

Other departures from dbus:
- no Request/Response, varlink sockets just block for however long is needed
- Sessions are not closed by an explicit call, they are closed by closing the varlink socket that session lives on
- while dbus can have any number of signals and calls going at a time, varlink sockets block when subscribed (i.e. `org.freedesktop.portal.Settings.SubscribeSettingChanged`), therefore the client architecture I imagine is keeping n+1 connections on the socket open at once where n is the number of blocked connections. Accordingly I tried to design with a large number of open sockets in mind

Depends on libvarlink including https://github.com/varlink/libvarlink/pull/89 and https://github.com/varlink/libvarlink/pull/92

LLM assistance was used to help me understand some of the glib-isms (my background is more in Qt and embedded) and to write tests

Obviously doubt this will be merged, was mainly intended as a learning experience for me with portals and varlink


swick:
Not going to happen. The entire advantage of using varlink would be to cut down the middle men. This just adds more code for no reason.

jthoward64:
> Not going to happen. The entire advantage of using varlink would be to cut down the middle men. This just adds more code for no reason.

Figured as much! Just wanted to play around with the idea

PR #2139, opened 2026-09-10
realtime: skip export when rtkit cannot provide its properties

### LINKED ISSUE #2110: Realtime portal permanently exports RTTimeUSecMax=0 if the one-shot startup read of rtkit properties fails — downstream clients get SIGKILLed by the kernel
## Summary

`desktop-portal/realtime.c` reads rtkit's `MaxRealtimePriority` / `MinNiceLevel` /
`RTTimeUSecMax` exactly once, at portal startup (`realtime_new` / `load_all_properties`).
If the rtkit proxy creation or a property `Get` fails at that moment, the code logs a
warning and deliberately keeps the Realtime interface exported anyway ("it will fail to do
anything") — but the gdbus skeleton properties remain at their default value **0**, forever.
There is no retry and no lazy re-read. Identical code in 1.22.1 and current main (post-1.22.1
changes to realtime.c/xdp-utils.c are refactors only).

The consequence is much worse than "fails to do anything": pipewire's module-rt (loaded in
ALL pipewire clients since 1.4.0) reads the exported `RTTimeUSecMax`, receives a *valid* 0,
and sets `RLIMIT_RTTIME` to {0,0} on the client process before realtime scheduling is
granted (companion pipewire issue: https://gitlab.freedesktop.org/pipewire/pipewire/-/work_items/5429). rtkit then grants RT (0 passes
its limit check), and the kernel SIGKILLs the entire client process on the first tick —
silently. Observed as: every JACK application (REAPER in my case) vanishing ~1s after
launch, no log, no core dump, 5/5 reproducible.

## Environment

- xdg-desktop-portal 1.22.1 (Arch Linux), rtkit-daemon running
- pipewire 1.6.8 / pipewire-jack client
- kernel 7.1.8-arch1-3, Wayland/Hyprland session
- Each client launch also logs:
  `Realtime error: Could not map tid X of pid Y: Process ids could not be found: X`
  (secondary symptom — by the time MakeThreadRealtime scans /proc/<pid>/task, the thread
  group is dead or dying from the SIGKILL described above)

## Evidence (client side)

```
# strace: pipewire module-rt inside the client, after reading RTTimeUSecMax=0 from the portal
prlimit64(0, RLIMIT_RTTIME, {rlim_cur=0, rlim_max=0}, NULL) = 0

# ftrace signal_generate: kernel RLIMIT_RTTIME enforcement killing the RT data-loop thread's
# whole process, SI_KERNEL
data-loop.0-51964 ... signal_generate: sig=9 errno=0 code=128 comm=data-loop.0 grp=1
```

## Suggested fix

One or more of:
1. Retry / lazily (re)load the rtkit properties on first method call or property read
   instead of one-shot at startup.
2. If the properties could not be loaded, do NOT export the interface with bogus zero
   values — either unexport it or return an error on property access.
3. At minimum, never advertise `RTTimeUSecMax = 0`, which downstream consumers reasonably
   treat as an actual budget and which turns into a guaranteed kernel SIGKILL.

Related (same error family, different sub-causes, none fixing this): #1167, #1653, #1756,
#747 / PR #1169, #1953.

ravi-arnan:
## Problem

When rtkit's D-Bus proxy creation or its `Properties.Get` calls fail at
startup, the Realtime portal keeps exporting the interface with the gdbus
skeleton's default value (0) for `RTTimeUSecMax` (and the other two).

This is worse than "does nothing": PipeWire's `module-rt` (loaded in all
clients since 1.4.0) reads the exported `RTTimeUSecMax`, sees a *valid*
0, sets `RLIMIT_RTTIME={0,0}`, and the kernel SIGKILLs the entire
client process on the first realtime tick. Reproduced by two independent
reporters (rtkit present but a startup race, and rtkit absent entirely).
Issue: #2110.

## Root cause

`realtime_new()` deliberately continues even when the rtkit proxy cannot be
created, with the comment "We continue on so the realtime interface
remains exported, however it will fail to do anything". The `!rtkit_proxy`
guard makes the method handlers return an error, but the exported properties stay
at their 0 defaults -- and properties are what PipeWire reads.

## Fix

Make the portal **not export at all** when rtkit cannot provide its
properties:

- `load_all_properties()` now returns whether every property loaded.
- `realtime_new()` returns `NULL` (cleanly finalized via `g_autoptr`) if
  the rtkit proxy cannot be created **or** any property fails to load.
- `init_realtime()` skips `xdp_context_take_and_export_portal()` when
  `realtime_new()` returns `NULL`.

This matches the existing pattern in `init_usb` / `init_location`, which
`return` without exporting when a required dependency is unavailable. A client
now gets an unambiguous `org.freedesktop.DBus.Error.ServiceUnknown` / unknown
method rather than a plausible-looking 0 that the kernel enforces as a
termination sentence.

The change is one file, `desktop-portal/realtime.c`, +19/-11.

## Verification

- Follows the exact guard style of `init_usb()` / `init_location()` in the
  same `desktop-portal/` directory.
- `g_autoptr(Realtime)` was already defined (line 44), so the early
  `return NULL` paths release the proxy correctly through `realtime_finalize()`.

**Automation disclosure**, separate from the commit trailer as the policy
requires: this pull request's changes and this summary were prepared with Claude
Code. I reviewed the diff and verification myself before submitting, and I am the
person accountable for it.

Fixes: https://github.com/flatpak/xdg-desktop-portal/issues/2110

### DIFF
--- a/desktop-portal/realtime.c
+++ b/desktop-portal/realtime.c
@@ -233,12 +233,13 @@ realtime_class_init (RealtimeClass *klass)
   object_class->finalize = realtime_finalize;
 }
 
-static void
+static gboolean
 load_all_properties (Realtime *realtime)
 {
   GDBusProxy *proxy = realtime->rtkit_proxy;
   const char * properties[] = { "MaxRealtimePriority", "MinNiceLevel", "RTTimeUSecMax" };
   enum prop_type { MAX_REALTIME_PRIORITY, MIN_NICE_LEVEL, RTTIME_USEC_MAX };
+  gboolean success = TRUE;
 
   for (guint i = 0; i < G_N_ELEMENTS (properties); ++i)
     {
@@ -250,14 +251,15 @@ load_all_properties (Realtime *realtime)
       result = g_dbus_proxy_call_sync (proxy,
                                        DBUS_DBUS_IFACE ".Properties.Get",
                                         g_steal_pointer (&parameters),
-                                        G_DBUS_CALL_FLAGS_NONE,
-                                        -1,
-                                        NULL,
-                                        &error);
+                                       G_DBUS_CALL_FLAGS_NONE,
+                                       -1,
+                                       NULL,
+                                       &error);
 
       if (!result)
         {
           g_warning ("Failed to load RealtimeKit property: %s", error->message);
+          success = FALSE;
         }
       else
         {
@@ -279,14 +281,16 @@ load_all_properties (Realtime *realtime)
           g_dbus_proxy_set_cached_property (proxy, properties[i], value);
         }
     }
+
+  return success;
 }
 
 static Realtime *
 realtime_new (void)
 {
-  Realtime *realtime;
   g_autoptr(GError) error = NULL;
   g_autoptr(GDBusProxy) rtkit_proxy = NULL;
+  g_autoptr(Realtime) realtime = NULL;
 
   rtkit_proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
                                                G_DBUS_PROXY_FLAGS_NONE,
@@ -298,9 +302,8 @@ realtime_new (void)
                                                &error);
   if (!rtkit_proxy)
     {
-      /* We continue on so the realtime interface remains exported,
-       * however it will fail to do anything */
       g_warning ("Failed to create RealtimeKit proxy: %s", error->message);
+      return NULL;
     }
 
   realtime = g_object_new (realtime_get_type (), NULL);
@@ -308,10 +311,13 @@ realtime_new (void)
 
   xdp_dbus_realtime_set_version (XDP_DBUS_REALTIME (realtime), 1);
 
-  if (realtime->rtkit_proxy)
-    load_all_properties (realtime);
+  if (!load_all_properties (realtime))
+    {
+      g_warning ("Failed to load RealtimeKit properties at startup");
+      return NULL;
+    }
 
-  return realtime;
+  return g_steal_pointer (&realtime);
 }
 
 void
@@ -320,6 +326,8 @@ init_realtime (XdpContext *context)
   g_autoptr(Realtime) realtime = NULL;
 
   realtime = realtime_new ();
+  if (realtime == NULL)
+    return;
 
   xdp_context_take_and_export_portal (context,
                                       G_DBUS_INTERFACE_SKELETON (g_steal_pointer (&realtime)),


swick:
I don't like that it just doesn't export the interface. Is 0 really a valid value? Why do clients use it if it's not? What could we set it, for clients to understand that it means "unsupported"?

I'm really missing explanations to those questions.

ravi-arnan:
> I don't like that it just doesn't export the interface. Is 0 really a valid
> value? Why do clients use it if it's not? What could we set it, for clients
> to understand that it means "unsupported"?

Answering in order, with the sources.

**Both failure modes already export 0 today.** If the proxy creation fails,
`realtime_new` warns and deliberately keeps exporting ("We continue on so the
realtime interface remains exported, however it will fail to do anything"), and
if a property `Get` fails, `load_all_properties` warns and leaves that property
at the skeleton default, which is also 0. So this is not a new behavior I am
inventing for one path; it is the status quo on both.

**The "fail to do anything" premise is what #2110 disproves.** The methods do
fail without a proxy, but the properties do not fail: they succeed with 0, and
0 is consumable. `RTTimeUSecMax` is a usec budget with no out-of-band
"unknown" state, and per rtkit's own README it is a *minimum* threshold: a
client must have chosen at least that much `RLIMIT_RTTIME` before rtkit grants
scheduling. A 0 satisfies the minimum trivially, so rtkit grants, and the
client — PipeWire's `module-rt`, which copies the exported value as its own
limit — gets `RLIMIT_RTTIME={0,0}` and is SIGKILLed by the kernel on its first
realtime tick. That chain is not hypothesized: #2110 carries the strace
(`prlimit64(0, RLIMIT_RTTIME, {0,0})`) and the ftrace `signal_generate sig=9`,
5/5 reproducible, every JACK client vanishing ~1s after launch. The clients use
0 because the protocol tells them to trust it; they cannot distinguish "the
portal measured 0" from "the portal failed to measure".

**There is no in-band value that reads as "unsupported".** The candidates:

- A huge value ("unlimited") fails open: it would hand every sandboxed client a
  realtime budget the host never approved. Strictly worse than the bug.
- Erroring on the property `Get` is mechanically possible, but it asks every
  client to interpret a failure mode none of them handles today, in exchange
  for keeping an export whose methods already fail. Absence is the only signal
  with pre-existing defined semantics: `ServiceUnknown`.

And absence-when-the-backend-is-dead is this codebase's own pattern:
`init_usb` returns before exporting when there is no impl config, and
`init_location` does the same without an access impl. A half-exported interface
lies to clients; neither of those two tells the lie.

One alternative I did not take, from the issue's own list: lazily (re)loading
the properties on first method call instead of one-shot at startup. That keeps
the export for a rtkit that recovers, but it is a larger change and it still
exports zeros until the first call lands. Happy to do that version instead if
you prefer the export kept; this PR is the minimal fail-closed fix.

smcv:
> PipeWire's module-rt (loaded in all clients since 1.4.0) reads the exported RTTimeUSecMax, sees a valid 0, sets RLIMIT_RTTIME={0,0}, and the kernel SIGKILLs the entire client process on the first realtime tick.

Why is this a useful thing for Pipewire to be doing? Shouldn't Pipewire interpret RTTimeUSecMax=0 as meaning: I am not allowed to have any real-time scheduling, so don't try to be scheduled as real-time?

ravi-arnan:
> Why is this a useful thing for Pipewire to be doing? Shouldn't Pipewire
> interpret RTTimeUSecMax=0 as meaning: I am not allowed to have any real-time
> scheduling, so don't try to be scheduled as real-time?

**First, a correction to my comment above.** I called `RTTimeUSecMax` a minimum
threshold. That wording came from rtkit's README ("...need to have chosen at
minimum"), but the daemon does the opposite: `verify_process_rttime()` rejects
the process unless its `RLIMIT_RTTIME` *hard* limit is `<= rttime_usec_max`
(`/* Enforce that clients have RLIMIT_RTTIME set to a value <= this */`), and
re-checks it after granting. rtkit's own `--help` says "not greater than this".
It is a ceiling, not a floor. My mechanism wording was wrong; the outcome is the
same, and it matters for the next part.

**So no, it is not a useful thing for PipeWire to be doing.** The property isn't
a hint, it's an input: rtkit refuses RT to any process that has not already set
`RLIMIT_RTTIME`, so the client must pick a value before it asks, and this
property is what it is allowed to pick. `module-rt` does exactly that in
`rtkit_finish_init()`:

    impl->rl.rlim_cur = SPA_MIN (impl->rl.rlim_cur, rttime_max);
    impl->rl.rlim_max = SPA_MIN (impl->rl.rlim_max, rttime_max);

i.e. "cap my own limit at the advertised ceiling". With a ceiling of 0 you get
`RLIMIT_RTTIME={0,0}`, rtkit is satisfied (`0 <= 0`), grants RT, and the kernel
kills the thread on the first non-blocking RT tick. 0 is not out of contract
here — it is in range and means "zero CPU budget", which is what the client does
with it.

**And 0 still cannot mean "unsupported".** The property is a plain budget: any
larger value fails *open* (handing out a budget the host never approved), and 0
is already a legal in-range value. There is no in-band sentinel left.

**Yes, PipeWire should also guard against 0, and I will file that against
module-rt separately** ("if the ceiling is 0, don't set the rlimit and don't ask
for RT"). But it cannot be the whole fix, because the portal manufactures this 0
when a read fails, with nothing for the client to tell it apart from a real 0. A
client-side guard is guessing which zero it got.

**And absence is already handled by module-rt — better than 0 is.** Verified
against its source: with the interface gone, the `Properties.Get` calls error
out, the query results stay invalid, `SELECT()` falls back to defaults, and the
later `MakeThreadRealtimeWithPID` fails with a D-Bus error, so it logs "could not
make thread realtime" and the client simply runs without RT. Missing interface →
degrades to non-RT; exported 0 → app is killed. That asymmetry is the whole
reason for the shape of this PR, and `init_usb()` / `init_location()` in this
same file already skip the export when their backend is missing.

If you would rather the interface stayed exported, the alternative is to keep
the object and make the property `Get` fail instead of answering 0 when nothing
was read. Same result for the client, one more moving part in the skeleton. Tell
me which you prefer and I will do that version.

swick:
> nothing for the client to tell it apart from a real 0

Is a real 0 a useful value?

ravi-arnan:
No, and for this property it is not even reachable. rtkit is its only producer,
and rtkit's own option parser refuses to be configured with it:

    case ARG_RTTIME_USEC_MAX:
        rttime_usec_max = strtoull(optarg, &e, 0);
        if (errno != 0 || !e || *e || rttime_usec_max <= 0) {
            fprintf(stderr, "--rttime-usec-max parameter invalid.\n");
            return -1;
        }

`--rttime-usec-max=0` is rejected; the default is 200000. So the daemon this
interface proxies can never export `RTTimeUSecMax = 0`, which is why I stopped
hedging about the two zeros: the only way that property reads 0 today is a read
that failed.

And if some other implementation did send 0, it still would not mean "no
realtime". rtkit grants RT to a process whose hard limit is `<=` the advertised
value, so a ceiling of 0 is satisfied by `0 <= 0` and the client is made
realtime with a zero CPU budget, which is the value that gets it killed. A
ceiling cannot express "do not schedule me"; the absent interface and the
failing method call are the channels that can.

The other two properties are the reason this looks like a 0 problem in general
and is not one: `MaxRealtimePriority = 0` and `MinNiceLevel = 0` are ordinary
values that only clamp a client conservatively. A budget of 0 is the one that is
lethal instead of merely restrictive, and it is the one whose only reachable
source is the failure case.

One more data point for the consumer-side guard I offered to file: `module-rt`
already declines to trust a 0 of this kind, since `get_rtkit_priority_range()`
coerces a max priority below 1 up to 1. It simply has no equivalent clamp for
the rttime ceiling. That guard belongs on the client side whatever the portal
exports.

ravi-arnan:
Correction to my last paragraph: I said I would file the PipeWire side separately, but it
already exists and I should have checked before writing that.

- [pipewire#5429](https://gitlab.freedesktop.org/pipewire/pipewire/-/work_items/5429),
  "module-rt: RTTimeUSecMax=0 from portal is accepted, sets RLIMIT_RTTIME to {0,0}, client
  is SIGKILLed by kernel the moment RT is granted", opened 2026-08-15 by Sequoyah Kennedy,
  still open. Same chain as #2110 and the same suggested guard: treat a ceiling of 0 as
  invalid instead of clamping to it.
- I re-read `src/modules/module-rt.c` on master today. The `SPA_MIN` clamp in
  `rtkit_finish_init()` still has no zero check, so it is not fixed there.

That report predates this thread by a month, so the client side is already tracked as a
public bug and nothing here is news to it. It also says "companion xdg-desktop-portal issue,
link here after filing" and never got the link; #2110 and this PR are that companion, which
is why I am putting it in this thread. None of that changes the PR: the portal still should
not manufacture that 0. Sorry for the churn.

swick:
My instinct here is to fix callers of the portal, and to let them handle 0 as a sentinel for not supported.

ravi-arnan:
I checked the sentinel against the sources before answering, and it does not hold
for this interface. Three things, all from current master.

**0 is a legal, reachable value for one of the three properties.** rtkit parses
MinNiceLevel with `i < PRIO_MIN || i >= PRIO_MAX` (rtkit-daemon.c:2370), and
PRIO_MIN/PRIO_MAX are -20/20, so `--min-nice-level=0` is accepted. It is also
meaningful rather than degenerate: the daemon denies a request when
`nice_level < min_nice_level` (rtkit-daemon.c:828), so a minimum of 0 means
"nobody may take a negative nice level", which is a reasonable conservative
setting. A client that reads 0 as "unsupported" would stop touching niceness on a
host where rtkit is working exactly as configured.

That breaks the rule for one of the three:

- `RTTimeUSecMax`: `rttime_usec_max <= 0` is rejected (rtkit-daemon.c:2383),
  default 200000. 0 is unreachable.
- `MaxRealtimePriority`: rejected below `sched_get_priority_min(sched_policy)`
  (rtkit-daemon.c:2356). The policy is SCHED_RR by default and the option accepts
  only `rr` or `fifo` (rtkit-daemon.c:165, :2309), both of which have a floor of
  1, so 0 is unreachable here too.
- `MinNiceLevel`: 0 is reachable, as above.

**0 is already a fallback value on the consumer side, not a sentinel.** In
module-rt.c on master, `rtkit_finish_init` has
`const int min_nice_level = SELECT(min_nice_level, 0);` (line 982), so
module-rt's own default for that property is 0. The absent-interface case walks
the same path and degrades well: the property reads never produce a result, the
results stay at `INVALID_PROP_VALUE`, `SELECT` substitutes a default, and the
later `MakeThreadRealtimeWithPID` fails with a D-Bus error that module-rt logs
and survives. Missing interface, client runs without RT. Exported 0, the client
sets `RLIMIT_RTTIME={0,0}` (module-rt.c:1014, no zero check on `rttime_max`) and
the kernel kills it.

**Upstream's own convention is a separate channel, not an in-band value.** The
same three properties read through librtkit return `err < 0 ? err : retval`
(rtkit.c:141, :159, :163), so unavailability is signalled out of band and 0 is
just a value. "0 means unsupported" would be a new convention introduced for
this interface, and its own documentation says it "is analogous to the
org.freedesktop.RealtimeKit1 interface and will proxy requests there".

**On partial failure, which is what decides whether a sentinel could work at
all.** On main, `load_all_properties` sets each property independently inside its
loop, so one failed `Get` leaves that property at the skeleton default while the
other two carry real values. The interface can answer with a mixture, and a
client cannot separate "we failed to read this one" from "rtkit reported this
one". Making a sentinel reliable would mean zeroing all three on any failure,
which is where the `MinNiceLevel = 0` case above turns into a wrong answer
rather than a conservative one.

I am not attached to the shape of this PR. Two versions keep the export:

- Keep the object and make the property `Get` fail when nothing was read, so the
  failure travels on the D-Bus error channel. One more moving part in the
  skeleton, same outcome for clients.
- Lazily (re)load the properties on first method call instead of one-shot at
  startup, which the issue suggested. Larger change, and it still exports zeros
  until the first call lands.

Either is fine by me, say which and I will switch this PR to it. I went with
absence because it reuses semantics D-Bus already defines (`ServiceUnknown`) and
because the two portals in this same file that depend on a backend already do
it: `init_usb` returns before exporting when there is no impl config (usb.c:1535),
and `init_location` returns when there is no access impl (location.c:769).

PR #2148, opened 2026-09-18
tests: Add initial tests for MemoryMonitor portal

TheEvilSkeleton:


GeorgesStavracas (approved):

tests/test_memorymonitor.py
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# SPDX-FileCopyrightText: Copyright © the xdg-desktop-portal contributors
```
zeldaAhmed, line 2: New files require "Para-Real Ltd" as the copyright, i think

zeldaAhmed (approved):

tests/test_memorymonitor.py
zeldaAhmed: @swick am i correct in thinking that?

TheEvilSkeleton: Thanks, you're right about that

swick: They do not; we're tracking copyright via commits (if you change a file in a significant way, you also now are proud copyright owner). You might want to commit using a different (non-personal) mail address or use `--signoff` with a different address to track things properly.

PR #2152, opened 2026-09-22
Fuse file locking

swick:


smcv:
I haven't reviewed this in detail or anything, but here is some cursed knowledge that will likely be essential for a proper review. I'm sure @swick already knows this but reviewers might not.

There are several orthogonal, incompatible advisory locking APIs.

* `flock(2)` (`LOCK_SH`, `LOCK_EX`, `LOCK_UN`) ("BSD locks") (held by an open file description, which is the kernel object that fds point to)
* `fcntl(2)` `F_SETLK`, `F_SETLKW`, `F_GETLK` (POSIX "process-associated" locks) (locks are held by a *process* which implies semantics that x-d-p probably can't implement)
* `fcntl(2)` `F_OFD_SETLK`, `F_OFD_SETLKW`, `F_OFD_GETLK` (Linux "open file description" locks) (locks are held by an OFD, like `flock(2)`
* `lockf(3)`

On Linux, `lockf(3)` is just `fcntl(2)` `F_SETLK` with a different hat, I think.

Just to troll you further, it is mostly unspecified whether taking a lock with one of these APIs prevents taking an incompatible lock with a different one of these APIs. For example if process A takes an exclusive (write) flock(2) lock, and process B tries to take a `F_SETLK` lock, it's unspecified whether B will fail (prevented by A's lock) or succeed (both locks are held, independently and orthogonally).

According to my notes in steam-runtime-tools, F_OFD_SETLK and F_SETLK *are* documented to conflict with each other (but I can't remember what reference I got that from).

smcv:
> According to my notes in steam-runtime-tools, F_OFD_SETLK and F_SETLK are documented to conflict with each other (but I can't remember what reference I got that from).

Ah, here we are:

https://manpages.debian.org/trixie/manpages-dev/fcntl.2.en.html#Open_file_description_locks_(non-POSIX)

"Conflicting lock combinations (i.e., a read lock and a write lock or two write locks) where one lock is an open file description lock and the other is a traditional record lock conflict even when they are acquired by the same process on the same file descriptor."

and https://manpages.debian.org/trixie/manpages-dev/fcntl.2.en.html#NOTES

"Since Linux 2.0, there is no interaction between the types of lock placed by [flock(2)](https://manpages.debian.org/trixie/manpages-dev/flock.2.en.html) and fcntl()."

So:

* `flock(2)` vs. `fcntl(2) F_SETLK`: no interaction, either or both flavours of lock can be held
* `flock(2)` vs. `fcntl(2) F_OFD_SETLK`: no interaction, either or both flavours of lock can be held
* `fcntl(2) F_SETLK` vs. `fcntl(2) F_OFD_SETLK`: each lock excludes the other (unless they are both read/shared locks)

swick:
> fcntl(2) F_SETLK, F_SETLKW, F_GETLK (POSIX "process-associated" locks) (locks are held by a process which implies semantics that x-d-p probably can't implement)

You'd think so, but the actual rules are:
1. when there is an explicit unlock
2. when any file descriptor referring to the file gets closed (note that this says file descriptor, not file description)
3. when the process dies

AFAIU when a process dies, it closes all its file descriptors, and because a process needed an open file descriptor to lock the file, case 3 is really included in case 2. Fuse lowlevel calls flush for each file descriptor(!) which gets closed. So dropping all locks for the `lock_owner` (which is the process for "process-associated" locks) in flush nicely solves this all.

Maybe also worth noting: the fuse kernel driver lowers the OFD locks to "process-associated" locks with a unique `lock_owner`. The same code also works just fine here.

> So:
> 
>     flock(2) vs. fcntl(2) F_SETLK: no interaction, either or both flavours of lock can be held
>     flock(2) vs. fcntl(2) F_OFD_SETLK: no interaction, either or both flavours of lock can be held
>     fcntl(2) F_SETLK vs. fcntl(2) F_OFD_SETLK: each lock excludes the other (unless they are both read/shared locks)

Yes, and because fuse lowers F_OFD_SETLK to F_SETLK, we use F_OFD_SETLK for both, so they naturally exclude each other, but do not exclude flock which we implement via flock.

